Issue #3, opened 2022-05-05
Add document describing steps to add a new provider


bpradipt:
Create a basic guide with prerequisites and steps to add a new provider support.
Aspects like - golang binding for the provider APIs, code changes required etc

PR #19, opened 2022-05-17
Basic info to add a new provider

bpradipt:
Fixes: #3

### DIFF
--- a/docs/addnewprovider.md
+++ b/docs/addnewprovider.md
@@ -1,0 +1,56 @@
+# :memo: Adding support for a new provider
+
+### Step 1: Add provider specific options 
+
+Provider specific options goes under `cmd/cloud-api-adaptor`
+- [Options](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L48)
+- [Parsing](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/parse.go#L21)
+- [Calling the specific provider](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L103)
+
+
+### Step 2: Add provider specific code 
+
+- The code goes under `pkg/adaptor/hypervisor/<provider>`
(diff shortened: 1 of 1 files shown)


docs/addnewprovider.md
```diff
@@ -0,0 +1,56 @@
+# :memo: Adding support for a new provider
+
+### Step 1: Add provider specific options 
+
+Provider specific options goes under `cmd/cloud-api-adaptor`
+- [Options](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L48)
+- [Parsing](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/parse.go#L21)
+- [Calling the specific provider](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L103)
```
yoheiueda, line 8: Do we need to use permalinks here?

yoheiueda (commented):
This documentation looks good explaining the current design.

I have a question about the current design. As described in Step2, separate binaries for each provider are generated using build tag. However, there are switch statements to check hypervisor type at runtime in main.go. I think we can eliminate these runtime checks using build tags. 

Do you have any specific reason to use runtime checks instead of build tags in main.go?

docs/addnewprovider.md
bpradipt: > This documentation looks good explaining the current design.
> 
> I have a question about the current design. As described in Step2, separate binaries for each provider are generated using build tag. However, there are switch statements to check hypervisor type at runtime in main.go. I think we can eliminate these runtime checks using build tags.
> 
I was not aware of a way to eliminate these checks without having separate main files. If we can avoid that it'll be good.
So once the changes are there in the code, the same can be updated in this doc.

bpradipt: > Do we need to use permalinks here?

Not really, but I thought its better to point to specific example. Do you have any other suggestion? Happy to change.

yoheiueda: I see. Let's merge this patch for now. We will update this doc when we change the code.

yoheiueda (approved):
LGTM

PR #20, opened 2022-05-19
go: use github.com/yoheiueda/kata-containers tree

wainersm:
Currently go.mod instructs to replace the
github.com/kata-containers/kata-containers/src/runtime module with one
found on localhost (../kata-containers/src/runtime), which means the
developer must clone the repository manually and place on the right
directory at go root.

However, in practice, the tree github.com/yoheiueda/kata-containers/tree/CCv0-peerpod has been
used as the "official" kata-containers/src/runtime fork at the moment. So to simplify
the project's bootstrap on developers side as well as for a future CI, this
changed the go.mod so that Yohei's tree is explicitly referenced.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,6 @@ require (
 )
 
 replace (
-	github.com/kata-containers/kata-containers/src/runtime => ../kata-containers/src/runtime
+	github.com/kata-containers/kata-containers/src/runtime => github.com/yoheiueda/kata-containers/src/runtime v0.0.0-20220429130351-22a2aa8a867d
 	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8
 )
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@wainersm - it looks fine to me, but I've requested @yoheiueda's review and it's his `CCv0-peerpod` branch that we are explicitly calling now.

stevenhorsman (approved):

wainersm:
> @wainersm - it looks fine to me, but I've requested @yoheiueda's review and it's his `CCv0-peerpod` branch that we are explicitly calling now.

sure, makes sense. thanks for the review @stevenhorsman !

go.mod
```diff
@@ -26,6 +26,6 @@ require (
 )
 
 replace (
-	github.com/kata-containers/kata-containers/src/runtime => ../kata-containers/src/runtime
+	github.com/kata-containers/kata-containers/src/runtime => github.com/yoheiueda/kata-containers/src/runtime v0.0.0-20220429130351-22a2aa8a867d
```
bpradipt, line 29: I think it's better to avoid this change for now
Instead let's update the dev doc - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/docs/DEVELOPMENT.md#clone-source-code

bpradipt: This avoids explicit reference and keep the developer flow intact. Otherwise for any change we need to explicitly modify the go.mod.
Sometime back I had created a dev guide which (my bad) is still sitting on my private google doc. Here is a link to the same - https://hackmd.io/nBPF7ounQoapOZTelkmXKA
Please see if it helps. Will create a PR once you confirm

yoheiueda: Yes, the `replace` statement is intentionally used to avoid explicit reference to my personal repository. When the changes are merged to the upstream kata containers repo, we can update `go.mod` not to use `replace`.

I agree that the current setup procedure is somewhat error prone due to this `replace` statement in go.mod. If there is  a better way to refer the patched kata containers repo, we can switch to it.

Issue #24, opened 2022-05-23
Add creation of the IS SSH key to the cluster Terraform template in x86 peer pod demo


jtumber-ibm:
## Description
The public cloud-api-adaptor repo contains a [Terraform template](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/ibmcloud/terraform/cluster) for setting up the VSI instances for the demo environment for running the cloud api adaptor and peer pods demo. This template requires a public SSH to be added to the region the VSI instances will run in in IBM Cloud, which currently has to be done manually in IBM Cloud outside of Terraform.

To get the full one-click experience when setting up the demo environment Terraform should create this SSH key as well. You'd do this by making Terraform create an instance of the `ibm_is_ssh_key` [resource](https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/is_ssh_key)

## Task List
- [ ] Read the docs for the VPC infrastructure SSH key in the IBM Cloud Terraform provider, https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/is_ssh_key, to understand how to create an instance of this resource and how to access the data values associated with this resource once it has been created
- [ ] Update the Terraform template in `ibmcloud/terraform/cluster` so that the template creates a new SSH key instance rather than using a pre-existing one
- [ ] Add the name of the SSH key to the template's `locals` (e.g., construct it using `var.cluster_name` like the other `locals` for the template)
- [ ] Make the creation of the instance template in this Terraform template use the newly created SSH key resource
- [ ] Add a new variable for the SSH public key value to the Terraform template and remove the SSH key name variables
- [ ] Test the updated Terraform template for the create the Kubernetes cluster - you will need to run the `ibmcloud/terraform/common` Terraform template before `ibmcloud/terraform/cluster` to create the VPC, subnets and security groups etc
- [ ] Update the README.md file in `ibmcloud/` to reflect the SSH key is added to IBM Cloud automatically and explain how to add your public SSH key to the Terraform template as an input variable

## Security Considerations
It is the public SSH key that should be added to `terraform.tfvars`, not the private key. As this is a public key there would be not bad security implications of disclosing this public key, but we want to make sure we do not accidentally write the private key to a file then carelessly upload it to source control.

## Performance Considerations
None

## Acceptance Criteria

**Given** I want a one-click experience when creating the Kubernetes cluster for the peer pod demo environment 
[...]

PR #26, opened 2022-05-23
Add creation of the IS SSH key to the cluster Terraform template

jtumber-ibm:
Have switched the cluster Terraform template to use the [ibm_is_ssh_key resource](https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/is_ssh_key) over the old data source definition. This reduces the manual effort on the part of the developer.

- [x] Update the Terraform template in `ibmcloud/terraform/cluster` so that the template creates a new SSH key instance rather than using a pre-existing one
- [x] Add the name of the SSH key to the template's `locals` (e.g., construct it using `var.cluster_name` like the other `locals` for the template)
- [x] Make the creation of the instance template in this Terraform template use the newly created SSH key resource
- [x] Add a new variable for the SSH public key value to the Terraform template and remove the SSH key name variables
- [x] Test the updated Terraform template for the create the Kubernetes cluster - you will need to run the `ibmcloud/terraform/common` Terraform template before `ibmcloud/terraform/cluster` to create the VPC, subnets and security groups etc
- [x] Update the README.md file in `ibmcloud/` to reflect the SSH key is added to IBM Cloud automatically and explain how to add your public SSH key to the Terraform template as an input variable

Fixes: #24 

Signed-off-by: jtumber-ibm <james.tumber@ibm.com>

### DIFF
--- a/ibmcloud/README.md
+++ b/ibmcloud/README.md
@@ -92,6 +92,10 @@ ibmcloud_api_key = "<your API key>"
 ssh_key_name = "<your SSH key name>"
 cluster_name = "<cluster name>"
 ```
+If you don't have your public key already configured in IBM Cloud you can add
+```
+ssh_pub_key = "<your SSH public key>"
+```
 > **Hint:** In order to create the cluster based on a different type of VSI image you can overwrite more parameters here e.g. to create a **s390x** based cluster add follow two lines to the `terraform.tfvars` file
 >
 >
@@ -101,8 +105,9 @@ cluster_name = "<cluster name>"
 
 > **Notes:**
 > - `ibmcloud_api_key` is your IBM Cloud API Key that you just created at [https://cloud.ibm.com/iam/apikeys](https://cloud.ibm.com/iam/apikeys).
-> - `ssh_key_name` is a name of your SSH key registered in IBM Cloud. It is used to access a Generation 2 virtual server instance. You can add your SSH key at [https://cloud.ibm.com/vpc-ext/compute/sshKeys](https://cloud.ibm.com/vpc-ext/compute/sshKeys). This ssh key will be installed on control-plane and worker nodes. For more information, about SSH key, see [managing SSH Keys](https://cloud.ibm.com/docs/vpc?topic=vpc-ssh-keys).
->
+> - An SSH Key is used to access a Generation 2 virtual server instance. This ssh key will be installed on control-plane and worker nodes.
+>   - `ssh_key_name` is the name of your SSH key registered in IBM Cloud or the name of a new SSH key if a public key is also provided. You can add your SSH key at [https://cloud.ibm.com/vpc-ext/compute/sshKeys](https://cloud.ibm.com/vpc-ext/compute/sshKeys). For more information, about SSH key, see [managing SSH Keys](https://cloud.ibm.com/docs/vpc?topic=vpc-ssh-keys).
+>   - `ssh_pub_key` is an **optional** field for a SSH public key which has **not** been registered in IBM Cloud, terraform will manage this key instead.
 > - `cluster_name` is a name of a Kubernetes cluster. This name is used for the prefix of the names of control-plane and worker nodes. If you want to create another cluster in the same VPC, you need to use a different name for the new cluster.
 > - `instance_profile_name` is a name of IBM Cloud virtual server instance profile. This name is used to create IBM Cloud virtual server instance. For more information, about virtual server instance profile, see [instance profiles](https://cloud.ibm.com/docs/vpc?topic=vpc-profiles).
 > - `image_name` is a name of IBM Cloud Infrastructure image. This name is used to create IBM Cloud virtual server instance. For more information, about VPC custom images, see [IBM Cloud Importing and managing custom images](https://cloud.ibm.com/docs/vpc?topic=vpc-managing-images).
@@ -119,9 +124,14 @@ $ terraform apply
 
 > **Tip:** You can check the status of provisioned Kubernetes node VM instances at [https://cloud.ibm.com/vpc-ext/compute/vs](https://cloud.ibm.com/vpc-ext/compute/vs).
 
+The SSH key installed on control-plane and worker nodes will be output at the end of the terraform configuration.
+
+```bash
+Outputs:
+ssh_key_name = <SSH key name>
+```
 
 This Terraform configuration also triggers execution of an Ansible playbook to set up Kubernetes and other prerequisite software in the two nodes. Please check [ansible/playbook.yml](terraform/cluster/ansible/playbook.yml) for the details.
-
 
 If ansible fails for some reason, you can rerun the ansible playbook as follows.
 ```bash
@@ -445,3 +455,22 @@ groups, subnet and gateway with:
 $ cd ibmcloud/terraform/common
 $ terraform destroy
 ```
+
+## Troubleshooting
+
+### Cluster creation when switching use `ssh_pub_key`
+
+If you provide your SSH public key via the `ssh_pub_key` Terraform variable and, having already run `terraform apply`, you run `terraform apply` again then you may see the following error regarding your SSH key:
+> Error: Error deleting SSH Key : SSH-key still in use
+
+This is caused by the IBM Cloud terraform provider forcing the replacement of the SSH key resource. As such Terraform may try to delete the existing SSH key while the Virtual Server Instances are using it.
+
+To resolve this issue:
+
+- Use `terraform state rm ibm_is_ssh_key.created_ssh_key[0]` to remove the SSH key resource from Terraform's internal state.
+- Delete the `ssh_pub_key` variable from the `terraform.tfvars` file
+- Then run `terraform apply` again
+
+This will remove the SSH Key from the list of resources whose lifecycle is managed by Terraform.
+
+When you delete the cluster you will need to manually delete the SSH Key in your IBM Cloud VPC Infrastructure.
--- a/ibmcloud/terraform/cluster/main.tf
+++ b/ibmcloud/terraform/cluster/main.tf
@@ -9,21 +9,26 @@ locals {
   controlplane_floating_ip_name = "${local.controlplane_name}-ip"
   worker_name = "${var.cluster_name}-worker"
   worker_floating_ip_name = "${local.worker_name}-ip"
-  ssh_key_name = var.user_ssh_key_name == "" ? "${var.cluster_name}-ssh-key" : "${var.user_ssh_key_name}"
   controlplane_ip = resource.ibm_is_instance.controlplane.primary_network_interface[0].primary_ipv4_address
   worker_ip = resource.ibm_is_instance.worker.primary_network_interface[0].primary_ipv4_address
   bastion_ip = resource.ibm_is_floating_ip.worker.address
 }
 
 resource "ibm_is_ssh_key" "created_ssh_key" {
+  # Create the ssh key only if the public key is set
   count = var.ssh_pub_key == "" ? 0 : 1
-  name = local.ssh_key_name
+  name = var.ssh_key_name
   public_key = var.ssh_pub_key
 }
 
 data "ibm_is_ssh_key" "ssh_key" {
+  # Wait if the key needs creating first
   depends_on = [ibm_is_ssh_key.created_ssh_key]
-  name = local.ssh_key_name
+  name = var.ssh_key_name
+}
+
+output "ssh_key_name" {
+  value = var.ssh_key_name
 }
 
 data "ibm_is_image" "k8s_node" {
--- a/ibmcloud/terraform/cluster/main.tf
+++ b/ibmcloud/terraform/cluster/main.tf
@@ -9,13 +9,21 @@ locals {
   controlplane_floating_ip_name = "${local.controlplane_name}-ip"
   worker_name = "${var.cluster_name}-worker"
   worker_floating_ip_name = "${local.worker_name}-ip"
+  ssh_key_name = var.user_ssh_key_name == "" ? "${var.cluster_name}-ssh-key" : "${var.user_ssh_key_name}"
   controlplane_ip = resource.ibm_is_instance.controlplane.primary_network_interface[0].primary_ipv4_address
   worker_ip = resource.ibm_is_instance.worker.primary_network_interface[0].primary_ipv4_address
   bastion_ip = resource.ibm_is_floating_ip.worker.address
 }
 
+resource "ibm_is_ssh_key" "created_ssh_key" {
+  count = var.ssh_pub_key == "" ? 0 : 1
+  name = local.ssh_key_name
+  public_key = var.ssh_pub_key
+}
+
 data "ibm_is_ssh_key" "ssh_key" {
-  name = var.ssh_key_name
+  depends_on = [ibm_is_ssh_key.created_ssh_key]
+  name = local.ssh_key_name
 }
 
 data "ibm_is_image" "k8s_node" {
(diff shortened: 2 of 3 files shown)


stevenhorsman (changes requested):

stevenhorsman:
Also, if you could add a `Fixes: #24` to the commit bodies so match the [required format](https://github.com/confidential-containers/community/blob/main/CONTRIBUTING.md#main-patch) that would be great.

ibmcloud/README.md
```diff
@@ -89,9 +89,18 @@ As usual, you need to create `terraform.tfvars` to specify parameter values. The
...
 cluster_name = "<cluster name>"
 ```
+
+> **Hint:** To use a previously imported SSH key, use `terraform import ibm_is_ssh_key.ssh_key <ssh_key.ID>` after you run `terraform init`
+>
+>> You can find your key's ID using the following command `ibmcloud is keys --json | jq '.[] | select(.name == "<key_name>").id'` requires `jq`
+>>
+>> Or visit [https://cloud.ibm.com/vpc-ext/compute/sshKeys](https://cloud.ibm.com/vpc-ext/compute/sshKeys) to manage your keys
+>
+> Addtionally, to avoid the ssh key being deleted use `terraform state rm ibm_is_ssh_key.ssh_key` before `terraform destroy`
```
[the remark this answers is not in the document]
liudalibj: I would like we use the `ssh_key_name` or `ssh_key` here, refer to the document https://cloud.ibm.com/docs/vpc?topic=vpc-ssh-keys and https://github.com/IBM-Cloud/terraform-provider-ibm/search?q=ssh_key
```
ssh_key = "<your SSH public key for VSIs>"
```

jtumber-ibm:
@stevenhorsman and @liudalibj Thanks for your comments, I've taken an alternative approach to allow for either a public key or the key name to be specified

ibmcloud/README.md
```diff
@@ -89,9 +89,16 @@ As usual, you need to create `terraform.tfvars` to specify parameter values. The
...
 ```
 ibmcloud_api_key = "<your API key>"
-ssh_key_name = "<your SSH key name>"
+user_ssh_key_name = "<your SSH key name>"
 cluster_name = "<cluster name>"
 ```
+Or
+```
+ibmcloud_api_key = "<your API key>"
+ssh_pub_key = "<your SSH public key>"
```
stevenhorsman, line 98: With this approach we still have the problem that we need to use the `ssh_key_name` later and can't set it. Can we take take the approach of setting the key name always and then using the presence of absence of the `ssh_pub_key` field to tell whether to create a new one, or check an existing one (which I assume is the switch here?)

stevenhorsman (commented):
I'm not sure if this was supposed to be re-reviewed yet, but I have a couple of comments so far

ibmcloud/README.md
```diff
@@ -89,9 +89,16 @@ As usual, you need to create `terraform.tfvars` to specify parameter values. The
 
 ```
 ibmcloud_api_key = "<your API key>"
-ssh_key_name = "<your SSH key name>"
+user_ssh_key_name = "<your SSH key name>"
```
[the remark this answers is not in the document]
jtumber-ibm: That's just a miss on my part to update line 112.

ibmcloud/README.md
jtumber-ibm: I'm not sure what you mean by 
> we need to use the ssh_key_name later and can't set it

the rest of the terraform code needs the ssh ID which it either fetches from the data source, if provided a key_name or from the resource if provided a public key (which has created a new entry)

stevenhorsman: The 'Launch cloud-api-adaptor` section requires the user to pass in the `ssh_key_name` in order to work. Currewntly the user would know the value of this as they specify it in the cluster creation, with this PR we don't let them set it, so are expecting them to have to look it up and haven't provided instructions for this? Does that make sense?

ibmcloud/README.md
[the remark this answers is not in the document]
MartinCocks: I think this variable could keep the original name. 
The only thing to be careful of is to avoid confusion between `var.ssh_key_name` and `local.ssh_key_name` in the code. Perhaps change the local variable name instead so that the external name is shorter?

ibmcloud/README.md
jtumber-ibm: Ah yes, thanks for pointing that out. I'd only been focused on the cluster creation section. Didn't realise it would be required elsewhere. Your earlier comment makes more sense now. I definitely think it should be possible to have the ssh_key_name be required and just switch the logic based upon the existance of the pub_key. I'll double check that and update the code with those changes. Thanks again.

MartinCocks: I would counter that point. If I get terraform to create and manage the key for me, why should I have to choose its name? If another programatic stage needs to know it [terraform output variables](https://www.terraform.io/language/values/outputs) are probably a better option.

jtumber-ibm: Or as the naming pattern is based upon the cluster-name variable could just update the reference to use that in the docs
i.e.
```
$ image_name=<pod VM image name>
$ ssh_key_name=<your cluster name>-ssh-key
$ vpc_name=tok-vpc
```
??

jtumber-ibm: @stevenhorsman This maybe going too far beyond the scope of this issue but as there are a lot of parameters that need extracting for launch step is it worth adding them directly to the output that Martin has mentioned. 
The step requires
- key-id
- image-id
- profile-name
- zone-name
- primary-subnet-id
- primary-security-group-id
- vpc-id
which could all be outputted by terraform?

stevenhorsman: @jtumber-ibm - so after having a chat with Martin is seems like the best thing to do is create an output variable for the `ssh_key_name` in this step and longer-term is to use that as the input of the `Launch cloud-api-adaptor` submodule, once we've stitched all the terraform templates together.
Until we are in the state, in the short term, I think we should avoid trying to be too clever with defaults and just continue the pattern of prompting the user to enter the same value multiple times and then review and clean these all up later. Based on that my suggestion would just be to get the user to supply `ssh_key_name` always during the cluster creation and if they set `ssh_pub_key` use it as the name of the new key we generate.
Sorry to muck you around as we've been thinking and learning out loud. Does this make sense for the tactical step?

jtumber-ibm: Sure, I'll make those changes and update the documentation to mention the output variable too.

stevenhorsman: > @stevenhorsman This maybe going too far beyond the scope of this issue but as there are a lot of parameters that need extracting for launch step is it worth adding them directly to the output that Martin has mentioned. The step requires
> 
> * key-id
> * image-id
> * profile-name
> * zone-name
> * primary-subnet-id
> * primary-security-group-id
> * vpc-id
>   which could all be outputted by terraform?

Sorry, I hadn't refreshed whilst typing, so hadn't see that message. I think your point matches my suggestion though - there is lots of room to improve the 'chaining' of variables, but that's outside the scope of this PR :)

stevenhorsman (commented):
These changes are looking good - I tested both the scenario of having and existing key and creating a new one and they both worked without errors. I've left a couple of optional comments, but you can choose to ignore them if you wish :)
The other things that is required are:
- Rebasing on the latest staging
- Doing a squash of related commits together
- Adding `Fixes: #24` to those commit messages
Thanks!

ibmcloud/README.md
```diff
@@ -92,6 +92,13 @@ ibmcloud_api_key = "<your API key>"
 ssh_key_name = "<your SSH key name>"
 cluster_name = "<cluster name>"
 ```
+Or, if you don't have your public key already configured
```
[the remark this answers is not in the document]
jtumber-ibm: Thanks, yes I agree I think that does make it clearer

jtumber-ibm:
Hi @stevenhorsman 

Changes complete, and rebased to staging including adding the fixes line to the commits. Please review at your convenience.

Cheers JT

ibmcloud/README.md
```diff
@@ -119,9 +124,14 @@ $ terraform apply
 
 > **Tip:** You can check the status of provisioned Kubernetes node VM instances at [https://cloud.ibm.com/vpc-ext/compute/vs](https://cloud.ibm.com/vpc-ext/compute/vs).
 
+The SSH key installed on control-plane and worker nodes will be output at the end of the terraform configuration.
```
mattarnoatibm, line 127: Because of the way the IBM Cloud Terraform provider works with the `ibm_is_ssh_key` resource [reference here](https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/is_ssh_key#argument-reference) the resource is forced to be recreated every time you run `terraform apply`.

If you provide the `ssh_pub_key` variable to let Terraform create the SSH key in VPC and run `terraform apply` again after running `terraform apply` once, the second `terraform apply` fails with an error like this:

```
ibm_is_ssh_key.created_ssh_key[0]: Destroying... [id=r038-06cafbfb-42b9-41ac-a718-57c55446ff53]
╷
│ Error: Error deleting SSH Key : SSH-key still in use by 2 instance(s). First instance ID: 02q7_16201105-5828-4673-ab88-2ad1625aa079
│ {
```

this is because the SSK key cannot be destroyed and replaced, as the Terraform provider wants to do, because the key is being used by the two Virtual Server instances.

I found a work around which involves one low-level Terraform state manipulation command, which nonetheless allows `terraform apply` to be re-ran, so I think we should document this work around for people who are providing the `ssh_pub_key` to this Terraform template. The work around is:

* Use `terraform state rm` to remove the created SSH key resource from Terraform's internal state:
```
terraform state rm ibm_is_ssh_key.created_ssh_key[0]
```
* Delete the `ssh_pub_key` variable from the `terraform.tfvars` file
* Run `terraform state list`, which should show the following output:

```
data.ibm_is_image.k8s_node
data.ibm_is_security_group.primary
data.ibm_is_ssh_key.ssh_key
data.ibm_is_subnet.primary
data.ibm_is_vpc.vpc
ibm_is_floating_ip.controlplane
ibm_is_floating_ip.worker
ibm_is_instance.controlplane
ibm_is_instance.worker
ibm_is_instance_template.k8s_node
local_file.inventory
null_resource.ansible
```

Then out can run `terraform apply` again. When I did that I got the output:

```
$ terraform apply -auto-approve
ibm_is_instance_template.k8s_node: Refreshing state... [id=02q7-c5ff2ead-15c1-4ace-818b-51b9776b4595]
ibm_is_instance.worker: Refreshing state... [id=02q7_0ccb2896-fb38-4c48-a4ec-0240c72ee61b]
ibm_is_instance.controlplane: Refreshing state... [id=02q7_f8746938-7e4e-43f7-a4b8-0723d7b5b8c6]
ibm_is_floating_ip.worker: Refreshing state... [id=r038-2c60be66-ee57-4fad-8eb9-3cfb353d8ec9]
ibm_is_floating_ip.controlplane: Refreshing state... [id=r038-e56b765c-b6b6-4a4c-875a-4f1cf11316fb]
local_file.inventory: Refreshing state... [id=8c4ea2f90ec12f5eff3012ed72ce72f506ca61c6]
null_resource.ansible: Refreshing state... [id=1766717801830837328]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

ssh_key_name = "toronto-key"
```

ibmcloud/terraform/cluster/main.tf
```diff
@@ -14,10 +14,23 @@ locals {
...
+  public_key = var.ssh_pub_key
+}
+
 data "ibm_is_ssh_key" "ssh_key" {
+  # Wait if the key needs creating first
+  depends_on = [ibm_is_ssh_key.created_ssh_key]
   name = var.ssh_key_name
 }
 
+output "ssh_key_name" {
```
mattarnoatibm, line 30: This is an optional change, but it's recommended to put Terraform `output`s into a separate `outputs.tf` file, [per the source here](https://learn.hashicorp.com/tutorials/terraform/outputs#output-vpc-and-load-balancer-information).

mattarnoatibm (changes requested):
Hi James

I think a workaround when the user supplies the `ssh_pub_key` variable and wants to run `terraform apply` again having ran it once is needed.

Otherwise, I've tested your changes with and without using `ssh_pub_key` and they work, and look good. Thanks.

ibmcloud/README.md
jtumber-ibm: Hi Matt, 

I've just tried running `terraform apply` multiple times with the ssh-pub-key set and it runs successfully without any workaround. Potentially something we can discuss separately.

stevenhorsman: After Matt told me how he hit this I also managed to trigger this. My total case flow was:
- Don’t provide `ssh_pub_key` and use an existing key -> `terraform apply` which worked then `terraform destroy` to clean it up
- Provide  `ssh_pub_key` , but use the name of an existing key -> `terraform apply` which failed (correctly) with `[DEBUG] Create SSH Key Key with name already exists`
- Delete the existing key manually in IBM and re-run it and` terraform apply` which worked
- Re-run `terraform apply` without the destroy first, and then got:
```
ibm_is_ssh_key.created_ssh_key[0]: Destroying... [id=r022-727dba5f-bc61-42a9-9664-bb637444fdc3]
ibm_is_ssh_key.created_ssh_key[0]: Still destroying... [id=r022-727dba5f-bc61-42a9-9664-bb637444fdc3, 10s elapsed]
╷
│ Error: Error deleting SSH Key : SSH-key still in use by 2 instance(s). First instance ID: 02f7_0b218b6c-660e-484b-ba1d-785fd7fe1782
```
in case that helps?

ibmcloud/README.md
```diff
@@ -444,4 +454,25 @@ groups, subnet and gateway with:
...
+> Error: Error deleting SSH Key : SSH-key still in use
+
+Which happens when you have provided your SSH public key via the `ssh_pub_key` Terraform variable and successfully running `terraform apply` but running `terraform apply` subsequently results in the above error.
+
+This is caused by the IBM Cloud terraform provider forcing the replacement of the SSH key resource. As such Terraform may try to delete the existing SSH key while the Virtual Server Instances are using it.
+
+To resolve this issue:
+
+- Use `terraform state rm ibm_is_ssh_key.created_ssh_key[0]` to remove the SSH key resource from Terraform's internal state.
+- Delete the ssh_pub_key variable from the terraform.tfvars file
```
mattarnoatibm, line 473: To be consistent with elsewhere we would want to put ssh_pub_key and terraform.tfvars in backticks to format them as code in Markdown.

ibmcloud/README.md
```diff
@@ -444,4 +454,25 @@ groups, subnet and gateway with:
...
 $ terraform destroy
-```
\ No newline at end of file
+```
+
+## Troubleshooting
+
+### Cluster creation when switching use `ssh_pub_key`
+
+If during the process of setting up the cluster you may experience the following error regarding your SSH key
```
mattarnoatibm, line 463: I'd combine the first two paragraphs like this:

If you provide your SSH public key via the `ssh_pub_key` Terraform variable and, having already run `terraform apply`, you run `terraform apply` again then you may see the following error regarding your SSH key:

> `Error: Error deleting SSH Key : SSH-key still in use`

mattarnoatibm (changes requested):
I think the new troubleshooting section needs a couple of updates. Otherwise this PR looks good.

mattarnoatibm (approved):
Looks good to me.

PR #27, opened 2022-05-25
Ability to run peer pod enablement on an existing cluster

### LINKED ISSUE #21: doc | ibmcloud: Inconsistency in default values for `cos_service_instance_name `
From https://github.com/confidential-containers/cloud-api-adaptor/pull/18#discussion_r875737240:
> In the next section the doc says:
> > cos_service_instance_name is the COS Service Instance Name, optional, default to value based on your cluster_name > tfvar if not set, ${cluster_name}-cos-service-instance.
>
> I think we should make the defaults of these line up, otherwise it will get confusing.

We now have inconsistency in the default values of the `cos_service_instance_name` in the two sections it is used:
- In creating the COS instance we have:
  > cos_service_instance_name in variables.tf has the default value 
- In building the pod VM image and uploading it we have:
  > cos_service_instance_name is the COS Service Instance Name, optional, default to value based on your cluster_name tfvar if not set, ${cluster_name}-cos-service-instance.
  
This is not a good user experience, so we have a few choices:
- Either get rid of the defaults and stop them being optional altogether
- Change both defaults to `cos-image-instance`. I don't think there is a requirement for the cos service instance name to be unique across users, so this should work
- Change both defaults to `<cluster_name>-cos-image-instance`. Unless there is a way to read the `cluster_name` value from the cluster terraform.tfvars, I don't think this is a good solution, as there isn't another reason for cluster name to be required in the cos instance creation, so it's letting a user not supply one value by making them supply another
- Any other solution someone smarter than me can think of?

### LINKED ISSUE #22: As a peer pods tester I want the ability to run peer pod enablement on an existing cluster
## Description
We want to split the Ansible playbook in `terraform/cluster/ansible` into two new Ansible playbooks so we have the benefit running one for the initial cluster setup, and another for Peer Pod enablement. 

## Use case
_As a peer pods tester I want the ability to run peer pod enablement on an existing cluster. This will be the setup for an operator that runs on a Kube cluster._

## Task List
- [x] In the `terraform/cluster/ansible` playbook, split the initial cluster setup code from the Peer Pod enablement code into two playbooks.
- [x] Determine how the two new playbooks are going to run, as to not disrupt the current demo instructions.
- [x] Change the Terraform template to run the two playbooks in the correct order.
- [x] Determine how the README in `terraform/`should be updated and update the README.
- [x] Raise a PR with the changes. The PR must be approved and merged before this story is closed.

## Performance Considerations
Increasing the number of manual steps goes against the idea of automation, when we are splitting a playbook into two to have control over which functionality we want, ensure that the steps to run the demo aren't made more complicated.

## Security Considerations 
API keys must not be shared or pushed to this repository.

### Questions
1. Are the two Ansible playbooks generated going to run back to back?

2. How would this issue be documented in the `/terraform/README`. We need to decide if the instructions to run the README to setup the demo will evolve into something new as the functionality of the Terraform scripts evolve.
 
3. How do we capture the benefits of one without increasing the number of steps for the demo? 

## Acceptance Criteria
**Given** I want to run Peer Pod enablement on an existing cluster
**When** I split the existing `/cluster/ansible/`  playbook up into separate functionality
**Then** the existing Terraform template can run the new Peer Pod enablement Ansible playbook against an existing cluster without running the initial cluster setup steps

**Scenario: Regression testing**
**Given** I want to create a new Kubernetes cluster and enable it with the peer-pod capability
**When** I follow the `Create a Kubernetes cluster` and peer pod enablement documentation 
**Then** I have a new, working Kubernetes cluster and it has the kata-shim, cloud-api-adaptor and all the required dependencies installed on it

**Given** I want to be able to create an x86 and s390x cluster
**When** the existing Ansible playbook is split into two and I run them
**Then** I maintain the option to choose an x86 or s390x cluster with peer-pod enablement for creation

Jonah-F:
automation: split cluster playbook into two
    
Completed work to split cluster/ansible playbook into
initial cluster setup and peer pod enablement.
I split dependencies up after flagging them as kube or
kata related, or both.
Runbooks execute consecutively so no changes to README.
    
Fixes #22
    
Signed-off-by: [Jonah-Farrow] <jonah.farrow@ibm.com>

### DIFF
--- a/ibmcloud/terraform/cluster/ansible/kata-playbook.yml
+++ b/ibmcloud/terraform/cluster/ansible/kata-playbook.yml
@@ -1,0 +1,280 @@
+#
+# (C) Copyright IBM Corp. 2022.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+- hosts: all
+  remote_user: root
+  tasks:
+    - name: Install deb packages
+      apt:
+        name:
+          - "linux-modules-extra-{{ ansible_kernel }}"
+          - build-essential
+          - jq
+          - qemu-utils
+          - libgpgme-dev
+          - libassuan-dev
+          - libdevmapper-dev
+          - libseccomp-dev
+          - pkg-config
+          - git
+          - kpartx
+          - protobuf-compiler
+          - musl-tools
+
+    - name: Install deb packages for Ubuntu 20.04 or later
+      apt:
+        name:
+          - libbtrfs-dev
+      when:
+        - ansible_facts['distribution'] == "Ubuntu"
+        - ansible_facts['distribution_major_version'] | int >= 20
+
+    - name: Install deb packages for Ubuntu 18.04 or before
+      apt:
+        name:
+          - btrfs-tools
+      when:
+        - ansible_facts['distribution'] == "Ubuntu"
+        - ansible_facts['distribution_major_version'] | int < 20
+
+    - name: Configure kernel modules to load at boot
+      copy:
+        dest: "{{ item.path }}"
+        content: "{{ item.content }}"
+      with_items:
+        - path: /etc/modules-load.d/peerpod.conf
+          content: vrf
+
+    - name: Load kernel modules
+      modprobe:
+        name: "{{ item }}"
+      with_items:
+        - vrf
+
+    - name: Install Go
+      shell: |
+        set -o errexit -o pipefail
+        arch="{{ ansible_architecture }}"
+        # go1.18 cannot compile containerd v1.6.1. Use go1.17.8 until the issue is fixed.
+        # https://github.com/containerd/containerd/issues/6586
+        # gover=$(curl -sL 'https://golang.org/VERSION?m=text')
+        gover=go1.17.8
+        curl -sL "https://go.dev/dl/$gover.linux-${arch/x86_64/amd64}.tar.gz" | tar -xzf - -C /usr/local
+
+        if ! grep -q '^PATH=/usr/local/go/bin:\$PATH$' /root/.bashrc; then
+          echo 'PATH=/usr/local/go/bin:$PATH' >> /root/.bashrc
+        fi
+      args:
+        executable: /bin/bash
+        creates: /usr/local/go/bin/go
+
+    - name: Install containerd
+      shell: |
+        set -o errexit -o pipefail
+        containerd_release_tag=$(curl -sL https://api.github.com/repos/containerd/containerd/releases/latest | jq -r .tag_name)
+        rm -fr /tmp/containerd
+        git clone -b CC-main https://github.com/confidential-containers/containerd.git /tmp/containerd
+        (cd /tmp/containerd && make && make install)
+        rm -fr /tmp/containerd
+      environment:
+        PATH: /usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+      args:
+        executable: /bin/bash
+        creates: /usr/local/bin/containerd
+    
+    - name: Ensure /etc/containerd directory exists
+      file:
+        path: /etc/containerd
+        state: directory
+
+    - name: Copy containerd config file
+      copy:
+        dest: /etc/containerd/config.toml
+        content: |
+          version = 2
+          root = "/var/lib/containerd" 
+          state = "/run/containerd"
+          oom_score = -999
+
+          [grpc]
+            address = "/run/containerd/containerd.sock"
+            uid = 0
+            gid = 0
+
+          [debug]
+            address = "/run/containerd/debug.sock"
+            uid = 0
+            gid = 0
+            level = "debug"
+
+          [plugins]
+            [plugins."io.containerd.runtime.v1.linux"]
+              shim_debug = true
+            [plugins."io.containerd.grpc.v1.cri"]
+              [plugins."io.containerd.grpc.v1.cri".containerd]
+                default_runtime_name = "runc"
+                [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
+                  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+                    runtime_type = "io.containerd.runc.v2"
+                  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata]
+                    runtime_type = "io.containerd.kata.v2"
+                    cri_handler = "cc"
+
+    - name: Copy systemd unit file for containerd
+      copy:
+        dest: /etc/systemd/system/containerd.service
+        content: |
+          [Unit]
+          Description=containerd container runtime
+          Documentation=https://containerd.io
+          After=network.target
+
+          [Service]
+          ExecStartPre=-/sbin/modprobe overlay
+          ExecStart=/usr/local/bin/containerd --config /etc/containerd/config.toml --log-level debug
+          Delegate=yes
+          KillMode=process
+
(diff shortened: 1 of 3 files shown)


ibmcloud/terraform/cluster/ansible/kube-playbook.yml
```diff
@@ -0,0 +1,270 @@
...
+
+          [Install]
+          WantedBy=multi-user.target
+
+    - name: Start fix-resolvconf service
+      systemd:
+        name: fix-resolvconf.service
+        state: started
+
+    - name: Install IBM Cloud CLI
```
mattarnoatibm, line 241: We should test if the playbooks work when the Install IBM Cloud CLI and Install IBM Cloud CLI plugins tasks are only in the kata playbook.

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
```diff
@@ -0,0 +1,270 @@
...
+
+          [Install]
+          WantedBy=multi-user.target
+
+    - name: Start containerd service
+      systemd:
+        name: containerd.service
+        state: started
+    
+    - name: Copy systemd unit file for fix-resolvconf
```
mattarnoatibm, line 219: We should test if the playbooks work when the Copy systemd unit file for fix-resolvconf and Start fix-resolveconf service tasks are only in the kata playbook

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
```diff
@@ -0,0 +1,270 @@
...
+        - path: /etc/modules-load.d/k8s.conf
+          content: br_netfilter
+        - path: /etc/modules-load.d/peerpod.conf
+          content: vrf
+
+    - name: Load kernel modules
+      modprobe:
+        name: "{{ item }}"
+      with_items:
+        - br_netfilter
```
mattarnoatibm, line 100: We should test the playbooks work if the Configure kernel modules to load at boot and load kernel modules task in the kubernetes playbook only sets up the br_netfilter kernel module.

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
```diff
@@ -0,0 +1,270 @@
...
+        - vrf
```
mattarnoatibm, line 101: We should test the playbooks work if the Configure kernel modules to load at boot and load kernel modules task in the kata playbook only sets up the vrf kernel module.

mattarnoatibm (changes requested):
For some of the duplicates, where the tasks are in both the Kubernetes and Kata playbooks we should test if the duplicates are needing in both playbooks. We established the containerd tasks do need to be duplicated. For the other duplicated tasks, which I've marked in comments, we should test if the duplication is needed or if the task can be in one playbook or the other only.

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
Jonah-F: Tested and successfully reran Terraform template.
Remove vrf from Kube-playbook, and br_netfilter from Kata-playbook under Load kernel modules . Also, remove vrf configuration from Kube-playbook, and br_netfilter configuration from Kata-playbook under Configure kernel modules to load at boot.

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
Jonah-F: Tested and successfully reran Terraform template.
Remove Copy systemd unit file for fix-resolvconf and Start fix-resolvconf service  (which uses systemd) from Kube-playbook

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
Jonah-F: Tested and successfully reran Terraform template.
Remove IBMCloud CLI and plugin install from Kube-playbook

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
Jonah-F: Removed vrf configuration from Kube-playbook, and br_netfilter configuration from Kata-playbook under Configure kernel modules to load at boot.

ibmcloud/terraform/cluster/ansible/kata-playbook.yml
```diff
@@ -27,10 +18,6 @@
           - libdevmapper-dev
           - libseccomp-dev
           - pkg-config
-          - kubelet
-          - kubeadm
-          - kubectl
-          - kubernetes-cni
           - cri-tools
```
mattarnoatibm, line 21: The cri-tools package is only in the https://apt.kubernetes.io kubernetes-xenial/main apt repository, so if the kata enablement playbook needs to install the cri tools it needs to keep the 'Add Kubernetes apt repository signing key' and 'Add Kubernetes apt repository' tasks.

If cri-tools is used to debug the kubelet, does installing the cri-tools package only belong in he Kubernetes enablement playbook?

mattarnoatibm (approved):
I've tested the two Ansible playbooks with the duplication removed. I've made one comment that if we want to keep installation of the cri-tools package in the kata playbook we need to keep adding the kubernetes apt repository in that playbook. The other comments I made about removing duplication from the playbook have been addressed.

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
mattarnoatibm: I've tested the new playbooks and they work with installation of the IBM Cloud CLI only being in the kata playbook

ibmcloud/terraform/cluster/ansible/kube-playbook.yml
mattarnoatibm: I've tested the new playbooks and they work with enabling the br_netfilter kernel module only being in the kube playbook.

mattarnoatibm (changes requested):
Sorry, meant to request a change.

ibmcloud/terraform/cluster/scripts/keygen.sh
```diff
@@ -68,7 +68,7 @@ for host in "${hosts[@]}"; do
         trap stop 0
 
         ssh "${opts[@]}" -o ExitOnForwardFailure=yes -S "$ssh_ctl_sock" -L "$port:$host:22" -M -f -N -l root "$bastion"
-        ssh-copy-id "${opts[@]}" -i "$key.pub" -p "$port" root@localhost
+        ssh-copy-id -f "${opts[@]}" -i "$key.pub" -p "$port" root@localhost
```
mattarnoatibm, line 71: Changing the keygen.sh script is outside of the scope of the issue this PR is fixing, so we should be making this change under a separate PR and issue

Jonah-F: Thanks. I have reverted this change.

ibmcloud/terraform/cluster/ansible/kata-playbook.yml
Jonah-F: Tested and successfully reran Terraform template.
Removed the `cri-tools` dependency from the kata-playbook.
Test 1 result:
   Apply complete!

mattarnoatibm (approved):
Looks good to me.

stevenhorsman (approved):
I've run through creating a cluster using this code based on `x86` and it seemed to work without issues. Thanks Jonah!

PR #33, opened 2022-05-26
aws: fix checking of pointer in service.go#getIPs()

wainersm:
The `addr` was already checked for `nil`, it is the result of
`net.ParseIP()` (i.e. `ip`) that should be checked too.


bpradipt (approved):
Good catch @wainersm. Thanks

PR #36, opened 2022-05-29
Fix incorrect error checking for empty namespace in annotation

bpradipt:


liudalibj (approved):
LGTM

PR #37, opened 2022-05-30
doc: update manual ansible playbook step

### LINKED ISSUE #21: doc | ibmcloud: Inconsistency in default values for `cos_service_instance_name `

Jonah-F:
Updated README to reflect splitting of ansible playbook.
Step now runs two new playbooks consecutively.

Fixes #21

Signed-off-by: [Jonah-Farrow] <jonah.farrow@ibm.com>

### DIFF
--- a/ibmcloud/README.md
+++ b/ibmcloud/README.md
@@ -131,12 +131,12 @@ Outputs:
 ssh_key_name = <SSH key name>
 ```
 
-This Terraform configuration also triggers execution of an Ansible playbook to set up Kubernetes and other prerequisite software in the two nodes. Please check [ansible/playbook.yml](terraform/cluster/ansible/playbook.yml) for the details.
+This Terraform configuration also triggers execution of two Ansible playbooks to set up Kubernetes and other prerequisite software in the two nodes. Please check [ansible/kube-playbook.yml](terraform/cluster/ansible/kube-playbook.yml) and [ansible/kata-playbook.yml](terraform/cluster/ansible/kata-playbook.yml) for the details.
 
-If ansible fails for some reason, you can rerun the ansible playbook as follows.
+If ansible fails for some reason, you can rerun the Ansible playbooks as follows.
 ```bash
 $ cd ansible
-$ ansible-playbook -i ./inventory -u root ./playbook.yml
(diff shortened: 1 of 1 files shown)


ibmcloud/README.md
```diff
@@ -136,7 +136,7 @@ This Terraform configuration also triggers execution of an Ansible playbook to s
 If ansible fails for some reason, you can rerun the ansible playbook as follows.
```
jtumber-ibm, line 136: Small note but should this been updated to `ansible playbooks`?

jtumber-ibm (changes requested):
Line 134, also has a link to the playbook which is no longer there.

ibmcloud/README.md
Jonah-F: Thanks, made both changes.

jtumber-ibm: LGTM

jtumber-ibm (approved):

Jonah-F:
**Commit message references wrong issue for fix**.
The commit `doc: update manual ansible playbook step` should reference issue 22, not issue 21, as to what the change fixes.

PR #43, opened 2022-05-31
Fix linter errors in aws implementation

bpradipt:


wainersm (approved):
Great! I double-checked locally that it has no more lint errors. Thanks @bpradipt !

Issue #52, opened 2022-06-07
podnetwork: TestRouting and TestKeepAlive fails on Github's runner


wainersm:
The podnetwork's TestRouting and TestKeepAlive tests (in `pkg/podnetwork/tunneler/routing`) fail int the Ubuntu runner of Github (even though running as root).

The error is like:
```
=== RUN   TestKeepAlive
    keepalive_test.go:32: failed to add vrf ppvrf1: failed to create an interface of vrf: ppvrf1:  operation not supported
--- FAIL: TestKeepAlive (0.01s)
```

ariel-adam:
@wainersm is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

Issue #48, opened 2022-06-03
ibmcloud: unit tests don't compile


wainersm:
While working on https://github.com/confidential-containers/cloud-api-adaptor/issues/47 I realized the tests `pkg/adaptor/hypervisor/ibmcloud/server_test.go` and `pkg/adaptor/hypervisor/ibmcloud/shim_test.go` don't compile.

The build errors:
```
[wmoschet@wainer-laptop ibmcloud]$ go test
# github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/ibmcloud [github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/ibmcloud.test]
./server_test.go:89:58: undefined: Server
./server_test.go:138:64: cannot use newAgentService() (type *agentService) as type grpc.AgentServiceService in argument to grpc.RegisterAgentServiceService:
	*agentService does not implement grpc.AgentServiceService (missing CreateSandbox method)
./server_test.go:159:58: undefined: Server
./server_test.go:163:19: too many arguments in call to NewServer
./server_test.go:163:47: undefined: ServiceConfig
./server_test.go:193:41: undefined: Server
./shim_test.go:444:63: undefined: grpc.StartTracingRequest
./shim_test.go:448:62: undefined: grpc.StopTracingRequest
./shim_test.go:456:59: undefined: grpc.CreateVMRequest
./shim_test.go:460:60: undefined: grpc.DestroyVMRequest
./server_test.go:163:19: too many errors
FAIL	github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/ibmcloud [build failed]
```

I did a couple of fixes as shown in the diff below:
```
diff --git a/pkg/adaptor/hypervisor/ibmcloud/server_test.go b/pkg/adaptor/hypervisor/ibmcloud/server_test.go
index c1ada6e..f98fd80 100644
--- a/pkg/adaptor/hypervisor/ibmcloud/server_test.go
+++ b/pkg/adaptor/hypervisor/ibmcloud/server_test.go
@@ -1,5 +1,6 @@
 // (C) Copyright IBM Corp. 2022.
 // SPDX-License-Identifier: Apache-2.0
+// +build ibmcloud
 
 package ibmcloud
 
@@ -21,6 +22,7 @@ import (
        "github.com/IBM/vpc-go-sdk/vpcv1"
 
[...]

PR #53, opened 2022-06-07
github: run unit tests workflow

### LINKED ISSUE #47: github: add workflow to run the unit tests
There are already a couple of unit tests implemented in the project but they aren't executed on the current [build workflow](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/.github/workflows/build.yaml).

Let's extend the build workflow to all run the unit tests afterwards.

wainersm:
This adds a step on the build workflow in order to run the existing unit tests.

The tests for ibmcloud aren't compiling. I opened the issue #48 and disabled them all from CI.

The tests that need root privileges are refactored so that they use a common utility function for skipping. Even though running the CI pipeline as `sudo` some tests are failing to create network reosurces; those were also disabled from CI (see the issue #52).

I was working to get the test results nicely displayed on Github UI. I investigated the use of either https://github.com/marketplace/actions/junit-report-action or https://github.com/dorny/test-reporter but they both have limitations that didn't make me comfortable to enable them on the project. So for now they aren't used but I decided to keep the code that convert the raw output of the tests into JUnit XML with the hope that one day I could go back to this reporting problem.

All in all, at this point the CI pipelines should be all green but some unit tests are disabled.

Fixes #47

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,7 +1,7 @@
 # (C) Copyright Red Hat 2022.
 # SPDX-License-Identifier: Apache-2.0
 #
-# Build the project for each cloud provider.
+# Build and run unit tests the project for each cloud provider.
 ---
 name: build
 on: [pull_request]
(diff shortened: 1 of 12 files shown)


wainersm:
Updated this PR to fix a commit that was missing the SOB. Now DCO bot is happy.

wainersm:
Cc @bpradipt @stevenhorsman

.github/workflows/build.yaml
```diff
@@ -19,6 +17,10 @@ jobs:
           - aws
           - ibmcloud
           - libvirt
+        runner:
+          - ubuntu-latest
```
stevenhorsman, line 21: Just a thought, but I wonder if hooking this into the `runs-on` field e.g. `runs-on: ${{ matrix.runner }}`, might be more intuitive if we decided to add support for more OSs in future, and would remove a duplication of `ubuntu-latest`

stevenhorsman (approved):
@wainersm - I made an optional, minor comment, but this looks good. Thanks for raising #48 - I'll add it to the TODO list!

.github/workflows/build.yaml
wainersm: It does make sense. I will change that.

bpradipt (approved):
/lgtm
Thanks @wainersm

Issue #60, opened 2022-06-27
ibmcloud: ssh-key comment forces update when there is no change


jtumber-ibm:
As described in the [troubleshooting documentation](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/ibmcloud/TROUBLESHOOTING.md) there are occasions when the IBM Cloud terraform plugin will try to update (destroy and create) the ssh-key. Even though there have been no changes to the public key variable.

This is caused by the comment of the ssh-key being stripped by IBM Cloud, and terraform doing a full string comparision between the pub-key variable and the stripped key stored in IBM Cloud.

Steps to reproduce

1. Use the `cluster` module or create a dummy terraform module that only creates an ssh-key
2. Add a public ssh key to the `ssh_pub_key` variable with the comment included e.g. `ssh_pub_key = "ssh-rsa AAA...Z test@test.com`
3. Run `terraform apply` and accept
4. Run `terraform plan` and see that it will try to replace the ssh-key (destroy and recreate). 
5. Remove the comment e.g. `test@test.com` from the public key 
6. Run `terraform plan` and see that there is no change required 

> The comparison includes whitespace so ensure that the space is removed from the end too

Example Output (With Comment)

```bash

  # ibm_is_ssh_key.ssh_key must be replaced
-/+ resource "ibm_is_ssh_key" "ssh_key" {
     ....
      ~ public_key              = "ssh-rsa AAAA...=" -> "ssh-rsa AAAA= test@test.com" # forces replacement
```

Example Output (Without Comment)
```bash
root@chute1:~/workspace/ssh-key-investigation# terraform apply
ibm_is_ssh_key.ssh_key: Refreshing state... 

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

PR #61, opened 2022-06-27
ibmcloud: Strip ssh-key comment

tumberino:
Removes the comment at the end of ssh public keys which causes unnecessary replacement of the ssh-key.

Fixes: #60

Signed-off-by: jtumber-ibm <james.tumber@ibm.com>


jtumber-ibm:
This bug has already been fixed in later versions of the provider

Current version `1.34.0`
Fixed version `1.41.0-beta0`

yoheiueda:
@jtumber-ibm 
The latest version of ibmcloud terraform provider is v1.43.0 as of now.
https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.43.0

So, it is better to bump the version to v1.43.0 instead of using this proposed code change?

yoheiueda:
I noticed that 1.34.0 is too old to support the macOS Arm architecture.

Issue #5, opened 2022-05-05
Deploying, configuring and running cloud-api-adaptor on the worker node


bpradipt:
Currently the cloud-api-adaptor is manually executed by passing the relevant options. Going forward we should figure out a way to deploy it via the operator.

PR #62, opened 2022-06-29
deploy: run cloud-api-adaptor from within a pod

snir911:
Supports only libvirt and aws cloud providers
                                                                          
## Build and install with cloud-api-adaptor running in a pod

* set CLOUD_PROVIDER
    ```
    export CLOUD_PROVIDER=<aws|ibmcloud|libvirt>
    ```

* `make image` builds the container image and push it to `$registry`
    * configure install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml with your own settings
* `make delete` deletes the pod deployment from the configured cluster
                                                                         

Fixes: #5

**This was tested with aws (single noded) only**
As image was not pushed yet to repo to quickly deploy and test on an aws cluster:
1. Set your credentials on install/overlays/aws/kustomization.yaml (and newImage to: `quay.io/snir/cloud-api-adaptor-aws`)
2. `CLOUD_PROVIDER=aws make deploy`

NOTE:
In this implementation the caa pod is not deployed by the operator itself, we may prefer some other implementation or to use the pre-installtion script to deploy the caa pod, however, it would complicate a bit the development and CI (because of the multiple images and deployment nesting), hence, i'd suggest to use this implementation for now and work on more advanced implementation in different PR when required.

Depends-on: #64
Depends-on: #66 
Depends-on: https://github.com/confidential-containers/operator/pull/59
(operator image for testing:  quay.io/jensfr/confidential-containers-operator:fix-issue-58)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM golang:1.18 AS builder
+ARG CLOUD_PROVIDER
+ENV CLOUD_PROVIDER=${CLOUD_PROVIDER}
+COPY . cloud-api-adaptor
+RUN git clone -b CCv0-peerpod https://github.com/yoheiueda/kata-containers
+WORKDIR cloud-api-adaptor
+RUN make clean
+RUN if [ "$CLOUD_PROVIDER" = "libvirt" ] ; then apt-get update -y && apt-get install -y libvirt-dev libvirt0 genisoimage && apt-get clean; fi
+RUN make
+RUN install cloud-api-adaptor /usr/local/bin/cloud-api-adaptor-$CLOUD_PROVIDER
+RUN install entrypoint.sh /usr/local/bin/
+RUN make clean
+CMD entrypoint.sh
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-.PHONY: all build check fmt vet clean
+.PHONY: all build check fmt vet clean image deploy delete
 ifndef CLOUD_PROVIDER
 $(error CLOUD_PROVIDER is not set)
 endif
@@ -23,6 +23,18 @@ ifeq ($(CLOUD_PROVIDER),libvirt)
 else
 	CGO_ENABLED=0 go build $(GOFLAGS) -o "$@" "cmd/$@/main.go"
 endif
+
+# Build and push docker image to $regestry
+image:
+	hack/build.sh
+
+# Deploy cloud-api-adaptor using the operator, according to install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml
+deploy:
+	kubectl apply -f install/yamls/deploy.yaml
+	kubectl apply -k install/overlays/$(CLOUD_PROVIDER)
+
+delete:
+	kubectl delete -k install/overlays/$(CLOUD_PROVIDER)
 
 test:
 	# Note: sending stderr to stdout so that tools like go-junit-report can
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+CLOUD_PROVIDER=${1:-$CLOUD_PROVIDER}
+CRI_RUNTIME_ENDPOINT=${CRI_RUNTIME_ENDPOINT:-/run/peerpod/cri-runtime.sock}
+optionals+=""
+
+if [[ -S ${CRI_RUNTIME_ENDPOINT} ]]; then # will skip if socket isn't exist in the container
+	optionals+="-cri-runtime-endpoint ${CRI_RUNTIME_ENDPOINT} "
+fi
+
+test_vars() {
+        for i in $@; do
+                [ -z ${!i} ] && echo "\$$i is NOT set" && EXT=1
+        done
+        [[ -n $EXT ]] && exit 1
+}
+
+aws() {
+#test_vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION
+set -x
+cloud-api-adaptor-aws aws \
+	-aws-access-key-id ${AWS_ACCESS_KEY_ID} \
+	-aws-secret-key ${AWS_SECRET_ACCESS_KEY} \
+	-aws-region ${AWS_REGION} \
+	-pods-dir /run/peerpod/pods \
+	-aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME:-kata} \
+	${optionals} \
+	-socket /run/peerpod/hypervisor.sock
+}
+
+libvirt() {
+test_vars LIBVIRT_URI
+set -x
+cloud-api-adaptor-libvirt libvirt \
+	-uri ${LIBVIRT_URI} \
+	-data-dir /opt/data-dir \
+	-pods-dir /run/peerpod/pods \
+	-network-name ${LIBVIRT_NET:-default} \
+	-pool-name ${LIBVIRT_POOL:-default} \
+	${optionals} \
+	-socket /run/peerpod/hypervisor.sock
+}
+
+help_msg() {
+	cat <<EOF
+Usage:
+	CLOUD_PROVIDER=aws|libvirt $0
+or
+	$0 aws|libvirt
+in addition all cloud provider specific env variables must be set and valid
+(CLOUD_PROVIDER is currently set to "$CLOUD_PROVIDER")
+EOF
+}
+
+if [[ "$CLOUD_PROVIDER" == "aws" ]]; then
+	aws
+elif [[ "$CLOUD_PROVIDER" == "libvirt" ]]; then
+	libvirt
+else
+	help_msg
+fi
--- a/install/overlays/aws/kustomization.yaml
+++ b/install/overlays/aws/kustomization.yaml
@@ -1,0 +1,26 @@
+bases:
+- ../../yamls
+nameSuffix: -aws
+
+images:
+- name: kustomize.this/cloud-api-adaptor/image:url
+  newName: quay.io/confidential-containers/cloud-api-adaptor-aws # change image if needed
+  newTag: latest
+
+configMapGenerator:
+- name: peer-pods-cm
+  namespace: confidential-containers-system
+  literals:
+  - CLOUD_PROVIDER="aws"
+  - PODVM_LAUNCHTEMPLATE_NAME="kata"
+
+secretGenerator:
+- name: peer-pods-secret
+  namespace: confidential-containers-system
+  literals:
+  - AWS_ACCESS_KEY_ID="" # set
+  - AWS_SECRET_ACCESS_KEY="" # set
+  - AWS_REGION="" #set
+
+patchesStrategicMerge:
+  #- cri_runtime_endpoint.yaml # set (modify host's runtime cri socket path in the file, default is /run/containerd/containerd.sock)
--- a/install/overlays/libvirt/kustomization.yaml
+++ b/install/overlays/libvirt/kustomization.yaml
@@ -1,0 +1,25 @@
+bases:
+- ../../yamls
+nameSuffix: -libvirt
+
+images:
+- name: kustomize.this/cloud-api-adaptor/image:url
+  newName: quay.io/confidential-containers/cloud-api-adaptor-libvirt # change image if needed
+  newTag: latest
+
+configMapGenerator:
+- name: peer-pods-cm
+  namespace: confidential-containers-system
+  literals:
+  - CLOUD_PROVIDER="libvirt"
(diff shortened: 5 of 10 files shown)


Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN git clone -b ccv0-peer-pods https://github.com/bpradipt/kata-containers
+RUN apt-get update && apt-get install -y libvirt-dev libvirt0 genisoimage # other requirements?
+WORKDIR cloud-api-adaptor
+# generic build
+ENV CLOUD_PROVIDER=aws
+RUN make
+RUN install cloud-api-adaptor /usr/local/bin/cloud-api-adaptor
+RUN make clean
+# libvirt build
+ENV CLOUD_PROVIDER=libvirt
```
bpradipt, line 13: How about adding a Makefile target for cloud-api-adaptor `docker-build` and `docker-push` with `CLOUD_PROVIDER` as a build time ARG. ? This will be more flexible imho

cloud-api-adaptor.yaml
```diff
@@ -0,0 +1,105 @@
...
+kind: ConfigMap
+metadata:
+  name: peer-pods-config
+  namespace: peer-pods
+data:
+  CLOUD_PROVIDER: "libvirt"
+  LIBVIRT_URI: "qemu+ssh://root@192.168.122.1/system"
+  LIBVIRT_NET: "default"
+  LIBVIRT_POOL: "default"
+
```
bpradipt, line 63: nice :-)

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN make
+RUN install cloud-api-adaptor /usr/local/bin/cloud-api-adaptor-libvirt
```
bpradipt, line 15: For libvirt we don't have a static build yet so we need the dependencies (libvirt-dev etc). However for the cloud providers we have a static build. So we can use a distroless base image (eg `gcr.io/distroless/static`) for the actual binary.

snir911: I see, i can do so and also have a makefile target for each provider as you suggested but it will require us to maintain 2 container images (and different yamls?) instead of a single container which contains both binaries.
Is it preferable? WDYT?

bpradipt: Let's avoid maintaining multiple container images for now and stick to only one.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.17.3
+
+ADD . cloud-api-adaptor
+RUN git clone -b CCv0-peerpod https://github.com/yoheiueda/kata-containers
+RUN apt-get update && apt-get install -y libvirt-dev libvirt0 genisoimage # other requirements?
+WORKDIR cloud-api-adaptor
+# generic build
+ENV CLOUD_PROVIDER=aws
+RUN make
+RUN install cloud-api-adaptor /usr/local/bin/cloud-api-adaptor
```
bpradipt, line 10: nit: /usr/local/bin/cloud-api-adaptor -> /usr/local/bin/cloud-api-adaptor-aws

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.17.3
```
bpradipt, line 1: Better to use the same golang version used in go.mod

Makefile
```diff
@@ -14,6 +14,14 @@ SOURCEDIRS := ./cmd ./pkg
 PACKAGES   := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
 SOURCES    := $(shell find $(SOURCEDIRS) -name '*.go' -print)
 
+IMAGE_BUILD_CMD ?= podman build
```
bpradipt, line 17: Can we use a similar approach like - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/install/runtime-payload/Makefile#L4-L5

This will be uniform and help us create multi-arch images going forward, which will be useful

install/yamls/ccruntime-peer-pods-in-pod.yaml
```diff
@@ -0,0 +1,105 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cloud-api-adaptor-deployment
+  namespace: peer-pods
```
bpradipt, line 5: Better to use the same namespace as the CRD - `confidential-containers-system`

install/yamls/ccruntime-peer-pods-in-pod.yaml
```diff
@@ -0,0 +1,105 @@
...
+          type: ""
+        name: systemd
+      - hostPath:
+          path: /usr/local/bin/
+          type: ""
+        name: local-bin
+    installCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "install"]
+    uninstallCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "cleanup"]
+    cleanupCmd: ["/opt/kata-artifacts/scripts/kata-deploy.sh", "reset"]
+    # can we preInstall the caa deployment from above here?
```
bpradipt, line 97: Do you want to include it in the payLoad image or preInstall image ?

Dockerfile
snir911: It was like that because i implemented also the ibmcloud part, once other providers are supported they all can use this binary (except libvirt), i can change until then though

Dockerfile
snir911: I think i used this one because it's mentioned in the libvirt instructions maybe we should bump?

install/yamls/ccruntime-peer-pods-in-pod.yaml
snir911: Yes i wondered about it, i'd also prefer it.. because  ATM it's not really manged by the operator, it's installed "on the side" with the runtime i didn't do so, If that's not a problem I'll change that

install/yamls/ccruntime-peer-pods-in-pod.yaml
snir911: Not needed ATM, I'll add a comment about it in the PR issue

install/yamls/ccruntime-peer-pods-in-pod.yaml
bpradipt: Let's change it to use `confidential-containers-system` to ensure all peer-pods resources are in the same namespace.

Dockerfile
bpradipt: We are bumping the version to 1.18. So you can change this to `1.18`

Dockerfile
bpradipt: cloud-api-adaptor binary is specific to the provider. So if we add additional provider binaries to the same image, then we need to change the binary name. If we create separate images for each provider then we can keep the binary name as-is.
I believe it's better to create single image per provider. 
What you can do is take the provider name as an argument 
Something like this

```
ARG CLOUD_PROVIDER
ENV CLOUD_PROVIDER=${CLOUD_PROVIDER}

RUN if [ "$CLOUD_PROVIDER" = "libvirt" ] ; then apt-get update -y && apt-get install -y libvirt-dev libvirt0 genisoimage; fi

...

Since the make is called with `CLOUD_PROVIDER=<value>` the same can be passed to run the docker build and you get the right image for the provider.

Comment (bpradipt) [reply Dockerfile:10]: Additionally the `CLOUD_PROVIDER` value can be suffixed to the image name. eg `quay.io/confidential-containers/cloud-api-adaptor-aws:<timestamp>`

File: entrypoint.sh [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+set -x
+cloud-api-adaptor-libvirt libvirt \
+	-uri ${LIBVIRT_URI} \
+	-data-dir /opt/data-dir \
+	-pods-dir /run/peerpod/pods \
+	-network-name ${LIBVIRT_NET:-default} \
+	-pool-name ${LIBVIRT_POOL:-default} \
+	${CRI_RUNTIME_ENDPOINT_OPTION} \
+	-socket /run/peerpod/hypervisor.sock
+}
```
bpradipt, line 40: @snir911 can you please add ibmcloud option as well
```
/opt/confidential-containers/bin/cloud-api-adaptor-ibmcloud ibmcloud \
    -api-key "${IBMCLOUD_API_KEY}" \
    -key-id "${SSH_KEY_ID}" \
    -image-id "${IMAGE_ID}" \
    -profile-name "${INSTANCE_PROFILE}" \
    -zone-name ${IBMCLOUD_ZONE} \
    -primary-subnet-id "${VPC_PRIMARY_SUBNET_ID}" \
    -secondary-subnet-id "${VPC_SECONDARY_SUBNET_ID}" \
    -primary-security-group-id "${VPC_PRIMARY_SECURITY_GROUP_ID}" \
    -secondary-security-group-id "${VPC_SECONDARY_SECURITY_GROUP_ID}" \
    -vpc-id "${VPC_ID}" \
    -pods-dir /run/peerpod/pods \
    -socket /run/peerpod/hypervisor.sock \
    -cri-runtime-endpoint "${CRI_RUNTIME_ENDPOINT}"
```

entrypoint.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+CLOUD_PROVIDER=${1:-$CLOUD_PROVIDER}
+CRI_RUNTIME_ENDPOINT=${CRI_RUNTIME_ENDPOINT:-/run/peerpod/cri-runtime.sock}
```
bpradipt, line 4: please change the default to `/run/containerd/containerd.sock`

install/overlays/libvirt/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+  literals:
+  - CLOUD_PROVIDER="libvirt"
+  - LIBVIRT_URI="qemu+ssh://root@192.168.122.1/system" #set
+  - LIBVIRT_NET="default" # set
+  - LIBVIRT_POOL="default" # set
+
+secretGenerator:
+- name: peer-pods-secret
+  namespace: confidential-containers-system
+  literals:
```
bpradipt, line 22: please also add overlay for ibmcloud

config literals
```
          CLOUD_PROVIDER="ibmcloud"         
	  IMAGE_ID=""
	  INSTANCE_PROFILE=""
	  IBMCLOUD_ZONE=""
	  VPC_ID=""
	  VPC_PRIMARY_SUBNET_ID=""
	  VPC_SECONDARY_SUBNET_ID=""
	  VPC_PRIMARY_SECURITY_GROUP_ID=""
	  VPC_SECONDARY_SECURITY_GROUP_ID=""
	  CRI_RUNTIME_ENDPOINT="/run/containerd/containerd.sock"
```
and secret
```
 IBMCLOUD_API_KEY=""
 SSH_KEY_ID=""
```

install/overlays/aws/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+images:
+- name: kustomize.this/cloud-api-adaptor/image:url
+  newName: quay.io/confidential-containers/cloud-api-adaptor-aws # change image if needed
+  newTag: latest
+
+configMapGenerator:
+- name: peer-pods-cm
+  namespace: confidential-containers-system
+  literals:
+  - CLOUD_PROVIDER="aws"
```
bpradipt, line 14: please add the following additional literal
```
PODVM_LAUNCHTEMPLATE_NAME="kata"
```

entrypoint.sh
```diff
@@ -0,0 +1,59 @@
...
+
+aws() {
+#test_vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION
+set -x
+cloud-api-adaptor-aws aws \
+	-aws-access-key-id ${AWS_ACCESS_KEY_ID} \
+	-aws-secret-key ${AWS_SECRET_ACCESS_KEY} \
+	-aws-region ${AWS_REGION} \
+	-pods-dir /run/peerpod/pods \
+	${CRI_RUNTIME_ENDPOINT_OPTION} \
```
bpradipt, line 25: please add the following additional option - 
```
-aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} 
```

bpradipt:
@snir911 thanks so much for working on this one. Will help with CI and also make the deployment easier. I added a few comments and once those are addressed it should be good to be merged.

entrypoint.sh
snir911: We used to have this one, however as IBM ppl don't have cycles to test ATM, i discussed with @huoqifeng to split the ibmcloud to a different PR which he'll test later, I'll do this later

entrypoint.sh
snir911: note that this is the socket path which is exposed in the containers, the host's socket path default is `/run/containerd/containerd.sock`

bpradipt: I see.. thanks for the clarification

bpradipt (approved):
/lgtm
Thanks @snir911

PR #68, opened 2022-06-30
pkg: Do not set MTU greater than 1450 for VXLAN

### LINKED ISSUE #67: MTU size of VXLAN interface should not be greater than 1450
MTU of a VXLAN interface in a pod VM becomes greater than 1450, which is the maximum allowed size for VxLAN packets, when a CNI plugin such as Calico IPIP uses a MTU greater than 1450. This may cause undesirable packet fragmentation. 

https://github.com/confidential-containers/cloud-api-adaptor/blob/142b61254e2c06e6ab52fcdac25a0d4e90f534c6/pkg/podnetwork/tunneler/vxlan/podnode.go#L66-L67

yoheiueda:
MTU of a VXLAN interface in a pod VM becomes greater than 1450,
which is the maximum allowed size for VxLAN packets, when a CNI
plugin such as Calico IPIP uses a MTU greater than 1450. This may
cause undesirable packet fragmentation. This patch fixes the issue.

Fixes #67

Signed-off-by: Yohei Ueda <yohei@jp.ibm.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.15.0
 	github.com/aws/aws-sdk-go-v2/credentials v1.10.0
 	github.com/aws/aws-sdk-go-v2/service/ec2 v1.31.0
-	github.com/containerd/containerd v1.6.2
+	github.com/containerd/containerd v1.6.6
 	github.com/containerd/ttrpc v1.1.0
 	github.com/containernetworking/plugins v1.1.1
 	github.com/coreos/go-iptables v0.6.0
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):
The code changes look good to me, and I've tested it out with the ibmcloud provider and was able to successfully create an nginx peer pod

bpradipt (approved):
/lgtm

huoqifeng:
@yoheiueda is it possible to accept the MTU as a parameter rather than a hard-coded const? i think Calico MTU is configurable.

yoheiueda:
@huoqifeng MTU is not hard-coded.

MTU of the interface created by a CNI plugin is detected at a worker node. Then, the detected MTU is set to the interface of the pod VM.

When a detected MTU is greater than 1450, which is the maximum allowed MTU for VXLAN, MTU of the interface of the pod VM is set to 1450.

https://github.com/confidential-containers/cloud-api-adaptor/blob/41b75cc1335e444a04b482d290120257e7ffccc4/pkg/podnetwork/workernode.go#L114-L118

https://github.com/confidential-containers/cloud-api-adaptor/blob/41b75cc1335e444a04b482d290120257e7ffccc4/pkg/podnetwork/tunneler/vxlan/podnode.go#L67-L71

PR #69, opened 2022-07-01
ibmcloud: Add resource group parameter

huoqifeng:
Fixes: non default resource group problem

image

Signed-off-by: Qi Feng Huo <huoqif@cn.ibm.com>

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,42 @@
+# (C) Copyright Red Hat 2022.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Run linting tools on the sources of the project.
+---
+name: lint
+on: [pull_request]
+jobs:
+  golangci-lint_job:
+    name: golangci-lint
+    runs-on: ubuntu-latest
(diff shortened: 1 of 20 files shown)


huoqifeng:
@liudalibj @yoheiueda would you kindly review the change, I just found it's possible that resource group will be used by mstake when creating the PeerPod when the worker belongs to non-default resource group. So added an additional and optional parameter: `resource-group-id`

liudalibj (changes requested):
The codes LGTM, there is a failed in pipeline, please fix the commit title.
It should something like:
`ibmcloud: Add resource group parameter for ibmcloud provider`
And maybe please combine the four commits to one only?

huoqifeng:
> The codes LGTM, there is a failed in pipeline, please fix the commit title. It should something like: `ibmcloud: Add resource group parameter for ibmcloud provider` And maybe please combine the four commits to one only?

Thanks @liudalibj is there a method to rerun the commit check after my change?

huoqifeng:
Close it and use a new PR to clean the mesh

PR #70, opened 2022-07-01
Add initial implementation of mutating admission controller webhook

bpradipt:
Fixes: #73 


webhook/pkg/mutating_webhook/remove-resourcespec.go
```diff
@@ -0,0 +1,59 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	DEFAULT_RUNTIME_CLASS_NAME  = "kata-remote-cc"
+	VM_ANNOTATION_INSTANCE_TYPE = "kata.peerpods.io/instance_type"
+	VM_INSTANCE_TYPE_DEFAULT    = "t2.small"
+	VM_EXTENDED_RESOURCE        = "kata.peerpods.io/vm"
```
stevenhorsman, line 14: Not a big problem for now, but I wonder if we should move away from kata to cc/confidential-containers?

stevenhorsman (approved):
Hey @bpradipt, I've run through the documentation and I think I have it working as expected. I've left some minor comments/suggestions.
In addition, I'm slightly concerned that the `fixes #7` comment in the commit will close the issue and I think there is probably more to do, so we'll have to remember to check & re-open it.
I'll also admit that I'm far from being a Kubernetes expert, so @yoheiueda, or his team might be able to provide more value in review.
Thanks!

webhook/README.md
```diff
@@ -0,0 +1,86 @@
...
+```
+2. Create the pod
+```sh
+kubectl apply -f hack/pod.yaml
+```
+3. View the mutated pod
+```sh
+kubectl get -f hack/pod.yaml -o yaml | grep kata.peerpods
+```
+
```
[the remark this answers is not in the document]
bpradipt: Thanks @stevenhorsman will update the doc

webhook/pkg/mutating_webhook/remove-resourcespec.go
bpradipt: Yeah, this is a good suggestion. I'll make the changes

webhook/README.md
```diff
@@ -0,0 +1,86 @@
...
+## Getting Started
+You’ll need a Kubernetes cluster to run against. You can use [KIND](https://sigs.k8s.io/kind) to get a local cluster for testing, or run against a remote cluster.
+
+You'll also need to advertise the extended resource `kata.peerpods.io/vm`.
+
+A simple daemonset is provided under the following [directory](./hack/extended-resources/).
+
+```
+./hack/extended-resources/setup.sh
```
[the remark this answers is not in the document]
bpradipt: @stevenhorsman good catch on the #7. I think its better to keep #7 open. I'll create a new issue specifically mentioning mutation admission controller implementation and link this PR.

webhook/pkg/mutating_webhook/remove-resourcespec.go
bpradipt: @stevenhorsman re-visiting the extended resource naming aspect, I'm thinking of leaving it as-is for now and get some more ideas before finalising the name :-)

yoheiueda (approved):
LGTM

PR #91, opened 2022-07-13
Stop using HTTP Upgrade for communication between CAA and APF

### LINKED ISSUE #90: Do not use HTTP Upgrade for communicaiton between CAA and APF
This is a prerequisite to fix https://github.com/confidential-containers/cloud-api-adaptor/issues/88

Due to a historical reason,  communication between cloud-api-adaptor and agent-protocol-forwarder is established using HTTP. Agent-protocol-forwarder launches an HTTP server, and cloud-api-adaptor creates a raw TCP connection using HTTP Upgrade technique. This mechanism allows us to support multiple protocols other than TTRPC between the two components while using a single server port.

The current implementation only uses TTRPC between the two components, so we do not have to use the HTTP Upgrade technique, but we can simply use TTRPC over TCP connections.

yoheiueda:
This PR changes the code to use usual RAW TCP connections for TTRPC between cloud-api-adaptor and agent protocol forwarder.

The HTTP Upgrade technique for communication between two components are no longer used, so related util packages are also removed.

Fixes #90


bpradipt (approved):
@yoheiueda, Nice work w.r.to removing the http upgrade mechanism. 
Thanks..

Issue #94, opened 2022-07-14
Run kata-agent process in root namespace on Peer Pod VM


genjuro214:
We are trying to run a csi driver container on Peer Pod VM, It requires to mount a volume with mount progatation as true:
```
volumeMounts:
        - mountPath: /data
          mountPropagation: Bidirectional
          name: kubelet-data-dir
...
volumes:
    - hostPath:
        path: /root/data
        type: Directory
      name: kubelet-data-dir
```
About [Mount propagation](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation), if it's set as **Bidirectional**, "all volume mounts created by the container will be propagated back to the host and to all containers of all pods that use the same volume."

However, Bidirectional doesn't work as expected. The CSI driver container mounts a device under the volume mount path /data, the mount info can't be found on host.

If creating it as a normal pod, Bidirectional works as expected, the device mount info is also visible on host.

----

According to our investigation, the root cause is currently **kata-agent** process on Peer Pod VM is running in new network namespace and mount namespace, because it's started with `ip netns exec`:
```
[Service]
ExecStart=ip netns exec podns /usr/local/bin/cgroup-exec /usr/local/bin/kata-agent
ExecStartPre=ip netns add podns
ExecStartPre=ip netns exec podns ip link set lo up
ExecStopPost=ip netns delete podns
Environment="KATA_AGENT_SERVER_ADDR=unix:///run/kata-containers/agent.sock"
```

namespaces on Peer Pod VM:
```
# lsns
        NS TYPE   NPROCS   PID USER             COMMAND
4026531835 cgroup    111     1 root             /sbin/init
4026531836 pid       109     1 root             /sbin/init
4026531837 user      111     1 root             /sbin/init
4026531838 uts       106     1 root             /sbin/init
4026531839 ipc       109     1 root             /sbin/init
4026531840 mnt       100     1 root             /sbin/init
4026532243 mnt         1   708 root             /usr/local/bin/kata-agent
4026532301 net         3   708 root             /usr/local/bin/kata-agent
4026532253 pid         1  1481 root             /home/ibm-csi-drivers/ibm-vpc-block-csi-driver
4026532254 mnt         1  1481 root             /home/ibm-csi-drivers/ibm-vpc-block-csi-driver
```

The problem is, `ip netns exec` will create a new mount namespace for kata-agent process, checking with **strace** below:
```
# strace -e unshare,mount  ip netns add ns1
[...]

PR #93, opened 2022-07-14
Implement agent protocol proxy

yoheiueda:
Currently, cloud-api-adaptor just forwards connections from kata-shim
to kata agent via agent-protocol-forwarder without recognizing connection
contents.

This patch implements a proxy server of agent protocol in cloud-api-adaptor
in order to support monitoring and manipulation of TTRPC calls.

Fixes #92

PR #95, opened 2022-07-14
Use nsenter instead of ip netns exec 

yoheiueda:
Currently `kata-agent` is launched in the `podns` network namespace using `ip netns exec`.

`ip netns exec` creates a new mount namespace in addition to a new network namespace. 

https://man7.org/linux/man-pages/man8/ip-netns.8.html#DESCRIPTION
> For applications that are aware of network namespaces, the
> convention is to look for global network configuration files
> first in /etc/netns/NAME/ then in /etc/.  For example, if you
> want a different version of /etc/resolv.conf for a network
> namespace used to isolate your vpn you would name it
> /etc/netns/myvpn/resolv.conf.
> 
> ip netns exec automates handling of this configuration, file
> convention for network namespace unaware applications, by
> creating a mount namespace and bind mounting all of the per
> network namespace configure files into their traditional location
> in /etc.

This behavior interferes accesses to resources on host mount namespace such as cgroupfs. To work around the issue, we needed to implement [/usr/local/bin/cgroup-exec](https://github.com/confidential-containers/cloud-api-adaptor/blob/4f6457773176a21f56dc2341a0e87dc86f046c0b/ibmcloud/image/files/usr/local/bin/cgroup-exec) to prepare cgroupfs on the `podns` network namespace.

This patch uses the `nsenter` command instead to avoid such issues.

This patch also fixes https://github.com/confidential-containers/cloud-api-adaptor/issues/94


yoheiueda:
I found a problem caused by this PR.

I confirmed that container images are successfully pulled and containers are successfully created with this PR, but `/etc/resolv.conf` on **host** is updated by `kata-agent`.

After containers are created on a pod VM, the content of `/etc/resolv.conf` looks like this.
```
search default.svc.cluster.local svc.cluster.local cluster.local
nameserver 10.96.0.10
options ndots:5
```

The DNS server `10.96.0.10` is CoreDNS running in the Kubernetes cluster, and not accessible from the host network namespace of the pod VM.

Name resolution becomes malfunction in the host network, but I am not sure whether this is a big problem or not.

`/etc/resolv.conf` is a tmpfs file, so we can unmount it to restore the original content.

```
# umount /etc/resolv.conf
# cat /etc/resolv.conf
nameserver 161.26.0.10
nameserver 161.26.0.11
```

I will investigate to fix this problem.

yoheiueda:
I found a way to stop updates of `/etc/resolv.conf`.

If `CreatetSandboxRequest.DNS` is empty, `/etc/resolv.conf` will not be updated.

[kata-containers/src/agent/src/rpc.rs#L1243](https://github.com/kata-containers/kata-containers/blob/575b5eb5f5ea06f54a5945cd8718fb32340bb76e/src/agent/src/rpc.rs#L1243)
[kata-containers/src/agent/src/network.rs#L43-L49](https://github.com/kata-containers/kata-containers/blob/575b5eb5f5ea06f54a5945cd8718fb32340bb76e/src/agent/src/network.rs#L43-L49)

`/etc/resolv.conf` on containers are provided as volumes, so updating `/etc/resolv.conf` on host is not necessary at least with Kubernetes.

We will be able implement this workaround after #93 is merged.

bpradipt:
@yoheiueda https://github.com/confidential-containers/cloud-api-adaptor/pull/93 is merged now. Please move this out of draft.

yoheiueda:
@bpradipt thank you! I'll add the work around in this patch, and make this PR ready-to-review.

yoheiueda:
@liudalibj @stevenhorsman @bpradipt I have marked this PR as ready for review again. Could you please take a look?

Please note that @genjuro214 has already tested this change in this environment. https://github.com/confidential-containers/cloud-api-adaptor/issues/94#issuecomment-1186666483

liudalibj (approved):
LGTM

PR #104, opened 2022-07-19
No dns update

### LINKED ISSUE #98: Stop updating /etc/resolv.conf on peer pod VM
Kata-agent updates `/etc/resolv.conf` on a sandbox VM with the content of `/etc/resolv.conf` of the `pause` container.

Related code:
[kata-containers/src/runtime/virtcontainers/kata_agent.go#L751](https://github.com/kata-containers/kata-containers/blob/da7ba2ef71152281452e667d4a37d81405af5095/src/runtime/virtcontainers/kata_agent.go#L751)
[kata-containers/src/runtime/virtcontainers/kata_agent.go#L683-L705](https://github.com/kata-containers/kata-containers/blob/da7ba2ef71152281452e667d4a37d81405af5095/src/runtime/virtcontainers/kata_agent.go#L683-L705)
[kata-containers/src/runtime/virtcontainers/sandbox.go#L2240-L2255](https://github.com/kata-containers/kata-containers/blob/da7ba2ef71152281452e667d4a37d81405af5095/src/runtime/virtcontainers/sandbox.go#L2240-L2255)

Looks like there is no strong reason to do so. 

https://github.com/kata-containers/runtime/issues/1603#issuecomment-520544975
> I see the only value for now in https://github.com/kata-containers/agent/pull/625 is just for completeness.

Actually, `/etc/resolv.conf` on Kubernetes containers are provided as volumes, so updating `/etc/resolv.conf` on host is not necessary at least with Kubernetes.

On the other hand, this behavior causes some problems with peer pod VMs.
* https://github.com/confidential-containers/cloud-api-adaptor/pull/95#issuecomment-1184616726
* Workaround fix for symlink `resolv.conf` [`ibmcloud/image/files/etc/systemd/system/fix-resolvconf.service`](https://github.com/confidential-containers/cloud-api-adaptor/blob/4f6457773176a21f56dc2341a0e87dc86f046c0b/ibmcloud/image/files/etc/systemd/system/fix-resolvconf.service)

We can disable this behavior by removing the `DNS` data in `CreatetSandboxRequest` of the agent protocol.

[kata-containers/src/agent/src/rpc.rs#L1243](https://github.com/kata-containers/kata-containers/blob/575b5eb5f5ea06f54a5945cd8718fb32340bb76e/src/agent/src/rpc.rs#L1243)
[kata-containers/src/agent/src/network.rs#L43-L49](https://github.com/kata-containers/kata-containers/blob/575b5eb5f5ea06f54a5945cd8718fb32340bb76e/src/agent/src/network.rs#L43-L49)

yoheiueda:
This PR adds a fix for #98 as well as the following additional changes.

- 

in the `CreateSandbox` metheod,

```
	if len(req.Dns) > 0 {
		req.Dns = nil
	}
}
```

Fixes #98


pkg/adaptor/proxy/service.go
```diff
@@ -309,17 +327,47 @@ func (s *proxyService) GetMetrics(ctx context.Context, req *pb.GetMetricsRequest
...
+		logger.Print("    Eliminated the DNS setting above from CreateSandboxRequest to stop updating /etc/resolv.conf on the peer pod VM")
+		logger.Print("    See https://github.com/confidential-containers/cloud-api-adaptor/issues/98 for the details.")
+		req.Dns = nil
+	}
+	if len(req.Storages) > 0 {
+		logger.Print("    storages:")
+		for _, s := range req.Storages {
+			logger.Printf("         mountpoint:%s source:%s fstype:%s driver:%s", s.MountPoint, s.Source, s.Fstype, s.Driver)
+		}
+	}
```
bpradipt, line 346: nit: we should make these printfs part of DEBUG log

bpradipt (approved):
/lgtm

pkg/adaptor/proxy/service.go
yoheiueda: I agree. We are using the logger in the standard Go library. It only supports print, fatal, and panic. We need to switch to a modern logging library.

Issue #100, opened 2022-07-18
Fix imagePullPolicy(always) limitation in PeerPod


huoqifeng:
Currently, `imagePullPolicy: Always` need be applied to container so that the image in peerpod can be pulled, it's OK for a workload but not convenient for a sidecar as the sidecar descriptor is usually set by admission controller which is out of the deployer's control. So, we'd better to fix the problem so that container can be run in peerpod when imagePullPolicy equals IfNotPresent

yoheiueda:
I think we need to modify the custom containerd (or Kubernetes) to fix this issue.

Kubernetes checks whether an image already exits or not using the`ImageStatus` CRI call by embeding pod annotations  in `ImageSpec`. So, when  containerd serves `ImageStatusRequest`, it can check pod annotations in `ImageSpec`, and returns NotFound if the request is for peer pod.

https://github.com/kubernetes/kubernetes/blob/f19a26a22ebe26de32319d57231da9812862aaf6/pkg/kubelet/images/image_manager.go#L104-L133
```
	var podAnnotations []kubecontainer.Annotation
	for k, v := range pod.GetAnnotations() {
		podAnnotations = append(podAnnotations, kubecontainer.Annotation{
			Name:  k,
			Value: v,
		})
	}


	spec := kubecontainer.ImageSpec{
		Image:       image,
		Annotations: podAnnotations,
	}
	imageRef, err := m.imageService.GetImageRef(spec)
	if err != nil {
		msg := fmt.Sprintf("Failed to inspect image %q: %v", container.Image, err)
		m.logIt(ref, v1.EventTypeWarning, events.FailedToInspectImage, logPrefix, msg, klog.Warning)
		return "", msg, ErrImageInspect
	}


	present := imageRef != ""
	if !shouldPullImage(container, present) {
		if present {
			msg := fmt.Sprintf("Container image %q already present on machine", container.Image)
			m.logIt(ref, v1.EventTypeNormal, events.PulledImage, logPrefix, msg, klog.Info)
			return imageRef, "", nil
		}
		msg := fmt.Sprintf("Container image %q is not present with pull policy of Never", container.Image)
		m.logIt(ref, v1.EventTypeWarning, events.ErrImageNeverPullPolicy, logPrefix, msg, klog.Warning)
		return "", msg, ErrImageNeverPull
	}
```

[...]

PR #110, opened 2022-07-21
Calling PullImage before CreateContainer to pull images on pod VMs

### LINKED ISSUE #88: Implement workaround for the pause image issue of custom containerd
The custom conatinerd has an issue at pulling a pause container image into a sandbox VM. https://github.com/kata-containers/kata-containers/issues/4477

We currently modified `kata-shim` to pull a pause image to work around the issue. https://github.com/kata-containers/kata-containers/pull/3697#discussion_r917627581

This modification is not relevant to the remote hypervisor proposal, so we should remove the work around fix from the PR for `kata-shim`.

Until the issue is fixed in the custom containerd, we need to work around it in cloud-api-adaptor and/or agent-protocl-forwarder.

yoheiueda:
The container runtime implementations such as containerd and cri-o do not call `PullImage` of kata CCv0 to pull images in pod VMs. So far, we need a custom version of containerd or kata shim to pull images in pod VMs.

With this patch, `cloud-api-adaptor` calls `PullImage` to pull an image in a pod VM, while processing a `CreateContainer` request.

I have tested this patch with both upstream and custom versions of containerd. In theory, this patch should also work with cri-o.

Fixes #88

A side effect of this patch is that pod definitions with `imagePullPolicy: IfNotPresent` now work with peer pod VMs.
- #100


liudalibj (approved):
LGTM

huoqifeng:
This is quite similar as what we did in MVP1 :-)

Issue #128, opened 2022-07-29
PVC volume mount source path is changed in Peer Pod


genjuro214:
We are trying to create a container with PVC volume in Peer Pod. The container (`nginx`) will be running with csi driver (`ibm-vpc-block-podvm-node-driver`) and csi wrapper (`hyperpcons-csi-podvm-wrapper`) together. The pod definition would be:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    app: nginx
  namespace: kube-system
spec:
  runtimeClassName: kata
  containers:
    - name: ibm-vpc-block-podvm-node-driver
      env:
        - name: KUBE_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
      envFrom:
        - configMapRef:
            name: ibm-vpc-block-csi-configmap
      image: gcr.io/k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver:master
      imagePullPolicy: Always
      securityContext:
        privileged: true
        runAsNonRoot: false
        runAsUser: 0
      ports:
        - containerPort: 9808
          name: healthz
          protocol: TCP
      volumeMounts:
        - name: kubelet-data-dir
          mountPath: /var/lib/kubelet
          mountPropagation: Bidirectional
        - mountPath: /tmp
          name: plugin-dir
        - mountPath: /dev
          name: device-dir
        - mountPath: /etc/udev
          name: etcudevpath
        - mountPath: /run/udev
          name: runudevpath
        - mountPath: /lib/udev
          name: libudevpath
        - mountPath: /sys
          name: syspath
        - mountPath: /etc/storage_ibmc
          name: customer-auth
          readOnly: true
    - name: hyperpcons-csi-podvm-wrapper
      env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: POD_NAME_SPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: POD_UID
          valueFrom:
            fieldRef:
              fieldPath: metadata.uid
        - name: POD_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
      args:
        - /usr/bin/hyperpcons-csi-podvm-wrapper
        - --v=5
        - --endpoint=/tmp/csi-podvm-wrapper.sock
        - --target-endpoint=/tmp/csi.sock
[...]

PR #129, opened 2022-07-29
ibmcloud: create kubelet dir for CSI node plugin

genjuro214:
Try to create the dir `/var/lib/kubelet` in peer pod VM in advance, so that it can be mounted to CSI driver node plugin running in peer pod.

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/128
Signed-off-by: Lei Li [cdlleili@cn.ibm.com](mailto:cdlleili@cn.ibm.com)

### DIFF
--- a/ibmcloud/image/build.sh
+++ b/ibmcloud/image/build.sh
@@ -195,6 +195,8 @@ chroot "$src_mnt" bash -c 'rm -rf /var/lib/apt/lists/*'
 
 cp -a "$files_dir"/* "$src_mnt"
 
+mkdir -p "$src_mnt/var/lib/kubelet"
+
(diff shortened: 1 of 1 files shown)


yoheiueda:
How about editing `build.sh` instead of modifying `kata-agent.service`?

```
mkdir -p "$src_mnt/var/lib/kubelet" 
```

We can insert the above line around here.

https://github.com/confidential-containers/cloud-api-adaptor/blob/b8416f381d3c9f79c9a20076b8966e73a5b3cd9e/ibmcloud/image/build.sh#L195-L197

yoheiueda (approved):
LGTM

Issue #126, opened 2022-07-27
Support multiple containers with same image in PeerPod


huoqifeng:
Create a deploy file `nginx-peerpod-2.yaml` like:
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      runtimeClassName: kata
      nodeSelector:
        ibm-cloud.kubernetes.io/worker-id: huoqif-istio10-zvsi-1
      containers:
      - name: nginx1
        image: nginx
        ports:
          - containerPort: 80
      - name: nginx2
        image: nginx
        ports:
          - containerPort: 8080
```

Deploy the Pod
```
kubectl apply -f nginx-peerpod-2.yaml
```

Check the logs in worker:
```
root@huoqif-istio10-zvsi-1:~# tail -f /var/log/cloud-api-adaptor.log
2022/07/13 12:34:53 [helper/hypervisor] store daemon.json at /run/peerpod/pods/581afd29910ef072db9d2d075ac19507f745680649559fd157cae90798a2ef7a
2022/07/13 12:34:56 [helper/hypervisor] created an instance huoqif-istio10-zvsi-1-default-nginx-568dd877d7-k7gcj-581afd29 for sandbox 581afd29910ef072db9d2d075ac19507f745680649559fd157cae90798a2ef7a
2022/07/13 12:35:00 [helper/hypervisor] podNodeIP[0]=10.244.0.18
2022/07/13 12:35:00 [tunneler/vxlan] vxlan ppvxlan1 (remote 10.244.0.18, id: 555024) created at /proc/39286/task/39293/ns/net
2022/07/13 12:35:00 [tunneler/vxlan] vxlan ppvxlan1 created at /proc/39286/task/39293/ns/net
2022/07/13 12:35:00 [tunneler/vxlan] vxlan ppvxlan1 is moved to /run/netns/cni-1fc35d5c-4853-25bb-a545-f99b394e2bfb
2022/07/13 12:35:00 [tunneler/vxlan] Add tc redirect filters between eth0 and vxlan1 on pod network namespace /var/run/netns/cni-1fc35d5c-4853-25bb-a545-f99b394e2bfb
[...]

PR #130, opened 2022-08-01
proxy: fetch imageName from digest

huoqifeng:
fixes: #126
Signed-off-by: huoqifeng <huoqif@cn.ibm.com>

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -57,7 +57,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&hypcfg.HypProvider, "provider", "aws", "Hypervisor provider")
 			flags.StringVar(&cfg.TunnelType, "tunnel-type", podnetwork.DefaultTunnelType, "Tunnel provider")
 			flags.StringVar(&cfg.HostInterface, "host-interface", "", "Host Interface")
-
+			flags.StringVar(&hypcfg.CriSocketPath, "cri-runtime-endpoint", "", "cri runtime uds endpoint")
 		})
 
 	case "ibmcloud":
@@ -80,6 +80,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&hypcfg.HypProvider, "provider", "ibmcloud", "Hypervisor provider")
 			flags.StringVar(&cfg.TunnelType, "tunnel-type", podnetwork.DefaultTunnelType, "Tunnel provider")
 			flags.StringVar(&cfg.HostInterface, "host-interface", "", "Host Interface")
+			flags.StringVar(&hypcfg.CriSocketPath, "cri-runtime-endpoint", "", "cri runtime uds endpoint")
 		})
 
 	case "libvirt":
@@ -93,6 +94,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&hypcfg.HypProvider, "provider", "libvirt", "Hypervisor provider")
 			flags.StringVar(&cfg.TunnelType, "tunnel-type", podnetwork.DefaultTunnelType, "Tunnel provider")
 			flags.StringVar(&cfg.HostInterface, "host-interface", "", "Host Interface")
+			flags.StringVar(&hypcfg.CriSocketPath, "cri-runtime-endpoint", "", "cri runtime uds endpoint")
 		})
 	default:
 		os.Exit(1)
--- a/pkg/adaptor/proxy/proxy.go
+++ b/pkg/adaptor/proxy/proxy.go
@@ -15,7 +15,11 @@ import (
 	"sync"
 
 	"github.com/containerd/ttrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
+	criapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
 const (
@@ -24,6 +28,10 @@ const (
 
 var logger = log.New(log.Writer(), "[adaptor/proxy] ", log.LstdFlags|log.Lmsgprefix)
 
+type criClient struct {
+	criapi.ImageServiceClient
+}
+
 type AgentProxy interface {
 	Start(ctx context.Context, serverURL *url.URL) error
 	Ready() chan struct{}
@@ -31,19 +39,43 @@ type AgentProxy interface {
 }
 
 type agentProxy struct {
-	readyCh    chan struct{}
-	stopCh     chan struct{}
-	stopOnce   sync.Once
-	socketPath string
+	readyCh       chan struct{}
+	stopCh        chan struct{}
+	stopOnce      sync.Once
+	socketPath    string
+	criSocketPath string
 }
 
-func NewAgentProxy(socketPath string) AgentProxy {
+func NewAgentProxy(socketPath, criSocketPath string) AgentProxy {
 
 	return &agentProxy{
-		socketPath: socketPath,
-		readyCh:    make(chan struct{}),
-		stopCh:     make(chan struct{}),
+		socketPath:    socketPath,
+		criSocketPath: criSocketPath,
+		readyCh:       make(chan struct{}),
+		stopCh:        make(chan struct{}),
 	}
+}
+
+func initCriClient(ctx context.Context, target string) (*criClient, error) {
+	if target != "" {
+		conn, err := grpc.DialContext(ctx, target, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock(),
+			grpc.WithContextDialer(func(ctx context.Context, target string) (net.Conn, error) {
+				return (&net.Dialer{}).DialContext(ctx, "unix", target)
+			}),
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		criClient := &criClient{
+			ImageServiceClient: criapi.NewImageServiceClient(conn),
+		}
+
+		logger.Printf("established cri uds connection to %s", target)
+		return criClient, nil
+	}
+
+	return nil, fmt.Errorf("cri runtime endpoint is not specified, it is used to get the image name from image digest.")
 }
 
 func (p *agentProxy) Start(ctx context.Context, serverURL *url.URL) error {
@@ -64,7 +96,13 @@ func (p *agentProxy) Start(ctx context.Context, serverURL *url.URL) error {
 
 	close(p.readyCh)
 
-	proxyService := newProxyService()
+	criClient, err := initCriClient(ctx, p.criSocketPath)
+	if err != nil {
+		// cri client is optional currently, we ignore any errors here
+		logger.Printf("failed to init cri client, the err: %v", err)
+	}
+
+	proxyService := newProxyService(criClient)
 
 	if err := proxyService.connect(ctx, serverURL.Host); err != nil {
 		return err
--- a/pkg/adaptor/proxy/service.go
+++ b/pkg/adaptor/proxy/service.go
@@ -7,15 +7,17 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/containerd/ttrpc"
 	"github.com/gogo/protobuf/types"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols"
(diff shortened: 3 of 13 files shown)


yoheiueda (commented):
Cloud you code a GRPC client for CRI? We need to add an option of the `cloud-api-adaptor` command to specify a Unix domain socket path of containerd/crio-o.

I don't think the `crictl` command and `/etc/crictl.yaml` are available for every worker node in general.

huoqifeng:
> Cloud you code a GRPC client for CRI? We need to add an option of the `cloud-api-adaptor` command to specify a Unix domain socket path of containerd/crio-o.
> 
> I don't think the `crictl` command and `/etc/crictl.yaml` are available for every worker node in general.

Yes, leverage cri service is more graceful.

huoqifeng:
@yoheiueda my concern of using cri service in adpator is: is it reasonable to call cri service in cloud-api-adaptor because the data flow is cri -> shimv2 -> cloud-api-adaptor -> kata-agent currently.  Will that break the data flow?

yoheiueda:
Yes, but the `crictl` command also uses the CRI endpoint, so both approaches break the data flow.

Dependency to the `crictl` command introduces yet another complication, since the command is not generally available in worker nodes. Even if the command is installed, we also need to specify a CRI endpoint like `crictl --image-endpoint ...`, unless `/etc/crictl.yaml` is configured to a CRI endpoint.

Among these complicated approaches, I think writing a GRPC client is the most simple one. What do you think?

huoqifeng:
Agree, we can use a cri client to retrieve it, eventually we need drop it and move the algorithm to containerd/cri-o so that both CC and PeerPod can leverage.

yoheiueda:
Could you make this feature optional? That means the feature is enabled only when the option to specify a CRI image service endpoint is set in command line arguments.

huoqifeng:
> Could you make this feature optional? That means the feature is enabled only when the option to specify a CRI image service endpoint is set in command line arguments.

Do you want to use an optional parameter as feature gate so that after CRI supported it, we don't need remove the code?

yoheiueda:
My intention of the previous comment was that the name conversion occurs only when the option is given in the command line.
```
cloud-api-adaptor --cri-image-endpoint /run/containerd/containerd.sock
```

Existing systemd service unit or launch script does not use the option yet, so we can keep the compatibility if we make it optional.

huoqifeng:
> My intention of the previous comment was that the name conversion occurs only when the option is given in the command line.
> 
> ```
> cloud-api-adaptor --cri-image-endpoint /run/containerd/containerd.sock
> ```
> 
> Existing systemd service unit or launch script does not use the option yet, so we can keep the compatibility if we make it optional.

This should not be a problem because only digest image ref will fall in the branch, and. I'm thinking we can use it as a feature gate also -- the optional parameter. So that we don't need drop the code immediately after CRI supported imageName.

bpradipt:
> Cloud you code a GRPC client for CRI? We need to add an option of the `cloud-api-adaptor` command to specify a Unix domain socket path of containerd/crio-o.
> 
> I don't think the `crictl` command and `/etc/crictl.yaml` are available for every worker node in general.

A very valid point. Depending on the existence of crictl command might not be a good approach.

huoqifeng:
@yoheiueda may you kindly help review the change?

pkg/adaptor/proxy/service.go
```diff
@@ -103,38 +108,81 @@ func (s *proxyService) redirect(ctx context.Context, fn func(c *client)) {
...
+	return fmt.Errorf("cri runtime endpoint is not specified, it is used to get the image name from image digest.")
+}
+
+// TODO: REMOVE ME AFTER cc supported pass image name rather than digest for existing image
+func (s *proxyService) getImageFromDigest(ctx context.Context, digest string) (string, error) {
+	if s.criClient == nil {
+		err := s.initCriClient(ctx)
+		if err != nil {
+			return "", err
+		}
```
yoheiueda, line 160: Race here. How about using `sync.Once` instead?

pkg/adaptor/proxy/service.go
```diff
@@ -5,20 +5,20 @@ package proxy
...
 
+	"github.com/containerd/containerd"
 	"github.com/containerd/ttrpc"
 	"github.com/gogo/protobuf/types"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols"
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
 
 	cri "github.com/containerd/containerd/pkg/cri/annotations"
 	crio "github.com/containers/podman/v4/pkg/annotations"
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
```
yoheiueda, line 21: How about `criapi` instead of `runtime`?

pkg/adaptor/proxy/service.go
```diff
@@ -44,13 +44,18 @@ func newClient(conn net.Conn) *client {
...
 type proxyService struct {
-	agentClient   *client
-	maxRetries    int
-	retryInterval time.Duration
+	agentClient    *client
+	maxRetries     int
+	retryInterval  time.Duration
+	criSocketPath  string
+	criClient      *containerd.Client
+	criImageClient runtime.ImageServiceClient
```
yoheiueda, line 52: We don't need to have two separated fields for criClient and criImageClient. We only need to have ImageServiceClient as criClient.

If we want to use RuntimeServiceClient in the future, we can define a new struct like this
```
type criClient struct {
	criapi.RuntimeServiceClient
	criapi.ImageServiceClient
}
```

cmd/cloud-api-adaptor/main.go
```diff
@@ -57,7 +57,8 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&hypcfg.HypProvider, "provider", "aws", "Hypervisor provider")
 			flags.StringVar(&cfg.TunnelType, "tunnel-type", podnetwork.DefaultTunnelType, "Tunnel provider")
 			flags.StringVar(&cfg.HostInterface, "host-interface", "", "Host Interface")
-
+			// TODO: REMOVE ME AFTER cc supported pass image name rather than digest for existing image
+			flags.StringVar(&hypcfg.CriSocketPath, "runtime-endpoint", "/var/run/crio/crio.sock", "cri runtime uds endpoint")
```
yoheiueda, line 61: Cloud you use the same default value for all three hypervisor types?

yoheiueda (commented):
Could you squash the multiple commits to a single one, and force-push to your branch?

pkg/adaptor/proxy/service.go
```diff
@@ -103,38 +108,81 @@ func (s *proxyService) redirect(ctx context.Context, fn func(c *client)) {
...
+				s.criClient = nil
+				err := fmt.Errorf("reaches max retry count. gave up establishing cri uds connection to %s: %w", s.criSocketPath, err)
+				logger.Print(err)
+				return err
+			}
+			logger.Printf("failed to establish cri uds connection to %s: %v. (retrying... %d/%d)", s.criSocketPath, err, count, maxRetries)
+
+			count++
+		}
+
```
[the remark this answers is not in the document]
huoqifeng: Good point, I'll change it.

pkg/adaptor/proxy/service.go
huoqifeng: Good point, there is a race here indeed.

pkg/adaptor/proxy/service.go
huoqifeng: Sure, it's OK to rename.

pkg/adaptor/proxy/service.go
huoqifeng: Yes, struct is more convenient for future extension.

cmd/cloud-api-adaptor/main.go
huoqifeng: @yoheiueda I was thinking aws and libvirt are using cri-o as CRI runtime, is that the case?

yoheiueda: Sorry, let me change my comment.

Can we change this feature opt-in as I said previously? That means we don't set any default value for this option.

pkg/adaptor/proxy/service.go
huoqifeng: Fixed

pkg/adaptor/proxy/service.go
huoqifeng: Fixed

pkg/adaptor/proxy/service.go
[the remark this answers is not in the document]
huoqifeng: Fixed

pkg/adaptor/proxy/service.go
huoqifeng: I talked with Lei this morning, he tends to use this cri grpc client also for CSI works. so, looks we can move it to initialization algorithm, did the change accordingly.

cmd/cloud-api-adaptor/main.go
huoqifeng: Looks, aws is using cri-o and ibmcloud is using containerd, while libvirt is mostly used for debug/develop. How about we leave as empty for libvirt? @yoheiueda I made the change in new commits.

huoqifeng:
@yoheiueda may you please check the updates?

pkg/adaptor/proxy/service.go
```diff
@@ -100,6 +112,50 @@ func (s *proxyService) redirect(ctx context.Context, fn func(c *client)) {
...
+		logger.Printf("established cri uds connection to %s", s.criSocketPath)
+		return nil
+	}
+
+	return fmt.Errorf("cri runtime endpoint is not specified, it is used to get the image name from image digest.")
+}
+
+func (s *proxyService) getImageFromDigest(ctx context.Context, digest string) (string, error) {
+	err := s.initCriClient(ctx)
```
huoqifeng, line 138: I'll modify this again..

cmd/cloud-api-adaptor/main.go
yoheiueda: My point is that this feature is a kind of hacks and has a complicated data flow as you pointed out. Users may use cri-o on IBM Cloud, and also may use containerd on AWS.

Instead of automatically fixing the issue, I think we should let users explicitly enable this feature when they need to fix the issue. What do you think?

pkg/adaptor/proxy/service.go
huoqifeng: Fixed

pkg/adaptor/proxy/proxy.go
```diff
@@ -64,12 +66,18 @@ func (p *agentProxy) Start(ctx context.Context, serverURL *url.URL) error {
...
 	if err := proxyService.connect(ctx, serverURL.Host); err != nil {
 		return err
 	}
 
+	err = proxyService.initCriClient(ctx)
+	if err != nil {
+		// cri client is optional currently, we ignore any errors here
+		logger.Printf("failed to init cri client, the err: %v", err)
+	}
+
```
yoheiueda, line 80: Instead of calling `initCriClient` from `agentProxy.Start`, we can directly initialize a GRPC client in `agentProxy.Start` using `criSocketPath`, and pass the client to `newProxyService` as its argument.

cmd/cloud-api-adaptor/main.go
huoqifeng: Got your point, yeah, lets leave it empty and as an optional "opt-in" parameters as this is a workaround.

pkg/adaptor/proxy/proxy.go
huoqifeng: Agreed.

huoqifeng: @yoheiueda comments addressed, thanks for the review and very good suggestions!

yoheiueda (approved):
LGTM.

Please also raise another PR to update `kata-agent.service`, when you think it is ready.

huoqifeng:
> LGTM.
> 
> Please also raise another PR to update `kata-agent.service`, when you think it is ready.

@yoheiueda what do you mean `kata-agent.service` change?

yoheiueda:
@huoqifeng 

Sorry. Please forget it.

I meant that we need to update files like this one.

https://github.com/confidential-containers/cloud-api-adaptor/blob/d2e061496c77caf9ecf33332a66600afd9c85f3d/ibmcloud/terraform/start-cloud-api-adaptor/ansible/apply_playbook.yml#L25-L38

huoqifeng:
@yoheiueda That's also my understanding, I opened PR here https://github.com/confidential-containers/cloud-api-adaptor/pull/140 to address it already :-)

Issue #131, opened 2022-08-01
Hypervisor implementations are not using the caller context


bpradipt:
An example from the aws implementation - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/hypervisor/aws/service.go#L187

bpradipt:
Reopening this to include fixes for ibmcloud and libvirt provider

PR #132, opened 2022-08-01
aws: Use caller's context

bpradipt:
Fixes: #131


bpradipt:
@yoheiueda kindly review it. If this looks fine, then I'll make suitable changes for other providers.

pkg/adaptor/hypervisor/aws/service.go
```diff
@@ -229,7 +229,7 @@ func (s *hypervisorService) StartVM(ctx context.Context, req *pb.StartVMRequest)
 	go func() {
 		defer close(errCh)
 
-		if err := sandbox.agentProxy.Start(context.Background(), serverURL); err != nil {
+		if err := sandbox.agentProxy.Start(ctx, serverURL); err != nil {
```
yoheiueda, line 232: Here, we want to keep Agent Proxy running even after `StartVM` exits, since `StartVM` is a TTRPC endpoint, and returns when processing of `StartVMRequest` is done.

If we pass `ctx` from `StartVM` to `agentProxy.Start`, the Agent Proxy is possibly canceled when `StartVM` exits.

So, I think we need to use `context.Background()` here.

If we want to cancel the Agent Proxy only when `StartVM` fails, we need another logic to do so.

See https://go.dev/play/p/HceKHFzQMhE for an example.

yoheiueda: https://github.com/confidential-containers/cloud-api-adaptor/blob/bd9183b17e5efc5e9bb6b9113e4ab08bbf6531ab/pkg/adaptor/hypervisor/aws/service.go#L239-L240

We can insert the following line here to cancel the Agent Proxy when timeout occurs.
```
sandbox.agentProxy.Shutdown()
```

bpradipt: > If we pass `ctx` from `StartVM` to `agentProxy.Start`, the Agent Proxy is possibly canceled when `StartVM` exits.

This part was not clear to me. Since the agentProxy.Start is called after successful instantiation of the pod VM and testing was successful, I was thinking using the same context should be fine. Thanks for clarifying.

yoheiueda: The live of the context object is between beginning and end of a StartVM RPC call. The goroutine of agent proxy lives even after the end of a StartVM RPC call. In other words, a StartVM call and the agent proxy goroutine have different lives.

I guess your testing succeeded because the current implementation of TTRPC server does not use `context.WithTimeout`, `context.WithDeadline`, or `context.WithCancel`.  There is, however, a possibility that the implementation will use `cancel()` in the future.

As described in the doc, `cancel()` must always be called at the end of live, even after successful calls.
https://pkg.go.dev/context#pkg-overview

The agent proxy goroutine will be canceled at the end of a StartVM call, if they share the same context object.

bpradipt: @yoheiueda thanks a lot for the explanation

yoheiueda (approved):
LGTM

Issue #139, opened 2022-08-04
Add cri_runtime_endpoint variables for ibmcloud provider


huoqifeng:
In https://github.com/confidential-containers/cloud-api-adaptor/issues/126, a new "opt-in" parameter `cri-runtime-endpoint` was added to workaround the image pull problem when it's a digest rather than a tag.

This issue will add the parameters in ibmcloud provider.

PR #140, opened 2022-08-04
ibmcloud: add cri_runtime_endpoint variable

huoqifeng:
Fixes: #139
Signed-off-by: huoqifeng <huoqif@cn.ibm.com>

### DIFF
--- a/ibmcloud/terraform/start-cloud-api-adaptor/ansible/apply_playbook.yml
+++ b/ibmcloud/terraform/start-cloud-api-adaptor/ansible/apply_playbook.yml
@@ -35,5 +35,6 @@
               -primary-security-group-id "{{ ibmcloud_vpc_security_group_id }}" \
               -vpc-id "{{ ibmcloud_vpc_id }}" \
               -pods-dir /run/peerpod/pods \
+              -cri-runtime-endpoint "{{ ibmcloud_cri_runtime_endpoint }}" \
               -socket /run/peerpod/hypervisor.sock \
               > /var/log/cloud-api-adaptor.log 2>&1 &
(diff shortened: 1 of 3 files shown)


ibmcloud/terraform/cluster/variables.tf
```diff
@@ -64,4 +64,9 @@ variable "ansible_dir" {
 variable "scripts_dir" {
     description = "Subdirectory for shell scripts"
     default = "./scripts"
+}
+
+variable "cri_runtime_endpoint" {
+    description = "cri-runtime-endpoint for Ansible playbook, inventory and vars files"
+    default = "/run/containerd/containerd.sock"
```
yoheiueda, line 71: The variable `cri_runtime_endpoint` is defined in `start-cloud-api-adaptor/variables.tf`.

Is this definition necessary also in `cluster/variables.tf`?

huoqifeng: @yoheiueda I'm not experts for terraform, but I noticed the variables are all duplicated in https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/ibmcloud/terraform/cluster/variables.tf and https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/ibmcloud/terraform/start-cloud-api-adaptor/variables.tf, is that because `cluster` or `start-cloud-api-adaptor` can be run separately?

yoheiueda: I think variables that to be used in a directory should be defined in the same directory.

Regarding the `cluster` directory, the variable `cri_runtime_endpoint` is not used in the same directory.

huoqifeng: Read the terraform components carefully, yes, I think we don't need it in `cluster`. Revise a little in `start-cloud-api-adaptor` again. @yoheiueda

yoheiueda (approved):
LGTM

PR #144, opened 2022-08-04
ibmcloud: Run udpate-grub to update GRUB configuration

### LINKED ISSUE #143: ibmcloud: GRUB configuration is not updated
The `build.sh` script modifies the GRUB setting, but does not execute the `update-grub` command. Because of this, the updated setting is not reflected on a booted pod VM.

https://github.com/confidential-containers/cloud-api-adaptor/blob/cf5021eacf66bb174b6d14aa2b9e2d3da2d3505d/ibmcloud/image/build.sh#L185

yoheiueda:
This PR adds the following code changes.

## Run update-grub

Run the `udpate-grub` command in `build.sh` to update GRUB configuration.

Fixes #143

## Add SUDO variable

We can build a pod VM image with a non-root user as follows.
```
      make SUDO=sudo build
```
## Add --workdir option

Add an option to change working directory to store temporary QCOW2 images.  We can speed up image builds by storing temporary images on tmpfs.

### DIFF
--- a/ibmcloud/image/build.sh
+++ b/ibmcloud/image/build.sh
@@ -186,7 +186,6 @@ case "$(uname -m)" in
         ;;
     *)
         echo -e "\nUpdating initramfs\n"
-        sed -i -r -e 's|^GRUB_CMDLINE_LINUX=""$|GRUB_CMDLINE_LINUX="nomodeset nofb vga=normal console=ttyS0"|' "$src_mnt/etc/default/grub"
         echo -e "virtio_pci\nvirtio_blk" >> "$src_mnt/etc/initramfs-tools/modules"
         chroot "$src_mnt" update-initramfs  -u
         ;;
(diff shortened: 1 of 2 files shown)


ibmcloud/image/build.sh
```diff
@@ -224,19 +233,27 @@ echo "targetblocksize=512"
...
-
-        umount "$dst_mnt/proc"
-        umount "$dst_mnt/dev"
         rm "$helper"
         ;;
+    *)
+        echo -e "\nUpdating GRUB settings\n"
+        sed -i -r -e 's|^GRUB_CMDLINE_LINUX=|#\0"|' "$dst_mnt/etc/default/grub"
+        cat <<END >> "$dst_mnt/etc/default/grub"
+
```
huoqifeng, line 243: Is there special reason leave empty line in `/etc/default/grub` while not in `zip`l?

yoheiueda: In the case of zipl, we create a new temporary file.

For grub, we need to append several lines to `/etc/default/grub`, so I leave an empty line as a separator.

huoqifeng (approved):
LGTM

PR #149, opened 2022-08-07
Use packer to create qcow2 image for libvirt

bpradipt:


libvirt/README.md
```diff
@@ -26,13 +34,27 @@ If you want to set default password for debugging then you can use guestfish to
...
-     -network-name default  \
-     -pool-name default \
-     -uri "qemu+ssh://root@192.168.122.1/system" \
-     -data-dir "/var/lib/libvirt" 
+    -uri ${LIBVIRT_URI}  \
+    -data-dir /opt/data-dir \
+    -pods-dir /run/peerpod/pods \
+    -network-name ${LIBVIRT_NET} \
+    -pool-name ${LIBVIRT_POOL} \
+    -socket /run/peerpod/hypervisor.sock
```
snir911, line 57: cri-runtime-endpoint was added

huoqifeng: > cri-runtime-endpoint

Yes, this is an optional parameter for CAA, so that we can retrieve the images tag from a digest if an image is pulled 2nd time in guest.

bpradipt: Right - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/proxy/proxy.go#L149

I think for now we can avoid changing the README since its optional param.

snir911 (approved):
Thanks! works well depends on #163

Issue #109, opened 2022-07-20
Make kata-agent create a new network namespace for containers


yoheiueda:
In the current implementation, a `kata-agent` process runs in a fixed network namespace `podns` of a pod VM, and 
creates new containers in the `podns` network namespace.

https://github.com/confidential-containers/cloud-api-adaptor/blob/b5ecb837f63dab71abf2b22d9f4cd729ea132271/ibmcloud/image/files/etc/systemd/system/kata-agent.service#L5-L10

This work around is necessary, since `kata-shim` drops network namespace specification in the OCI runtime spec configuration of a newly created container, and containers created by `kata-agent` runs on the same network namespace where `kata-agent` runs. This just works, since the original `kata-agent` does not initiate any network activities.

https://github.com/kata-containers/kata-containers/blob/6d56cdb9acc9cd39165ab0f7380226d28104128a/src/runtime/virtcontainers/kata_agent.go#L943-L959
```
	// Disable network namespace since it is already handled on the host by
	// virtcontainers. The network is a complex part which cannot be simply
	// passed to the agent.
	// Every other namespaces's paths have to be emptied. This way, there
	// is no confusion from the agent, trying to find an existing namespace
	// on the guest.
	var tmpNamespaces []grpc.LinuxNamespace
	for _, ns := range grpcSpec.Linux.Namespaces {
		switch ns.Type {
		case string(specs.CgroupNamespace):
		case string(specs.NetworkNamespace):
		default:
			ns.Path = ""
			tmpNamespaces = append(tmpNamespaces, ns)
		}
	}
	grpcSpec.Linux.Namespaces = tmpNamespaces
```

However, there are some security and operational concerns with this approach, since the following network communication is done in the `podns`, and is possibly interfered by containers. 
- Image pulling on a pod VM (work around b5ecb837f63dab71abf2b22d9f4cd729ea132271)
- Abstract Unix domain socket communication between `kata-agent` and `agent-protocol-forwarder`

If `kata-agent` can create a new network namespace for containers, `kata-agent` and newly created containers run on separated network namespaces, and concerns mentioned above do not occur.

yoheiueda:
A possible solution of this issue is to use the agent protocol proxy in `cloud-api-adaptor` 4dd33d9b02848aff87179df03a531fd24f29f351.

An OCI runtime spec configuration is embedded in a CreateContainerRequest as follows.

[...]

PR #151, opened 2022-08-08
Implement agent protocol redirector as a common library package

### LINKED ISSUE #150: Implement a common library creating a proxy of agent protocol calls
We need to create a common library package to implement a proxy of agent protocol calls, since we need similar proxy services in both `cloud-api-adaptor` and `agent-protocol-forwarder` to fix #109.

We can add a new package under `pkg/util`, and use the package from `cloud-api-adaptor` and `agent-protocol-forwarder`.

This is required to fix https://github.com/confidential-containers/cloud-api-adaptor/issues/109#issuecomment-1205295195.

yoheiueda:
This PR add a  common library package for proxy service of agent protocol.

Previously, it is implemented in `cloud-api-adaptor`, but `agent-protocol-forwarder` will also need to have a similar service, so we need a common library that can be used in the two components. https://github.com/confidential-containers/cloud-api-adaptor/issues/109#issuecomment-1205295195

Fixes #150


huoqifeng (approved):
LGTM

Issue #89, opened 2022-07-12
ibmcloud: Add skip option for verify console check after podvm-image build


jtumber-ibm:
The current verify step for the ibmcloud podvm image build process has some issues when running as part of a CI pipeline under our service account. 

By adding a flag to skip the console log output and instead just check that the image reaches a status of running we can improve the reliability of the build process.

This could be expanded to update the verify logic to check that other conditions of the images are met.

Adding more details

The `verify.sh` file contains a python script which attaches to the console output of the newly created instance using the `instance-console` command and checking the logs produce certain outputs from that command. However when running under our service account instead we get a `Error during reading message from serial console: websocket: close 1006 (abnormal closure): unexpected EOF`.

To enable us to run this using the service account, our tactical solution is to add an option to skip console log check. In this case we still want to check that the instance reaches a running state using our newly created image.

Tasks:

- [ ] Add `skip_console_verify` (renaming as required) to terraform variables for E2E and podvm-build module
- [ ] Forward the `skip_console_verify` to ansible podvm build scripts
- [ ] Update `verify.sh` to exit early if `skip_console_verify` is set

PR #154, opened 2022-08-08
ibmcloud: add skip_verify_console variable

GeorginaKin:
Allows skipping of console output during verify

Fixes: #89
Signed-off-by: Georgina Kinge <georgina.kinge@ibm.com>

### DIFF
--- a/ibmcloud/image/verify.sh
+++ b/ibmcloud/image/verify.sh
@@ -54,6 +54,11 @@ while ! ibmcloud is instance $id --output json | jq -e '.status == "running"' > 
     sleep 5
 done
 
+if [[ -n "${SKIP_VERIFY_CONSOLE:-}" ]]; then
+    echo "SKIP_VERIFY_CONSOLE is set. Skipping console check..."
+    exit 0
+fi
+
 echo "Watch console log..."
 
 python3 - "$id" <<'END'
--- a/ibmcloud/terraform/podvm-build/ansible/playbook.yml
+++ b/ibmcloud/terraform/podvm-build/ansible/playbook.yml
@@ -36,6 +36,7 @@
         IBMCLOUD_VPC_SUBNET_NAME: "{{ ibmcloud_vpc_subnet_name }}"
         IBMCLOUD_VPC_REGION: "{{ ibmcloud_region_name }}"
         IBMCLOUD_VPC_ZONE: "{{ ibmcloud_vpc_zone }}"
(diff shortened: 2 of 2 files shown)


ibmcloud/image/verify.sh
```diff
@@ -54,6 +54,10 @@ while ! ibmcloud is instance $id --output json | jq -e '.status == "running"' >
     sleep 5
 done
 
+skip_verify_console=$SKIP_VERIFY_CONSOLE
+
+if ["$skip_verify_console" = false] ; then
+
```
yoheiueda, line 60: Could you change the code to skip console checking only when the variable is set?

The code will be something like this.

```
if [[ -n "${SKIP_VERIFY_CONSOLE:-}" ]]; then
	echo 'SKIP_VERIFY_CONSOLE is set. skip console checking'
	exit 0
fi
```

`"${SKIP_VERIFY_CONSOLE:-}"` will suppress the "unbound variable" error, even when an older version of bash is used.

ibmcloud/image/verify.sh
```diff
@@ -99,3 +103,4 @@ os.write(fd, b"\035\n") # send ^[
 os.write(1, b"\n")
 sys.exit(exit_status)
 END
+fi
```
yoheiueda, line 106: Then, remove this line.

yoheiueda (changes requested):
Thank you very much for contributing this patch!

ibmcloud/image/verify.sh
GeorginaKin: Thank you so much for reviewing these changes! I've made some further changes based on your comments :)

ibmcloud/image/verify.sh
```diff
@@ -54,6 +54,13 @@ while ! ibmcloud is instance $id --output json | jq -e '.status == "running"' >
     sleep 5
 done
 
+skip_verify_console=$SKIP_VERIFY_CONSOLE
+
```
yoheiueda, line 58: This line is no longer necessary.

yoheiueda (changes requested):

ibmcloud/image/verify.sh
GeorginaKin: Ah because we won't have the unbound variable issue now of course - I've taken out this line too, thanks!

yoheiueda (approved):
LGTM. Thank you very much!

PR #155, opened 2022-08-09
Launch kata-agent without using nsenter

### LINKED ISSUE #109: Make kata-agent create a new network namespace for containers

yoheiueda:
This PR makes kata-agent create containers in a specified network namespace. Without this patch, kata-agent creates containers in the host network namespace.

We no longer requires the `nsenter` workaround (#95), and cloud-api-adaptor becomes ready for transition to image.rs.

In addition to the above change, this PR add the following changes.

*  Add agent proto logging
*  Move the DNS workaround code to forwarder (#104)
*  Remove unused option (-host-interface)

Fixes #109


pkg/forwarder/interceptor/interceptor.go
```diff
@@ -46,7 +53,73 @@ func NewInterceptor(agentSocket, nsPath string) Interceptor {
...
+}
+
+func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (*types.Empty, error) {
+
+	logger.Printf("CreateContainer: containerID:%s", req.ContainerId)
+
+	res, err := i.Redirector.CreateContainer(ctx, req)
+
+	if err != nil {
+		logger.Printf("CreateContainer fails: %v", err)
```
bpradipt, line 62: nit: wondering if replacing "fails" with "failed with error" be better ?

bpradipt (approved):
/lgtm
Thanks @yoheiueda

pkg/forwarder/interceptor/interceptor.go
yoheiueda: OK. I'll change the message.

bpradipt (approved):
/lgtm

Issue #160, opened 2022-08-10
IBM cloud and AWS implementations not redacting fields from service log


bpradipt:
Example log from ibmcloud provider
```
./cloud-api-adaptor ibmcloud -api-key 12345 -image-id s -key-id k  -zone-name t -vpc-id v -resource-group-id r
cfg: &main.daemonConfig{helperDaemonRoot:"", httpTunnelTimeout:"", TunnelType:"vxlan", HostInterface:""}; hypcfg: hypervisor.Config{SocketPath:"/run/peerpod/hypervisor.sock", CriSocketPath:"", PodsDir:"/run/peerpod/pods", HypProvider:"ibmcloud"}
2022/08/10 04:46:17 [helper/hypervisor] hypervisor config {/run/peerpod/hypervisor.sock  /run/peerpod/pods ibmcloud}
2022/08/10 04:46:17 [helper/hypervisor] cloud config {********** https://iam.cloud.ibm.com/identity/token https://jp-tok.iaas.cloud.ibm.com/v1 r  t s     k v}
2022/08/10 04:46:17 [helper/hypervisor] service config &{12345 https://iam.cloud.ibm.com/identity/token https://jp-tok.iaas.cloud.ibm.com/v1 r  t s     k v}
```

Example log from aws provider
```
 ./cloud-api-adaptor aws  -aws-access-key-id "12345"
cfg: &main.daemonConfig{helperDaemonRoot:"", httpTunnelTimeout:"", TunnelType:"vxlan", HostInterface:""}; hypcfg: hypervisor.Config{SocketPath:"/run/peerpod/hypervisor.sock", CriSocketPath:"", PodsDir:"/run/peerpod/pods", HypProvider:"aws"}
2022/08/10 04:38:46 [helper/hypervisor] hypervisor config {/run/peerpod/hypervisor.sock  /run/peerpod/pods aws}
2022/08/10 04:38:46 [helper/hypervisor] cloud config {12345 ********** ap-south-1 test}
2022/08/10 04:38:46 [helper/hypervisor] service config &{12345  ap-south-1 test}
```

PR #161, opened 2022-08-10
Fix redaction of sensitive fields

bpradipt:
IBM cloud and AWS implementations not redacting fields from service log. This series fixes the issue.
Fixes: #160


huoqifeng (approved):
I saw `Redact` is defined but not used in ibmcloud provider.

LGTM

Issue #165, opened 2022-08-12
Executing sudo causing issues with github action


bpradipt:
The build workflow is using `sudo` for running tests and that's resulting in failure - https://github.com/confidential-containers/cloud-api-adaptor/runs/7800126934?check_suite_focus=true

This is most likely due to the mismatch of the golang version when using `sudo`.

bpradipt:
Did a small experiment on my local dev setup having two different golang versions 

```
$ go version
go version go1.19 linux/amd64

$ sudo go version
go version go1.18.1 linux/amd64

$ sudo -E go version
go version go1.18.1 linux/amd64

$ sudo env "PATH=$PATH" go version
go version go1.19 linux/amd64
```

PR #166, opened 2022-08-12
ci: fix test issue due to sudo usage

bpradipt:
Fixes: #165

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,7 +53,7 @@ jobs:
         run: |
           go install github.com/jstemmer/go-junit-report@v1.0.0
           export CI="true"
-          sudo -E make CLOUD_PROVIDER=${{ matrix.provider }} test | tee tests_report.txt
+          sudo -E env PATH="$PATH" make CLOUD_PROVIDER=${{ matrix.provider }} test | tee tests_report.txt
           sudo chmod o+rw tests_report.txt
           cat tests_report.txt | $(go env GOPATH)/bin/go-junit-report -set-exit-code > tests_report_junit.xml
         shell: bash


.github/workflows/build.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
         run: |
           go install github.com/jstemmer/go-junit-report@v1.0.0
           export CI="true"
-          sudo -E make CLOUD_PROVIDER=${{ matrix.provider }} test | tee tests_report.txt
+          sudo -E env PATH="$PATH" make CLOUD_PROVIDER=${{ matrix.provider }} test | tee tests_report.txt
```
stevenhorsman, line 56: Sorry for my ignorance, but is `sudo -E env PATH="$PATH"` doing anything that `sudo -E PATH="$PATH"` wouldn't? That the pattern I see a lot more commonly in our code

stevenhorsman (commented):
Quick question due to my lack of understanding!

.github/workflows/build.yaml
bpradipt: My understanding was also that `sudo -E PATH="$PATH"` should do what we want. Even `-E` should have been sufficient since it's supposed to preserve environment. But looks like it doesn't. See the examples - https://github.com/confidential-containers/cloud-api-adaptor/issues/165
It might be to do with version of sudo, shell etc. Honestly I don't know :-(

bpradipt: Ok little more digging and google helped.

The behaviour is due to the `env_reset` and `secure_path` options in the `/etc/sudoers` file.   This is a good thread on the same - https://unix.stackexchange.com/questions/83191/how-to-make-sudo-preserve-path

stevenhorsman: It seems like there is a little bit of confusion as to why `sudo -E PATH="$PATH"` isn't working and it might be due to the sudoers entry that we can't control in the github actions, so given that this approach seems to be working for Pradipta, we can go with it now and iterate on it later if necessary - thanks for the extra explanations!

stevenhorsman (approved):
Based on the conversations and evidence supplied from Pradipta - I'm happy to approve. Thanks!

Issue #184, opened 2022-08-29
adapt kata-deploy script to install on crio runtime environment


snir911:
currently installation using the operator works only when runtime engine is containerd

PR #185, opened 2022-08-29
install: enable installation when crio is used

snir911:
set its configuration files etc..

Fixes: #184

* io.kubernetes.cri.sandbox-name annotation doesn't propagate with crio, expected to be fixed in crio-1.24.3, meanwhile can be added manually

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -13,78 +13,20 @@
     kubectl label node $NODENAME node-role.kubernetes.io/worker=
     ```
 
-* **Deploy the `Confidential Containers` operator**
+## Build and install with cloud-api-adaptor running in a pod
 
+* set CLOUD_PROVIDER
     ```
-    kubectl apply -f https://raw.githubusercontent.com/confidential-containers/cloud-api-adaptor/staging/install/yamls/deploy.yaml
-    ```
-
-* **Create a `ConfigMap` with the cloud provider settings**
-
-  Example configmap providing parameters for the AWS provider.
-  Note the `name`, `namespace` and usage of `hyp.env` as the file name in the  data section. These values must not be changed.
-
-    ```
-    apiVersion: v1
-    kind: ConfigMap
-    metadata:
-      name: hyp-env-cm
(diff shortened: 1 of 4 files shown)


install/README.md
```diff
@@ -113,13 +56,12 @@
     kata-qemu   kata-qemu   6m7s
     ```
 
-* Login to the worker node and verify the status remote hypervisor service
+* View cloud-api-adaptor logs
 
     ```
-    sudo systemctl status remote-hyp.service
+    kubectl logs pod/cloud-api-adaptor-deployment-<aws|libvirt>-7c66d68484-zpnnw -n confidential-containers-system
```
bpradipt, line 62: For the example, better to use the same pod name as displayed in the status above

bpradipt (commented):
nit: typo installtion -> installation

bpradipt (approved):
/lgtm

install/runtime-payload/scripts/kata-deploy.sh
```diff
@@ -250,6 +303,9 @@ function main() {
 			[ -x $(command -v containerd) ]; then
 			containerd config default > "$containerd_conf_file"
 		fi
+	# CRI-O isn't consistent with the naming -- let's use crio to match the service file
+	elif [ "$runtime" == "cri-o" ]; then
+		runtime="crio"
```
littlejawa, line 308: Maybe this part of the code should belong to the "get_container_runtime" function?
This would make the function reliably return the runtime's name, wherever it is called

littlejawa: I'm not sure for the rest of the code : at this place, we do what the runtime requires based on the received runtime name, but we don't modify the runtime name.
The code you add does nothing but fix an unreliable runtime name returned by the function, that's why I think it belongs to the function.

Anyway, no big deal, and it certainly can wait :-)

bpradipt (approved):
/lgtm
Thanks @snir911

PR #191, opened 2022-09-06
Update webhook sdk, dependencies and deployment instructions

bpradipt:


bpradipt:
Successfully tested the latest webhook on AWS

install/README.md
```diff
@@ -13,17 +13,21 @@
...
     ```
     export CLOUD_PROVIDER=<aws|ibmcloud|libvirt>
     ```
 
-* `make image` builds the container image and push it to `$registry`
     * configure install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml with your own settings
+
 * `make delete` deletes the pod deployment from the configured cluster
```
snir911, line 31: for consistency maybe we should also switch it to undeploy? could be done in separate of course

webhook/config/manager/kustomization.yaml
```diff
@@ -12,4 +12,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/bpradipt/peer-pods-webhook
+  newName: quay.io/confidential-containers/peer-pods-webhook
```
snir911, line 15: maybe better to have implicit tag? default is latest?

snir911 (approved):
minor comments, however LGTM

webhook/config/manager/kustomization.yaml
bpradipt: Yeah I agree. The current webhook build targets don't have support to add tags like we do for the caa container. I'll add this logic in a separate PR.

PR #212, opened 2022-09-12
ibmcloud: Use variables for source code repos

### LINKED ISSUE #211: ibmcloud: Remove hard-coded source code repos and use variables instead
Currently, Ansible scripts use hard-coded URLs and branch names for external source code repositories.

To improve customizability, we need to introduce variables for such repositories in Terraform configuration, and pass them to the Ansible scripts.

yoheiueda:
This patch remove hard-coded source code repository URLs and branch names, and introduce variables to specify them. The default values are the original hard-coded ones.

Fixes #211


mattarnoatibm (changes requested):
Thanks for the update to set repo branches and URLs.

Can we add the same new 6 variables to `ibmcloud/terraform/variables.tf` so it's possible to set the variables to custom values when running the E2E Terraform configuration.

Can we also pass the variable into `module "cluster" { ...` in `ibmcloud/terraform/main.tf`.

Can we update `ibmcloud/README.md` to define and describe the new variables.

yoheiueda:
@mattarnoatibm thank you for your comments! I've updated the code. could you take a look?

mattarnoatibm (approved):
LGTM

PR #217, opened 2022-09-13
vsphere: Initial CAA implementation

caavery:
This pull request contains the patch which is the initial peer pod cloud-api-adapter for vsphere vcenter.

### DIFF
--- a/pkg/adaptor/hypervisor/vsphere/vsphere.go
+++ b/pkg/adaptor/hypervisor/vsphere/vsphere.go
@@ -1,0 +1,275 @@
+package vsphere
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"net"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/session/cache"
+	"github.com/vmware/govmomi/vim25"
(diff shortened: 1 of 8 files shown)


yoheiueda:
Thank you very much for contributing this patch.

Could you follow the commit message format described here?
https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#main-patch

pkg/adaptor/hypervisor/vsphere/vsphere.go
```diff
@@ -0,0 +1,275 @@
...
+		// vm.Destroy will fail if the VM is still powered on.
+		_ = task.Wait(ctx)
+	}
+
+	task, err = vm.Destroy(ctx)
+	if err != nil {
+		return err
+	}
+
+	task.Wait(ctx)
```
bpradipt, line 272: @caavery is this needed ?

caavery: In all the vsphere go code I've looked at it is common practice to wait for the destroy to complete

caavery: So would this fix it? 
_ = task.Wait(ctx)

bpradipt: Yeah you can use the above

yoheiueda (approved):
LGTM

PR #222, opened 2022-09-14
Add support for creating POD VM in AWS without using a pre-created launch template

bpradipt:


entrypoint.sh
```diff
@@ -23,7 +23,12 @@ cloud-api-adaptor-aws aws \
 	-aws-secret-key ${AWS_SECRET_ACCESS_KEY} \
 	-aws-region ${AWS_REGION} \
 	-pods-dir /run/peerpod/pods \
-	-aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME:-kata} \
+	-use-lt false \
+	-imageid ${PODVM_AMI_ID} \
```
snir911, line 27: I wonder whether we should provide some prebuilt public image at some point..
I suppose the instructions we already have to crate the image are still valid, however i just recalled they are missing the requirement to install aws-cli maybe it would be worth to add it as part of this series

entrypoint.sh
```diff
@@ -23,7 +23,12 @@ cloud-api-adaptor-aws aws \
...
+	-instance-type ${PODVM_INSTANCE_TYPE:-t3.small} \
+	-securitygroupid ${AWS_SG_ID} \
+	-keyname ${SSH_KP_NAME} \
+	-subnetid ${AWS_SUBNET_ID} \
```
snir911, line 31: better to update also aws/kustomization.yaml so user will know to set the new variables

entrypoint.sh
bpradipt: > provide some prebuilt public image at some point..
Do you mean AMI ?

entrypoint.sh
bpradipt: Right. Thanks

bpradipt:
@snir911 my local branch got accidentally deleted. Recreating a new PR

entrypoint.sh
snir911: yes

bpradipt: `aws-cli` is not mandatory. One can create the required AWS infra (VPC, SG etc) using the UI as well. Packer doesn't uses `aws-cli`.

Issue #224, opened 2022-09-15
Create cloud-api-adaptor image for vsphere provider


bpradipt:
With the vsphere provider added, we should also create container image for vsphere provider

PR #227, opened 2022-09-15
vsphere: Add container image support for deployment

bpradipt:
Fixes: #224

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -41,12 +41,26 @@ cloud-api-adaptor-libvirt libvirt \
 	-socket /run/peerpod/hypervisor.sock
 }
 
+vsphere() {
+set -x
+cloud-api-adaptor-vsphere vsphere \
+        -vcenter-url ${GOVC_URL}  \
+	-user-name ${GOVC_USERNAME} \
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -41,12 +41,26 @@ cloud-api-adaptor-libvirt libvirt \
...
+vsphere() {
+set -x
+cloud-api-adaptor-vsphere vsphere \
+        -vcenter-url ${GOVC_URL}  \
+	-user-name ${GOVC_USERNAME} \
+	-password ${GOVC_PASSWORD} \
+	-data-store datastore2 \
+	-pods-dir /run/peerpod/pods \
+	-deploy-folder peerpods \
+	-template podvm-new-template \
```
snir911, line 53: should template be configurable? any idea what is data store?

snir911 (approved):
LGEM (couldn't test though), added a minor comment/question

entrypoint.sh
bpradipt: `template` is equivalent to ec2 launch template. It is configurable. For this iteration let's go with default and we'll change it later on.

PR #241, opened 2022-09-20
Allow ibmcloud provider to be deployed as a POD

bpradipt:


install/overlays/ibmcloud/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
...
+secretGenerator:
+- name: peer-pods-secret
+  namespace: confidential-containers-system
+  literals:
+  - IBMCLOUD_API_KEY="" # set
+  - IBMCLOUD_IAM_ENDPOINT="" #set
+  - IBMCLOUD_ZONE="" #set
+
+patchesStrategicMerge:
+  - cri_runtime_endpoint.yaml # set (modify host's runtime cri socket path in the file, default is /run/containerd/containerd.sock)
```
snir911, line 38: Do we want it enabled by default? (than it would be require to modify configuration when crio is used)

entrypoint.sh
```diff
@@ -33,6 +33,26 @@ cloud-api-adaptor-aws aws \
 	-socket /run/peerpod/hypervisor.sock
 }
 
+ibmcloud() {
```
snir911, line 36: nit: should we test_vars here? also i just noticed it's disabled on aws

bpradipt: I don't think it's needed. Will remove it.

install/overlays/ibmcloud/kustomization.yaml
bpradipt: hmm, I believe ibmcloud is using containerd by default. I'll let @stevenhorsman @yoheiueda and others to comment

inatatsu:
1. The [type field](https://github.com/bpradipt/cloud-api-adaptor/blob/57992202f26b94cac1faf807a5ef945dbef4a12f/install/overlays/ibmcloud/cri_runtime_endpoint.yaml#L18) of a volume causes the following error:
```
Error from server (BadRequest): error when creating "install/overlays/ibmcloud": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.volumes[0].type"
```
2. Deployment of `cloud-api-adaptor-deployment-ibmcloud` fails since the image pull policy is always but the target image is not yet available on `quay.io`.

bpradipt:
> 1. The [type field](https://github.com/bpradipt/cloud-api-adaptor/blob/57992202f26b94cac1faf807a5ef945dbef4a12f/install/overlays/ibmcloud/cri_runtime_endpoint.yaml#L18) of a volume causes the following error:
> 
> ```
> Error from server (BadRequest): error when creating "install/overlays/ibmcloud": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.volumes[0].type"
> ```
> 
> 2. Deployment of `cloud-api-adaptor-deployment-ibmcloud` fails since the image pull policy is always but the target image is not yet available on `quay.io`.

@inatatsu thanks for providing the feedback. Can you please retry and let me know.

snir911 (approved):

yoheiueda (approved):
LGTM

PR #251, opened 2022-09-22
podnetwork: Parameterize VXLAN port and ID

### LINKED ISSUE #244: Pod IP is unreachable when Calico VXLAN mode is set to Always
When pod network is set up with Calico, and its inter-node communication is configured to use VXLAN, the pod IP on a pod VM is unreachable from the worker node.

```
# curl http://172.20.189.195
curl: (7) Failed to connect to 172.20.189.195 port 80: No route to host
```

This problem does not occur with Flannel even though Flannel also uses VXLAN.

The used VXLAN IDs (VNIs) are as follows, so there are no conflicts.

| | VXLAN ID |
| --- | --- |
| Peer Pod | 555000 |
| Calico VXLAN | 4096 |
| Flannel | 1 |

yoheiueda:
Fixes #244

### DIFF
--- a/pkg/podnetwork/tunneler/vxlan/workernode.go
+++ b/pkg/podnetwork/tunneler/vxlan/workernode.go
@@ -18,8 +18,8 @@ import (
 var logger = log.New(log.Writer(), "[tunneler/vxlan] ", log.LstdFlags|log.Lmsgprefix)
 
 const (
-	vxlanPort                = 4789
-	vxlanMinID               = 555000
+	DefaultVXLANPort         = 4789
+	DefaultVXLANMinID        = 555000
 	hostVxlanInterfacePrefix = "ppvxlan"
 	secondPodInterface       = "vxlan1"
 )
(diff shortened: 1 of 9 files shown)


bpradipt (approved):
/lgtm

@yoheiueda were you able to fix the issue in ibmcloud ?

pkg/podnetwork/tunneler/vxlan/workernode.go
```diff
@@ -18,8 +18,8 @@ import (
 var logger = log.New(log.Writer(), "[tunneler/vxlan] ", log.LstdFlags|log.Lmsgprefix)
 
 const (
-	vxlanPort                = 4789
-	vxlanMinID               = 555000
+	DefaultVXLANPort         = 4789
```
huoqifeng, line 21: Should we change default port to another one? so that it's not conflict with calico without passing VxLAN port? I know 4789 is well known, that's why it's possible to conflict.

yoheiueda: In this case, Calico behaves wrongly. The VXLAN port is 4789, which is defined in RFC, and we can distinguish VXLAN packets by VXLAN ID (VNI).

Calico should not drop VXLAN packets that have VXLAN IDs other than 4096, which is the VXLAN ID for packets encapsulated by Calico.

I will raise an issue at Calico's GitHub repo.

yoheiueda:
> /lgtm
> 
> @yoheiueda were you able to fix the issue in ibmcloud ?

Yes. The issue is fixed in IBM Cloud.

Issue #254, opened 2022-09-22
quay complains about caa-pod security risks


snir911:
see:
https://quay.io/repository/confidential-containers/cloud-api-adaptor-aws?tab=tags

PR #255, opened 2022-09-22
caa-pod: run in fedora container

snir911:
to avoid old pkgs security risks

Fixes: #254

**It seems Debian (non-rawhide) based images has a lot of security risk warnings emits by quay's security scanner ([see](https://quay.io/repository/confidential-containers/cloud-api-adaptor-aws?tab=tags)), running in fedora container (build is still on debian) seem to solve all of them AFAIU that's should be safe for the static caa builds and i think it should be fine also for the libvirt build as the libs are installed.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,10 @@ WORKDIR cloud-api-adaptor
 RUN if [ "$CLOUD_PROVIDER" = "libvirt" ] ; then apt-get update -y && apt-get install -y libvirt-dev; fi
 RUN make
 
-FROM golang:1.18
+FROM fedora:36
 ARG CLOUD_PROVIDER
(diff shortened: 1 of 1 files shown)


bpradipt:
@snir911 wondering if using a different image will result in the  missingcertificate issue that we faced earlier?

snir911:
Not sure, I think this issue is more common in slim distributions

bpradipt (dismissed):
/lgtm
I didn't get time to test it.
@snir911 did you test this for any provider ?

snir911:
no, let's not merge, I'll verify tomrrow.

snir911:
adding ssh client for libvirt case
tested both on aws and libivrt @bpradipt

PR #257, opened 2022-09-26
vsphere: Authenticate during StartVM and StopVM

### LINKED ISSUE #253: vsphere fix authentications in case of timeouts before create or delete
Move authentications to create and delete

caavery:
Fixes #253

### DIFF
--- a/pkg/adaptor/hypervisor/vsphere/service.go
+++ b/pkg/adaptor/hypervisor/vsphere/service.go
@@ -145,6 +145,12 @@ func (s *hypervisorService) CreateVM(ctx context.Context, req *pb.CreateVMReques
 
 func (s *hypervisorService) StartVM(ctx context.Context, req *pb.StartVMRequest) (*pb.StartVMResponse, error) {
 
+	sess, lerr := VcenterLogin(s.vim25Client, s.serviceConfig)
+	if lerr != nil {
+		logger.Printf("StartVM login error  %s %v", lerr, sess)
+		return nil, lerr
+	}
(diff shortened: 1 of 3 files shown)


pkg/adaptor/hypervisor/vsphere/service.go
```diff
@@ -145,6 +145,12 @@ func (s *hypervisorService) CreateVM(ctx context.Context, req *pb.CreateVMReques
 
 func (s *hypervisorService) StartVM(ctx context.Context, req *pb.StartVMRequest) (*pb.StartVMResponse, error) {
 
+	sess, lerr := VcenterLogin(s.vim25Client, s.serviceConfig)
```
bpradipt, line 148: Hmm do you really need login/logout for every API call? I feel it'll slow down all the operator and don't really see any use for it.
Can't the login/logout be tied to server start and shutdown (in server.go) ?

caavery: Thats the way it was originally. Unfortunately the login session can timeout in between a create and destroy VM which produces an error. This is a stop gap hack until I figure out how the session manager works and come up with something better. I don't see how it can slow down the operator much as how often do we expect VMs to be created and destroyed?

caavery: Plus logging in is a fraction of the total time it takes to create the VM. Cloning the template is the long pole.

caavery:
I am withdrawing this pull request.

PR #258, opened 2022-09-27
vsphere: Add optional command line settings as defined by kustomization…

### LINKED ISSUE #250: vsphere cloud-api-adaptor fix command line settings for options not included in the kustomization.yaml
entrypoint.sh needs to check if all options have been set by the kustomization.yaml before constructing the command line. Otherwise the command line is not parsed correctly.

caavery:
….yaml

Fixes #250

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -73,15 +73,36 @@ cloud-api-adaptor-libvirt libvirt \
 
 vsphere() {
 set -x
+
+if [[ "${GOVC_TEMPLATE}" ]]; then
+    optionals+="-template ${GOVC_TEMPLATE} "
+fi
+
+if [[ "${GOVC_DATACENTER}" ]]; then
+    optionals+="-data-center ${GOVC_DATACENTER} "
+fi
+
+if [[ "${GOVC_VCLUSTER}" ]]; then
+    optionals+="-vcluster ${GOVC_VCLUSTER} "
+fi
+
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm

entrypoint.sh
```diff
@@ -73,15 +73,36 @@ cloud-api-adaptor-libvirt libvirt \
...
 cloud-api-adaptor-vsphere vsphere \
-        -vcenter-url "${GOVC_URL}"  \
-	-user-name "${GOVC_USERNAME}" \
-	-password "${GOVC_PASSWORD}" \
-	-data-store datastore2 \
-	-pods-dir /run/peerpod/pods \
-	-deploy-folder peerpods \
-	-template podvm-new-template \
-	"${optionals}" \
+	-vcenter-url ${GOVC_URL}  \
```
littlejawa, line 102: I don't see the environment variables being set anywhere. Am I missing it?
Does that mean entrypoint.sh needs to be called differently now, or are you relying on default values in cloud-api-adaptor-vsphere?
(just curious, not a blocking question)

caavery: They are set in kustomization.yaml. If they are not set then then a null is passed to vcenter which uses its defaults. See the comments in https://github.com/confidential-containers/cloud-api-adaptor/issues/250

littlejawa: Got it. Thank you!

littlejawa (approved):

PR #260, opened 2022-09-28
Fix pull image errors when the custom containerd for CCv0 is used with the latest CCv0

### LINKED ISSUE #259: Invalid bundle path is generated due to a pull request by the custom containerd
The protobuf definition of PullImageRequest in the custom containerd for CCv0 does not contain a container ID.

https://github.com/confidential-containers/containerd/blob/4b12e77e79dccebba54e0ec91cc24d6a51a85fa6/runtime/v2/task/image.proto#L28-L31
```
message PullImageRequest {
    // Name of the image.
    string image = 1;
}
```

So, the CRI plugin does not send a container ID in a PullImageRequest.

https://github.com/confidential-containers/containerd/blob/4b12e77e79dccebba54e0ec91cc24d6a51a85fa6/pkg/cri/services/cc/service.go#L989-L992
```
	req := &task.PullImageRequest{}
	if r.Image != nil {
		req.Image = r.Image.Image
	}
```

The protobuf definition of PullImageRequest in the  CCv0 `kata-agent` **does** contain a container ID.

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/libs/protocols/protos/image.proto#L18-L25
```
message PullImageRequest {
    // Image name (e.g. docker.io/library/busybox:latest).
    string image = 1;
    // Unique image identifier, used to avoid duplication when unpacking the image layers.
    string container_id = 2;
    // Use USERNAME[:PASSWORD] for accessing the registry
    string source_creds = 3;
}
```

But, `kata-shim` does **not** send a container ID to `kata-agent`.

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/runtime/virtcontainers/kata_agent.go#L2284-L2286
```
	r := &grpc.PullImageRequest{
		Image: req.Image,
	}
```

The CCv0 `kata-agent` behaves differently based on whether a container ID is supplied or not.

When a container ID is not supplied, the image name is used as a container ID.

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/image_rpc.rs#L226-L236
```
        if cid.is_empty() {
            let v: Vec<&str> = image.rsplit('/').collect();
            if !v[0].is_empty() {
                // ':' have special meaning for umoci during upack
                cid = v[0].replace(':', "_");
            } else {
                return Err(anyhow!("Invalid image name. {}", image));
            }
        } else {
            verify_cid(&cid)?;
        }
```

Then, a container image bundle path is generated from the container ID as follows.

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/image_rpc.rs#L285
```
            let bundle_path = Path::new(CONTAINER_BASE).join(&cid);
```

The `cloud-api-adaptor` sends a PullImageRequest with a container ID just before sending a CreateContainerRequest.

https://github.com/confidential-containers/cloud-api-adaptor/blob/241448c33e983a49ebb3af8d9c296394149ad058/pkg/adaptor/proxy/service.go#L125-L128
```
		pullImageReq := &pb.PullImageRequest{
			Image:       imageName,
			ContainerId: req.ContainerId,
		}
```

If the custom containerd sends a PullImageRequest before sending CreateContainerRequest, kata-agent recieves two PullImageRequests. One without a contiaierID and another one with a container ID. This behavior confuses `kata-agent`,and causes the following error.

```
2022/09/26 07:40:20 [adaptor/proxy] CreateContainer: calling PullImage for "nginx:latest" before CreateContainer
2022/09/26 07:40:22 [adaptor/proxy] CreateContainer: successfully pulled image "nginx:latest"
2022/09/26 07:40:22 [adaptor/proxy] CreateContainer fails: rpc error: code = Internal desc = load image bundle

Caused by:
    No such file or directory (os error 2)
```

yoheiueda:
This PR fixes pull image errors when the custom containerd for CCv0 is used with the latest CCv0.

Fixes #259

### DIFF
--- a/pkg/adaptor/proxy/service.go
+++ b/pkg/adaptor/proxy/service.go
@@ -120,11 +120,10 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 			}
 		}
 
-		logger.Printf("CreateContainer: calling PullImage for %q before CreateContainer", imageName)
+		logger.Printf("CreateContainer: calling PullImage for %q before CreateContainer (cid: %q)", imageName, req.ContainerId)
 
 		pullImageReq := &pb.PullImageRequest{
-			Image:       imageName,
-			ContainerId: req.ContainerId,
+			Image: imageName,
 		}
 
 		pullImageRes, pullImageErr := s.Redirector.PullImage(ctx, pullImageReq)
@@ -134,6 +133,10 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 		} else {
 			logger.Printf("CreateContainer: successfully pulled image %q", pullImageRes.ImageRef)
 		}
+
+		// kata-agent uses this annotation to fix the image bundle path
+		// https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L694-L696
+		req.OCI.Annotations[cri.ImageName] = imageName
 	}
 
 	res, err := s.Redirector.CreateContainer(ctx, req)
--- a/pkg/forwarder/interceptor/interceptor.go
+++ b/pkg/forwarder/interceptor/interceptor.go
@@ -182,3 +182,16 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
 
 	return res, err
 }
+
+func (i *interceptor) PullImage(ctx context.Context, req *pb.PullImageRequest) (*pb.PullImageResponse, error) {
+
+	logger.Printf("PullImage: image: %q, containerID: %q", req.Image, req.ContainerId)
+
+	res, err := i.Redirector.PullImage(ctx, req)
+
+	if err != nil {
+		logger.Printf("PullImage failed with error: %v", err)
+	}
+
+	return res, err
+}


yoheiueda:
I confirmed that this patch works with the following combinations.

- Custom containerd and pod VM image with skepeo
- Custom containerd and pod VM image without skepeo
- Upstream containerd and pod VM image with skepeo
- Upstream containerd and pod VM image without skepeo

bpradipt (approved):
/lgtm
I'll test with crio later on.

snir911:
hi, it fails with crio on aws, not sure what's wrong i get:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: failed to mount "/run/kata-containers/shared/containers/3cf1732550e7a392557ec115c7d4ac585f6c45b25da197f70de37d33be48893c/rootfs" to "/run/kata-containers/3cf1732550e7a392557ec115c7d4ac585f6c45b25da197f70de37d33be48893c/rootfs", with error: ENOENT: No such file or directory: unknown
```
and same in caa:
```
CreateContainer: successfully pulled image "registry.k8s.io/pause:3.6"                                                                                    
2022/09/28 12:14:20 [adaptor/proxy] CreateContainer fails: rpc error: code = Internal desc = failed to mount "/run/kata-containers/shared/containers/42d14b521080dc9014ffb3b00202f2204bde3058e
f23ee2a87519aa1881f473a/rootfs" to "/run/kata-containers/42d14b521080dc9014ffb3b00202f2204bde3058ef23ee2a87519aa1881f473a/rootfs", with error: ENOENT: No such file or directory              
2022/09/28 12:14:20 [adaptor/proxy] DestroySandbox                                                                                                                                            
2022/09/28 12:14:20 [adaptor/proxy] shutting down socket forwarder
```

I built and used the following caa image: `quay.io/snir/cloud-api-adaptor-aws:pull-error`

yoheiueda:
Thank you @bpradipt  and @snir911 

I'll investigate the issue with cri-o.

yoheiueda:
@bpradipt and @snir911 I identified the cause of the error, and have updated the patch to fix the problem. Could you review the updated patch?

The cause of the error is that kata-agent does not recognize cri-o's annotation name for sandbox name.

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L90
```
const ANNO_K8S_IMAGE_NAME: &str = "io.kubernetes.cri.image-name";
```

https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L694-L696
```
        if let Some(image_name) = container_oci
            .annotations
            .get(&ANNO_K8S_IMAGE_NAME.to_string())
```

`kata-agent` recognizes sandbox name by checking annotation `io.kubernetes.cri.image-name`.

https://github.com/containers/podman/blob/29c137ff665314f18a65cf55ba55522e702987b3/pkg/annotations/annotations.go#L76-L77
```
	// SandboxName is the sandbox name annotation
	SandboxName = "io.kubernetes.cri-o.SandboxName"
```

However, cri-o uses another annocation name `io.kubernetes.cri-o.SandboxName`. `kata-agent` can't recognize this annocation name, and causes the error.

I think this can be fixed in kata-agent, but we can also work around the error in cloud-api-adaptor.

https://github.com/confidential-containers/cloud-api-adaptor/blob/931a412aceaf4e23c56da1d5f581fc0dbccf0544/pkg/adaptor/proxy/service.go#L137-L139

I added a workaround that adds annotation ``io.kubernetes.cri.image-name` to every CreateContainerRequest.

yoheiueda:
I confirmed that this patch works with the following combinations.

- Custom containerd and pod VM image with skepeo
- Custom containerd and pod VM image without skepeo
- Upstream containerd and pod VM image with skepeo
- Upstream containerd and pod VM image without skepeo
- cri-o and pod VM image with skepeo
- cri-o and pod VM image without skepeo

bpradipt:
> @bpradipt and @snir911 I identified the cause of the error, and have updated the patch to fix the problem. Could you review the updated patch?
> 
> The cause of the error is that kata-agent does not recognize cri-o's annotation name for sandbox name.
> 
> https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L90
> 
> ```
> const ANNO_K8S_IMAGE_NAME: &str = "io.kubernetes.cri.image-name";
> ```
> 
> https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L694-L696
> 
> ```
>         if let Some(image_name) = container_oci
>             .annotations
>             .get(&ANNO_K8S_IMAGE_NAME.to_string())
> ```
> 
> `kata-agent` recognizes sandbox name by checking annotation `io.kubernetes.cri.image-name`.
> 
> https://github.com/containers/podman/blob/29c137ff665314f18a65cf55ba55522e702987b3/pkg/annotations/annotations.go#L76-L77
> 
> ```
> 	// SandboxName is the sandbox name annotation
> 	SandboxName = "io.kubernetes.cri-o.SandboxName"
> ```
> 
> However, cri-o uses another annocation name `io.kubernetes.cri-o.SandboxName`. `kata-agent` can't recognize this annocation name, and causes the error.
> 
> I think this can be fixed in kata-agent, but we can also work around the error in cloud-api-adaptor.
> 
> https://github.com/confidential-containers/cloud-api-adaptor/blob/931a412aceaf4e23c56da1d5f581fc0dbccf0544/pkg/adaptor/proxy/service.go#L137-L139
> 
> I added a workaround that adds annotation ``io.kubernetes.cri.image-name` to every CreateContainerRequest.

Nice analysis @yoheiueda  I think this should be fixed in kata-agent, but for now we can use the workaround in caa.

Did you create a tracking issue in kata repo ? 
cc  @snir911

yoheiueda:
In [my previous comment](https://github.com/confidential-containers/cloud-api-adaptor/pull/260#issuecomment-1261769158), I referred `SandboxName` instead of `ImageName` by mistake. I've corrected the original comment to refer  `ImageName` as follows.

> `kata-agent` recognizes sandbox name by checking annotation `io.kubernetes.cri.image-name`.
>
> https://github.com/containers/podman/blob/29c137ff665314f18a65cf55ba55522e702987b3/pkg/annotations/annotations.go#L37-L38
>  ```	// ImageName is the container image name annotation
> 	ImageName = "io.kubernetes.cri-o.ImageName"
> ```

bpradipt (approved):
/lgtm

Issue #266, opened 2022-10-06
Fix optionals param usage in the entrypoint.sh


bpradipt:
The `optionals` param in the entrypoint.sh script incorrectly gets converted to a string literal resulting in invocation failure.
https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/entrypoint.sh#L70

PR #267, opened 2022-10-06
install: fix optionals var expansion in entrypoint.sh

bpradipt:
Fixes: #266


yoheiueda (approved):
LGTM.

This works since the values do not contain whitespaces.

In general, we can use a bash array instead of a scalar variable, and refer it here as `"${optionals[@]}"`, if we need to deal with values that include whitespaces.

PR #274, opened 2022-10-07
Sanitize strings when generating VM name

### LINKED ISSUE #265: ibmcloud: sanitise nodeName if it contains none valid characters
The rules for naming vsi instances in ibmcloud are:

1. must contain lowercase alphanumeric characters and hyphens only (without spaces).
2. must have a lowercase letter for the first character.

Currently if you have a nodeName which doesn't match these rules. i.e. it is set to an ip address `10.0.0.1` it will fail to create a instance with the name `10.0.0.1-default-nginx-9348e9fa` for example.

Proposed solution is to replace all none valid characters with `-` resulting in  `10-0-0-1-default-nginx-9348e9fa`. We must then ensure it is prefixed with an lowercase letter. This could be `ip-` or `node-` or `anything-`.

yoheiueda:
This PR sanitizes input string values, and then generates a VM name.

This function generates a VM instance name from sanitized values of node name, pod namespace, pod name, and sandbox ID.

Fixes #265

### DIFF
--- a/pkg/util/hvutil/hvutil.go
+++ b/pkg/util/hvutil/hvutil.go
@@ -1,6 +1,7 @@
 package hvutil
 
 import (
+	"fmt"
 	"strings"
 
 	cri "github.com/containerd/containerd/pkg/cri/annotations"
@@ -22,4 +23,30 @@ func GetPodName(annotations map[string]string) string {
 func GetPodNamespace(annotations map[string]string) string {
 
 	return annotations[cri.SandboxNamespace]
+}
+
+func sanitize(input string) string {
+
+	var output string
+
+	for _, c := range strings.ToLower(input) {
+		if !(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '-') {
+			c = '-'
+		}
+		output += string(c)
+	}
+
+	return output
+}
+
+func CreateInstanceName(nodeName, podNamespace, podName, sandboxID string) string {
+
+	nodeName = sanitize(nodeName)
+	podNamespace = sanitize(podNamespace)
+	podName = sanitize(podName)
+	sandboxID = sanitize(sandboxID)
+
+	vmName := fmt.Sprintf("podvm-%s-%s-%s-%.8s", nodeName, podNamespace, podName, sandboxID)
+
+	return vmName
 }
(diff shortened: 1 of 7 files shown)


littlejawa (dismissed):

pkg/util/hvutil/hvutil.go
```diff
@@ -23,3 +24,33 @@ func GetPodNamespace(annotations map[string]string) string {
...
+		if !(('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == '-') {
+			c = '-'
+		}
+		output += string(c)
+	}
+
+	return output
+}
+
+func GetInstanceName(nodeName, podNamespace, podName, sandboxID string) string {
```
bpradipt, line 42: nit: technically this is creating the name of the instance, so should this method be instead called `MakeInstanceName` or `CreateInstanceName` ?

pkg/util/hvutil/hvutil.go
```diff
@@ -23,3 +24,33 @@ func GetPodNamespace(annotations map[string]string) string {
...
+
+	nodeName = sanitize(nodeName)
+	podNamespace = sanitize(podNamespace)
+	podName = sanitize(podName)
+	sandboxID = sanitize(sandboxID)
+
+	vmName := fmt.Sprintf("%s-%s-%s-%.8s", nodeName, podNamespace, podName, sandboxID)
+
+	if c := vmName[0]; !('a' <= c && c <= 'z') {
+		vmName = "ip-" + vmName
```
bpradipt, line 52: nit: how about using a prefix unique to caa (eg. `podvm` or `caa-vm-`) to make it easier for querying ?
Also should this method also restrict the name to a specified length ?

yoheiueda: OK. I'll use prefix 'podvm`.

pkg/util/hvutil/hvutil.go
yoheiueda: OK. I'll rename to `CreateInstanceName`

bpradipt (approved):
/lgtm

littlejawa (approved):
lgtm
Thanks!

Issue #275, opened 2022-10-07
azure: VM resources might not be garbage collected on timeout


mkulke:
Currently there can be a state in which the creation of a VM has been triggered but the respective api call hasn't returned yet. If a create-vm call is then cancelled at the call site (remote hypervisor) due to a timeout, the deployed VM resources will not be garbage collected, because the sandbox's vm name has not been populated yet. If the system runs into timeouts continuously that can be quite problematic.

https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/hypervisor/azure/service.go#L266

PR #276, opened 2022-10-07
azure: populate sandbox vm name earlier

mkulke:
Fixes: #275

There might be a state in which the creation of a VM has been triggered but the respective api call hasn't returned yet. If a create-vm call is then cancelled at the call site due to a timeout, the VM resources will not be garbage collected, because the sandbox's vm name has not been populated yet.


bpradipt (approved):
/lgtm
Thanks @mkulke

stevenhorsman (approved):
LGTM, Thanks Magnus!

Issue #83, opened 2022-07-07
Feature Request - can we obfuscate the API key in cloud-api-adaptor ?


davidhay1969:
At present the API key is visible in the process listing: -

`ps -ef | grep cloud-api-adaptor`

```text
root     22197     1  0 Jul05 ?        00:00:11 /usr/local/bin/cloud-api-adaptor ibmcloud -api-key <<REDACTED>> -iam-service-url https://iam.cloud.ibm.com/identity/token -vpc-service-url https://jp-tok.iaas.cloud.ibm.com/v1 -key-id r022-80727076-b53b-4f7c-a7b3-96ef9b367ce5 -image-id r022-fe40596b-7cb4-4a37-afee-539c1fb25b61 -profile-name bz2-2x8 -zone-name jp-tok-2 -primary-subnet-id 02f7-13efe926-25f5-4ec6-b09c-aaeed55bebcb -primary-security-group-id r022-b48f040f-4ffc-45b1-8bd2-ac58b08bbaf6 -vpc-id r022-95078f9c-8ba8-482d-b9af-cfed7cf70abf -pods-dir /run/peerpod/pods -socket /run/peerpod/hypervisor.sock
```

and the logs: -

`cat /var/log/cloud-api-adaptor.log`

```text
cfg: &main.daemonConfig{helperDaemonRoot:"", httpTunnelTimeout:"", TunnelType:"vxlan", HostInterface:""}; hypcfg: hypervisor.Config{SocketPath:"/run/pe
erpod/hypervisor.sock", PodsDir:"/run/peerpod/pods", HypProvider:"ibmcloud"}
2022/07/05 20:09:57 [helper/hypervisor] hypervisor config {/run/peerpod/hypervisor.sock /run/peerpod/pods ibmcloud}
2022/07/05 20:09:57 [helper/hypervisor] cloud config {<<REDACTED>> https://iam.cloud.ibm.com/identity/token https://jp-
tok.iaas.cloud.ibm.com/v1  bz2-2x8 jp-tok-2 r022-fe40596b-7cb4-4a37-afee-539c1fb25b61 02f7-13efe926-25f5-4ec6-b09c-aaeed55bebcb r022-b48f040f-4ffc-45b1
-8bd2-ac58b08bbaf6   r022-80727076-b53b-4f7c-a7b3-96ef9b367ce5 r022-95078f9c-8ba8-482d-b9af-cfed7cf70abf}
[...]

PR #284, opened 2022-10-10
azure: mask sensitive fields in cloud config

mkulke:
Use redacted config in log output like in ibmcloud, aws adaptors

Partially Fixes #83

Signed-off-by: Magnus Kulke <magnuskulke@microsoft.com>

PR #285, opened 2022-10-10
cmd: Take sensitive defaults from environment variables

jtumber-ibm:
Adds `os.Getenv` option to each sensitive field.

Fixes: #83

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -55,8 +55,8 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 	switch os.Args[1] {
 	case "aws":
 		cmd.Parse("aws", os.Args[1:], func(flags *flag.FlagSet) {
-			flags.StringVar(&awscfg.AccessKeyId, "aws-access-key-id", "", "Access Key ID")
-			flags.StringVar(&awscfg.SecretKey, "aws-secret-key", "", "Secret Key")
+			flags.StringVar(&awscfg.AccessKeyId, "aws-access-key-id", "", "Access Key ID, defaults to `AWS_ACCESS_KEY_ID`")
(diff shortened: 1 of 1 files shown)


cmd/cloud-api-adaptor/main.go
```diff
@@ -79,8 +79,8 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
 	case "azure":
 		cmd.Parse("azure", os.Args[1:], func(flags *flag.FlagSet) {
-			flags.StringVar(&azurecfg.ClientId, "clientid", "", "Client Id")
-			flags.StringVar(&azurecfg.ClientSecret, "secret", "", "Client Secret")
+			flags.StringVar(&azurecfg.ClientId, "clientid", os.Getenv("AZURE_CLIENT_ID"), "Client Id")
+			flags.StringVar(&azurecfg.ClientSecret, "secret", os.Getenv("AZURE_CLIENT_SECRET"), "Client Secret")
 			flags.StringVar(&azurecfg.TenantId, "tenantid", "", "Tenant Id")
```
stevenhorsman, line 84: Hey JT - from https://github.com/confidential-containers/cloud-api-adaptor/pull/284/files it looks like `tenantid` might also be a sensitive field for Azure

jtumber-ibm:
Changed the approach because the sensitive values would be leaked in the command help e.g.
```
  -api-key IBMCLOUD_API_KEY
        IBM Cloud API key (default "my-api-key")
```
Unlikely that someone would be able to exploit that, if they already have access then they can just `echo $IBMCLOUD_API_KEY` but it could be accidentally left in a log during troubleshooting.
Moved the assignment out of the parsing to avoid that.

stevenhorsman (approved):
LGTM. Thanks for the update!

bpradipt (approved):
/lgtm
Thanks @jtumber-ibm

PR #302, opened 2022-10-14
Use unused IMAGE_NAME for image

knrt10:
Final image that is built should have a generic name instead of
something random like uuid.

Fixes: #300


surajssd (approved):

bpradipt (approved):
/lgtm

Issue #305, opened 2022-10-16
webhook resulting in deadlock


bpradipt:
When the node running the webhook is restarted, the webhook pod fails to start on another node as it is configured to intercept resources required to start its own pods

PR #306, opened 2022-10-16
webhook: update webhook manifest to avoid deadlock

bpradipt:
Fixes: #305

### DIFF
--- a/webhook/docs/INSTALL.md
+++ b/webhook/docs/INSTALL.md
@@ -64,3 +64,13 @@ kubectl set env deployment/peer-pods-webhook-controller-manager -n peer-pods-web
 ```
 
 The default Pod VM instance type is `t2.small` and can be changed by modifying the `POD_VM_INSTANCE_TYPE` environment variable.
+
+### Label namespace for mutation
+
+The mutation needs to be opt-in to avoid any errors or conflicts with critical control-plane pods. 
+For any namespace where you want to create peer pods, add the following label `peerpods: true`
+
+```
+export NAMESPACE="REPLACE_ME"
+kubectl label ns $NAMESPACE peerpods=true
+```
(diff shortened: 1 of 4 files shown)


webhook/docs/INSTALL.md
```diff
@@ -64,3 +64,8 @@ kubectl set env deployment/peer-pods-webhook-controller-manager -n peer-pods-web
 ```
 
 The default Pod VM instance type is `t2.small` and can be changed by modifying the `POD_VM_INSTANCE_TYPE` environment variable.
+
+### Label namespace for mutation
+
+The mutation needs to be opt-in to avoid any errors or conflicts with critical control-plane pods.
+For any namespace where you want to create peer-pods, add the following label `peerpods: true`
```
knrt10, line 71:
```suggestion
The mutation needs to be opt-in to avoid any errors or conflicts with critical control-plane pods. For any namespace where you want to create peer-pods, add the following label `peerpods: true`
```

knrt10, line 71: Also perhaps we could add the step for the user. WDYT?

```bash
export $NODENAME="REPLACE_ME"
kubectl label node $NODENAME peerpods=true
```

knrt10 (changes requested):
Small NIT. Rest LGTM

webhook/docs/INSTALL.md
bpradipt: Thanks @knrt10. Will add the instruction

bpradipt: Just a small caveat; the label will be for the namespace where we want to create peerpods.

bpradipt: Breaking up the long line will facilitate the diffs going forward, hence kept it separate.

littlejawa (approved):

Issue #205, opened 2022-09-09
Add github action to build container image for cloud-api-adaptor


bpradipt:

ariel-adam:
@bpradipt is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

bpradipt:
@knrt10 assigning this issue to you based on your comment above.

@ariel-adam this is a relevant issue.

ariel-adam:
Added the issue to the CoCo releases project drop V0.2.0 priority medium

PR #311, opened 2022-10-18
Actions: Add jobs for building caa images

knrt10:
When the code is pushed to the staging branch, we need container images for different cloud providers to be built automatically.

Fixes: #205

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,0 +1,49 @@
+# (C) Copyright Confidential Containers Contributors 2022.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Build and push container images for each cloud provider.
+---
+name: image
+on:
+  push:
+    branches:
+      - 'staging'
+jobs:
+  build_push_job:
+    name: build and push
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        provider:
+          # Please keep this list in alphabetical order.
+          - aws
+          - azure
+          - ibmcloud
+          - libvirt
+          - vsphere
+        runner:
+          - ubuntu-latest
+        go_version:
+          - 1.18
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v3
+      - name: Setup Golang version ${{ matrix.go_version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yaml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        provider:
+          # Please keep this list in alphabetical order.
+          - aws
+          - azure
+          - ibmcloud
+          - libvirt
```
stevenhorsman, line 23: Should we also build the vsphere image here?

knrt10: I am not sure regarding it because I don't have the context. I checked the [images](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/.github/workflows/build.yaml#L15) we were testing builds for and copied the list or the same. If it makes sense, I guess we should add it. @bpradipt can give more context.

stevenhorsman: Ok, no worries. I just looked on quay and saw there was an image that Pradipta built yesterday for it:
image
, so mentioned it, but lets wait for Pradipta assessment before making changes. Thanks!

bpradipt: Yeah, let's build the image for vsphere as well.

knrt10: Done

.github/workflows/image.yaml
```diff
@@ -0,0 +1,49 @@
+# (C) Copyright Red Hat 2022.
```
bpradipt, line 1: nit: Is the above copyright coming from the github action ?
If not then better to change to "Confidential Containers Contributors"

bpradipt (approved):
/lgtm
Just a nit on the copyright statement

.github/workflows/image.yaml
knrt10: I was also confused initially so copied the same from other github actions yaml in the repo. Let me update.

stevenhorsman (approved):
LGTM. Thanks!

bpradipt (approved):
/lgtm

Issue #313, opened 2022-10-18
All secrets are visible in the container log of CAA POD


bpradipt:
The entrypoint.sh uses `set -x` which prints all the secrets in the container log negating the redaction used in the cloud-api-adaptor binary.
This should be removed.

@jtumber-ibm, any other thoughts from you ?

jtumber-ibm:
We should be able to remove the sensitive args from entrypoint.sh cloud-api-adaptor call.

i.e. 
```
cloud-api-adaptor-ibmcloud ibmcloud \
        -api-key "${IBMCLOUD_API_KEY}" \
        -iam-service-url "${IBMCLOUD_IAM_ENDPOINT}" \
```
`-api-key` option shouldn't be required anymore as it should get the correct value from `IBMCLOUD_API_KEY` env var now. Same should be true for the other providers.

PR #318, opened 2022-10-19
all: remove secrets from entrypoint.sh

jtumber-ibm:
These secrets can be passed in as environment variables.

Fixes: #313

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -25,6 +25,7 @@ test_vars() {
 }
 
 aws() {
+test_vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
 set -x
 
 if [[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]]; then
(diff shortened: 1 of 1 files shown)


snir911 (approved):
LGTM, thanks! one question, do we get any useful error if one of these fields is missing? if not i'd consider using the `test_vars` in entrypoint.sh for these variables until we'll have a proper checking&erroring for this in the caa binary.

jtumber-ibm:
> LGTM, thanks! one question, do we get any useful error if one of these fields is missing? if not i'd consider using the `test_vars` in entrypoint.sh for these variables until we'll have a proper checking&erroring for this in the caa binary.

Good point, I've added in the test for each provider. Pretty sure that there wouldn't be an error message until it tried to authenticate.

snir911 (approved):
Thanks! LGTM, if tested (i didn't test, I'm assuming vars should be available for the executed caa binary )

jtumber-ibm:
> Thanks! LGTM, if tested (i didn't test, I'm assuming vars should be available for the executed caa binary )

Yes they are, tested for ibmcloud.

knrt10 (approved):

PR #322, opened 2022-10-25
CI: enhance with go check and escapes detect

SamYuan1990:
Signed-off-by: Sam Yuan <yy19902439@126.com>

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,3 +29,28 @@ jobs:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
           version: v1.46.2
 
+  go_check:
+    strategy:
+      matrix:
+        provider:
+          # Please keep this list in alphabetical order.
+          - aws
+          - azure
+          - ibmcloud
+          - libvirt
+          - vsphere
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - uses: actions/setup-go@main
+        with:
+          go-version: 1.18
+      - name: Install build dependencies
+        if: ${{ matrix.provider == 'libvirt' }}
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libvirt-dev
+      - name: check
+        run: make CLOUD_PROVIDER=${{ matrix.provider }} check
+      - name: escapes
+        run: make CLOUD_PROVIDER=${{ matrix.provider }} escapes
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,16 @@ endif
 
 ##@ Development
 
+.PHONY: escapes
+escapes: ## golang memeory escapes check
+ifeq ($(CLOUD_PROVIDER),libvirt)
+	go build $(GOFLAGS) -gcflags="-m -l" -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go | grep "escapes to heap" || true
+	go build $(GOFLAGS) -gcflags="-m -l" -o agent-protocol-forwarder cmd/agent-protocol-forwarder/main.go | grep "escapes to heap" || true
+else
+	CGO_ENABLED=0 go build -gcflags="-m -l" $(GOFLAGS) -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go | grep "escapes to heap" || true
+	CGO_ENABLED=0 go build -gcflags="-m -l" $(GOFLAGS) -o agent-protocol-forwarder cmd/agent-protocol-forwarder/main.go | grep "escapes to heap" || true
+endif
+
 .PHONY: test
 test: ## Run tests.
 	# Note: sending stderr to stdout so that tools like go-junit-report can
@@ -52,11 +62,11 @@ check: fmt vet ## Run go vet and go vet against the code.
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
-	find $(SOURCEDIRS) -name '*.go' -print0 | xargs -0 gofmt -l -d
+	find $(SOURCEDIRS) -name '*.go' -print0 | xargs -0 gofmt -l -s -w
 
 .PHONY: vet
 vet: ## Run go vet against code.
-	go vet $(PACKAGES)
+	go vet $(GOFLAGS) $(PACKAGES)
 
 .PHONY: clean
 clean: ## Remove binaries.
@@ -78,5 +88,3 @@ deploy: ## Deploy cloud-api-adaptor using the operator, according to install/ove
 .PHONY: delete
 delete: ## Delete cloud-api-adaptor using the operator, according to install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml file.
 	kubectl delete -k install/overlays/$(CLOUD_PROVIDER)
-
-
--- a/pkg/adaptor/hypervisor/libvirt/libvirt.go
+++ b/pkg/adaptor/hypervisor/libvirt/libvirt.go
@@ -47,7 +47,7 @@ func createCloudInitISO(v *vmConfig, libvirtClient *libvirtClient) string {
 
 	// Set VM Hostname
 	if err := os.MkdirAll(libvirtClient.dataDir, os.ModePerm); err != nil {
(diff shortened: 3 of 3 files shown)


SamYuan1990:
```
# github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/registry
Error: pkg/adaptor/hypervisor/registry/register.go:9:9: undefined: newServer
make: *** [Makefile:59: vet] Error 2
```

I am confused a little bit for where is the newServer function?

SamYuan1990:
memory escapes can be found at https://github.com/SamYuan1990/cloud-api-adaptor/actions/runs/3321361525/jobs/5488906203

SamYuan1990:
> > ```
> > # github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/registry
> > Error: pkg/adaptor/hypervisor/registry/register.go:9:9: undefined: newServer
> > make: *** [Makefile:59: vet] Error 2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am confused a little bit for where is the newServer function?
> 
> https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L194, this code may helps. Maybe fix could be https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/hypervisor/registry/azure.go#L12

I suppose it may a code bug? not sure, but I just remove this function, as in fact we use build tag to control the different cloud provider... hence, I suppose for local development it can be removed safety and replace by return `nil`? 

Also nit fix as result of `go fmt`

Please help enable CI as I am 1st time contributor. :-)

SamYuan1990:
Hi @wainersm and @stevenhorsman, could you please help review this pr?

Makefile
```diff
@@ -39,6 +39,16 @@ else
 	CGO_ENABLED=0 go build $(GOFLAGS) -o "$@" "cmd/$@/main.go"
 endif
 
+.PHONY: escapes
+escapes:
```
knrt10, line 43: Can you please add a help comment here like other make targets? Context: https://github.com/confidential-containers/cloud-api-adaptor/blob/d9caf989940fd77b0d16401ced272d573f5d8760/Makefile#L20

pkg/adaptor/hypervisor/registry/register.go
```diff
@@ -6,5 +6,5 @@ import (
 )
 
 func NewServer(cfg hypervisor.Config, cloudConfig interface{}, workerNode podnetwork.WorkerNode, daemonPort string) hypervisor.Server {
-	return newServer(cfg, cloudConfig, workerNode, daemonPort)
+	return nil
```
knrt10, line 9: Can you please add this is a different commit? Also why is it returning nil?

knrt10 (changes requested):
Some suggestions

pkg/adaptor/hypervisor/registry/register.go
SamYuan1990: the `newServer` function, in fact is useless. 
as you can see ref to @huoqifeng 's comment above. https://github.com/confidential-containers/cloud-api-adaptor/pull/322#issuecomment-1290744174

the `NewServer` function is invoked base on build tag, and if we go through build tag this `newServer` function is not been invoked. or even if this file is able to be deleted? which I am not sure, hence I keep this file.

the reason I make this line to return `nil`
- `hypervisor.Server` is an interface, as the `NewServer` function seems not been invoked according to build tag is not set, I suppose the default value of an interface is `nil`
- as this file without build tag, during go vet process it will try to find `newServer` function as original code. But we don't have any `newServer` function implemented without any build tag. which breaks vet check.

knrt10: Thank you for the explaining. Makes sense

Makefile
SamYuan1990: updated, please review.

SamYuan1990:
sorry for mistake as remove @wainersm from review list.

huoqifeng (dismissed):
/lgtm and thanks @SamYuan1990 contribution!

bpradipt:
/hold
Kindly test this change with different providers before merge.

pkg/adaptor/hypervisor/registry/register.go
bpradipt: @SamYuan1990 Looks like I'm missing something.
The `NewServer` method is the public method called from main. This invokes provider specific`newServer` based on build tag. If NewServer returns nil then how is it going to work?

huoqifeng: > @SamYuan1990 Looks like I'm missing something. The `NewServer` method is the public method called from main. This invokes provider specific`newServer` based on build tag. If NewServer returns nil then how is it going to work?

@bpradipt @SamYuan1990 I'm sorry I misunderstood it, right, it could not return nil. maybe, should we change registry to a factory and return concrete server in https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/main.go#L194? -- I got the same error report when build the repo in my pipeline system.

bpradipt: @huoqifeng I would avoid changing it unless there is a good reason behind it.
Although to be honest, I don't understand this change :-( hence my question to @SamYuan1990 .
Btw, can you point me to the error that is reported ?

.github/workflows/lint.yaml
```diff
@@ -29,3 +29,27 @@ jobs:
...
 
+  go_check:
+    strategy:
+      matrix:
+        provider:
+          # Please keep this list in alphabetical order.
+          - aws
+          - azure
+          - ibmcloud
+          - libvirt
```
bpradipt, line 40: vsphere provider is missing

bpradipt, line 40: vsphere provider is missing

pkg/adaptor/hypervisor/registry/register.go
SamYuan1990: wait a min, @huoqifeng , @bpradipt , if this function return nil breaks pipeline for build, why the build pipeline with tag successes?

SamYuan1990: https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/hypervisor/ibmcloud/service.go#L55 em... will update this pr.

SamYuan1990: @huoqifeng ， @bpradipt so far, I committed with a mock implementation which return nil when build tag is not given as an implementation of `newServer` to avoid error happen as go vet will not read build tag, as
```
# github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor/registry
Error: pkg/adaptor/hypervisor/registry/register.go:9:9: undefined: newServer
make: *** [Makefile:59: vet] Error 2
```

.github/workflows/lint.yaml
SamYuan1990: added

pkg/adaptor/hypervisor/registry/register.go
huoqifeng: > @huoqifeng I would avoid changing it unless there is a good reason behind it. Although to be honest, I don't understand this change :-( hence my question to @SamYuan1990 . Btw, can you point me to the error that is reported ?

@bpradipt my company requires code scan before using it, the scan reported similar errors. Let me create an issue to address those issues reported separately.

SamYuan1990: thanks @huoqifeng , I just updated code as adding build tag in go vet process. as you mentioned at https://github.com/confidential-containers/cloud-api-adaptor/issues/327

huoqifeng (approved):
/lgtm

bpradipt:
Thanks @SamYuan1990..This looks much safer now and we can run it and gradually fix the issues without breaking the working code.

@wainersm @stevenhorsman @knrt10 please take a look and provide your ack so that we can move this forward

knrt10 (approved):
LGTM

Issue #315, opened 2022-10-19
aws: derive the subnet of the worker node automatically to be used for pod VM creation


bpradipt:
Currently for the aws provider we need to explicitly specify the subnet-id when creating the pod VM. When using worker nodes spread across multiple zones, this becomes tricky.

The subnet id should be derived from the worker node on which CAA runs.

PR #325, opened 2022-10-26
aws: retrieve instance metadata from IMDS

snir911:
if not explicitly set, retrieve subnet-id, region and key-name from
AWS Instance Metadata Service

Fixes: #315

(we may fetch also SGs later, this was in mind when created)

to test on aws: quay.io/snir/cloud-api-adaptor-aws:315

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -57,7 +57,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		cmd.Parse("aws", os.Args[1:], func(flags *flag.FlagSet) {
 			flags.StringVar(&awscfg.AccessKeyId, "aws-access-key-id", "", "Access Key ID, defaults to `AWS_ACCESS_KEY_ID`")
 			flags.StringVar(&awscfg.SecretKey, "aws-secret-key", "", "Secret Key, defaults to `AWS_SECRET_ACCESS_KEY`")
-			flags.StringVar(&awscfg.Region, "aws-region", "ap-south-1", "Region")
+			flags.StringVar(&awscfg.Region, "aws-region", "", "Region")
 			flags.StringVar(&awscfg.LoginProfile, "aws-profile", "test", "AWS Login Profile")
 			flags.StringVar(&awscfg.LaunchTemplateName, "aws-lt-name", "kata", "AWS Launch Template Name")
 			flags.BoolVar(&awscfg.UseLaunchTemplate, "use-lt", false, "Use EC2 Launch Template for the Pod VMs")
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -63,7 +63,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.BoolVar(&awscfg.UseLaunchTemplate, "use-lt", false, "Use EC2 Launch Template for the Pod VMs")
 			flags.StringVar(&awscfg.ImageId, "imageid", "", "Pod VM ami id")
 			flags.StringVar(&awscfg.InstanceType, "instance-type", "t3.small", "Pod VM instance type")
-			flags.StringVar(&awscfg.SecurityGroupId, "securitygroupid", "", "Security Group Id to be used for the Pod VM")
+			flags.Var(&awscfg.SecurityGroupIds, "securitygroupids", "Security Group Ids to be used for the Pod VM, comma separated")
 			flags.StringVar(&awscfg.KeyName, "keyname", "", "SSH Keypair name to be used with the Pod VM")
 			flags.StringVar(&awscfg.SubnetId, "subnetid", "", "Subnet ID to be used for the Pod VMs")
 			flags.StringVar(&hypcfg.SocketPath, "socket", hypervisor.DefaultSocketPath, "Unix domain socket path of remote hypervisor service")
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -27,15 +27,12 @@ test_vars() {
 aws() {
 test_vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
 
-if [[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]]; then
-	optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} "
-else
-	optionals+="-imageid ${PODVM_AMI_ID} "
-	optionals+="-instance-type ${PODVM_INSTANCE_TYPE:-t3.small} "
-	[[ "${AWS_SG_ID}" ]] && optionals+="-securitygroupid ${AWS_SG_ID} " # if not set retrieved from IMDS
-	[[ "${SSH_KP_NAME}" ]] && optionals+="-keyname ${SSH_KP_NAME} " # if not retrieved from IMDS
-	[[ "${AWS_SUBNET_ID}" ]] && optionals+="-subnetid ${AWS_SUBNET_ID} " # if not set retrieved from IMDS
-fi
+[[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]] && optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} " # has precedence if set
+[[ "${AWS_SG_ID}" ]] && optionals+="-securitygroupid ${AWS_SG_ID} " # MUST if template is not used
+[[ "${PODVM_AMI_ID}" ]] && optionals+="-imageid ${PODVM_AMI_ID} " # MUST if template is not used
+[[ "${PODVM_INSTANCE_TYPE}" ]] && optionals+="-instance-type ${PODVM_INSTANCE_TYPE} " # default t3.small
+[[ "${SSH_KP_NAME}" ]] && optionals+="-keyname ${SSH_KP_NAME} " # if not retrieved from IMDS
+[[ "${AWS_SUBNET_ID}" ]] && optionals+="-subnetid ${AWS_SUBNET_ID} " # if not set retrieved from IMDS
 [[ "${AWS_REGION}" ]] && optionals+="-aws-region ${AWS_REGION} " # if not set retrieved from IMDS
 
 set -x
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -26,18 +26,19 @@ test_vars() {
 
 aws() {
 test_vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
-set -x
 
 if [[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]]; then
 	optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} "
 else
 	optionals+="-imageid ${PODVM_AMI_ID} "
 	optionals+="-instance-type ${PODVM_INSTANCE_TYPE:-t3.small} "
-	optionals+="-securitygroupid ${AWS_SG_ID} "
-	optionals+="-keyname ${SSH_KP_NAME} "
-	optionals+="-subnetid ${AWS_SUBNET_ID} "
+	[[ "${AWS_SG_ID}" ]] && optionals+="-securitygroupid ${AWS_SG_ID} " # if not set retrieved from IMDS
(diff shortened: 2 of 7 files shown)


bpradipt:
@snir911 can you please split the entrypoint changes into a new PR

snir911:
removed, will open the PR later

entrypoint.sh
```diff
@@ -26,18 +26,16 @@ test_vars() {
...
-	optionals+="-keyname ${SSH_KP_NAME} "
-	optionals+="-subnetid ${AWS_SUBNET_ID} "
-fi
+[[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]] && optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} " # has precedence if set
+[[ "${AWS_SG_ID}" ]] && optionals+="-securitygroupid ${AWS_SG_ID} " # MUST if template is not used
+[[ "${PODVM_AMI_ID}" ]] && optionals+="-imageid ${PODVM_AMI_ID} " # MUST if template is not used
+[[ "${PODVM_INSTANCE_TYPE}" ]] && optionals+="-instance-type ${PODVM_INSTANCE_TYPE} " # default t3.small
+[[ "${SSH_KP_NAME}" ]] && optionals+="-keyname ${SSH_KP_NAME} " # if not retrieved from IMDS
+[[ "${AWS_SUBNET_ID}" ]] && optionals+="-subnetid ${AWS_SUBNET_ID} " # if not set retrieved from IMDS
+[[ "${AWS_REGION}" ]] && optionals+="-aws-region ${AWS_REGION} " # if not set retrieved from IMDS
```
bpradipt, line 36: hmm doesn't the above changes the logic? Earlier the ami, instance etc were only needed when launch template was not used. Now it's not the case..

pkg/adaptor/hypervisor/aws/ec2.go
```diff
@@ -76,3 +76,42 @@ func NewEC2Client(cloudCfg Config) (*ec2.Client, error) {
 	client := ec2.NewFromConfig(cfg)
 	return client, nil
 }
+
+func RetrieveMissingConfig(cfg *Config) error {
```
bpradipt, line 80: nit: Does it make sense to name this method as `AddAWSConfigs` ?

pkg/adaptor/hypervisor/aws/ec2.go
```diff
@@ -76,3 +76,42 @@ func NewEC2Client(cloudCfg Config) (*ec2.Client, error) {
...
+	mdr := NewMetadataRetriever()
```
bpradipt, line 81: checking for mdr ?

pkg/adaptor/hypervisor/aws/types.go
```diff
@@ -19,3 +24,35 @@ type Config struct {
...
+	client *imds.Client
+	mac    string // save it as it commonly used in path
+}
+
+func NewMetadataRetriever() *MetadataRetriever {
+	var r MetadataRetriever
+	r.client = imds.New(imds.Options{ClientEnableState: imds.ClientDefaultEnableState}) // use imds.ClientEnabled to enforce enabling
+	mac, err := r.Get("mac")
+	if err != nil {
+		// print warn // will fail on Get
```
bpradipt, line 38: Better to log the error to help with debugging.

entrypoint.sh
snir911: yes, variables passing logic has changed but caa logic is the same,  i.e. if you choose to set these ami, subnet.. vars in addition to launch template they will be passed to caa but will be ignored in favor of using the template.
I can change it but i think it makes more sense to avoid doing non-mandatory logic (which is already done in the caa code) in the entrypoint.sh, additionally, it will allow to have template fallback if we wish (in the future)

pkg/adaptor/hypervisor/aws/ec2.go
snir911: maybe it does, the question is whether we should mention this configs are retrieved?

pkg/adaptor/hypervisor/aws/ec2.go
snir911: it fails only if it cannot fetch the mac address, which is not always required and if some field does require it will fail on Get
(imds client creating also doesn't seem to fail)

pkg/adaptor/hypervisor/aws/types.go
snir911: right, i missed that, thanks!

pkg/adaptor/hypervisor/aws/ec2.go
bpradipt: Yeah you are right. I couldn't think of a better name other than "RetrieveMissingConfig"

entrypoint.sh
bpradipt: I see.. Thanks for the explanation.

pkg/adaptor/hypervisor/aws/ec2.go
bpradipt: ok. May be add a comment just above this call for future use.

snir911:
changes:
- changed function names to lower case as it should be used from within the aws pkg
- fixes as commented (note: in types.go i used to fmt as logger is not available for all build tags)

pkg/adaptor/hypervisor/aws/types.go
```diff
@@ -19,3 +25,36 @@ type Config struct {
...
 	return *util.RedactStruct(&c, "AccessKeyId", "SecretKey").(*Config)
 }
+
+type MetadataRetriever struct {
+	client *imds.Client
+	mac    string // save it as it commonly used in path
+}
+
+// no error return, if something is wrong, will fail on Get()
+func newMetadataRetriever() *MetadataRetriever {
```
bpradipt, line 35: @snir911 as discussed offline, let's move the imds methods to it's own file.

snir911:
changes:
* moved imds to a new file (which allows to use logger now)
* added SGs retrieve (also adding WIP as it was not tested yet)
also, as multiple SGs are allowed and i cannot tell which one should be used by the caa, it fetches all the SGs the node is using, is this security compromise or does it make sense?

bpradipt:
> changes:
> 
> * moved imds to a new file (which allows to use logger now)
> * added SGs retrieve (also adding WIP as it was not tested yet)
>   also, as multiple SGs are allowed and i cannot tell which one should be used by the caa, it fetches all the SGs the node is using, is this security compromise or does it make sense?

I think it should be fine. If a specific SG is required then it should be specified in the config

snir911:
tested (quay.io/snir/cloud-api-adaptor-aws:315)

cmd/cloud-api-adaptor/main.go
```diff
@@ -57,13 +57,13 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
 			flags.StringVar(&awscfg.SecretKey, "aws-secret-key", "", "Secret Key, defaults to `AWS_SECRET_ACCESS_KEY`")
-			flags.StringVar(&awscfg.Region, "aws-region", "ap-south-1", "Region")
+			flags.StringVar(&awscfg.Region, "aws-region", "", "Region")
 			flags.StringVar(&awscfg.LoginProfile, "aws-profile", "test", "AWS Login Profile")
 			flags.StringVar(&awscfg.LaunchTemplateName, "aws-lt-name", "kata", "AWS Launch Template Name")
 			flags.BoolVar(&awscfg.UseLaunchTemplate, "use-lt", false, "Use EC2 Launch Template for the Pod VMs")
 			flags.StringVar(&awscfg.ImageId, "imageid", "", "Pod VM ami id")
 			flags.StringVar(&awscfg.InstanceType, "instance-type", "t3.small", "Pod VM instance type")
-			flags.StringVar(&awscfg.SecurityGroupId, "securitygroupid", "", "Security Group Id to be used for the Pod VM")
+			flags.Var(&awscfg.SecurityGroupIds, "securitygroupid", "Security Group Id to be used for the Pod VM")
```
bpradipt, line 66: The short name be changed should also be changed to `securitygroupids`

entrypoint.sh
```diff
@@ -26,18 +26,16 @@ test_vars() {
...
-	optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} "
-else
-	optionals+="-imageid ${PODVM_AMI_ID} "
-	optionals+="-instance-type ${PODVM_INSTANCE_TYPE:-t3.small} "
-	optionals+="-securitygroupid ${AWS_SG_ID} "
-	optionals+="-keyname ${SSH_KP_NAME} "
-	optionals+="-subnetid ${AWS_SUBNET_ID} "
-fi
+[[ "${PODVM_LAUNCHTEMPLATE_NAME}" ]] && optionals+="-use-lt -aws-lt-name ${PODVM_LAUNCHTEMPLATE_NAME} " # has precedence if set
+[[ "${AWS_SG_ID}" ]] && optionals+="-securitygroupid ${AWS_SG_ID} " # MUST if template is not used
```
bpradipt, line 31: The short name be changed should also be changed to `securitygroupids`  include the variable AWS_SG_IDS

install/overlays/aws/kustomization.yaml
```diff
@@ -19,18 +19,18 @@ configMapGenerator:
   #- PODVM_LAUNCHTEMPLATE_NAME="" # Uncomment and set if you want to use launch template
   # Comment out all the following variables if using launch template
   - PODVM_AMI_ID="" #set
-  - PODVM_INSTANCE_TYPE="t3.small" #set
-  - AWS_SG_ID="" #set
-  - SSH_KP_NAME="" #set
-  - AWS_SUBNET_ID="" #set
+  #- PODVM_INSTANCE_TYPE="t3.small" # caa defaults to t3.small
+  #- AWS_SG_ID="" # if not set retrieved from IMDS (NOTE: if not set it will use all node SGs)
```
bpradipt, line 23: AWS_SG_IDS

cmd/cloud-api-adaptor/main.go
bpradipt: Another aspect is how the multiple values should be passed from the cli

snir911: Currently cmdline supports only single SG, the reason i added support to multiple SGs when getting it from IDMS is only because i couldn't tell which i should choose.
I couldn't think of a use-case where multiple values are needed for the cmdline but i'll add support it also if it just the sake of consistency

bpradipt (approved):
/lgtm
Thanks @snir911

littlejawa (approved):

cmd/cloud-api-adaptor/main.go
```diff
@@ -57,13 +57,13 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
 			flags.StringVar(&awscfg.SecretKey, "aws-secret-key", "", "Secret Key, defaults to `AWS_SECRET_ACCESS_KEY`")
-			flags.StringVar(&awscfg.Region, "aws-region", "ap-south-1", "Region")
+			flags.StringVar(&awscfg.Region, "aws-region", "", "Region")
 			flags.StringVar(&awscfg.LoginProfile, "aws-profile", "test", "AWS Login Profile")
 			flags.StringVar(&awscfg.LaunchTemplateName, "aws-lt-name", "kata", "AWS Launch Template Name")
 			flags.BoolVar(&awscfg.UseLaunchTemplate, "use-lt", false, "Use EC2 Launch Template for the Pod VMs")
 			flags.StringVar(&awscfg.ImageId, "imageid", "", "Pod VM ami id")
 			flags.StringVar(&awscfg.InstanceType, "instance-type", "t3.small", "Pod VM instance type")
-			flags.StringVar(&awscfg.SecurityGroupId, "securitygroupid", "", "Security Group Id to be used for the Pod VM")
+			flags.Var(&awscfg.SecurityGroupIds, "securitygroupids", "Security Group Ids to be used for the Pod VM, comma separated")
```
huoqifeng, line 66: Will this cause other providers code change also? comma separated array looks has less impact but seems the key changed from `securitygroupid` to `securitygroupids`.

bpradipt: It won't have impact to other providers since the code is specific to aws. However the other providers can use the same logic to add support for inputs which can take multiple values.

bpradipt: @huoqifeng hope this addresses your question ?

huoqifeng (approved):
LGTM

yoheiueda (approved):
LGTM

PR #336, opened 2022-11-02
vsphere: use packer to create the podvm template

whitebrandy:
v2: 
Added a few changes as pointed out in the reviews
 - Added the option to configure disk size by the user 
 - Cleaned up clutter in the Makefile and moved information gathering to a script
 - Renamed data/user-data.pkrtpl.hcl to data/user-data.pkr.hcl
 - Removed settings.auto.pkrvars.hcl, this will be created when building the template by the above script
 - Added *auto*.hcl files to .gitignore
 - Added a README
 - Fix for #339 (included in a previous push)

Mostly, derived from its libvirt counterpart, this script creates an esx guest from a standard ubuntu iso and then converts it into a template.

Standard settings are in settings.auto.pkrvars.hcl. vsphere config is expected in vsphere.auto.pkrvals.hcl which will be created by the Makefile if not present; user-data.pkrtpl.hcl contains the autoinstall template.

The automated input of characters at boot to start autoinstall is kind of flaky, the currently working sequence is defined in boot_command of the main script.

Fixes: #337 

### DIFF
--- a/vsphere/image/files/etc/systemd/system/kata-agent.service
+++ b/vsphere/image/files/etc/systemd/system/kata-agent.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Kata Agent
+After=network.target
+
+[Service]
+ExecStart=/usr/local/bin/kata-agent --config /etc/agent-config.toml
+ExecStartPre=ip netns add podns
+ExecStartPre=ip netns exec podns ip link set lo up
+ExecStopPost=ip netns delete podns
+#Environment="KATA_AGENT_SERVER_ADDR=unix:///run/kata-containers/agent.sock"
+SyslogIdentifier=kata-agent
+
+[Install]
+WantedBy=multi-user.target
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,7 @@ skopeo
 
 # IDEs
 .vscode
+
+# Files
+vcenter.auto.pkrvars.hcl
+settings.auto.pkrvars.hcl
--- a/vsphere/README.md
+++ b/vsphere/README.md
@@ -1,0 +1,38 @@
+# Setup instructions
+
+## Prerequisites:
+
+- A configured vCenter instance
+- datacenter, datastore and cluster are required fields
+
+## Creating the vsphere template:
+- Install packer on your system by following the instructions in the following [link](https://learn.hashicorp.com/tutorials/packer/get-started-install-cli)
+- Create the vsphere template
+```
+cd vsphere/image && make
+```
+If successful, you can find the image on your vcenter deployment
+
+## File structure
+- **vcenter.auto.pkrvars.hcl** - The main template.
+- **settings.auto.pkrvars.hcl** - Guest settings referenced by the template file
+- **vcenter.auto.pkrvars.hcl** - Your vCenter config
+
+If settings.auto.pkrvars.hcl and vcenter.auto.pkrvars.hcl are not present, they will be created
+when you first run make. You can also choose to create these files if you prefer.
+
+Please take a look at variables.pkr.hcl for all the variables supported.
+
+Here are a few that you might find useful to configure the guest:
+- *vm_cpu_count*
+  This field configures the number of vcpus for the guest.
+- *vm_guest_size*
+  Determines the amount of guest memory.
+- *vm_disk_size*
+  The size of the virtual hard disk. Please note that installation might fail if the disk size is too small.
+- *vm_network_name*
+  The virtualized network adapter to use. vmxnet3 is the default.
+
+## Potential issues
+The start of the installation uses automated keyboard input. Timing issues may prevent entering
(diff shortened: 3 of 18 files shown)


whitebrandy:
@caavery Please take a look and review when you get a chance.

caavery:
Will do. I'm in the process of trying it out.

vsphere/image/files/etc/systemd/system/kata-agent.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Kata Agent
+After=network.target
+
+[Service]
+ExecStart=/usr/local/bin/kata-agent
```
fidencio, line 6: Something similar to https://github.com/confidential-containers/cloud-api-adaptor/pull/339 wil have to be done here, as in, passing the agent configuration and also creating the configuration file.

whitebrandy: Thanks for the review and the useful info. I updated kata-agent.service to reflect this change based on changes made in #339

caavery:
Tested on my vsphere rig.

lgtm

vsphere/image/data/user-data.pkrtpl.hcl
```diff
@@ -0,0 +1,198 @@
+#cloud-config
```
bpradipt, line 1: nit: can we keep the naming consistent with other packer files used in the project: `<name>.pkr.hcl` ? 
Ref: https://developer.hashicorp.com/packer/docs/templates/hcl_templates#code-organization)

This file could be just `user-data.pkr.hcl`

vsphere/image/settings.auto.pkrvars.hcl
```diff
@@ -0,0 +1,33 @@
+/*
+    DESCRIPTION:
```
bpradipt, line 2: can some of the values be moved to `variables.pkr.hcl` as default values and this file be removed ?

vsphere/image/settings.auto.pkrvars.hcl
```diff
@@ -0,0 +1,33 @@
...
+vm_guest_os_type = "ubuntu64Guest"
+
+
+// Virtual Machine Hardware Settings
+vm_firmware              = "efi"
+vm_cpu_count             = 2
+vm_mem_size              = 2048
+vm_disk_size             = 40960
+vm_disk_thin_provisioned = true
+vm_interface_name         = "vmxnet3"
```
bpradipt, line 19: is the interface_name constant or depends on env ?

whitebrandy: Sorry, this misnamed variable actually represents the virtualized network card. vmxnet3 is the most recent paravirtualized network card and is default in my setup but I am not sure if it really is! I believe it's best to specify to make sure we are consistent. Maybe, I can rename this variable to vm_network_card ?

vsphere/image/settings.auto.pkrvars.hcl
whitebrandy: The way I envisioned the design for this is that the *auto*.hcl files contain the user facing name=value settings that are most likely to change for a custom template build. Ideally, I would think that users should not have to alter the variables file to edit default values which has a more "structured" syntax and IMO is easier to break. The examples I followed from https://github.com/vmware-samples/packer-examples-for-vsphere also seem to follow this approach. Do you see any disadvantages to this ?

vsphere/image/data/user-data.pkrtpl.hcl
whitebrandy: Will do, thanks for the review.

vsphere/image/settings.auto.pkrvars.hcl
bpradipt: Fair point. I was going by the pattern where any custom values will be passed as env variables and handled in the Makefile target. Example - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/aws/image/Makefile#L73
If having the *.auto.pkrvars.hcl files then we need to decide whether to track it or add it to `.gitignore`. 

Another aspect to consider is the number of input params required. For a large number of params `*.auto.pkrvars.hcl` seems more convenient. Hence I was thinking if some of the defaults are common then those can be added to `variables.pkr.hcl` 
Anyways, I'll leave the decision to you. Both approaches are fine :-)

vsphere/image/settings.auto.pkrvars.hcl
bpradipt: You can add a brief explanation in the README

vsphere/image/settings.auto.pkrvars.hcl
whitebrandy: Thanks @bpradipt I pushed new changes and updated the cover letter to summarize them.

snir911 (approved):
I'm not a packer expert nor vsphere user, however from what i do understand, LGTM

caavery (approved):

PR #351, opened 2022-11-15
aws, azure: Update install instructions

knrt10:
Instead of using the binary and manually updating command line flags when running CAA, users can make use of daemonset to run CAA now.

Fixes: #294


bpradipt (approved):
/lgtm
thanks @knrt10

snir911 (approved):
Thanks!

Issue #357, opened 2022-11-17
Issues when handling multiple IP addresses on the primary network interface


bpradipt:
CAA doesn't handle multiple IP addresses on the primary network interface and consequently POD creation fails.

PR #358, opened 2022-11-17
podnetwork: Handle network interface with multiple addresses

bpradipt:
Fixes: #357

### DIFF
--- a/pkg/podnetwork/workernode.go
+++ b/pkg/podnetwork/workernode.go
@@ -89,8 +89,10 @@ func (n *workerNode) Inspect(nsPath string) (*tunneler.Config, error) {
 		return nil, fmt.Errorf("failed to get IP address on %s (netns: %s): %w", hostInterface, hostNS.Path, err)
 	}
 	if len(addrs) != 1 {
-		return nil, fmt.Errorf("more than one IP address assigned on %s (netns: %s)", hostInterface, hostNS.Path)
+		logger.Printf("more than one IP address (%v) assigned on %s (netns: %s)", addrs, hostInterface, hostNS.Path)
 	}
+	// Use the first IP as the workerNodeIP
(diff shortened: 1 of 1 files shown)


snir911 (approved):
Is it possible addrs[0] will contain wrong IP? anyhow, i guess for now it's good enough, thanks!

knrt10 (approved):

bpradipt:
> Is it possible addrs[0] will contain wrong IP? anyhow, i guess for now it's good enough, thanks!

Possible. We need an improved mechanism to get the worker node IP.

huoqifeng:
> > Is it possible addrs[0] will contain wrong IP? anyhow, i guess for now it's good enough, thanks!
> 
> Possible. We need an improved mechanism to get the worker node IP.

I'm also worry about it, maybe @yoheiueda can comment on that?

bpradipt:
> > > Is it possible addrs[0] will contain wrong IP? anyhow, i guess for now it's good enough, thanks!
> > 
> > 
> > Possible. We need an improved mechanism to get the worker node IP.
> 
> I'm also worry about it, maybe @yoheiueda can comment on that?

Just to be clear, the existing code already used the first IP address of the network device. Just that it returned error when there were more than one ip addresses in the network device, which effectively meant that the code can only handle network interface with single IP address. So this PR just corrects that.

We do need a better (and faster) mechanism to get the worker node IP. Since we run cloud-api-adaptor as a pod, we might be able to use `downward api` to get the worker node primary IP and use the same. I can do that in a separate PR.

yoheiueda (approved):
LGTM.

I think my original implementation is too conservative to handle various network configurations. The intention of the code was to notify users if there is ambiguity when cloud-api-adaptor decides worker node IP address. I think showing a warning message is enough for now.

Issue #359, opened 2022-11-18
Azure: Get rid of the manual route requirement for CAA


knrt10:
When the CAA deploys a pod VM and spawns workload on the remote machine, we cannot reach out to the workload from the worker machine. The pod traffic doesn't work, unless we create routes manually.

This manual route creation requirement is specific to Azure, other clouds don't need this.

Docs: https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/azure#running-cloud-api-adaptor

PR #360, opened 2022-11-18
azure: remove manual route creation

knrt10:
When using Calico CNI, it was unexpectedly dropping VXLAN packets unrelated to calico. To avoid manual route creation one has to configure VXLAN encapsulation on calico and a new VXLAN UDP port is used rather than the default one.

Fixes: #359

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -47,6 +47,7 @@ exec cloud-api-adaptor-azure azure \
   -region "${AZURE_REGION}" \
   -instance-size "${AZURE_INSTANCE_SIZE}" \
   -resourcegroup "${AZURE_RESOURCE_GROUP}" \
+  -vxlan-port 8472 \
   -subnetid "/subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP}/providers/Microsoft.Network/virtualNetworks/${AZURE_VM_NAME}VNET/subnets/${AZURE_VM_NAME}Subnet" \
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -47,6 +47,7 @@ exec cloud-api-adaptor-azure azure \
   -region "${AZURE_REGION}" \
   -instance-size "${AZURE_INSTANCE_SIZE}" \
   -resourcegroup "${AZURE_RESOURCE_GROUP}" \
+  -vxlan-port 8472 \
```
bpradipt, line 50: While the change is good as-is, I wanted to ask you if want to make it configurable across all providers.?

I had to use a similar mechanism for aws when using a ovn cni. I recall @yoheiueda had to use it in ibmcloud as well. So I was planning to add  vxlan-port in the `kustomization.yaml` file, but since you have this PR I wanted to check with you before raising a new PR:-)

knrt10: I think making it configurable among other providers make sense. So should I close this PR?

bpradipt: @knrt10 I will need some time before raising a new PR. So let's go ahead with this one so that you are unblocked. 
Hope it's fine ?

bpradipt (approved):
/lgtm
Thanks @knrt10.

entrypoint.sh
knrt10: Cool. Thanks

mkulke (approved):
lgtm

PR #345, opened 2022-11-09
libvirt: add e2e demo code and documentation

wainersm:
In https://github.com/confidential-containers/cloud-api-adaptor/issues/16#issuecomment-1292524184 I drafted a framework that we could be using for running e2e tests on CI. To continue that work I want to create an implementation for libvirt because developers can run on their own workstation and this should allow others to test the framework.

But soon I realized the information to build an environment to develop for libvirt is spread in the repository or simply missing. Then I started fixing some stuffs and automating some steps. This pull request contain that work. Apart from using the changes I propose here on CI, I think it will be useful for developers setting up their environments for testing locally.

Issue #16, opened 2022-05-12
ci: Start thinking about CI/CD options for the cloud-api-adaptor/peer-pods solution


stevenhorsman:
This is (initially at least), not going to be a proper plan, but just a braindump of some things related to CI/CD that we might have to think of, please feel free to add any thoughts or suggestion directly to the description, or as comments.

We've already hit a few occasions when code changes to enable/improve one of the CSP implementations has broken one of the others, so it's definitely something we are going to want to think about sooner, rather than later to avoid regressions and/or lots of manual testing.

- The big obstacle (which I don't think we want to solve here, but will hopefully soon come out of some of the community release discussions) is where confidential-containers CI/CD will run in general. Up until now we've been running any CI and testing on the [kata-containers jenkins](http://jenkins.katacontainers.io/) as the jobs are all versions of existing kata ones, but I'm not sure if that will be the case for the cloud-api-adaptor code. I'm wondering if we can get GHA workflows for some of the build and test steps in the short term and avoid having to have dedicated servers?
- The different providers are a challenge as already have AWS and IBM Cloud implementations that would require access to accounts in order to test, which cost money. I'm not sure if we already have access to these anywhere that we can use? The other option we might want to take is use the libvirt support in open source and then leave the other provider testing to be done in house based on that?
- Work a proper build process. At the moment, the IBM Cloud provider implementation (the only one I've personally tried) has some [documentation](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/ibmcloud/README.md) available, but the process is still a PoC where the custom peer pod VM image is created on the test worker node and then pushed to IBM Cloud and the components required on the worker node to get peer-pods working (`cloud-api-adaptor`, `kata-shim` confidential-containerd fork of `containerd` are all built and installed directly on the worker). I think we probably want to separate out the build of the peer pod VM image (I think Pradipta has created https://github.com/confidential-containers/cloud-api-adaptor/issues/6 that covers the AWS equivalent of this?) and the `cloud-api-adaptor` binary as separate build artefacts?
- Work out a 'deployment' strategy. If we build the artefacts independently of the test environment we'll want to install them into a Kubernetes cluster (maybe using a single-node cluster like the Kata pipeline does?) I think this is something that we can use an extension of the coco-operator to do?

I'm sure there are other things that I've not thought of now and it's probably quite IBM Cloud centric, but @bpradipt and @wainersm please feel free to chip in your experience with CI/CD, the other providers and the operator

[...]

PR #362, opened 2022-11-18
WIP: tests/e2e: Introduce a e2e framework for testing

wainersm:
[WORK IN PROGRESS but I'd like to hear anyone comments/suggestions before moving on]

Cc @stevenhorsman @jtumber-ibm 

### What I'm trying to achieve...

... is to find a way to reused the existing code to run the *same* end-to-end tests regardless of cloud provider specific details.

@stevenhorsman  start the CI discussion on https://github.com/confidential-containers/cloud-api-adaptor/issues/16 . In https://github.com/confidential-containers/cloud-api-adaptor/issues/16#issuecomment-1292524184 I shared some snippets of code with the ideas i had in mind. Now, I am sending a code that although not complete, it is functional so that we could continue the discussion in something concrete. 

###  How I see this being used

I envision the e2e tests being executed on the developers' workstation with a simple `CLOUD_PROVIDER=awesome make test-e2e` command. He/she would need to export some environment variables with information to run the pipeline on the *awesome* provider; for example, the user and authorization token.

Likewise on CI, whether Github actions or Jenkins, the right environment variables are exported on the environment by the CI system.  Same `make test-e2e` command is executed.

On CI I expect that the framework will be able to install dependent software (e.g. `packer`, `terraform`...) as well. It is is a work to be done (and designed), see next section.

### What I did not addressed (yet)

  - Installation of dependent software. Examples: `packer` to build podvm images, `terraform` to provision VPC, and so on. I would like to install software only and only if running on CI context, or if the developer set a flag allow to perform installations on his/her machine.
  - Effectively run the tests once the test environment is completed (i.e. peers pods installed)
  - pipeline teardown. At the end of the execution the resources created are destroyed. However, I would like to have flag (`DEBUG=true`) that prevents the teardown from being executed; this way it keeps the environment intact for debugging failure or even allowing the developer use to implement features.   

### Why Python?

I wrote a draft this framework in bash (https://github.com/confidential-containers/cloud-api-adaptor/issues/16#issuecomment-1292524184) then I realized I was using object oriented programming (OOP) concepts that would be better implement with a native OOP language.

I thought re-write it in golang which is the peer pods language, but two things hold me back:
 1. Golang is not OOP
 2. Peer poods can end being re-written on Rust as has happened recently with Kata Containers. I didn't want to re-implement this framework although it would be cool to learn Rust :)

In a remote past I did some stuffs in Python, so finally I decided for using that language. I am not an expert on Python and I understand that the vast community members might not be either, but I don´t believe the pipeline implementers will need to learn deep Python to extend the framework; I do believe that the framework will be just a glue for scripts (bash, ansible...etc) that actually will perform the tasks. 

### Few more words

 - Don't pay attention to the small implementation details, I am more concerned about higher level details at this point.
 - Don't hesitate to criticize hard this proposal
 - This code can be threw to the garbage if we think it is too complex or not what we collectively agree on 'what is wanted' 
 - This PR depends on https://github.com/confidential-containers/cloud-api-adaptor/pull/345

### DIFF
--- a/libvirt/README.md
+++ b/libvirt/README.md
@@ -58,3 +58,229 @@ mkdir -p /opt/data-dir
 
 ```
 
+# Creating an end-to-end environment for testing and development
+
+In this section you will learn how to setup an environment in your local machine to run peer pods with
+the libvirt cloud API adaptor. Bear in mind that many different tools can be used to setup the environment
+and here we just make suggestions of tools that seems used by most of the peer pods developers.
+
+## Requirements
+
+You must have a Linux/KVM system with libvirt installed and the following tools:
+
+- docker (or podman-docker)
+- [kubectl](https://kubernetes.io/docs/reference/kubectl/)
+- [kcli](https://kcli.readthedocs.io/en/latest/)
+
+Assume that you have a 'default' network and storage pools created in libvirtd system instance (`qemu:///system`). However,
+if you have a different pool name then the scripts should be able to handle it properly.
+
+## Create the Kubernetes cluster
+
+Use the [`kcli_cluster.sh`](./kcli_cluster.sh) script to create a simple two VMs (one master and one worker) cluster
+with the kcli tool, as:
+
+```
+./kcli_cluster.sh create
+```
+
+With `kcli_cluster.sh` you can configure the libvirt network and storage pools that the cluster VMs will be created, among
+other parameters. Run `./kcli_cluster.sh -h` to see the help for further information.
+
+If everything goes well you will be able to see the cluster running after setting your Kubernetes config with:
+
+`export KUBECONFIG=$HOME/.kcli/clusters/peer-pods/auth/kubeconfig`
+
+For example, shown below:
+
+```
+$ kcli list kube
++-----------+---------+-----------+---------------------------------------+
+|  Cluster  |   Type  |    Plan   |                  Vms                  |
++-----------+---------+-----------+---------------------------------------+
+| peer-pods | generic | peer-pods | peer-pods-master-0,peer-pods-worker-0 |
++-----------+---------+-----------+---------------------------------------+
+$ kubectl get nodes
+NAME                 STATUS   ROLES                  AGE     VERSION
+peer-pods-master-0   Ready    control-plane,master   6m8s    v1.25.3
+peer-pods-worker-0   Ready    worker                 2m47s   v1.25.3
+```
+
+## Prepare the Pod VM volume
+
+In order to build the Pod VM without installing the build tools, you should use the `docker-build` Makefile target that will
+run the entire process inside a container.
+
+Ensure that the Kata Containers repository is cloned side-by-side with the cloud-api-provider's, as both repositories will be
+mounted as volumes in the container environment. Refer to [docs/DEVELOPMENT.md](/docs/DEVELOPMENT.md) for further
+information about setup of the sources.
+
+Then run:
+
+```
+$ cd image
+$ make docker-build
+```
+
+The qcow2 image file will be created at the `output` directory as shown below:
+
+```
+$ ls output/
+podvm-7da706d-dirty-amd64.qcow2
+```
+
+Next you will need to create a volume on libvirt's system storage and upload the image content. That volume is used by
+the cloud-api-adaptor program to instantiate a new Pod VM. Still on the `image` directory, you should run `make push` just
+like below:
+
+```
+$ make push
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+100  5394  100  5394    0     0   6421      0 --:--:-- --:--:-- --:--:--  6421
+virsh -c qemu:///system vol-create-as --pool default --name podvm-base.qcow2 \
+	--capacity 107374182400 --allocation 2361393152 --prealloc-metadata \
+	--format qcow2
+Vol podvm-base.qcow2 created
+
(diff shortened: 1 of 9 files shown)


bpradipt:
@wainersm thanks for starting the discussion on e2e test framework. Really helpful and much needed.

Getting to the specifics, personally I personally think it's better to re-use an e2e test framework that is common across the K8s ecosystem. 
And that is the kubernetes e2e test framework:
https://kubernetes.io/blog/2019/03/22/kubernetes-end-to-end-testing-for-everyone/
https://github.com/kubernetes-sigs/e2e-framework

Listing down the basis of my thinking:
1. Re-use or adapt existing test cases with minimal effort from other k8s projects. In addition to the CAA code, we also need to test CRD, webhook, potential CSI drivers.  Building on shared knowledge in k8s ecosystem will be faster and maintainable in the long run. 
2. Tap on to the shared knowledge of multiple developers in the k8s ecosystem who are experienced with the same framework.
3. Enable developers who are already experienced with the framework and test cases related to k8s ecosystem to contribute to this project. The contributions could be specific to test cases itself to start with.
4. Synergy between different pieces of the code (CAA, CRD, webhook, CSI etc), tests and  test frameworks.

bpradipt:
I missed explaining point 4 (synergy between different pieces... ).

What I meant was having the code, test cases and the test framework based on single language (golang in this case).
While technically it's not needed, the framework can be based on a different language. However choice of framework should be thought out well looking at our requirements as well as the framework's fitment, current usage in different projects, documentation, help availability etc.

stevenhorsman:
So my immediate thoughts is that it look decent and addresses the issues we currently need as discussed in the issue. I have done a bit of python hacking in my youth 😉, so that doesn't scare me either. That being said I think it is definitely worth a look into the Kubernetes e2e framework, even if go might not be as easy to get going with if there is already an existing technology and community we can build on.

wainersm:
> @wainersm thanks for starting the discussion on e2e test framework. Really helpful and much needed.
>

And thanks for the feedback @bpradipt :)
 
> Getting to the specifics, personally I personally think it's better to re-use an e2e test framework that is common across the K8s ecosystem. And that is the kubernetes e2e test framework: https://kubernetes.io/blog/2019/03/22/kubernetes-end-to-end-testing-for-everyone/ https://github.com/kubernetes-sigs/e2e-framework
>

I'm totally in favor of re-using existing e2e test frameworks. I wasn't aware of that **kubernetes-sigs/e2e-framework**; I always got the impression that every project develop its own framework (indeed, kubernetes' e2e_node and e2e_kubeadm have their own tools).

I spent the last days learning and playing with **kubernetes-sigs/e2e-framework**. It got some very interesting features that we could be using. However I'd like to see how other projects use it. Do you know of a reference project?

Here is a pilot of using the framework to test a simple peer pod can be created in libvirt (yes, this test is real, not fake):

```
$ CLOUD_PROVIDER=libvirt make test-e2e
go test -v -tags=libvirt -count=1 ./pkg/internal/testing/e2e
do setup
=== RUN   TestLibvirtCreateSimplePod
=== RUN   TestLibvirtCreateSimplePod/Simple_Peer_Pod
=== RUN   TestLibvirtCreateSimplePod/Simple_Peer_Pod/PodVM_is_created
--- PASS: TestLibvirtCreateSimplePod (55.09s)
    --- PASS: TestLibvirtCreateSimplePod/Simple_Peer_Pod (55.09s)
        --- PASS: TestLibvirtCreateSimplePod/Simple_Peer_Pod/PodVM_is_created (0.05s)
PASS
ok      github.com/confidential-containers/cloud-api-adaptor/pkg/internal/testing/e2e   55.111s
```

The framework extends Go's `testing.TestMain()` function, i.e., kind of build a test harness upon the standard `testing` package. Example of a customization, notice on the output above the ** do setup** message, it is from a method (`Setup()`) which run once before the tests execution. In theory we could run the pipeline I sent on this PR (but re-witten in Go) to setup the environment (build podvm, instantiate VPC...etc) although there is a chicken-and-eggs problem (don't want to extend the discussion here).     

Talking about the setup of the environment, that's IMHO the hardest part. **kubernetes-sigs/e2e-framework** provides the code to start a Kind cluster which is not very useful to us (maybe perhaps with the libvirt e2e). I still not sure how to connect the pre-test environment setup with the framework, and what other k8s Go packages could be used (ideas?).

> Listing down the basis of my thinking:
> 
>     1. Re-use or adapt existing test cases with minimal effort from other k8s projects. In addition to the CAA code, we also need to test CRD, webhook, potential CSI drivers.  Building on shared knowledge in k8s ecosystem will be faster and maintainable in the long run.
> 

Do you know if those test cases are exported Go packages? Or the re-use here is more like copy-paste-adapt the code?

>     2. Tap on to the shared knowledge of multiple developers in the k8s ecosystem who are experienced with the same framework.
> 
>     3. Enable developers who are already experienced with the framework and test cases related to k8s ecosystem to contribute to this project. The contributions could be specific to test cases itself to start with.
>

That would be awesome.
 
>     4. Synergy between different pieces of the code (CAA, CRD, webhook, CSI etc), tests and  test frameworks.

jtumber-ibm:
So are we thinking that each provider would include an implementation of the `CreateCluster` in the k8s e2e-framework? Could there be benefits this implementation being added as contributions back to the e2e project as mentioned in https://github.com/kubernetes-sigs/e2e-framework/issues/131 ?

I do prefer the go framework approach, just for reuse ability, but the python framework looks great too.

One question with the python framework could we split the `step_install_peerpods` in two. A `step_kustomize_peerpods` step, setting all the provider specific settings, in the overlay and then a `step_install_peerpods` could be common across the providers?

wainersm:
This PR was superseded by https://github.com/confidential-containers/cloud-api-adaptor/pull/386

PR #364, opened 2022-11-23
Allow vxlan-port to be set via kustomize

bpradipt:


huoqifeng (approved):
/lgtm
I did not try deploy caa via the container yet, but the code change looks good, thank you @bpradipt

snir911 (approved):
LGTM

PR #381, opened 2022-11-29
WIP:  inject passed auth.json credentials file to podvm

snir911:


install/yamls/ccruntime-peer-pods-in-pod.yaml
```diff
@@ -31,6 +31,9 @@ spec:
             name: peer-pods-cm
         command: ["/usr/local/bin/entrypoint.sh"]
         volumeMounts:
+        - name: auth-json
+          mountPath: "/root/containers/" # hardcoded
```
bpradipt, line 35: Should it be full path like `/root/containers/auth.json` ?

snir911: i thought so as well but then it was mounted at `/root/containers/auth.json/auth.json `:P

bpradipt (approved):
/lgtm
Thanks @snir911

snir911:
Thank you @stevenhorsman , i'm thinking about creating the offline_fs_kbc-resources.json file on the caa side and then pass it to the podvm to be injected to the right path, would it make sense? Can we assume passing the auth.json file this way will be done only for peer-pods users and not for CC users?
(as if user will pass the file and CC+offline_fs_kbc are used the existing offline_fs_kbc-resources.json file will be override by a file that contains only the "Credential" field provided by the user)

stevenhorsman:
> Thank you @stevenhorsman , i'm thinking about creating the offline_fs_kbc-resources.json file on the caa side and then pass it to the podvm to be injected to the right path, would it make sense?

Yes, I think that should work

> Can we assume passing the auth.json file this way will be done only for peer-pods users and not for CC users? (as if user will pass the file and CC+offline_fs_kbc are used the existing offline_fs_kbc-resources.json file will be override by a file that contains only the "Credential" field provided by the user)

@snir, Yes - as my understanding is that the credentials will be accessible on the worker node, so the Kubernetes admin can get them in the clear. This does bring up the question of whether we can provide a locked down mode that prevent this happening and an admin overriding the credentials in the image though. As a knee-jerk (bad) suggestion, could be make it so that if the `aa-offline_fs_kbc-resources.json` was provided in the peer pod already then we don't overwrite it and then that gives a bit extra safety? I think that would still work with your plans to enable one image to be used my multiple different credentials, but they can only get updated once on start up anyway? Then if we want to lock it down for a CC & peer pods scenario we'd pre-build in the `aa-offline_fs_kbc-resources.json` file and it can't be updated by a Kubernetes admin? Just a very quick idea, so let me know what the wholes are. \cc @bpradipt

snir911:
Yes, it make sense, not straight forward though, as it allows only to append to an existing file: https://cloudinit.readthedocs.io/en/latest/topics/modules.html#write-files
I think we'll have to run some sort of script on the podvm side, I'll try it out, thanks!

snir911:
I sent a new version which supports image-rs authentication, few notes:
* currently auth.json is converted to the kbc resources json format **before** file is passed to the podvm, we'll drop it later in favor of a image integrated script doing the conversion in the podvm (or some other method)
* I removed previous instructions (although it's still supported) in all providers but ibmcloud (as AFAIU caa in pod is still not used)
* would it make sense to have the offline kbc as default?

snir911:
> I'm not sure if the notes were open to comments, but here are a couple of questions/thoughts I have:

Sure ;) 

> So, it looks like you have removed the instructions we added for statically setting the credentials into the AWS image. I'm not sure if it is worth keeping for the CC case where we don't want to allow dynamic offline_fs_kbc resource updates?
> 
np, I'll leave it, i also thought about having it in separate doc file which is common to all providers, AFAIU this part is identical

> Sorry, I think I'm missing something obvious here - what is the 'caa' in pod that isn't used for ibm cloud?

Maybe i got it wrong but I understood running caa from within a pod was not tested yet with ibmcloud (i.e. process is running natively on the node), hence, passing to it auth.json as a secret won't work.

> Assuming this means adding `AA_KBC ?= "offline_fs_kbc"` in the `podvm/Makefile.inc`. We did think about this and having the attestation-agent built into the peer pod VM by default, but wanted to let the other cloud providers migrate over to use it and test it first, but I'm open to other suggestions.

I can vouch only for aws and libvirt :)

stevenhorsman:
> np, I'll leave it, i also thought about having it in separate doc file which is common to all providers, AFAIU this part is identical

Yes - as we are starting to merge stuff together (which is great), consolidated doc makes a lot of sense

> Maybe i got it wrong but I understood running caa from within a pod was not tested yet with ibmcloud (i.e. process is running natively on the node), hence, passing to it auth.json as a secret won't work.

Oh, are you talking about the cloud-api-adaptor process running as a container on the worker node as deployed by the operator? That makes sense now. It's a fair statement to say it's not fully tested - we'd had it running, but not merged that into the open-source code and updated our process to use the operator as the mode of deployment until we can have a standalone packer based s390x peer pod VM created, so we can keep our other method for now and then switch over as part of our doc update later.

snir911:
moved the instructions to a separate file and updated the copy-files.sh scripts

docs/registries-authentication.md
```diff
@@ -0,0 +1,29 @@
+## Container Image Registries Authentication
+
+To authenticate with private container image registry you are required to provide
+[registry authentication file](https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md) to your podvm in order
+to allow the image to be pulled directly.
+
+Registry authentication file can be provided either statically or at runtime.
```
bpradipt, line 7: @stevenhorsman anything to be mentioned specific to CC case and attestation ?

bpradipt (dismissed):
/lgtm
Thanks @snir911 
I just have a question for which I tagged @stevenhorsman.

ibmcloud/README.md
```diff
@@ -360,23 +360,7 @@ $ terraform apply
 >    ```
 
 ### Enabling Attestation agent and Authenticated Registry
-> **Prerequisites:**
```
Jordan9500, line 363: Please could we re-add the pre-requisites to the file, as we have to be ssh'd into the worker to get the authenticated changes to work.

pkg/adaptor/hypervisor/aws/service.go
```diff
@@ -173,6 +173,20 @@ func (s *hypervisorService) StartVM(ctx context.Context, req *pb.StartVMRequest)
 		},
 	}
 
+	if authJSON, err := os.ReadFile(cloudinit.DefaultAuthfileSrcPath); err == nil {
```
Jordan9500, line 176: nit: Should this be moved into a single file? OR is the plan to add this into the refactor steps being done in a different PR.

Jordan9500 (changes requested):
Just a couple of notes:

docs/registries-authentication.md
stevenhorsman: I don't think so. At the moment we don't have proper CC support and I wouldn't want to indicate that it was, so I think we'll update the doc once we've firmed that up.

ibmcloud/README.md
snir911: please TAL

stevenhorsman (approved):
I'm not able to run it as we're a bit behind on ibm cloud support with the operator, but based on the code and principles it LGTM. Thanks @snir911!

Jordan9500 (approved):
LGTM, Thanks @snir911!

bpradipt (approved):
/lgtm
Thanks @snir911

Issue #384, opened 2022-11-30
Add a builder to create a rhel podvm image


whitebrandy:
Have the option to create a rhel podvm image as long as a valid rhel cloud image is specified.

jtumber-ibm:
@bpradipt Is it worth looking at making the generation of qcow images a common task (no provider specific). Also relates to #372. 

Should providers like libvirt and ibmcloud (as we also generate the qcow image and then import rather than building the image using a base ibmcloud image) be able to share the code for creating the qcow image?

Will need to consider cross-platform for this too.

PR #349, opened 2022-11-11
all: introduce common image directory

jtumber-ibm:
- Adds Makefile which has the common code between providers
- Common files, copy-files.sh, services, etc

Each providers Makefile will be a child of their provider `aws/Makefile` vs `aws/image/Makefile`
Therefore I have updated the relative paths for this

Each provider can implement their own targets as there is variation (build, push, etc)

Fixes #314
Signed-off-by: James Tumber <james.tumber@ibm.com>

PR #385, opened 2022-11-30
Rhel packer template

whitebrandy:
v3: 
  Keep Makefile target common
  Still depends on #389 
v2: 
* Modified agent-protocol-forwarder.service for all service providers and moved it to #389 Unforunately the modification still shows up in this MR :/
Again, this MR depends on #389 now.
* Modified libvirt/image/Makefile to avoid duplicate code.
* Created separate dirs for rhel and ubuntu packer scripts.
* Added a new selinux_relabel.sh and removed copy-rhel-files.sh to be consistent and make it easy to adapt to #349 

An initial RHEL based packer template. Unfortunately, the user has to manually provide a url/local path to a cloud image before the builder can continue.

Fixes: #384 


libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
```diff
@@ -4,6 +4,9 @@ After=network.target
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
+Restart=on-failure
+RestartSec=5s
+
```
bpradipt, line 9: Is this needed ?
If it's must, we should also change the common files under `podvm`

libvirt/image/qemu-linux.pkr.hcl
```diff
@@ -0,0 +1,127 @@
...
+    destination = "~/misc-settings.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    inline = [
+      "sudo bash ~/misc-settings.sh"
+    ]
+  }
+}
```
bpradipt, line 127: I think it's better to have a separate packer file for centos/rhel

libvirt/image/Makefile
```diff
@@ -68,17 +69,42 @@ STATIC_LIB_BUILDER = ../../../kata-containers/ci/install_libseccomp.sh
...
 	echo ${UBUNTU_IMAGE_CHECKSUM}
 	rm -fr output
 	rm -f cloud-init.img
 	cloud-localds cloud-init.img userdata.cfg
 	mkdir -p toupload
-	packer build -var qemu_image_name=${IMAGE_FILE} \
+	packer build -var qemu_image_name=${UBUNTU_IMAGE_FILE} \
 		-var cloud_image_url=${UBUNTU_IMAGE_URL} \
-		-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM} .
+		-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM} -only=qemu.ubuntu .
```
bpradipt, line 107: how about still using the same target but using the template file based on provided option

```
packer build ...opts.. $TEMPLATE
```

Default TEMPLATE set to qemu-ubuntu.pkr.hcl

libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
whitebrandy: Yes, without this change, I ended up with a rhel image where agent-protocol-forwarder had exited with failure because of missing /peerpod/daemon.json. Can you please clarify which common files are you referring to ? I could add the restart clauses to other services but from testing, it doesn't look like they are required.

bpradipt: The common https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/podvm/files/etc
Further, I think this is a good change to make for all the providers. Can you move the "Restart" setting to a separate PR  for all the providers ?

whitebrandy: Yeah, I just saw the newer commit, will do. Thanks!

libvirt/image/qemu-linux.pkr.hcl
whitebrandy: @bpradipt  From my understanding, there are a few restrictions here. You can only specify a specific template file if your variables are also defined in the same file. So, everything in variable*.hcl has to move to both qemu-ubuntu*.hcl and qemu-rhel*.hcl.  The other option is to create separate directories for ubuntu and rhel and copy over the variables file to both. Having both builders in the same file seems to be the convenient option here (since variables are common) but please let me know what you think.

bpradipt: Is there a need to have different variable*.hcl file ? Can't we have a single variable file and instantiate the required values in the Makefile ? If not, then putting it in separate directories makes sense.

bpradipt: You can also explicitly specify the var file in the CLI - https://developer.hashicorp.com/packer/guides/hcl/variables
Otoh, having separate directories might be cleaner.  

It'll be good to discuss the different approaches in the related issue or over slack as the change impacts all the providers

whitebrandy: If you are referring to the "-var-file" switch, I think that is meant for variable assignments  not definitions.  Either way, tbh, I was just trying out with separate dirs (rhel/ubuntu) and it isn't that bad. The build just works without any modifications.

libvirt/image/Makefile
```diff
@@ -68,17 +68,38 @@ STATIC_LIB_BUILDER = ../../../kata-containers/ci/install_libseccomp.sh
...
+	-var cloud_image_checksum=${RHEL_IMAGE_CHECKSUM} rhel)
+
+rhel-build: rhel-opts $(IMAGE_FILE)
+
+ubuntu-opts:
+	$(eval OPTS := -var qemu_image_name=${IMAGE_FILE} \
+	-var cloud_image_url=${UBUNTU_IMAGE_URL} \
+	-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM} ubuntu)
+
+ubuntu-build: ubuntu-opts $(IMAGE_FILE)
```
bpradipt, line 95: @whitebrandy I feel this change will make it harder to adopt the common Makefile/files that was introduced by @jtumber-ibm to avoid duplication.

Can we avoid introducing new build targets. How about adding a variable like PODVM_DISTRO  and based on it's value (ubuntu/rhel/centos/debian etc) set the required options for packer build. 

The packer target is modified to use the options (as you did it in this PR)

aws/image/files/etc/systemd/system/agent-protocol-forwarder.service
```diff
@@ -4,6 +4,9 @@ After=network.target
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
+Restart=on-failure
+RestartSec=5s
+
```
bpradipt, line 9: The new issue created by you https://github.com/confidential-containers/cloud-api-adaptor/issues/388 provides the reasoning behind the failure.
Let's remove this change from this PR and tackle it as part of #388

bpradipt (approved):
/lgtm
Thanks @whitebrandy

caavery (approved):
lgtm

Although I am by no means an expert

Issue #388, opened 2022-11-30
Restart agent-protocol-forwarder on failure


whitebrandy:
agent-protocol-forwarder in the podvm image might fail if it races with cloudinit configuration to complete. Although not common, I hit this consistently with a rhel podvm image.

PR #389, opened 2022-11-30
podvm: restart agent-protocol-forwarder on failure

whitebrandy:
RHEL 9 podvm image is encountering an issue where cloudconfig init hasn't completed before agent-protocol-forwarder starts and it fails because it can't find /peerpods/daemon.json. Restart the service on failure.

Fixes: #388

Suggested-by: Pradipta Banerjee <pradipta.banerjee@gmail.com>

### DIFF
--- a/aws/image/files/etc/systemd/system/agent-protocol-forwarder.service
+++ b/aws/image/files/etc/systemd/system/agent-protocol-forwarder.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=Agent Protocol Forwarder
-After=network.target
+After=cloud-init.target
+Wants=cloud-init.target
+DefaultDependencies=no
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
--- a/azure/image/files/etc/systemd/system/agent-protocol-forwarder.service
+++ b/azure/image/files/etc/systemd/system/agent-protocol-forwarder.service
@@ -1,6 +1,9 @@
 [Unit]
 Description=Agent Protocol Forwarder
-After=network.target
+After=cloud-init.target
+Wants=cloud-init.target
+DefaultDependencies=no
+
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
--- a/libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
+++ b/libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
@@ -1,6 +1,9 @@
 [Unit]
 Description=Agent Protocol Forwarder
-After=network.target
+After=cloud-init.target
+Wants=cloud-init.target
+DefaultDependencies=no
+
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
--- a/podvm/files/etc/systemd/system/agent-protocol-forwarder.service
+++ b/podvm/files/etc/systemd/system/agent-protocol-forwarder.service
@@ -1,6 +1,9 @@
 [Unit]
 Description=Agent Protocol Forwarder
-After=network.target
+After=cloud-init.target
+Wants=cloud-init.target
+DefaultDependencies=no
+
 
 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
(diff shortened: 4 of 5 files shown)


yoheiueda:
Thank you very much for investigating this issue.

Is it possible to introduce an appropriate dependency to a cloud-init boot stage in the agent-protocol-forwarder service definition, instead of restarting the service?

https://cloudinit.readthedocs.io/en/latest/topics/boot.html

I am not sure which target we should use, and how to write such dependency in the service definition file.

`/etc/cloud/cloud.cfg` in Ubuntu contains the following settings. This indicates that `write-files` is called at the `init` stage.

```
# The modules that run in the 'init' stage
cloud_init_modules:
 - migrator
 - seed_random
 - bootcmd
 - write-files
 - growpart
 - resizefs
 - disk_setup
 - mounts
 - set_hostname
 - update_hostname
 - update_etc_hosts
 - ca-certs
 - rsyslog
 - users-groups
 - ssh
```

https://cloudinit.readthedocs.io/en/latest/topics/modules.html#write-files

The attached SVG file is generated by `systemd-analyze plot` at a Ubuntu pod VM. 

 Is there any difference between Ubuntu and RHEL?

systemd

whitebrandy:
> Thank you very much for investigating this issue.
...
Hmm.. yeah, the restart method is a little ugly.

Thanks for the handy command. I ran it in the rhel podvm as well. 
rhelplot

I can't claim to understand systemd well enough but one thing that sticks out is that in the ubuntu plot, cloud-init.service runs a little before agent-protocol-forwarder in the timeline. However, with rhel, it's after. 

So, something like this seems to work:
--- a/libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
+++ b/libvirt/image/files/etc/systemd/system/agent-protocol-forwarder.service
@@ -1,11 +1,9 @@
 [Unit]
 Description=Agent Protocol Forwarder
-After=network.target
+After=network.target cloud-final.service

 [Service]
 ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock
-Restart=on-failure
-RestartSec=5s

The problem I see is that both cloud-init.target and cloud-final.service have RemainAfterExit=yes which makes it a little scary to use the After clause. But since we have WantedBy=multi-user.target later, it works but I am not sure if multi-user.target exists for all distros. 

The only thing I would add to be on the safe side is a DefaultDependencies=no.

Coming to the error with agent-protocol-forwarder.service, it seems to be failing on the path to daemon.json:
Dec 01 01:07:20 podvm-nginx-c9902f85 systemd[1]: Started Agent Protocol Forwarder.
Dec 01 01:07:20 podvm-nginx-c9902f85 agent-protocol-forwarder[744]: /usr/local/bin/agent-protocol-forwarder: failed to open /peerpod/daemon.json: open /peerpod/daemon.json: no such file or directory
Dec 01 01:07:20 podvm-nginx-c9902f85 systemd[1]: agent-protocol-forwarder.service: Main process exited, code=exited, status=1/FAILURE

Thoughts or any other suggestions ?

yoheiueda:
I am not an export of systemd, but I found the following description in `/usr/lib/systemd/system/cloud-init.target`.

```
# cloud-init.target is a  synchronization point when all cloud-init's initial
# system configuration tasks have completed. To order a service after cloud-init
# is done, add the directives as applicable:
#  After=cloud-init.target and Wants=cloud-init.target
[Unit]
Description=Cloud-init target
After=multi-user.target
```

So, I think we can add the following lines instead of `After=network.target cloud-final.service`

```
After=cloud-init.target
Wants=cloud-init.target
```

The file `daemon.json` is generated by `cloud-init` using `write-file`. `agent-protocol-forwarder` loads this file, so it needs to be executed after `cloud-init` completes generating the file.

https://github.com/confidential-containers/cloud-api-adaptor/blob/7acbf62939405c04fa98ac4c4904209bdaf9de19/pkg/adaptor/hypervisor/ibmcloud/service.go#L179-L188

whitebrandy:
> So, I think we can add the following lines instead of `After=network.target cloud-final.service`
> 
> ```
> After=cloud-init.target
> Wants=cloud-init.target
> ```
> 
Unfortunately, this does not work and ends up in an ordering cycle on RHEL.
```
○ agent-protocol-forwarder.service - Agent Protocol Forwarder
     Loaded: loaded (/etc/systemd/system/agent-protocol-forwarder.service; enabled; vendor preset: disabled)
     Active: inactive (dead)

Dec 02 19:44:07 podvm systemd[1]: multi-user.target: Found ordering cycle on agent-protocol-forwarder.service/start
Dec 02 19:44:07 podvm systemd[1]: multi-user.target: Found dependency on cloud-init.target/start
Dec 02 19:44:07 podvm systemd[1]: multi-user.target: Found dependency on multi-user.target/start
Dec 02 19:44:07 podvm systemd[1]: multi-user.target: Job agent-protocol-forwarder.service/start deleted to break ordering cycle starting with multi-user.target/start
```
From the plot above, it looks like the ordering of cloud-init.target wrt multi-user.target is not the same for Ubuntu and RHEL. I suggest we disable implicit dependencies by specifying DefaultDependencies=no as I did in this code update but I am open to other ideas. Thanks again for the review.

bpradipt (approved):
/lgtm
Thanks @whitebrandy

Issue #397, opened 2022-12-02
RHEL podvm image for vsphere


whitebrandy:
Provide a packer template to create a rhel podvm image. We can just reuse the packer script for ubuntu because:
1. autostart/kickstart differences
2. RHEL subscription

PR #398, opened 2022-12-02
RHEL vsphere template

whitebrandy:
v2: 
Based on testing info from @caavery where she discovered there might be a race between packer calling shutdown and vmtoolsd initializing properly.
 - Remove subscribe.sh and install required packages from the kickstart since we already have a dvd
 - Make sure cloud-init isn't started during Packer's provisioning stage because first run of cloud-init resets ssh keys and packer is never able to login 
 
RHEL podvm image from an installable iso.

Please update vsphere/image/rhel/scripts/subscribe.sh with subscription info or else package installation will fail.
This means that open-vm-tools will never get installed vm ip discovery would never work.

Fixes: #397 


whitebrandy:
@caavery please take a look when you can!

caavery:
I approved the commit and somehow github thinks I asked myself to review it again. Not sure how to delete that.

bpradipt (approved):
/lgtm
Thanks @whitebrandy

Issue #403, opened 2022-12-05
Allow for specifying cross-build of podvm images


jtumber-ibm:
This will require being able to specify a different target architecture to the host for the agent-forwarder, kata-agent (and any other binaries that maybe included) and qcow image itself.

PR #396, opened 2022-12-02
Create qcow2 image for all providers

bpradipt:
This PR adds support for creating qcow2 image for all providers. Further it introduces container builds for Pod VM image.
The builder image concept is taken from the PR by @wainersm

PR #404, opened 2022-12-05
all: Add Cross-build options for image binaries

jtumber-ibm:
- Add GOOS/GOARCH setting
- Add RUST cross-compile options (CC/ARCH)

Fixes: #403

Holding until #396 has been merged so can do updates for building cross arch podvm images too.

### DIFF
--- a/podvm/Makefile.inc
+++ b/podvm/Makefile.inc
@@ -13,7 +13,19 @@ UMOCI_VERSION      = 0.4.7
 
 IMAGE_PREFIX := podvm
 
-ARCH := $(subst x86_64,amd64,$(shell uname -m))
+HOST_ARCH   := $(shell uname -m)
+ARCH        ?= $(HOST_ARCH)
+DEB_ARCH    := $(subst x86_64,amd64,$(ARCH))
+LIBC        := $(if $(findstring s390x,$(ARCH)),gnu,musl)
+RUST_TARGET := $(ARCH)-unknown-linux-$(LIBC)
+
+ifneq ($(HOST_ARCH),$(ARCH))
+    ifeq ($(CC),cc)
+		CC := $(ARCH)-linux-$(LIBC)-gcc
+        $(warning A foreign ARCH was passed, but no CC alternative. Using $(CC) as best guess)
+	endif
+RUST_FLAGS := CC=$(CC) ARCH=$(ARCH)
+endif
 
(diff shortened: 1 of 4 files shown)


podvm/Makefile.inc
```diff
@@ -26,10 +38,10 @@ COMMIT := $(shell	commit=$$(git describe --match '' --dirty --always) && \
...
 endif
-IMAGE_NAME := $(IMAGE_PREFIX)-$(COMMIT)-$(ARCH)
+IMAGE_NAME := $(IMAGE_PREFIX)-$(COMMIT)-$(DEB_ARCH)
 endif
 
-IMAGE_SUFFIX ?= 
+IMAGE_SUFFIX ?=
 IMAGE_FILE = $(IMAGE_NAME)$(IMAGE_SUFFIX)
 
 LIBC_ARCH := $(shell uname -m)
```
Jordan9500, line 47: We should be able to remove these lines as `LIBC` is now defined above and uses `ARCH` instead.

Jordan9500 (changes requested):

stevenhorsman (approved):
LGTM.
This was the (slightly convoluted) testing I did to check this worked:
- Create a new x86 ibmcloud cluster with the IBM e2e process (in order to have a machine with all the required pre-reqs installed and configured)
- Logged into the x86 worker node
- Installed the extra s390x tools needed:
```
rustup target add s390x-unknown-linux-gnu
sudo apt-get install gcc-s390x-linux-gnu
```
- removed the x86 versions of the libseccomp caches built into the kata-agent directory with:
```
pushd  ~/kata-containers/src/agent
make clean
popd
```
(this can be done more surgically to remove the specific cache(s) for libsecomp)
- Deleted the CAA x86 version of the libsecomp binaries so the makefile can built them with s390x
```
cd cloud-api-adaptor/ibmcloud/image/
rm -rf staticlib/
```
- Built the binaries in s390x:
```
ARCH=s390x CLOUD_PROVIDER=ibmcloud make binaries
```
This looked like it all worked.

Whilst this looks a little messy with the libsecomp binaries and caches being deleted, this is just as I'm re-using a system that has previously build it for a different architecture. If it was built in a isolated environment like a container, or a automation controlled VM this shouldn't be an issue, so I'm happy with this work.
Thanks @jtumber-ibm for the PR and help with debugging!

Jordan9500 (approved):
LGTM! Following the same steps done by Steven above I was able to get a successful build going. 

Thank you @jtumber-ibm :)

PR #405, opened 2022-12-06
Add setup steps for the attestation agent to the podvm makefile

Jordan9500:
- Add attestation agent if AA_KBC has been set in the makefile
- Update ibmcloud docs to include steps for setting up the attestation agent with authenticated registry support
- Update AWS docs to remove the skopeo specific steps and replace it with using the attestation agent with authenticated registry support

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -20,10 +20,20 @@ create create new VPC with public internet access and set also the following env
 export VPC_ID="REPLACE_ME"
 export SUBNET_ID="REPLACE_ME"
 ```
-
+   
 - Create a custom AMI based on Ubuntu 20.04 having kata-agent and other dependencies
-        - **NOTE:** If authentication against container image registries is required copy your JSON format credentials
-        file to [podvm/files/](../podvm/files/)auth.json and add also `USE_SKOPEO=true` before `make image`
+- **Setting up authenticated registry support:** 
+  - If authentication against container image registries is required base64 encode your JSON format credentials file (`auth.json`): `cat auth.json | base64 -w 0`
+  - Then export the base64 encoded version of the file: `export AUTHFILE=<base64 encoded auth.json file>`
+  - Create the necessary resource JSON using the above exported variable in the [podvm/files/etc/](../podvm/files/etc/) directory: 
+  ```
+  cat <<EOF | tee aa-offline_fs_kbc-resources.json
+  {
+    "Credential": "${AUTHFILE}"
+  } 
+  EOF
+  ```
+  - Add `AA_KBC="offline_fs_kbc"` prior to the make image step.
 ```
 cd image
 CLOUD_PROVIDER=aws make image
--- a/ibmcloud/README.md
+++ b/ibmcloud/README.md
@@ -359,6 +359,44 @@ $ terraform apply
 >    $ ibmcloud is images --visibility=private
 >    ```
 
+### Enabling Attestation agent and Authenticated Registry
+> **Prerequisites:**
+> - An ibmcloud worker node using the cloud-api-adaptor
+> - A `auth.json` file with your credentials in, based off: (containers-auth.json.5.md)[https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md]
+> - SSH'd into the worker node: `ssh root@floating-ip-of-worker-node`
+
+**Setting up the attestation agent:**
+- `cd ~/cloud-api-adaptor/podvm/files/etc`
+- Base64 your `auth.json`, this can be done by doing `cat auth.json | base64 -w 0`
+- Export the base64 encoded file `export AUTHFILE=<base64-encoded-auth.json>`
+- Create and Add the base64 encoded auth file into the `aa-offline_fs_kbc-resources.json` like so:
+```
+cat <<EOF | tee aa-offline_fs_kbc-resources.json
+{
+  "Credential": "${AUTHFILE}"
+} 
+EOF
+```
+
+**Building the image:**
+- `cd ~/cloud-api-adaptor/ibmcloud/image`
+- Export these variables:
+```
+export CLOUD_PROVIDER=ibmcloud
+export IMAGE_NAME=<An image name with arch included e.g. image-amd64>
+export IBMCLOUD_COS_REGION=<the region your IKS cluster and zVSI is running on>
+export IBMCLOUD_VPC_REGION=$IBMCLOUD_COS_REGION
+export IBMCLOUD_VPC_NAME=<your generated vpc name from the terraform>
+export IBMCLOUD_VPC_SUBNET_NAME=<the linked subnet name to the above vpc>
+export IBMCLOUD_COS_SERVICE_ENDPOINT="https://s3.${IBMCLOUD_COS_REGION}.cloud-object-storage.appdomain.cloud"
+export IBMCLOUD_COS_SERVICE_INSTANCE=<your cos instance>
+export IBMCLOUD_COS_BUCKET=<your cos bucket in the region your cluster runs in>
+export IBMCLOUD_API_KEY=<your ibmcloud apikey>
+export IBMCLOUD_API_ENDPOINT="https://cloud.ibm.com"
+export GOPATH=/root/go
+```
+- run `AA_KBC="offline_fs_kbc" make push` 
+> **Note:** The image ID can be found at the end of the logs when the make steps have completed. This can then be substituted for the image id in the start cloud api adaptor steps below.
 
 ## Install custom Kata shim
 
--- a/podvm/Makefile.inc
+++ b/podvm/Makefile.inc
@@ -32,17 +32,26 @@ endif
 IMAGE_SUFFIX ?= 
 IMAGE_FILE = $(IMAGE_NAME)$(IMAGE_SUFFIX)
 
+LIBC_ARCH := $(shell uname -m)
+LIBC := $(if $(findstring s390x,$(LIBC_ARCH)),gnu,musl)
 
 AGENT_PROTOCOL_FORWARDER = $(FILES_DIR)/usr/local/bin/agent-protocol-forwarder
 KATA_AGENT    = $(FILES_DIR)/usr/local/bin/kata-agent
 PAUSE         = $(FILES_DIR)/$(PAUSE_BUNDLE)/rootfs/pause
 SKOPEO    = $(FILES_DIR)/usr/bin/skopeo
 UMOCI     = $(FILES_DIR)/usr/local/bin/umoci
+ATTESTATION_AGENT = $(FILES_DIR)/usr/local/bin/attestation-agent
 
 BINARIES = $(AGENT_PROTOCOL_FORWARDER) $(KATA_AGENT) $(PAUSE)
 
 ifdef USE_SKOPEO
 BINARIES += $(SKOPEO) $(UMOCI)
+endif
+
+KBC_URI ?= "null"
+ifdef AA_KBC
+BINARIES += $(ATTESTATION_AGENT)
+$(shell sed -i "s/\(aa_kbc_params = \)\"[^\"]*\"/\1\"${AA_KBC}::${KBC_URI}\"/g" $(FILES_DIR)/etc/agent-config.toml)
 endif
 
 AGENT_PROTOCOL_FORWARDER_SRC = ../..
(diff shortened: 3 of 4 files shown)


aws/README.md
```diff
@@ -20,10 +20,20 @@ create create new VPC with public internet access and set also the following env
 export VPC_ID="REPLACE_ME"
 export SUBNET_ID="REPLACE_ME"
 ```
-
+   
 - Create a custom AMI based on Ubuntu 20.04 having kata-agent and other dependencies
-        - **NOTE:** If authentication against container image registries is required copy your JSON format credentials
-        file to [podvm/files/](../podvm/files/)auth.json and add also `USE_SKOPEO=true` before `make image`
+- **Setting up authenticated registry support:** 
+  - If authentication against container image registries is required copy your JSON format credentials file to [podvm/files/etc/](../podvm/files/etc/)auth.json
```
snir911, line 26: it seems podvm/copy-files.sh doesn't copy this file, isn't it? i guess i missing something

ibmcloud/README.md
```diff
@@ -359,6 +359,44 @@ $ terraform apply
...
 
+### Enabling Attestation agent and Authenticated Registry
+> **Prerequisites:**
+> - An ibmcloud worker node using the cloud-api-adaptor
+> - A `auth.json` file with your credentials in, based off: (containers-auth.json.5.md)[https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md]
+> - SSH'd into the worker node: `ssh root@floating-ip-of-worker-node`
+
+**Setting up the attestation agent:**
+- `cd ~/cloud-api-adaptor/podvm/files/etc`
+- Base64 your `auth.json`, this can be done by doing `cat auth.json | base64 -w 0`
```
snir911, line 370: Hi, it seems podvm/copy-files.sh doesn't copy this file, does it? maybe i missed something..

aws/README.md
snir911, line 26: Hi, `files/etc/`? it seems copy-files.sh doesn't copy the file from this location, does it? maybe i missed something..

stevenhorsman: Hi Snir, as discussed earlier the auth.json file isn't needed to be copied into the peer pod vm image, so it doesn't need to go into this exact place, it was just a suggestion for simplicity, but @Jordan9500 maybe we should clarify that, or move the instructions so it is clear that `aa-offline_fs_kbc-resources.json` goes into `podvm/files/etc/`.

Jordan9500: Hi Snir, 
I have updated the lines to better reflect that the auth.json file is no longer required to be inside the `podvm/files/` directory. And made it clear the offline kbc file should be in the `podvm/files/etc/` location.

snir911: The aws instructions are currently explains how to set the auth.json file so that it will be used by skopeo in the podvm, AFAIU as long as it is under `podvm/files/etc/ ` it will not be copied by copy-files.sh to the path expected by skopeo.
I think you can just remove this note and mention just to use the user auth.json file, instructions are expected to be changed soon anyway

stevenhorsman: > I think you can just remove this note and mention just to use the user auth.json file, instructions are expected to be changed soon anyway

Hey Snir, sorry, I'm not sure I understand this bit. This PR changes the instructions from using skopeo, to image-rs (with the attestation-agent), so the location of `auth.json` isn't relevant as we won't be using it directly in the peer podvm image any more?

snir911: right, sorry, i misread, i thought it implies the location must be on this path, LGTM

podvm/files/etc/agent-config.toml
```diff
@@ -5,6 +5,10 @@ enable_signature_verification=false
 # When using the agent-config.toml the KATA_AGENT_SERVER_ADDR env var seems to be ignored, so set it here
 server_addr="unix:///run/kata-containers/agent.sock"
 
+# This field sets up the KBC that attestation agent uses
+# This is replaced in the makefile steps so do not set it manually
+aa_kbc_params = ""
```
huoqifeng, line 10: Is it reasonable to use `offline_fs_kbc: null` as default value here as this is the most common use case today? We can replace it via real settings from makefile and podvm build scripts.

aws/README.md
```diff
@@ -20,10 +20,20 @@ create create new VPC with public internet access and set also the following env
...
 ```
-
+   
 - Create a custom AMI based on Ubuntu 20.04 having kata-agent and other dependencies
-        - **NOTE:** If authentication against container image registries is required copy your JSON format credentials
-        file to [podvm/files/](../podvm/files/)auth.json and add also `USE_SKOPEO=true` before `make image`
+- **Setting up authenticated registry support:** 
+  - If authentication against container image registries is required copy your JSON format credentials file to [podvm/files/etc/](../podvm/files/etc/)auth.json 
+  - Then export the base64 encoded version of the file: ```export AUTHFILE=`cat auth.json | base64 -w 0` ```
+  - Create the necassary resource JSON using the above exported variable in the [podvm/files/etc/](../podvm/files/etc/) directory:
```
bpradipt, line 28: nit: necassary -> necessary

podvm/Makefile.inc
```diff
@@ -69,6 +78,10 @@ STATIC_LIBSECCOMP_BUILDER = ../../../kata-containers/ci/install_libseccomp.sh
 STATIC_LIBSECCOMP_DIR     = $(abspath staticlib)
 STATIC_LIBSECCOMP         = $(STATIC_LIBSECCOMP_DIR)/kata-libseccomp/lib/libseccomp.a
 
+ATTESTATION_AGENT_SRC = ../../../attestation-agent
+ATTESTATION_AGENT_REPO = https://github.com/confidential-containers/attestation-agent
+ATTESTATION_AGENT_BRANCH = main
```
bpradipt, line 83: Does it make sense to use commit hash or tag instead of branch name ?

podvm/files/etc/agent-config.toml
stevenhorsman: Hey QiFeng, so I think the problem with this is that we aren't (currently) always building the attestation-agent into the peer pod vm, so if we default `aa_kbc_params = "offline_fs_kbc: null"` in the agent-config.toml then the kata-agent would try and launch the binary which might not exist and then fail. That's why I suggested leaving it blank for now, but I'm open to other suggestions.

podvm/Makefile.inc
stevenhorsman: Good question. I'm a little bit mixed on it at the moment.
- Pinning the attestation-agent makes sense to increase stability
- IIRC we are currently pulling in the latest (`CCv0`) version of the kata-agent, so if we pin one part and not the other then we have a chance of going out of sync and breaking
I guess the best solution would be to read the version of attestation-agent that our version of kata-containers uses from versions.yaml and use that, but that's significantly more involved, so maybe better as a separate PR?
I'm happy to discuss if `main` of a pinnned tag/commit is the best default in the short term though.

bpradipt: Ideally we should update the Makefile.inc to use the right commits or tags for all the sources. Unfortunately we don't do that today. And the latest from the default branch is downloaded. 
I think it's fine to pin the versions for all the source trees that we use in a separate PR. 
May be you can even remove the use of `main` branch for now since it's the default for AA. Will leave it to you. It's not a blocker to move ahead with this PR.

podvm/files/etc/agent-config.toml
huoqifeng: Thanks Steve, I found the algorithm here https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/image_rpc.rs#L101, I'm a little confused, why don't we use existing of `AA_PATH` as a condiftion but to combine `AA_PATH` and `aa_kbc_params` when launch attestation-agent? Is there a case both skopeo and attestation-agent binaries provided in the guest?

podvm/Makefile.inc
```diff
@@ -32,19 +32,28 @@ endif
...
 BINARIES = $(AGENT_PROTOCOL_FORWARDER) $(KATA_AGENT) $(PAUSE)
 
 ifdef USE_SKOPEO
 BINARIES += $(SKOPEO) $(UMOCI)
 endif
 
+KBC_URI ?= "null"
+ifdef AA_KBC
+BINARIES += $(ATTESTATION_AGENT)
+$(shell sed -i "s/\(aa_kbc_params = \)\"[^\"]*\"/\1\"${AA_KBC}::${KBC_URI}\"/g" $(FILES_DIR)/etc/agent-config.toml)
```
huoqifeng, line 54: Is it `\"${AA_KBC}::${KBC_URI}\"` or `\"${AA_KBC}:${KBC_URI}\"`?

Jordan9500: It is: ```\"${AA_KBC}::${KBC_URI}\"```

podvm/files/etc/agent-config.toml
stevenhorsman: Sorry to reply to myself, but I think the attestation-agent launch decision might be more interesting than I covered above, and I missed some thing, so my brain has been turning over the problem more, and it's worth me dumping this to 'paper' before making any decisions:

*Only check `aa_kbc_params` isn't empty*
This is the current case. I could lead to a scenario (particularly if we ever want to use image offload for non-confidential cases such as peer pods avoid nested virtualisation issues, or people first playing with CoCo with no attestation set) that the `aa_kbc_params` is configured, but the attestation-agent binary is not in the guest image.
This would result in an error on image pull (that may or may not be understandable to a user). There is an argument that this isn't great, but also one that says if a user thinks they are using an attestation-agent-kbc but aren't then we'd want to fail.

*Just check the presence of the attestation-agent in the guest image*
This breaks down into a couple of 'mis-configured' use cases:
- *The attestation-agent binary is in the guest image, but the `aa_kbc_param` hasn't been configured* - I don't think this will currently break anything as we've added check for `aa_kbc_param` before using the attestation-agent features, but in future this could be a trap that a dev falls into and tries to use the `attestation_agent_started` setting for something. I'm also not sure if this option could confuse the image-rs logic.
- *The attestation-agent binary isn't in the guest image, but the `aa_kbc_param` has been configured* - I think this is the worse case as a user has configured to use the a kbc, but the attestation-agent won't be started. Best case scenario the user thinks they have more security than they do, worst case (and I think the current situation with the code base) is things won't work as they will be configured to use attestation-agent config, but the binary isn't there

*Check the presence of the attestation-agent in the guest image and for offline_fs_kbc being set*
This breaks down into the following use cases:
- *The attestation-agent binary is in the guest image, and the `aa_kbc_param` is configured* - This is the good case, so we'd continue with no issues
- *The attestation-agent binary is in the guest image, but the `aa_kbc_param` hasn't been configured* - We'd not start the attestation-agent here, which I think is also a fine decision
- *The attestation-agent binary isn't in the guest image, but the `aa_kbc_param` has been configured* - This is the interesting case to consider here. Under this case we wouldn't start the attestation-agent, which would protect against the problem we would have in the current code base where the call to start the attestation-agent binary fails here, but there is a question of what the correct behaviour should be under this scenario. Do we want to proceed with no attestation-agent running (and end up with the later code that reads for config `aa_kbc_param` failing, should we add extra protection to those uses of `aa_kbc_param` to check the attestation-agent service is running (and risk ignoring the user thinking they have more security than they do, or should we bail out with a nice user message that this isn't a valid scenario?

I think I'm leaning towards keeping the check of just `aa_kbc_params` isn't empty before running the `init_attestation_agent`, but then within `init_attestation_agent` doing an up-front check for the binary and erroring in a nice way if it's not in the guest.

I'd love to hear others thoughts though as I might be overthinking this!

aws/README.md
Jordan9500: Fixed

podvm/files/etc/agent-config.toml
huoqifeng: >Yes - skopeo support for decrypting images relies on the attestation-agent to get the key, but we want to remove skopeo soon, so that shouldn't be a problem.

Thanks @stevenhorsman for explain the case!

>Do you think it's fair for us to proceed with this PR as is at the moment, or do you think we should make that kata-agent change and then think about defaulting to offline_fs_kbc in the agent-config.toml?

Yeah, I think it's perfectly OK to proceed with this PR and change agent-config.toml later. Thanks!

huoqifeng (dismissed):
/lgtm

podvm/Makefile.inc
Jordan9500: I have removed the branch logic from the makefile.

stevenhorsman: Yep, I think that maybe removing the branch selection logic is a good idea for now as it isn't usable for selecting a commit, but later we can/should work on version pinning as a more holistic solution for the kata-containers and attestation-agent source we are pulling in and whether the this common `Makefile.inc` should have the ability to override the other places we checkout repos.

huoqifeng (approved):
/lgtm thanks @Jordan9500

Issue #406, opened 2022-12-07
ibmcloud: Secure Execution Support for Peer Pods on IBM Cloud VPC


liudalibj:
Update the ibmcloud image build script to generate an IBM Secure Execution image for peer-pod

<img width="971" alt="Screenshot 2022-12-07 at 16 41 41" src="https://user-images.githubusercontent.com/39799378/206130190-aef1d37d-8f07-4d11-9a17-8b4d9c650c75.png">

liudalibj:
obtain the host key via https://www.ibm.com/docs/en/linux-on-systems?topic=execution-obtain-host-key-document

PR #407, opened 2022-12-07
ibmcloud: Secure Execution Support for Peer Pods on IBM Cloud VPC

liudalibj:
- add document for IBM Secure Execution support
- update build.sh script to generate an IBM Secure Execution image
- handle multiple host keys to generate SE image
- using LUKS encrypted root partition

fixes: #406

Signed-off-by: Da Li Liu <liudali@cn.ibm.com>
Co-authored-by: leilibj <cdlleili@cn.ibm.com>

### DIFF
--- a/ibmcloud/SECURE_EXECUTION.md
+++ b/ibmcloud/SECURE_EXECUTION.md
@@ -1,0 +1,150 @@
+# Secure Execution Support for Peer Pods on IBM Cloud VPC
+
+IBM Cloud VPC already supports to create Hyper Protect Virtual Server on LinuxONE. And Confidential Computing is enabled by using the [IBM Secure Execution](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-introduction) for Linux technology. For more information, please refer to [Confidential computing with LinuxONE](https://cloud.ibm.com/docs/vpc?topic=vpc-about-se)
+
+To support Secure Execution for Peer Pods on IBM Cloud VPC, we need build a Secure Execution enabled custom VM image for pod VMs, to replace the non-SE custom image. This document describes how to build a SE enabled custom image, based on the existing [image build script](./image/build.sh) for IBM Cloud VPC.
+
+> **Note**: In [the document](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-secure-workload) describe the details about execution secure workload, you can go through the document to get a simplified view about how your workload is protected.
+
+## Set up a demo environment without Secure Execution on your development machine
+
+Follow the [README.md](./README.md) to setup a demo environment on IBM Cloud VPC. We need create a **s390x** architecture based cluster, so please include the following two lines in the `terraform.tfvars` file:
+```
+instance_profile_name = "bz2-2x8"
+image_name = "ibm-ubuntu-18-04-1-minimal-s390x-3"
+```
+
+## Prepare to build SE enabled image
+
+1. **Build genprotimg on worker node**
+
+`genprotimg` is used to build an encrypted boot record from a given kernel, initial RAM disk, parameters, and public host-key document.
+
+The worker node is using Ubuntu 18.04, so we need build genprotimg from the source on GitHub:
+```bash
+ssh root@ip-of-your-worker-node
+git clone https://github.com/ibm-s390-linux/s390-tools.git
+cd s390-tools/genprotimg/
+apt install -y libglib2.0-dev libssl-dev libcurl4-openssl-dev
+make install
+```
+
+2. **Download host key document from Resource Link on your development machine**
+
+The host key must match the host system for which the image is prepared. You can download a host key document from Resource Link.
+
+As a registered user, access the search page:
+```
+https://www.ibm.com/servers/resourcelink/hom03010.nsf/pages/HKDSearch?OpenDocument
+```
+
+If you have never signed in to Resource Link, you need to register before you can access the host key document search page. Please refer to document [Obtaining a host key document from Resource Link](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-obtain-host-key-document#lxse_obtain_hkd) for details.
+
+Need input the **machine type** and **machine serial number**, which can be obtained from `/proc/sysinfo` of the s390x worker node. Check the values of **Type** and **Sequence Code**. Machine serial number would be the last 5 or 7 characters. Please refer to [this documentation](https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-find-machine-serial) for details.
+
+When you obtain the host key document, please copy the downloaded host key document `HKD-<type>-<serial>.crt` to the worker node host keys directory, the directory can be any, if you use `/root/hostkeys/` as the host keys directory, you can use follow commands:
+```bash
+ssh root@ip-of-peer-pod-worker-node mkdir /root/hostkeys/
+scp HKD-<type>-<serial>.crt root@ip-of-peer-pod-worker-node:/root/hostkeys/
+```
+
+> **Note**
+> - You can download multiple different host keys and prepare only one SE enabled image for different host systems.
+>
+> eg. If you use `/root/hostkeys/` as the hosy keys directory and you want to deploy the SE enabled image on two host systems which the machine type are `8562` the machine serial number are `1234567` and `7654321` then the hosy keys directory tree looks like as follow:
+> ```bash
+> ssh root@ip-of-peer-pod-worker-node tree /root/hostkeys/
+> ...
+> /root/hostkeys/
+> ├── HKD-8562-1234567.crt
+> └── HKD-8562-7654321.crt
+> 
+> 0 directories, 2 files
+> ```
+> - Get the **machine type** and **machine serial number** from worker node only works when there is only one s390x arch host system in the target zone, if there are multiple s390x arch host systems in the target zone, please try to create z vsis on the target s390x arch host system one by one then get the **machine type** and **machine serial number** from the created z vsis on diferent host systems
+> - Another way maybe you can contact the system admin to get the **machine type** and **machine serial number** for target host systems
+
+## Build one SE enabled custom image
+
+Please confirm the IBM Cloud API key, COS instance name and COS bucket name, VPC name, subnet name, region and zone when you set up the demo environment, set `SE_BOOT` as `1`, set `HOST_KEYS_DIR` with the host key documents directory (eg. `/root/hostkeys/`), set `IMAGE_NAME` as `se-<image_name>-s390x` (eg. `se-podvm-s390x`) then run follow commands:
+```
+export IBMCLOUD_API_KEY=<your_api_key>
+export IBMCLOUD_API_ENDPOINT="https://cloud.ibm.com"
+export IBMCLOUD_COS_REGION=<the region your IKS cluster and zVSI is running on>
+export IBMCLOUD_COS_SERVICE_INSTANCE=<your_cos_service_instance_name>
+export IBMCLOUD_COS_BUCKET=<your_cos_bucket_name>
+export IBMCLOUD_COS_SERVICE_ENDPOINT="https://s3.${IBMCLOUD_COS_REGION}.cloud-object-storage.appdomain.cloud"
+export IBMCLOUD_VPC_NAME=<your_VPC_name>
+export IBMCLOUD_VPC_SUBNET_NAME=<your_subnet_name>
+export IBMCLOUD_VPC_REGION=<the_region_name>
+export IBMCLOUD_VPC_ZONE=<the_zone_name>
+export SE_BOOT=1
+export HOST_KEYS_DIR=/root/hostkeys
+export IMAGE_NAME=se-podvm-s390x
+export CLOUD_PROVIDER=ibmcloud
+
+cd /root/cloud-api-adaptor/ibmcloud/image
+make push
+make verify
+```
+> **Note**
+> - `make push` will call the [build script](./image/build.sh) to build SE enabled qcow2 image first and then call the [push script](./image/push.sh) to upload the built se-enabled qcow2 image to cos bucket and then create one se-enabled custom image.
+- The [build script](./image/build.sh) is following the steps in [the document](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-workload-owner-tasks) to prepare SE enabled qcow2 image
+> - `make verify` will call the [verify script](./image/verify.sh) to create one SE enabled vsi and then delete it
+> - The `IMAGE_NAME` is must end with `-s390x` when `SE_BOOT=1`
+
+### Troubleshooting
+The build se-image will use `/dev/nbd0` and `/dev/nbd1`, please make sure they are avaliable to use
+```bash
+qemu-nbd --disconnect /dev/nbd0
+qemu-nbd --disconnect /dev/nbd1
+lsblk
+```
+If the device can't be disconnected:
+- When the `MOUNTPOINT` is not empty, please umount it
+```
+umount <MOUNTPOINT>
+```
+- When there is one `crypt` partation, plesse close it
+```
+lsblk
+...
+nbd1                                           43:32   0  100G  0 disk
+├─nbd1p1                                       43:33   0  255M  0 part
+└─nbd1p2                                       43:34   0 92.9G  0 part
+  └─LUKS-ad6e1db7-6833-4638-8eb1-08972554149b 253:0    0 92.9G  0 crypt
+
+cryptsetup close LUKS-ad6e1db7-6833-4638-8eb1-08972554149b
+```
+
+## Restart cloud-api-adaptor on worker node with new image ID
+
+When the se enabled custom image is created, get the image ID.
+
+Go the development machine to run terraform. We can use the terraform configuration [start-cloud-api-adaptor](./terraform/start-cloud-api-adaptor/) to start the `cloud-api-adaptor` process on the Kubernetes worker node.
+
+Create a `terraform.tfvars` file in the [configuration directory](./terraform/start-cloud-api-adaptor). The `terraform.tfvars` file should look like this:
+```
+ibmcloud_api_key = "<your API Key>"
+region_name = "<the_region_name>"
+cluster_name = "<cluster name>"
+ssh_key_name = "<your SSH key name>" OR ssh_key_id = "<your SSH key id>"
+vpc_name = "<name of your VPC>" OR vpc_id = "<ID of your VPC>"
+primary_subnet_name = "<name of your primary subnet>" OR primary_subnet_id = "<ID of your primary subnet>"
+primary_security_group_name = "<name of your primary security group>" OR primary_security_group_id = "<ID of your primary security group>"
+podvm_image_id = "<ID of your se enabled peer pod VM image>"
+instance_profile_name = "bz2e-2x8"
+```
+
+Please set `podvm_image_id` as the new custom image ID above. And set `instance_profile_name` as a valid confidential computing profile name(eg `bz2e-2x8`).
+
+Usually, we need destroy existing `cloud-api-adaptor` process first, and then start the new process with the new Pod VM image ID:
+```bash
+terraform init
+terraform destroy
+terraform apply
+```
+
+## Test
+
+We can follow guide [Deploy the nginx pod and sniff test nginx](./README.md#deploy-the-nginx-pod-and-sniff-test-nginx) to deploy a nginx pod. Verify the Peer Pod VM is created with the SE enabled custom image.
--- a/ibmcloud/image/build.sh
+++ b/ibmcloud/image/build.sh
@@ -34,6 +34,18 @@ if [[ -z "${base_img_path-}" || -z "${dst_img_path-}" || -z "${files_dir-}" ]]; 
     exit 1
 fi
 
+SE_BOOT=${SE_BOOT:-0}
+
+if [ "${SE_BOOT}" = "1" ]; then
+    if [[ -z "${HOST_KEYS_DIR-}" ]]; then
+        echo "HOST_KEYS_DIR is missed" 1>&2
+        echo "CLOUD_PROVIDER=ibmcloud SE_BOOT=1 HOST_KEYS_DIR=<host keys directory> make build"
+        exit 1
+    fi
+    umount ./rootkeys/ || true
+    rm -rf ./rootkeys/
+fi
+
 base_img_path=$(realpath "$base_img_path")
 src_img_path="$workdir/src.qcow2"
 tmp_img_path="$workdir/tmp.qcow2"
@@ -53,14 +65,22 @@ fi
 
 function cleanup () {
     msg=$1
-
+    if [ "${SE_BOOT}" = "1" ]; then
+        for mnt in "$dst_mnt/boot-se" "$dst_mnt/etc/keys" "$dst_mnt/sys"; do
+            mountpoint -q "$mnt" && umount "$mnt" || true
+            [[ -d "$mnt" ]] && rmdir "$mnt" 2> /dev/null || true
+        done
+        mountpoint -q ./rootkeys/ && umount ./rootkeys/ || true
+        [[ -d ./rootkeys/ ]] && rm -rf ./rootkeys/
+    fi
     for mnt in "$src_mnt/run" "$src_mnt/dev/pts" "$src_mnt/dev" \
                "$src_mnt/proc" "$src_mnt/sys" "$src_mnt" \
                "$dst_mnt/dev" "$dst_mnt/proc" "$dst_mnt"; do
         mountpoint -q "$mnt" && umount "$mnt"
         [[ -d "$mnt" ]] && rmdir "$mnt" 2> /dev/null || true
     done
-
+    sleep 1
+    [ -n "${LUKS_NAME:-}" ] && (cryptsetup close $LUKS_NAME 2> /dev/null || true)
     qemu-nbd --disconnect "$src_nbd"
     qemu-nbd --disconnect "$tmp_nbd"
     kpartx -dsv "$src_nbd"
@@ -76,8 +96,19 @@ function cleanup () {
 trap 'cleanup "image creation failed"' 0
 
 rm -f "$src_img_path" "$tmp_img_path"
+echo "Cleanuping build env"
+cleanup ""
+if [ "${SE_BOOT}" = "1" ]; then
+    echo "Finding host key files" 
+    host_keys=""
+    for i in $(ls ${HOST_KEYS_DIR}/*.crt); do
+        echo "found host key file: \"${i}\""
+        host_keys+="-k ${i} "
+    done
+    [[ -z $host_keys ]] && echo "Didn't find hosy key files, please set HOST_KEYS_DIR correctly "&& exit 1
+fi
 
-echo -e "\nCopying partitions from the base image $src_img_path\n"
+printf "\nCopying partitions from the base image $src_img_path\n"
 
 qemu-img create -f qcow2 -b "$base_img_path" "$src_img_path" $disksize
 qemu-img create -f qcow2 "$tmp_img_path" $disksize
@@ -143,21 +174,50 @@ done < <(parted -s -m "$src_nbd" unit B print)
 sleep 1
 resize2fs -f "${src_nbd}p1"
 
-case "$disklabel" in
-    gpt)
-        sgdisk "$src_nbd" -R "$tmp_nbd"
-        sgdisk -G "$tmp_nbd"
-        ;;
-    msdos)
-        sfdisk -d "$src_nbd" | sfdisk "$tmp_nbd"
-        ;;
-esac
+if [ "${SE_BOOT}" = "1" ]; then
+    echo "Creating boot-se and root partitions" 
+    parted -a optimal $tmp_nbd mklabel gpt \
+        mkpart boot-se ext4 1MiB 256MiB \
+        mkpart root 256MiB "${disksize}" \
+        set 1 boot on
+else
+    case "$disklabel" in
+        gpt)
+            sgdisk "$src_nbd" -R "$tmp_nbd"
+            sgdisk -G "$tmp_nbd"
+            ;;
+        msdos)
+            sfdisk -d "$src_nbd" | sfdisk "$tmp_nbd"
+            ;;
+    esac
+fi
 
-for part_number in "${parts[@]}"; do
-    dd if="${src_nbd}p$part_number" of="${tmp_nbd}p$part_number" bs=$((1024*1024))
-done
+if [ "${SE_BOOT}" = "1" ]; then
+    echo "Waiting for the two nbd partitions to show up"
+    while true; do
+    sleep 1
+    [ -e ${tmp_nbd}p2 ] && break
+    done
+    printf "\nFormatting boot-se partition\n"
+    mke2fs -t ext4 -L boot-se ${tmp_nbd}p1
+    export boot_uuid=$(blkid ${tmp_nbd}p1 -s PARTUUID -o value)
+    printf "\nSetting up encrypted root partition\n"
+    mkdir rootkeys || true
+    mount -t tmpfs rootkeys ./rootkeys
+    dd if=/dev/random of=./rootkeys/rootkey.bin bs=1 count=64 &> /dev/null
+    echo YES | cryptsetup luksFormat --type luks2 ${tmp_nbd}p2 --key-file ./rootkeys/rootkey.bin
+    echo "Setting LUKS name for root partition"
+    LUKS_NAME="LUKS-$(blkid -s UUID -o value ${tmp_nbd}p2)"
+    export LUKS_NAME
+    echo "LUKS name is: $LUKS_NAME"
+    cryptsetup open ${tmp_nbd}p2 $LUKS_NAME --key-file ./rootkeys/rootkey.bin
+else
+    for part_number in "${parts[@]}"; do
+        dd if="${src_nbd}p$part_number" of="${tmp_nbd}p$part_number" bs=$((1024*1024))
+    done
+fi
 
-echo -e "\nMounting the root partition\n"
+printf "\nMounting the root partition\n"
 
 src_part="${src_nbd}p1"
 dst_part="${tmp_nbd}p1"
@@ -176,7 +236,7 @@ cp /run/systemd/resolve/resolv.conf "$src_mnt/run/systemd/resolve/resolv.conf"
 cp /run/systemd/resolve/stub-resolv.conf "$src_mnt/run/systemd/resolve/stub-resolv.conf"
 
 if (( ${#packages[@]} )); then
-    echo -e "\nInstalling packages: ${packages[*]}\n"
+    printf "\nInstalling packages: ${packages[*]}\n"
     chroot "$src_mnt" apt-get update
     chroot "$src_mnt" apt-get install -y "${packages[@]}"
 fi
@@ -185,13 +245,14 @@ case "$(uname -m)" in
     s390x)
         ;;
     *)
-        echo -e "\nUpdating initramfs\n"
+        printf "\nUpdating initramfs\n"
         echo -e "virtio_pci\nvirtio_blk" >> "$src_mnt/etc/initramfs-tools/modules"
         chroot "$src_mnt" update-initramfs  -u
         ;;
 
 esac
 
+chroot "$src_mnt" apt-get remove unattended-upgrades -y
 chroot "$src_mnt" apt-get autoremove
 chroot "$src_mnt" apt-get clean
 chroot "$src_mnt" bash -c 'rm -rf /var/lib/apt/lists/*'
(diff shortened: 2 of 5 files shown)


genjuro214:
Add the document. @liudalibj maybe you can add your diagram here?

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,106 @@
...
+```
+instance_profile_name = "bz2-2x8"
+image_name = "ibm-ubuntu-18-04-1-minimal-s390x-3"
+```
+
+## Prepare worker node to build SE enabled image
+
+1. **Build genprotimg**
+
+`genprotimg` is used to build an encrypted boot record from a given kernel, initial RAM disk, parameters, and public host-key document.
```
hbrueckner, line 19: `genprotimg` should be already part of recent distributions -- so not necessarily to build it.

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,106 @@
...
+2. **Donwload host key document from Resource Link**
+
+The host key must match the host system for which the image is prepared. Please refer to document [Obtaining a host key document from Resource Link](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-obtain-host-key-document#lxse_obtain_hkd) for details.
+
+As a registered user, access the search page:
+```
+https://www.ibm.com/servers/resourcelink/hom03010.nsf/pages/HKDSearch?OpenDocument
+```
+
+Need input the **machine type** and **machine serial number**, which can be found in an IBM internal repository(https://github.ibm.com/ZaaS/iaas_configuration), according to current VPC region and zone. It would be the last 7 characters of the **CPCA** value.
```
hbrueckner, line 38: well.. external users cannot access this internal repository. I would rather remove it here.

If users have access to a VSI instance, they could obtain machine type and machine serial from `/proc/sysinfo`.

@stevenhorsman @magowan How to provide those VPC region/zone specific information in the cloud documentation?

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,106 @@
...
+```bash
+git clone https://github.com/ibm-s390-linux/s390-tools.git
+cd s390-tools/genprotimg/
+apt install -y libglib2.0-dev libssl-dev libcurl4-openssl-dev
+make install
+```
+
+2. **Donwload host key document from Resource Link**
+
+The host key must match the host system for which the image is prepared. Please refer to document [Obtaining a host key document from Resource Link](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-obtain-host-key-document#lxse_obtain_hkd) for details.
```
hbrueckner, line 31: This reads like there is only one host key possible. In fact, you can create an SE-enabled image (technically, SE header) wrapped by multiple different host keys.

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,106 @@
...
+3. **Install cryptsetup for LUKS**
+
+We'll encrypt the root partition with LUKS, please install **cryptsetup** in advance:
+```
+apt install -y cryptsetup
+```
+
+## Build the SE enabled qcow2 image
+
+Still use the [build script](./image/build.sh) for IBMCLOUD image. To build SE enabled image, please set `SE_BOOT` as 1, and set `HOST_KEY_FILE` with the path to host key documnet. For example:
```
hbrueckner: Can `HOST_KEY_FILE` handle multiple host keys?

ibmcloud/image/build.sh
```diff
@@ -34,6 +34,19 @@ if [[ -z "${base_img_path-}" || -z "${dst_img_path-}" || -z "${files_dir-}" ]];
...
 fi
 
+if [[ -z "${SE_BOOT-}" ]]; then
+    SE_BOOT=0
+fi
+
+if [ "${SE_BOOT}" = "1" ]; then
+    dst_img_path="se-"$dst_img_path
+    if [[ -z "${HOST_KEY_FILE-}" ]]; then
+        echo "HOST_KEY_FILE is missed"
```
hbrueckner, line 44: minor: Could you print error messages on `stderr`, e.g., `>&2`.

ibmcloud/image/build.sh
```diff
@@ -143,21 +162,53 @@ done < <(parted -s -m "$src_nbd" unit B print)
...
+    echo -e "\nFormatting boot-se partition"
+    mke2fs -t ext4 -L boot-se ${tmp_nbd}p1
+    export BOOT_UUID=$(blkid ${tmp_nbd}p1 -s PARTUUID -o value)
+    export ROOT_UUID=$(blkid ${tmp_nbd}p2 -s PARTUUID -o value)
+    echo UUIDs:
+    echo "Boot UUID=$BOOT_UUID"
+    echo "Root UUID=$ROOT_UUID"
+    echo -e "\nSetting up encrypted root partition"
+    dd if=/dev/random of=rootkey.bin bs=1 count=64 &> /dev/null
+    cat rootkey.bin | \
```
hbrueckner, line 198: nit: Any reason why you do a `cat` and do not pass `rootkey.bin` as argument to `--keyfile` directly?

ibmcloud/image/build.sh
```diff
@@ -143,21 +162,53 @@ done < <(parted -s -m "$src_nbd" unit B print)
...
+        cryptsetup luksFormat --type luks2 ${tmp_nbd}p2 --key-file -
+    echo "setting LUKS name for root partition"
+    LUKS_NAME="LUKS-$(blkid -s UUID -o value ${tmp_nbd}p2)"
+    export LUKS_NAME
+    echo "LUKS name is: $LUKS_NAME"
+    cat rootkey.bin | cryptsetup open ${tmp_nbd}p2 $LUKS_NAME --key-file -
```
hbrueckner, line 204: ditto

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,157 @@ umount "$src_mnt/dev"
...
+targetbase=/dev/vda
+targettype=scsi
+targetblocksize=512
+targetoffset=2048
+
+[linux]
+image = /boot-se/se.img
+END
+    echo "updating initial ram disk"
+    chroot "$dst_mnt" update-initramfs -u || echo -e "!!! Bootloader install errors prior to this line are intentional !!!!!"
```
hbrueckner, line 327: please echo to `stderr`.

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,157 @@ umount "$src_mnt/dev"
...
+    echo -e "\nGenerating an IBM Secure Execution image"
```
hbrueckner, line 328: nit: instead of `echo -e`, you can also use `printf` built-in (does not require `-e`)

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,157 @@ umount "$src_mnt/dev"
...
+    INITRD_FILE=$(readlink ${dst_mnt}/boot/initrd.img)
+    echo "using initrd: ${INITRD_FILE}"
+    SE_PARMLINE="root=/dev/mapper/$LUKS_NAME console=ttysclp0 quiet panic=0 rd.shell=0 blacklist=virtio_rng swiotlb=262144"
+    echo "${SE_PARMLINE}" > ${dst_mnt}/boot/parmfile
+    /usr/bin/genprotimg \
+        -i ${dst_mnt}/boot/${KERNEL_FILE} \
+        -r ${dst_mnt}/boot/${INITRD_FILE} \
+        -p ${dst_mnt}/boot/parmfile \
+        --no-verify \
+        -k ${HOST_KEY_FILE} \
```
hbrueckner, line 341: As mentioned in the documentation review, `-k` can be specified multiple times with different host keys.

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,157 @@ umount "$src_mnt/dev"
...
+    rm -rf ${dst_mnt}/boot/*
+    echo -e "\nRunning zipl to prepare boot partition"
+    chroot $dst_mnt zipl --targetbase $tmp_nbd \
+    --targettype scsi \
+    --targetblocksize 512 \
+    --targetoffset 2048 \
+    --target /boot-se \
+    --image /boot-se/se.img
+    echo -e "\nClean luks keyfile"
+    rm rootkey.bin
```
hbrueckner, line 355: Please use `shred -u` to (more securely) remove the key.

hbrueckner (commented):
@liudalibj Thanks a lot!

ibmcloud/SECURE_EXECUTION.md
genjuro214: Thanks for the info. I will update document to obtain machine serial from `/proc/sysinfo` on s390x worker.

ibmcloud/SECURE_EXECUTION.md
genjuro214: I have tried to install it via `apt install s390-tools`, but there is no `genprotimg` in the newest version (2.3.0-0ubuntu3.6) on Ubuntu 18.04

ibmcloud/image/build.sh
liudalibj: fixed, use the --keyfile directly

ibmcloud/image/build.sh
liudalibj: updated.

ibmcloud/image/build.sh
liudalibj: updated.

ibmcloud/image/build.sh
liudalibj: Updated, read host keys from `HOST_KEYS_DIR` to support multiple host

ibmcloud/image/build.sh
liudalibj: fixed.

liudalibj (commented):
Updated build script base on review comments.

ibmcloud/SECURE_EXECUTION.md
liudalibj: supported now.

ibmcloud/SECURE_EXECUTION.md
liudalibj: add a Note for one se-enabled image can support multiple different host via provide multiple different host keys

liudalibj:
@hbrueckner thanks your review comments, we updated the document and scripts base on them. please help to check them again. 
Thanks.

ibmcloud/SECURE_EXECUTION.md
hbrueckner: oh well.. I guess 18.04 is not that recent ;-) Let me check which distros where it is available.

liudalibj:
> Add the document. @liudalibj maybe you can add your diagram here?

In ibm public document, already have some diagrams.
Add two **Note** strings to the MD file.

> **Note**: In [the document](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-secure-workload) describe the details about execution secure workload, you can go through the document to get a simplified view about how your workload is protected.

...

> **Note** The [build script](./image/build.sh) is following the steps in [the document](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-workload-owner-tasks)

ibmcloud/image/build.sh
```diff
@@ -185,7 +235,7 @@ case "$(uname -m)" in
     s390x)
         ;;
     *)
-        echo -e "\nUpdating initramfs\n"
+        echo -e "\nUpdating initramfs"
```
hbrueckner, line 238: minor: `printf` could be used here. As this is more style, you might do a follow-up PR.

hbrueckner (commented):
@liudalibj Thanks a lot! LGTM

ibmcloud/SECURE_EXECUTION.md
hbrueckner: You could also reference this documentation: https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-find-machine-serial

ibmcloud/image/build.sh
liudalibj: fixed.

hbrueckner (approved):

hbrueckner (dismissed):
Thanks for adding the docu reference!
/lgtm

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,102 @@
...
+make install
+```
+
+2. **Download host key document from Resource Link**
+
+The host key must match the host system for which the image is prepared. Please refer to document [Obtaining a host key document from Resource Link](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-obtain-host-key-document#lxse_obtain_hkd) for details.
+
+As a registered user, access the search page:
+```
+https://www.ibm.com/servers/resourcelink/hom03010.nsf/pages/HKDSearch?OpenDocument
```
huoqifeng, line 37: Is this always available for registered user? I guess registered user means a user has w3id in IBM Cloud?

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,163 @@ umount "$src_mnt/dev"
...
 mount -t sysfs sysfs "$dst_mnt/sys"
 mount -t proc proc "$dst_mnt/proc"
 mount --bind /dev "$dst_mnt/dev"
 
-case "$(uname -m)" in
-    s390x)
-        echo -e "\nExecuting zipl\n"
-        helper="$dst_mnt/lib/s390-tools/zipl_helper.nbd"
-        cat <<END > "$helper"
+if [ "${SE_BOOT}" = "1" ]; then
```
huoqifeng, line 287: It's not necessary in this PR, but if we want to add similar function for libvirt, maybe need consider separate SE related function and place it somewhere can be reused like https://github.com/confidential-containers/cloud-api-adaptor/tree/9e976024cb9927d17e077f3de6a2a8e3de0a4725/podvm/qcow2? so that libvirt provider on LinuxONE can also leverage the function.

ibmcloud/SECURE_EXECUTION.md
genjuro214: As a customer, they may request customer access to resource link:
https://www-40.ibm.com/servers/resourcelink/hom03010.nsf?OpenDatabase&login

liudalibj: > Is this always available for registered user? I guess registered user means a user has w3id in IBM Cloud?

Anyone can get one IBMID via this page https://www.ibm.com/account/reg/us-en/signup?formid=urx-19776&target=https%3A%2F%2Flogin.ibm.com%2Fsaml%2Fsps%2Fauth

huoqifeng (approved):
/lgtm thanks @liudalibj @genjuro214

liudalibj:
@stevenhorsman @yoheiueda would you like provide your comments and approve when you are free?

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+Need input the **machine type** and **machine serial number**, which can be obtained from `/proc/sysinfo` of the s390x worker node. Check the values of **Type** and **Sequence Code**. Machine serial number would be the last 5 or 7 characters. Please refer to [this documentation](https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-find-machine-serial) for details.
+
+When you obtain the host key document, please send it to the worker node host keys documents directory.
```
Jordan9500, line 42: Could we potentially provide an example of copying the file across if that is how it is done?
And changing the wording slightly to be: 

>  please copy the files contents to the worker nodes host keys directory.

Could the directory be any? If it can should we mention that?

Jordan9500 (changes requested):

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,102 @@
...
+> **Note** The [build script](./image/build.sh) is following the steps in [the document](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-workload-owner-tasks)
+
+## Push the SE enabled qcow2 image to IBM Cloud COS bucket
+
+Use the [push script](./image/push.sh) to push the SE enabled qcow2 image to IBM Cloud COS bucket. Please confirm the IBM Cloud API key, COS instance name and COS bucket name when you set up the demo environment:
+```
+export IBMCLOUD_API_KEY=<your_api_key>
+export IBMCLOUD_COS_SERVICE_INSTANCE=<your_cos_service_instance_name>
+export IBMCLOUD_COS_BUCKET=<your_cos_bucket_name>
+./push.sh --name se-podvm-566065e-dirty-s390x.qcow2 --path se-podvm-566065e-dirty-s390x.qcow2
```
Jordan9500, line 62: The exported variables are only going to be valid if the user is running the demo environment inside `jp-tok-1` As this is the default for the script. 
I think adding a note to mention that exported variables will need to expanded on to include: 
```
export CLOUD_PROVIDER=ibmcloud
export IMAGE_NAME=<An image name with arch included e.g. image-amd64>
export IBMCLOUD_COS_REGION=<the region your IKS cluster and zVSI is running on>
export IBMCLOUD_VPC_REGION=$IBMCLOUD_COS_REGION
export IBMCLOUD_VPC_NAME=<your generated vpc name from the terraform>
export IBMCLOUD_VPC_SUBNET_NAME=<the linked subnet name to the above vpc>
export IBMCLOUD_COS_SERVICE_ENDPOINT="https://s3.${IBMCLOUD_COS_REGION}.cloud-object-storage.appdomain.cloud"
export IBMCLOUD_COS_SERVICE_INSTANCE=<your cos instance>
export IBMCLOUD_COS_BUCKET=<your cos bucket in the region your cluster runs in>
export IBMCLOUD_API_KEY=<your ibmcloud apikey>
export IBMCLOUD_API_ENDPOINT="https://cloud.ibm.com"
export GOPATH=/root/go
``` 
And I think that `./push.sh` command needs to remove the `.qcow2` from the name as it seems to fail due to it being an invalid name.

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,102 @@
...
+https://www.ibm.com/servers/resourcelink/hom03010.nsf/pages/HKDSearch?OpenDocument
+```
+
+Need input the **machine type** and **machine serial number**, which can be obtained from `/proc/sysinfo` of the s390x worker node. Check the values of **Type** and **Sequence Code**. Machine serial number would be the last 5 or 7 characters. Please refer to [this documentation](https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-find-machine-serial) for details.
+
+When you obtain the host key document, please send it to the worker node host keys documents directory.
+
+> **NOTE**: You can download multiple different host keys and prepare only one SE enabled image for different host systems.
+
+## Build the SE enabled qcow2 image
```
Jordan9500, line 46: Could you combine the steps by just running `make push` here instead, which will do both. 
From my understanding the build fails when you try and run this due to the addition of `se-` within `build.sh` on line 42.
This means when it reaches the push.sh file it can no longer find the path because it has added an extra prefix that is not included in `IMAGE_NAME`. 
Is this a required prefix or could we instead specify in the exports something similar to: 
```
export CLOUD_PROVIDER=ibmcloud
export IMAGE_NAME=se-<An image name with arch included e.g. image-s390x>
export IBMCLOUD_COS_REGION=<the region your IKS cluster and zVSI is running on>
export IBMCLOUD_VPC_REGION=$IBMCLOUD_COS_REGION
export IBMCLOUD_VPC_NAME=<your generated vpc name from the terraform>
export IBMCLOUD_VPC_SUBNET_NAME=<the linked subnet name to the above vpc>
export IBMCLOUD_COS_SERVICE_ENDPOINT="https://s3.${IBMCLOUD_COS_REGION}.cloud-object-storage.appdomain.cloud"
export IBMCLOUD_COS_SERVICE_INSTANCE=<your cos instance>
export IBMCLOUD_COS_BUCKET=<your cos bucket in the region your cluster runs in>
export IBMCLOUD_API_KEY=<your ibmcloud apikey>
export IBMCLOUD_API_ENDPOINT="https://cloud.ibm.com"
export GOPATH=/root/go

CLOUD_PROVIDER=ibmcloud SE_BOOT=1 HOST_KEYS_DIR=/root/ make push
```

Jordan9500 (changes requested):

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,163 @@ umount "$src_mnt/dev"
...
+    rm -rf ${dst_mnt}/boot/*
+    printf "\nRunning zipl to prepare boot partition\n"
+    chroot $dst_mnt zipl --targetbase $tmp_nbd \
+    --targettype scsi \
+    --targetblocksize 512 \
+    --targetoffset 2048 \
+    --target /boot-se \
+    --image /boot-se/se.img
+    printf "\nClean luks keyfile\n"
+    shred -u rootkey.bin
```
yoheiueda, line 361: If `rootkey.bin` is a very secret one,  how about using a temporary file on a tmpfs, instead of a regular file on the current directory?

A more secure way is to use a temporary variable, but doing it in bash looks complicated.
https://unix.stackexchange.com/a/278644

ibmcloud/image/build.sh
```diff
@@ -34,6 +34,19 @@ if [[ -z "${base_img_path-}" || -z "${dst_img_path-}" || -z "${files_dir-}" ]];
     exit 1
 fi
 
+if [[ -z "${SE_BOOT-}" ]]; then
+    SE_BOOT=0
+fi
```
yoheiueda, line 39: You can write this in a line as follows.

```
SE_BOOT=${SE_BOOT:-0}
```

ibmcloud/image/build.sh
```diff
@@ -143,21 +162,52 @@ done < <(parted -s -m "$src_nbd" unit B print)
...
+if [ "${SE_BOOT}" = "1" ]; then
+    echo "Waiting for the two nbd partitions to show up"
+    while true; do
+    sleep 1
+    [ -e ${tmp_nbd}p2 ] && break
+    done
+    printf "\nFormatting boot-se partition\n"
+    mke2fs -t ext4 -L boot-se ${tmp_nbd}p1
+    export BOOT_UUID=$(blkid ${tmp_nbd}p1 -s PARTUUID -o value)
+    export ROOT_UUID=$(blkid ${tmp_nbd}p2 -s PARTUUID -o value)
```
yoheiueda, line 192: I don't think we need to export `BOOT_UUID` and `ROOT_UUID`.

I prefer variable names with lower letters (`boot_uuid` and `root_uuid`) for internal variables.

yoheiueda (approved):
LGTM.

I have some minor comments, but I think it is OK to fix them later.

ibmcloud/SECURE_EXECUTION.md
genjuro214: OK, I will add it.

ibmcloud/SECURE_EXECUTION.md
genjuro214: OK, push.sh will directly create custom image after uploading the qcow2 file to COS bucket. Here I just need upload the file to bucket. I will change the document to use `multipart_upload.sh` instead.

Jordan9500: OK, so the multipart upload will only create the image inside the COS bucket. And we want the user to manually create the custom image in ibmcloud. 

The main difference I can see is the change of OS, to point to hyper-protect instead when creating the custom image. 
Could we instead look at adding something similar to: 
```
[ "$SE_BOOT" = "1" ] && os_name="hyper-protect-1-0-s390x" || os_name="ubuntu-20-04-${arch/x86_64/amd64}"
```
To this line: [ibmcloud/image/push.sh#L96](https://github.com/liudalibj/cloud-api-adaptor/blob/6af27a7789a8a7624daf58fccb90f309fda7281e/ibmcloud/image/push.sh#L96)

And remove the need to separate the different `build` then `upload` then `create` steps? And just run: `CLOUD_PROVIDER=ibmcloud SE_BOOT=1 HOST_KEYS_DIR=/root/ make push`. This would mean that the requested changes mentioned here: https://github.com/confidential-containers/cloud-api-adaptor/pull/407#discussion_r1046085254. Will also be needed.

liudalibj: I will updated it.

ibmcloud/image/build.sh
liudalibj: updated.

ibmcloud/image/build.sh
liudalibj: updated.

ibmcloud/SECURE_EXECUTION.md
liudalibj: Update the push.sh and verify.sh to support SE image.

ibmcloud/SECURE_EXECUTION.md
liudalibj: Updated the document

ibmcloud/image/Makefile
```diff
@@ -17,25 +17,25 @@ IMAGE_SUFFIX := .qcow2
 
 build: $(IMAGE_FILE)
 
-push: $(IMAGE_FILE)
+push:
```
Jordan9500, line 20: With the changes to the makefile why does the push now no longer do the build?

Jordan9500 (changes requested):

Jordan9500 (changes requested):
More of a wider question in regards to the use of the prefix `se-`. What is the needed use case for this to be a hard coded addition? 

From my understanding we could add a recommendation to the docs to use the existing variable `IMAGE_NAME` to also add a `se-` prefix. This would then remove the need for different additions related to solely adding the `se-` step.
Also the users image name that is defined will be overwritten with the `se-` prefix.

Jordan9500 (changes requested):

ibmcloud/image/build.sh
liudalibj: updated, use tmpfs here.

ibmcloud/image/build.sh
```diff
@@ -53,7 +66,13 @@ fi
 
 function cleanup () {
     msg=$1
-
+    if [ "${SE_BOOT}" = "1" ]; then
+        for mnt in "$dst_mnt/boot-se" "$dst_mnt/etc/keys" "$dst_mnt/sys"; do
+            mountpoint -q "$mnt" && umount "$mnt"
+            [[ -d "$mnt" ]] && rmdir "$mnt" 2> /dev/null || true
+        done
+        cryptsetup close $LUKS_NAME 2> /dev/null || true
```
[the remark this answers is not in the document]
jtumber-ibm: I think this would be an unnecessary change, the `|| true` means even if the command fails it won't affect further execution.

ibmcloud/image/build.sh
```diff
@@ -206,69 +254,165 @@ umount "$src_mnt/dev"
...
+    esac
+fi
 
 umount "$dst_mnt/dev"
 umount "$dst_mnt/proc"
 umount "$dst_mnt/sys"
-
 umount "$dst_mnt"
 
+if [ "${SE_BOOT}" = "1" ]; then
```
jtumber-ibm, line 398: Any reason to not leave this to the `cleanup` function called on L413?

ibmcloud/image/build.sh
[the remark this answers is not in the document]
Jordan9500: I think the addition of: `set -o errexit -o pipefail -o nounset` however means this command will fail on `LUKS_NAME: unbound variable`. Rather than avoiding the error.

ibmcloud/image/build.sh
liudalibj: need close encrypted root partition before `qemu-nbd --disconnect`
the cleanup function is the final step.

ibmcloud/image/build.sh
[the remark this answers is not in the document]
jtumber-ibm: Ah you're right `-o nounset` causes exit even when ignoring errors. Thanks

ibmcloud/image/Makefile
liudalibj: if we run `make build` and then `make push` with the old Makefile, the pod-vm-image will be build twice, so that remove the dependency here.

ibmcloud/SECURE_EXECUTION.md
liudalibj: use `IMAGE_NAME ` remove `se-` hardcode string.

ibmcloud/image/build.sh
```diff
@@ -206,69 +257,163 @@ umount "$src_mnt/dev"
...
-        helper="$dst_mnt/lib/s390-tools/zipl_helper.nbd"
-        cat <<END > "$helper"
+if [ "${SE_BOOT}" = "1" ]; then
+    printf "\nPreparing secure execution boot image\n"
+    echo "mounting tmpfs to /etc/keys"
+    mkdir -p ${dst_mnt}/etc/keys
+    mount -t tmpfs keys ${dst_mnt}/etc/keys
+    # ADD CONFIGURATION
+    echo "adding fstab"
+    echo "#This file was auto-generated" >  ${dst_mnt}/etc/fstab
```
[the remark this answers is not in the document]
liudalibj: done.

ibmcloud/image/build.sh
```diff
@@ -34,6 +34,17 @@ if [[ -z "${base_img_path-}" || -z "${dst_img_path-}" || -z "${files_dir-}" ]];
     exit 1
 fi
 
+SE_BOOT=${SE_BOOT:-0}
+
+if [ "${SE_BOOT}" = "1" ]; then
+    dst_img_path="se-"$dst_img_path
```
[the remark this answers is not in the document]
liudalibj: use `IMAGE_NAME` remove the `se-` hardcode string, updated the document/script.

ibmcloud/image/Makefile
Jordan9500: I think with the removal of the se- prefix it will no longer do that if there is an existing file that matches the IMAGE_NAME.

ibmcloud/image/build.sh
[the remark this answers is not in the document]
Jordan9500: I believe the rebuild uses the image name to check if the file already exists, if it does then it will no rebuild it.

ibmcloud/image/Makefile
liudalibj: yeah, verified, rollback the change for Makefile.

liudalibj:
@Jordan9500  @jtumber-ibm the document and script update base on your review comments, please help to check them again. Thanks!

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,124 @@
...
+export IBMCLOUD_COS_REGION=<the region your IKS cluster and zVSI is running on>
+export IBMCLOUD_COS_SERVICE_INSTANCE=<your_cos_service_instance_name>
+export IBMCLOUD_COS_BUCKET=<your_cos_bucket_name>
+export IBMCLOUD_COS_SERVICE_ENDPOINT="https://s3.${IBMCLOUD_COS_REGION}.cloud-object-storage.appdomain.cloud"
+export IBMCLOUD_VPC_NAME="<your_VPC_name>"
+export IBMCLOUD_VPC_SUBNET_NAME="<your_subnet_name>"
+export IBMCLOUD_VPC_REGION="<the_region_name>"
+export IBMCLOUD_VPC_ZONE="<the_zone_name>"
+
+CLOUD_PROVIDER=ibmcloud SE_BOOT=1 IMAGE_NAME=se-podvm-s390x make push
```
Jordan9500, line 91: nit: With the replacement for push now also redoing the build, could we instead combine the build steps above and just do the push?

Another thing is we could instead export `IMAGE_NAME` so that we keep consistent with other docs, and we decrease the chance of the user making a mistake between build, push and verify. 
So instead adding: 
```
export IMAGE_NAME=se-<image-name>-<arch>
```
to the list of exported variables above. And that way it should remain the same for each run.

ibmcloud/image/build.sh
```diff
@@ -34,6 +34,17 @@ if [[ -z "${base_img_path-}" || -z "${dst_img_path-}" || -z "${files_dir-}" ]];
...
 
+SE_BOOT=${SE_BOOT:-0}
+
+if [ "${SE_BOOT}" = "1" ]; then
+    if [[ -z "${HOST_KEYS_DIR-}" ]]; then
+        echo "HOST_KEYS_DIR is missed" 1>&2
+        echo "CLOUD_PROVIDER=ibmcloud SE_BOOT=1 HOST_KEYS_DIR=<host keys directory> make build"
+        exit 1
+    fi
+    rm -rf rootkeys
```
Jordan9500, line 45: If the build fails for any reason prior to the `shred`, `umount` and `rm` on line 362.
So I think adding something similar here and in the cleanup function (maybe line 72?):
```
umount ./rootkeys/ || true
rm -rf ./rootkeys/
```

Jordan9500 (changes requested):
Thank you for the added changes, just a couple more things.

ibmcloud/image/build.sh
liudalibj: updated.

ibmcloud/SECURE_EXECUTION.md
liudalibj: updated.
use exported variables.

liudalibj:
> Thank you for the added changes, just a couple more things.

Thanks, updated.

ibmcloud/image/build.sh
Jordan9500: Please could we add it at line 44 as well?

liudalibj: added

ibmcloud/image/build.sh
```diff
@@ -53,7 +65,15 @@ fi
 
 function cleanup () {
     msg=$1
-
+    if [ "${SE_BOOT}" = "1" ]; then
+        for mnt in "$dst_mnt/boot-se" "$dst_mnt/etc/keys" "$dst_mnt/sys"; do
+            mountpoint -q "$mnt" && umount "$mnt" || true
+            [[ -d "$mnt" ]] && rmdir "$mnt" 2> /dev/null || true
+        done
+        [ -n "${LUKS_NAME:-}" ] && cryptsetup close $LUKS_NAME 2> /dev/null || true
```
Jordan9500, line 73: After doing some failing testcases and running the script with the hostkey not in the correct directory it fails on the `genprotimg` step at line 347 (to be expected).

However it seems that the closing of the partition for `tmp_nbd` is not correct. When it finishes currently that partition stays open, and means the next build will fail until either it is:
1. Manually closed after the script fully finishes by doing: 
```
qemu-nbd --disconnect "/dev/nbd1/"
export LUKS_NAME="LUKS-$(blkid -s UUID -o value nbd1p2)"
cryptsetup close ${LUKS_NAME}
```
2. Or manually changing the `tmp_nbd` to point at a different instance. e.g. `/dev/nbd2` inside the `build.sh` file. 

I did find a **potential solution** of changing the line to be:
```
[ -n "${LUKS_NAME:-}" ] && (cryptsetup close $LUKS_NAME 2> /dev/null || true)
```
And moving the line to Line: [build.sh#L83](https://github.com/liudalibj/cloud-api-adaptor/blob/c3c684c9b9b54caf389ebcf78de58291d51c35df/ibmcloud/image/build.sh#L83)
I think this happens because the `||`  is seen as the other response of the inline statement. And the partition is still seen as being in use. After moving the line too after the other unmounts it seems to successfully remove them. 
I have a feeling that this is something to do with `${dst_mnt}` not properly unmounted until after that loop. So it cant successfully close those partitions. 


It might also be worth creating some form of troubleshoot area at the end of the file to provide context of the error and a potential solution if this happens again.
The error we see with this when attempting to run the `make build` again is: 
```
Formatting './tmp.qcow2', fmt=qcow2 size=107374182400 cluster_size=65536 lazy_refcounts=off refcount_bits=16
Failed to set NBD socket
Disconnect client, due to: Unexpected end-of-file before all bytes were read
```
And the checks to see if it related to the partition issue is to run `ls /dev/nbd1p*` which should result in:
```
/dev/nbd1p1
/dev/nbd1p2
```
If the error has occurred.

Jordan9500 (changes requested):

stevenhorsman (approved):
Apart from the couple of minor comments that I made it looks good to me and I was able to create an SE base nginx peer pod:
```
ibmcloud is ins
Listing instances in all resource groups and region ca-tor under account Steven Horsman's Account as user steven@uk.ibm.com...
ID                                          Name                                Status    Reserved IP    Floating IP    Profile    Image                                VPC                    Zone       Resource group
02r7_a1baabe5-387c-4a62-9713-aa51f60574e5   podvm-nginx-bfc40523                running   10.249.64.11   -              bz2e-2x8   se-podvm-s390x                       sh-pp-vpc-2022-12-15   ca-tor-2   default
```
image
Big thanks to @liudalibj @genjuro214 and all the reviewers who helped improve this!

ibmcloud/image/build.sh
liudalibj: Refine the build.sh to handle the bad path and add one Troubleshooting section to the document.

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,125 @@
...
+
+Need input the **machine type** and **machine serial number**, which can be obtained from `/proc/sysinfo` of the s390x worker node. Check the values of **Type** and **Sequence Code**. Machine serial number would be the last 5 or 7 characters. Please refer to [this documentation](https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-find-machine-serial) for details.
+
+When you obtain the host key document, please copy the downloaded host key document `HKD-<type>-<serial>.crt` to the worker node host keys directory, the directory can be any, if you use `/root/hostkeys/` as the host keys directory, you can use follow commands:
+```bash
+ssh root@ip-of-peer-pod-worker-node mkdir /root/hostkeys/
+scp HKD-<type>-<serial>.crt root@ip-of-peer-pod-worker-node:/root/hostkeys/
+```
+
+> **NOTEs**:
```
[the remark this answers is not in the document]
liudalibj: Updated.

ibmcloud/SECURE_EXECUTION.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Restart cloud-api-adaptor on worker node with new image ID
+
+When the se enabled custom image is created, get the image ID.
+
+Go the development machine to run terraform. We can use the terraform configuration [start-cloud-api-adaptor](./terraform/start-cloud-api-adaptor/) to start the `cloud-api-adaptor` process on the Kubernetes worker node.
+
+Create a `terraform.tfvars` file in the [configuration directory](./terraform/start-cloud-api-adaptor). The `terraform.tfvars` file should look like this:
+```
+ibmcloud_api_key = "<your API Key>"
```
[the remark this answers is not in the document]
liudalibj: added.

stevenhorsman (approved):
Sorry, I meant to approve, rather than comment yesterday, but thanks for the updates!

liudalibj:
@yoheiueda @hbrueckner @huoqifeng @Jordan9500 When you are free, would you like check this pr again and approve it? So we can merge it. Thanks!

huoqifeng (approved):
/lgtm

Jordan9500 (approved):
LGTM, thank you.

liudalibj:
Thanks you all for review and approve this PR.
We have Secure Execution Support for Peer Pods on IBM Cloud VPC now!

PR #416, opened 2022-12-12
vsphere: Add command line argument checking

### LINKED ISSUE #415: vsphere CAA Add command line argument checking
Check to make sure are required fields have been filled in during CAA Startup

GOVC_DATASTORE and GOVC_DATACENTER are for now mandatory and not left up to defaults as it was too confusing.

caavery:
Fixes #415

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -182,6 +182,10 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		})
 		defaultToEnv(&vspherecfg.UserName, "GOVC_USERNAME")
 		defaultToEnv(&vspherecfg.Password, "GOVC_PASSWORD")
+		err := validateArgs("vsphere", vspherecfg)
+		if err != nil {
+			return nil, err
+		}
 
 	default:
 		os.Exit(1)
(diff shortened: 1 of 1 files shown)


caavery:
> Thanks! IIUC you can have the same result by simply passing these vars to the **test_vars** function [here](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/entrypoint.sh#L93) (and if datacenter & datastore are mandatory i'd also remove them from being passed as "[optionals](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/entrypoint.sh#L96)" Having said that, i think checking these in the code would also make sense but only if there's a way to make it more cloud-provider generic

caavery:
OK accidentally closed the PR. 
@snir911 
If I add all of these to test_vars then they become mandatory? What happens if they are not included meaning what kind of error is seen?
The code is generic for all cloud providers. if hypProvider == "vsphere" . Each provider can add their own checking section. If you notice all of main.go provides this type of customization per specific provider. However I can move all the missing stuff within the  if hypProvider == "vsphere" brackets as that is what vpshere cares about. Or each can call a separate checking function

bpradipt:
@caavery @snir911 does it make sense to do the check here - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/parse.go#L11 ?

caavery:
@snir911 @bpradipt 
It does except for I could not check these values until after
defaultToEnv(&vspherecfg.UserName, "GOVC_USERNAME")
defaultToEnv(&vspherecfg.Password, "GOVC_PASSWORD")

snir911:
@caavery it would not even start the cloud-api-adaptor (container will fail to start with `var not set` [error](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/entrypoint.sh#L18)

PR #426, opened 2022-12-16
controller: make spec fields optional

### LINKED ISSUE #425: controller: make spec fields in peerpodconfig optional
nodeSelector, configMapName and cloudSecretName should have been made optional from the start.
It doesn't make sense for them to be required because users don't have to use those fields.

jensfr:
nodeSelector, configMapName and cloudSecretName should be optional, there's no hard requirement for them in the spec. 

Fixes #425 

### DIFF
--- a/peer-pod-controller/api/v1alpha1/peerpodconfig_types.go
+++ b/peer-pod-controller/api/v1alpha1/peerpodconfig_types.go
@@ -36,6 +36,7 @@ type PeerPodConfigSpec struct {
 
 	// NodeSelector selects the nodes to which the cca pods, the RuntimeClass and the MachineConfigs we use
 	// to deploy the full peer pod solution.
+	// +optional
 	NodeSelector *metav1.LabelSelector `json:"nodeSelector"`
 
 	// ConfigMapName is the name of the configmap that holds cloud provider specific environment Variables


bpradipt:
Would it make sense to have a default configmap and secret instead of making it optional ?
I'm missing how the controller will use the configmap and secret as defined here - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/install/overlays/aws/kustomization.yaml

jensfr:
> Would it make sense to have a default configmap and secret instead of making it optional ?

Maybe. Currently we have only documented what the names for the configmap and secret have to be and we use hard-coded names in the controll, but
we don't enforce it on the CRD level. The result is error messages in the log when one of them is not found.

If we said they are required, should we also set a default value for the names?

> I'm missing how the controller will use the configmap and secret as defined here - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/install/overlays/aws/kustomization.yaml

It looks up CLOUD_PROVIDER in the configmap to start the corresponding cloud-api-adaptor-$CLOUD_PROVIDER image
And it generates environment variables from both the secret and configmap. I think this may be changed to only create env variables from the config map, since we have moved CLOUD_PROVIDER from the secret to the config map. 

bottom line, I will change this PR to only make nodeSelector optional. In another PR I'll fix the hard-coded names to be read from the spec fields.

Issue #429, opened 2022-12-17
vsphere: fix kickstart for RHEL8 podvm build


whitebrandy:
RHEL8 kickstart seems to have trouble with the isencrypted flag for the user command. Use plaintext instead as a workaround.

PR #430, opened 2022-12-17
vsphere: template: workaround ks user bug

whitebrandy:
RHEL8 kickstarts do not seem to recognize the isencrypted flag to user command. Use plaintext instead that works universally. While at it, also make username/password configurable.

Fixes: #429


caavery (approved):
Tested on vcenter with rhel8 iso.

LGTM

bpradipt (approved):
/lgtm
Thanks @whitebrandy

Issue #431, opened 2022-12-17
vsphere: unmount misc cgroup in the podvm


whitebrandy:
Since podvm image is using legacy cgroup, umount /sys/fs/cgroup/misc before running kata-agent

PR #432, opened 2022-12-17
vsphere: template: unmount misc cgroup as its not handled by kata

whitebrandy:
Now that we are using legacy cgroups...
https://github.com/kata-containers/kata-containers/issues/4610

Fixes: #431


bpradipt (approved):
/lgtm 
Thanks @whitebrandy 
Going forward, we should move vsphere image generation to use the common files from podvm to avoid tracking changes in multiple places

snir911 (approved):
lgtm, thanks!

PR #434, opened 2022-12-19
Add support for container build on centos

bpradipt:


bpradipt:
@snir911 @whitebrandy can you review the approach for CentOS8-stream build. I think a similar approach should work for rhel8 qcow2 customization using container builds

podvm/Dockerfile.podvm.centos
```diff
@@ -0,0 +1,28 @@
...
+# Building binaries separately allows for reuse in case of error
+# with image building step
+RUN cd cloud-api-adaptor/podvm && \
+     CLOUD_PROVIDER=$CLOUD_PROVIDER LIBC=gnu make binaries
+
+RUN cd cloud-api-adaptor/podvm && \
+     CLOUD_PROVIDER=$CLOUD_PROVIDER PODVM_DISTRO=$PODVM_DISTRO make image
+
+FROM scratch
+COPY --from=podvm_builder /src/cloud-api-adaptor/podvm/output/*.qcow2 /
```
whitebrandy, line 28: Trying to understand the workflow, does this line copy from the previous build to / in the current container ? How do we finally retrieve the usable qcow2 ?

bpradipt: 1. Create a container using the image
2. Copy the qcow2 from the container to the host file system

bpradipt:
Tested building of the qcow2 image for both Ubuntu and CentOS. Haven't tested pod VM creation yet.

snir911:
I didn't have a chance to test yet but it LGTM overall.
also worth checking that it's actually ssh accessible with this user, i recently had issues to login with ubuntu, it seems to require some additional ssh configuration to allow password login

bpradipt:
> I didn't have a chance to test yet but it LGTM overall. also worth checking that it's actually ssh accessible with this user, i recently had issues to login with ubuntu, it seems to require some additional ssh configuration to allow password login

During the image creation process packer will access the VM using SSH and make the required changes like executing scripts etc. I tested image creation successfully for both ubuntu20.04 and centos8-stream.
Post the image creation process, this PR locks the user so that logins are disallowed when creating pod VM.
This is a security measure. If for debugging SSH access is needed, then guestfish can be used to enable password login for the user.

podvm/Dockerfile.podvm_builder.centos
```diff
@@ -0,0 +1,59 @@
+# Copyright Confidential Containers Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Creates a builder container image that should be used to build the Pod VM
+# disk inside a container.
+#
+FROM quay.io/centos/centos:stream8
+
+ARG GO_VERSION="1.18.7"
```
stevenhorsman, line 10: kata-containers has recently bumped to required a minimum version of go 1.19.3 to built the runtime, so it is probably worth bumping to that version?

podvm/qcow2/userdata.cfg
```diff
@@ -1,8 +1,15 @@
...
   expire: false
 
 locale: en_US.UTF-8
 locale_configfile: /etc/default/locale
+
+users:
+  - name: peerpod
+    #default password: PeerP0d
+    #openssl passwd -1 PeerP0d
+    passwd: $1$u4tXfju6$lvXwBnkFT4S0depRtcTjo/
```
stevenhorsman, line 13: Sorry, with my nominated security focal hat on, I just to want to double check - is this password just a newly generated one that is then locked out, so can't be used to ssh into the peer pod build from this?

stevenhorsman (dismissed):
Just a couple of comments/questions from me.

I've not run the whole process, but I've created a new ibm cloud image using the new Makefile.inc changes and it LGTM.

podvm/qcow2/userdata.cfg
bpradipt: This is the hash version of plaintext password `PeerP0d` that is used during image creation and post that locked out.
User can change this to anything else but it's a manual process before starting the image build.
I will add an additional comment clarifying the same.

podvm/Dockerfile.podvm_builder.centos
bpradipt: Is it ok if I bump up the golang version in a follow-on PR? Need to test the CAA binary once after building with golang-1.19 to ensure things are working fine.

stevenhorsman: Yeah, that's fine. From https://github.com/confidential-containers/cloud-api-adaptor/issues/440 you can see that I'm still a bit confused how the kata-shim is able to be built with go 1.18, but if it still works for now, then addressing it separately makes sense!

stevenhorsman (approved):
LGTM. Thanks @bpradipt!

bpradipt:
/hold
Testing the built image uncovered two issues with CentOS which requires some additional changes.
1. Explicitly turning selinux to permissive for CentOS
2. Ensuring agent-protocol-forwarder runs after cloud-init (this was added to the vsphere image building but not in the common podvm files)

whitebrandy:
> /hold Testing the built image uncovered two issues with CentOS which requires some additional changes.
> 
>     1. Explicitly turning selinux to permissive for CentOS
You can also just relabel the files you are copying without having to turn off selinux.

>     2. Ensuring agent-protocol-forwarder runs after cloud-init (this was added to the vsphere image building but not in the common podvm files)
I think the changes to agent-protocol-forwarder depending on cloud-init was added to all sites.
https://github.com/confidential-containers/cloud-api-adaptor/pull/389/commits/9c64364c21fadabd25d2dc1a86ddcf8735c601f1

bpradipt:
> > /hold Testing the built image uncovered two issues with CentOS which requires some additional changes.
> > ```
> > 1. Explicitly turning selinux to permissive for CentOS
> > ```
> 
> You can also just relabel the files you are copying without having to turn off selinux.
>

Right. Let me do that and check.
Btw what is this for in RHEL - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/vsphere/image/rhel/data/ks.pkr.hcl#L29 ?

> > ```
> > 2. Ensuring agent-protocol-forwarder runs after cloud-init (this was added to the vsphere image building but not in the common podvm files)
> > ```
> 
> I think the changes to agent-protocol-forwarder depending on cloud-init was added to all sites. [9c64364](https://github.com/confidential-containers/cloud-api-adaptor/commit/9c64364c21fadabd25d2dc1a86ddcf8735c601f1)

Hmm right. LEt me check why it didn't pickup for me

bpradipt:
Successfully tested with libvirt using both Ubuntu and CentOS podvm images

cc @snir911 @stevenhorsman

snir911 (approved):

stevenhorsman (approved):
LGTM. Thanks!

PR #444, opened 2022-12-21
controller: add volume to mount auth.json

### LINKED ISSUE #443: add support to the controller to add an auth.json file to cloud-api-adaptor pods
Similar to #381 create a volume from a secret that was created by the user and mount it to /root/containers/auth.json in the container.

jensfr:
This adds a volume and volumeMount to the cloud-api-adaptor pod definition. cloud-api-adaptor can then inject the auth.json file into the podvm and enable it to pull image from registries that require authentication.

Rename variable sshSecretOptional to be more generic and use it for all optional secrets

Fixes #443


snir911 (approved):
Thanks LGTM

bpradipt (approved):
/lgtm
@jensfr the subject still shows draft.

Issue #446, opened 2022-12-21
Move vsphere template generation to podvm


whitebrandy:
vsphere template still uses its own copy of files to copy to the podvm image. Avoid duplication by move the template generation under podvm.

PR #447, opened 2022-12-21
vsphere: template: consolidate bootstrap files

whitebrandy:
Use common files in podvm for generating ova templates. No functional change except for the build target
(CLOUD_PROVIDER=vsphere make image)

Fixes: #446


whitebrandy:
@bpradipt @caavery pls review

podvm/Makefile
```diff
@@ -17,6 +17,7 @@ KATA_AGENT_SRC := ../../kata-containers/src/agent
 STATIC_LIBSECCOMP_BUILDER := ../../kata-containers/ci/install_libseccomp.sh
 AGENT_PROTOCOL_FORWARDER_SRC := ../
 
+ifeq ($(CLOUD_PROVIDER),libvirt)
```
bpradipt, line 20: We are creating qcow2 for all the providers in addition to native images (eg. AMIs for AWS).

whitebrandy: How about I just create another target, eg. "make template" which is valid only for vsphere and remove the conditional on the "image" target ?

bpradipt: Can you simply have `ifeq ($(CLOUD_PROVIDER),vsphere)` and the specifics which are applicable only for vsphere ?

caavery: OK I'm confused. Perhaps I don't understand the objective.
 
Before these are the instructions in the README under vsphere for making a template

cd vsphere/image && make
```
For a rhel podvm image, try:
```
cd vsphere/image && PODVM_DISTRO=rhel make

Using the word image is confusing as it is used to build the CAA container image and pushes it to quay.

How do I distinguish between wanting a ubuntu template vs a rhel template?
Where do I put the RHEL8.iso I will be using to generate the template?

whitebrandy: Making sure I understand this - The default $(IMAGE_FILE) target isn't applicable for vsphere. For example, we don't create a cloudinit image with cloud-localds and we don't really generate an image file. So, to be able to do CLOUD_PROVIDER=vsphere make image, we have to specify a different "image" target. But then, if I understood your comment above correctly,  we can't create a vsphere specific qcow2 anymore.  So, the way to make it work is define a new target - template, which does image generation the vmware way and keep image as is so that `CLOUD_PROVIDER=vsphere make image`  provides a qcow2 and `CLOUD_PROVIDER=vsphere make template` creates a vcenter template.

whitebrandy:
@caavery Creating a new thread for your questions -
> OK I'm confused. Perhaps I don't understand the objective.
The objective is simply to use common code to create the podvm image.
...
> Using the word image (CLOUD_PROVIDER=vsphere make image) is confusing as it is used to build the CAA container  
> image and pushes it to quay.
This isn't specific to this MR. I just reused the existing target that we have for qcow2. 

> How do I distinguish between wanting a ubuntu template vs a rhel template?
Still the same, use PODVM_DISTRO.

> Where do I put the RHEL8.iso I will be using to generate the template?
You can put it anywhere you want, you have to make sure you specify the correct path in your settings*.auto.hcl file.

> Where do I run the make command?
cd podvm && CLOUD_PROVIDER=vsphere make image

bpradipt:
@whitebrandy Sorry, I'm bit confused with the creation of new directory `ova` under podvm and few other pieces in this PR.  

I was hoping you could do something similar like aws  - https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/aws/image

IOW, keep vsphere specific stuffs under `vsphere` and use the common files from podvm to start with. 
Under `vsphere/images` you can create a Makefile importing the common one and define any new target which is specific for vsphere (eg. template)

I don't think you need to change anything in `podvm`. Please correct me if I'm wrong.

whitebrandy:
> @whitebrandy Sorry, I'm bit confused with the creation of new directory `ova` under podvm and few other pieces in this PR.
> 
> I was hoping you could do something similar like aws - https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/aws/image
> 
> IOW, keep vsphere specific stuffs under `vsphere` and use the common files from podvm to start with. Under `vsphere/images` you can create a Makefile importing the common one and define any new target which is specific for vsphere (eg. template)
> 
> I don't think you need to change anything in `podvm`. Please correct me if I'm wrong.

Ah, sorry, I misunderstood the plan! And this addresses @caavery's concerns too in that nothing changes on the user front. Thanks for the aws link, I will try something similar.

caavery (approved):
Works fine.
LGTM

bpradipt (approved):
/lgtm
Thanks @whitebrandy

Issue #448, opened 2022-12-22
Ensure mutating webhook is called for all namespaces except it's own


bpradipt:
Currently the mutating webhook only mutates pods when the namespace have the label `peerpods: true`. This creates the following problem:

Peer pods created from namespaces without having the labels will still get created without any resource accounting. This is bad and requires a policy engine for enforcement. The policy engine will ensure only namespaces with labels `peerpods: true` are able to create peer pods. While policy engines are common in production deployments, and should be used; however having a hard dependency to ensure proper functioning of peer pods is not desirable.

One solution we can use is to ensure that the mutating webhook works for all regular namespaces by default. 
This provides a workable out-of-box configuration.  

Further, the cluster admin can always configure the inclusions/exclusions in the webhook deployment manifest, in addition to policy enforcement as needed.

PR #449, opened 2022-12-22
webhook: Ensure webhook gets called for pods created in all user namespaces

bpradipt:
Fixes: #448

### DIFF
--- a/webhook/config/webhook/manifests.yaml
+++ b/webhook/config/webhook/manifests.yaml
@@ -24,9 +24,12 @@ webhooks:
     - UPDATE
     resources:
     - pods
+    scope: Namespaced
   sideEffects: None
   namespaceSelector:
     matchExpressions:
-      - key: peerpods
-        operator: In
-        values:  ["true"]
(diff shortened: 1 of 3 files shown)


jensfr (approved):
Hey @bpradipt, lgtm, thank you! Just to summarize how I understand the changes,  the mutation is now opt-out instead of opt-in and the webhook will be run for all pods except those that are excluded by the namespaceSelector.  That prevents the problem that when a user forgets to label the namespace the pod will be started anyway but not be accounted for in the extended resources.

bpradipt:
> Hey @bpradipt, lgtm, thank you! Just to summarize how I understand the changes, the mutation is now opt-out instead of opt-in and the webhook will be run for all pods except those that are excluded by the namespaceSelector. That prevents the problem that when a user forgets to label the namespace the pod will be started anyway but not be accounted for in the extended resources.

Right

snir911 (approved):
LGTM, thanks also for the explanation  Jens ;)

PR #451, opened 2022-12-22
controller: set default values for name of config map and secret

### LINKED ISSUE #450: set default values for config map and secret names in PeerPodConfig CRD
We use `peer-pods-secret` and `peer-pods-cm` as the default names. If we use those also in the CRD then users don't have to explicitly set it.

jensfr:
use kubebuilder annotations in the API definition. The annotations will be used to set the default names in the PeerPodConfig CRD when it is generated via make manifests.

Fixes #450 


snir911 (approved):

bpradipt (approved):
/lgtm
Thanks @jensfr

Issue #455, opened 2022-12-22
aws: have an helper script to convert qcow/raw image to ami


snir911:
by uploading it to s3 and register it as snapshot.

PR #461, opened 2022-12-22
aws: helper script for ami creation

snir911:
it uploads a raw image to a s3 bucket and creates an ami based on it

Fixes: #455

### DIFF
--- a/aws/raw-to-ami.sh
+++ b/aws/raw-to-ami.sh
@@ -1,0 +1,176 @@
+#!/usr/bin/env bash
+
+[ "$DEBUG" == 'true' ] && set -x
+
+ARGC=$#
+if [ $ARGC -ne 3 ]; then
+    echo "USAGE: $(basename $0) <Image Path> <S3 Bucket Name> <Region>"
+    exit 1
+fi
+IMAGE_FILE_PATH=$1
+BUCKET_NAME=$2
+REGION=$3
+
+IMAGE_NAME="$(basename -- ${IMAGE_FILE_PATH})"
+AMI_NAME="${IMAGE_NAME%.*}"
+FORMAT="${IMAGE_NAME#*.}"
+
+TMPDIR=$(mktemp -d)
+IMAGE_IMPORT_JSON_FILE="${TMPDIR}/image-import.json"
+AMI_REGISTER_JSON_FILE="${TMPDIR}/register-ami.json"
+TRUST_POLICY_JSON_FILE="${TMPDIR}/trust-policy.json"
+ROLE_POLICY_JSON_FILE="${TMPDIR}/role-policy.json"
+BUCKET_POLICY_JSON_FILE="${TMPDIR}/bucket-policy.json"
+
+pre_checks() {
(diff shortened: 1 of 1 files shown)


aws/raw-to-ami.sh
```diff
@@ -0,0 +1,177 @@
...
+pre_checks() {
+	[[ -x "$(command -v jq)" ]] || { echo "jq is not installed" 1>&2 ; exit 1; }
+	[[ -x "$(command -v aws)" ]] || { echo "aws is not installed" 1>&2 ; exit 1; }
+	[[ "$FORMAT" == "raw"  ]] || { echo "image must be \"raw\", convert with: \"qemu-img convert ${IMAGE_NAME} ${AMI_NAME}.raw\"" 1>&2 ; exit 1; }
+	aws sts get-caller-identity &>/dev/null || { echo "awc cli missing credentials"; exit 1; }
+}
+
+create_bucket() {
+	echo "Create s3 Bucket"
+	aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
```
jensfr, line 35: when the bucket already exists, will this return an error? will the script exit immediately because of 'set -e'? Would it be appropriate to ignore the error if it means that the bucket already exists?

aws/raw-to-ami.sh
```diff
@@ -0,0 +1,177 @@
...
+			"Effect":"Allow",
+			"Principal":{ "Service":"vmie.amazonaws.com" },
+			"Action": "sts:AssumeRole",
+			"Condition":{"StringEquals":{"sts:Externalid":"vmimport"}}
+		}
+	]
+}
+EOF
+
+	aws iam create-role --role-name vmimport --assume-role-policy-document "file://${TRUST_POLICY_JSON_FILE}"
```
jensfr, line 54: same question here as for the bucket creation. what if the policy already exists? I guess in this case it may be fine that the script aborts and the user has to clean up the environment first.

aws/raw-to-ami.sh
snir911: Indeed, I'll address this

wainersm:
hi @snir911 , I am just curious: can this be done with terraform scripts?

wainersm (approved):
LGTM

snir911:
possibly I didn't look at terraform

aws/raw-to-ami.sh
```diff
@@ -0,0 +1,176 @@
...
+    "Description": "Peer Pod VM image",
+    "Format": "RAW",
+    "UserBucket": {
+	"S3Bucket": "${BUCKET_NAME}",
+	"S3Key": "${IMAGE_NAME}"
+    }
+}
+EOF
+
+	IMPORT_TASK_ID=$(aws ec2 import-snapshot --disk-container "file://${IMAGE_IMPORT_JSON_FILE}" | jq -r '.ImportTaskId')
```
snir911, line 113: TODO: missing `--output json` i think

bpradipt (approved):
/lgtm

Issue #464, opened 2022-12-27
controller, fetch CLOUD_PROVIDER from either configMap or Secret


snir911:
Currently it assumes it must be preset in a Configmap

PR #465, opened 2022-12-27
controller: fetch CLOUD_PROVIDER from either cm or secret

snir911:
instead of only ConfigMap

Fixes: #464

### DIFF
--- a/peer-pod-controller/controllers/peerpodconfig_controller.go
+++ b/peer-pod-controller/controllers/peerpodconfig_controller.go
@@ -121,21 +121,37 @@ func (r *PeerPodConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
-// Check if "peer-pods-secret" exists and has valid cloud provider name set in CLOUD_PROVIDER
+// Check if "peer-pods-secret" or "peer-pods-cm" exists and has valid cloud provider name set in CLOUD_PROVIDER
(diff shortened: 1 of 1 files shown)


peer-pod-controller/controllers/peerpodconfig_controller.go
```diff
@@ -121,21 +121,35 @@ func (r *PeerPodConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
-// Check if "peer-pods-secret" exists and has valid cloud provider name set in CLOUD_PROVIDER
+// Check if "peer-pods-secret" or "peer-pods-cm" exists and has valid cloud provider name set in CLOUD_PROVIDER
+// logic allows a valid CLOUD_PROVIDER value to be set only in one of the objects, if was set in both, value must be
+// equal and valid
 func (r *PeerPodConfigReconciler) peerpodCloudProviderIsValid() error {
 	peerpodscm := corev1.ConfigMap{}
+	peerpodssecret := corev1.Secret{}
```
bpradipt, line 129: Do we really need this? What are the issues if we continue to retrieve the provider from the configmap only ?

snir911: It's not really a problem, but the idea is to avoid the requirement to have both, it's only for simplification purposes

jensfr (approved):
lgtm, thank you @snir911

bpradipt (approved):
/lgtm
Thanks @snir911 
The WIP label in the title should be removed :-)

Issue #164, opened 2022-08-11
podvm: Remove skopeo and umoci from the peer pod VM image


stevenhorsman:
Once the skopeo to image-rs gaps have been resolved (such as signature verification in [https://github.com/kata-containers/kata-containers/issues/4888](https://github.com/kata-containers/kata-containers/issues/4888) and authenticated registry support in [https://github.com/kata-containers/kata-containers/issues/4601](https://github.com/kata-containers/kata-containers/issues/4601)) we should try removing all references to the umoci/skopeo binaries from the peer pod vm images (including the `USE_SKOPEO `, rather than just default to them not being used as per #256) and check that they still work as expected and there aren't any other skopeo specific features we were relying on in any cloud-providers.

The reason being that once we remove the kata-agent references to Skopeo and Umoci then we won't be able to use any skopeo/umoci functions, so want to understand this before we remove an upstream feature we require.

ariel-adam:
@stevenhorsman is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

stevenhorsman:
Yes, this is relevant and we will hopefully have it for the mid-November release. Yohei is already looking into it IIRC

ariel-adam:
I've mapped it to the CoCo release project, priority medium and drop V0.2.0 as the release

yoheiueda:
I raised a duplicated issue https://github.com/confidential-containers/cloud-api-adaptor/issues/256, and PR #256 has already closed the issue.

bpradipt:
@stevenhorsman any additional work pending w.r.to this issue ?

stevenhorsman:
@bpradipt - yes. As per the description in this - it is about removing all references to building skopeo and umoci in our peer pod VM image builds, rather than just changing the default as #263 did. I want this as a gate for removing the ability to use skopeo in the kata-agent as only by removing the ability to use skopeo (which we know people are still using for authenticated registry support) will we drive out the remaining cases that need skopeo. Does that make sense?

Issue #428, opened 2022-12-16
skopeo is getting built and included even without explicitly asking for it


bpradipt:
The default set of binaries to be included in the podvm image includes the pause image - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/Makefile.inc#L45

And that has a dependency on skopeo - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/Makefile.inc#L118

I think this needs to be de-coupled to completely switch to image-rs

cc @stevenhorsman @snir911

stevenhorsman:
So I might be wrong, but my understanding from a quick scan is that skopeo and umoci are built on the 'dev machine' to pull down the pause image, but in the pause image target they aren't built into the peer pod VM's file directory and therefore the image as it shouldn't add this target:
```
$(SKOPEO): $(SKOPEO_SRC)/bin/skopeo
	install -D --compare "$(SKOPEO_SRC)/bin/skopeo" "$@"
```
to the binary step, so I'm totally fine with skopeo and umoci used for dev (and that's the case in CoCo for adding the pause image to the guest image too.

It just means that I need to be more careful in https://github.com/confidential-containers/cloud-api-adaptor/issues/164 to not remove skopeo and umoci from being built, but just from being added to the binary list as happens in:
```
ifdef USE_SKOPEO
BINARIES += $(SKOPEO) $(UMOCI)
endif
```

If I'm correct then I think this issue can be closed as it's probably not worth the effort to get image-rs running standalone during the dev build at this point.

bpradipt:
I have to verify if we really need the pause image embedded in the peer pods case when using unmodified containerd and crio. 
If it's optional then I would add pause to the optional list as well and reduce the build time.

yoheiueda:
Skopeo and umoci are now only required for embedding the pause image, and are not included in the pod VM image.

When I implemented the support of embedding the pause image, kata-agent failed with an error when no pause image is available in the pod VM image. There was no option to disable it.
https://github.com/confidential-containers/cloud-api-adaptor/issues/196#issuecomment-1241524111

Issue #196, opened 2022-09-09
Integrate pause container image in pod VM image


yoheiueda:
Pause container image is now embedded in the rootfs image of the upstream Kata Containers.

https://github.com/kata-containers/kata-containers/issues/4768

We also need to embed a pause image in the pod VM image.

bpradipt:
> Pause container image is now embedded in the rootfs image of the upstream Kata Containers.
> 
> [kata-containers/kata-containers#4768](https://github.com/kata-containers/kata-containers/issues/4768)
> 
> We also need to embed a pause image in the pod VM image.

@yoheiueda is it really needed for peer-pods ? I understand this is needed when not using peer-pods, till complete support for image. offload is available in containerd/crio.

yoheiueda:
@bpradipt 

The current kata-agent always use the embedded pause image, when PullImage is called for pause image.

https://github.com/kata-containers/kata-containers/blob/55c8c7226d0b81d383e609c7a5a7f3da5652313b/src/agent/src/image_rpc.rs#L239-L240

If a pause image is not embedded, PullImage fails with error message `Pause image not present in rootfs`.
https://github.com/kata-containers/kata-containers/blob/55c8c7226d0b81d383e609c7a5a7f3da5652313b/src/agent/src/image_rpc.rs#L155

So, fixing it requires to modify kata-agent to make the feature optional.

bpradipt:
> @bpradipt
> 
> The current kata-agent always use the embedded pause image, when PullImage is called for pause image.
> 

Ah :-(.. Adds more dependencies for the VM image.

Anyways, looks like this need to be addressed within kata-containers and currently we have to live with it. 
@yoheiueda, thanks for the clarification.

yoheiueda:
@bpradipt can you also apply this change to VM images for other platforms such as AWS? Do I need to reopen this issue, or will you open another issue?

PR #476, opened 2023-01-03
image: Remove skopeo & umoci

stevenhorsman:
- Update Makefiles to remove the skopeo and umoci targets that install them into the peer pod image.
- Note: The skopeo and umoci source targets are still needed to copy and unpack the pause image for now.

Fixes: #164

### DIFF
--- a/azure/image/Makefile
+++ b/azure/image/Makefile
@@ -33,14 +33,8 @@ IMAGE_FILE := $(IMAGE_NAME)
 AGENT_PROTOCOL_FORWARDER = $(FILES_DIR)/usr/local/bin/agent-protocol-forwarder
 KATA_AGENT    = $(FILES_DIR)/usr/local/bin/kata-agent
 PAUSE         = $(FILES_DIR)/$(PAUSE_BUNDLE)/rootfs/pause
-SKOPEO    = $(FILES_DIR)/usr/bin/skopeo
-UMOCI     = $(FILES_DIR)/usr/local/bin/umoci
 
 BINARIES = $(AGENT_PROTOCOL_FORWARDER) $(KATA_AGENT) $(PAUSE)
-
-ifdef USE_SKOPEO
-BINARIES += $(SKOPEO) $(UMOCI)
-endif
 
 AGENT_PROTOCOL_FORWARDER_SRC = ../..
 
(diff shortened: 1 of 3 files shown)


stevenhorsman:
\cc @bpradipt @knrt10 - I've tested this with an s390x IBM Cloud peer pod manually, but it would be good to get a review to check that there aren't any issues with AWS, or Azure. Thanks!

knrt10:
Will check with Azure and update

knrt10 (approved):
Tested on Azure, works. +1

bpradipt:
@stevenhorsman I tested successfully on libvirt. I don't think there will be any issues with AWS. 

One suggestion though. Does it make sense to remove the code to handle `pause_bundle` in this PR?  

For peer pods we don't seem to require pause_bundle to be embedded into the podvm image as the pause image is always pulled as shown below:
```
2023/01/04 09:21:15 [adaptor/proxy] getImageName: check if its a sandbox type container
2023/01/04 09:21:15 [adaptor/proxy] getImageName: pause image: quay.io/bpradipt/okd-pause
2023/01/04 09:21:15 [adaptor/proxy] CreateContainer: calling PullImage for "quay.io/bpradipt/okd-pause" before CreateContainer (cid: "f737f189f0a9dfe047848c65250015ceb13286dbe6390280e21b6a80c0adf3eb")
2023/01/04 09:21:20 [adaptor/proxy] CreateContainer: successfully pulled image "quay.io/bpradipt/okd-pause"
2023/01/04 09:21:20 [adaptor/proxy] StartContainer: containerID:f737f189f0a9dfe047848c65250015ceb13286dbe6390280e21b6a80c0adf3eb
2023/01/04 09:21:20 [adaptor/proxy] CreateContainer: containerID:efe1969d0031cb7125b52c75393c1707ecfb599058321526d5aff1afa6773ad9
```

It's fine to handle it in a separate PR. But since this PR removes umoci/skopeo I thought about suggesting to remove even the pause image download/unpack/embed stuff and completely doing away with umoci/skopeo.

stevenhorsman:
> One suggestion though. Does it make sense to remove the code to handle `pause_bundle` in this PR?

I'm very happy to get rid of the pause bundle code in this PR, but I'd like to understand what has changed since https://github.com/confidential-containers/cloud-api-adaptor/issues/196#issuecomment-1241524111 when we did need the pause bundle as the agent code doesn't seem to have changed?

I'm wondering if the proxy log is a red herring as the agent would likely report to the proxy that it did the pull, even though it is done locally.

/cc @yoheiueda in case I've missed something.

stevenhorsman:
> I removed the pause_bundle from the pod VM image to verify. I'm using crio. Can you try by removing the pause_bundle from the podVM image in your ibmcloud setup if possible ?

Yeah, I'll set and environment up to test that.

bpradipt (approved):
/lgtm
Thanks @stevenhorsman

stevenhorsman:
> So there is definitely some permutations which works ;) but it's hard to tell without a deeper investigation. But your test have proved that we can't remove it yet. So let's go ahead with this PR and will investigate the pause image stuff later on.

Thanks, I guess we can use https://github.com/confidential-containers/cloud-api-adaptor/issues/428 to track this work?

Issue #477, opened 2023-01-03
Add support for -version to see the current version and commit


Jordan9500:
Add support for `-version` to work across all providers. Currently the command is hard-coded and does not work for other providers except, `ibmcloud` and `aws`.
Updating the message to be more concise, something similar to: 
```
/usr/local/bin/cloud-api-adaptor -version 
cloud-api-adaptor, version: unknown, commit: <commit-hash>
go: go<go-version>
```
This will allow for quicker debugging by narrowing down the version of code and commit in question.

PR #478, opened 2023-01-03
all: Add support for version command

Jordan9500:
When doing version it will display the commit and version 
Update gitignore with the new files generated by the make 
Update the display for when the args are less than 2 
Use a util file to store the Version and Commit Hash

Fixes: #477
Signed-off-by: Jordan Jackson <jordan.jackson@ibm.com>

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -35,8 +35,13 @@ type networkConfig struct {
 func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
 	if len(os.Args) < 2 {
-		fmt.Printf("%s aws|azure|ibmcloud|libvirt|vsphere <options>\n", os.Args[0])
+		fmt.Printf("%s \n<aws|azure|ibmcloud|libvirt|version> [options]\n", os.Args[0])
 		cmd.Exit(1)
+	}
+
+	if os.Args[1] == "version" {
+		cmd.ShowVersion(programName)
+		cmd.Exit(0)
 	}
 
 	//TODO: transition to better CLI library
--- a/cmd/parse.go
+++ b/cmd/parse.go
@@ -14,12 +14,12 @@ func Parse(programName string, args []string, fn func(flags *flag.FlagSet)) {
 
 	fn(flags)
 
-	if len(args) == 0 {
+	if len(args) == 1 {
 		flags.PrintDefaults()
 		Exit(1)
 	}
 
-	if err := flags.Parse(args); err != nil {
+	if err := flags.Parse(args[1:]); err != nil {
 		if errors.Is(err, flag.ErrHelp) {
 			Exit(0)
 		} else {
--- a/cmd/parse.go
+++ b/cmd/parse.go
@@ -9,39 +9,21 @@ import (
 )
 
 func Parse(programName string, args []string, fn func(flags *flag.FlagSet)) {
-
-	var versionFlag bool
-
 	flags := flag.NewFlagSet(programName, flag.ContinueOnError)
 	flags.SetOutput(flag.CommandLine.Output())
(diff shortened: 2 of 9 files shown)


Jordan9500:
With the changes being made in the refactor PR, I will wait before making further changes for that to be merged

Depends on: https://github.com/confidential-containers/cloud-api-adaptor/pull/329

yoheiueda (commented):
@Jordan9500 Could you update this PR if possible, since #329 has been merged.

Only providers IBM Cloud and AWS have been switched to the new code layout. It is OK If you want to wait until all of the providers need to be refactored.

Jordan9500:
I have updated the code with the latest changes. My plan with the version check was to implement it like: 
```
/usr/local/bin/cloud-api-adaptor version 
cloud-api-adaptor, version: unknown, commit: <current-commit-hash>
go: <go version>
```
I originally thought that adding the code to the top of the function using os.Args that was used before would allow the use across all providers. (Not just the ones moved to the new refactor)

In regards to the current failures, i'm looking into a different way that does not use the make file to generate a version file that gets used. Or atleast uses a file that has the variables defined as `unknown`.

cmd/parse.go
```diff
@@ -9,39 +9,21 @@ import (
...
-			Exit(1)
-		}
-
+	if len(args) == 0 {
+		flags.PrintDefaults()
+		Exit(1)
 	}
 
-	if err := flags.Parse(args[1:]); err != nil {
+	if err := flags.Parse(args); err != nil {
```
yoheiueda, line 22: Please keep to use `args[1:]` to suppress the unit test failures. `flags.Parse` assumes `args[0]` is a program name.

yoheiueda (changes requested):
Could you use the `-ldflags`technique to embed a version information in a binary, instead of modifying `version.go`? Modifying a git-managed file may cause confusion.

https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

In the future, we need to switch to a modern third-party command line parser library. I originally implemented this simple parser, but I don't think it is suitable for the current complicated scenario.

yoheiueda:
Thank you for updating the patch.

The second commit messages needs to be formatted as your first commit message. Could you update your commit message? Note that at least one of commits needs to have a `Fixes` line.

Jordan9500:
I have updated the PR to match the changes you have requested, and got the ldflags working. 
This is now the output, without the generated file steps from the previous commit.
```
/usr/local/bin/cloud-api-adaptor version
cloud-api-adaptor, version: unknown, commit: 3afd0f5bd1bc338dc44213ef300ed3bc05a00b32
go: go1.18.5
```

yoheiueda:
@Jordan9500 could you rebase to resolve the conflicts in main.go? Cloud providers for azure and libvirt have been merged.

Jordan9500:
I have rebased with the latest changes

yoheiueda (approved):
LGTM. Thank you!

bpradipt (approved):
/lgtm
Thanks @Jordan9500

cmd/cloud-api-adaptor/main.go
```diff
@@ -35,10 +35,15 @@ type networkConfig struct {
 func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
 	if len(os.Args) < 2 {
-		fmt.Printf("%s aws|azure|ibmcloud|libvirt|vsphere <options>\n", os.Args[0])
+		fmt.Printf("%s \n<aws|azure|ibmcloud|libvirt|version> [options]\n", os.Args[0])
```
stevenhorsman, line 38: @Jordan9500 - did you mean to remove vsphere here?

Jordan9500: Good spot, I did not mean to remove that from the print. 
I will make another PR to re add that command to the print. 
Thank you :)

PR #485, opened 2023-01-04
aws: fix bucket creation in helper script

snir911:
to avoid failure on some of the regions

Fixes: #455

### DIFF
--- a/aws/raw-to-ami.sh
+++ b/aws/raw-to-ami.sh
@@ -31,7 +31,11 @@ pre_checks() {
 
 create_bucket() {
 	echo "Create s3 Bucket"
-	aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
+	if [[ ${REGION} == us-east-1 ]]; then
+		aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
+	else
+		aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION}
+	fi
 }
 
 set_bucket_policies() {


aws/raw-to-ami.sh
```diff
@@ -31,7 +31,7 @@ pre_checks() {
 
 create_bucket() {
 	echo "Create s3 Bucket"
-	aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
+	aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION}
```
bpradipt, line 34: @snir911 looks like the `LocationConstraint` option doesn't work for `us-east-1`. So you'll need to use a conditional based on REGION

whitebrandy: Indeed, the docs at https://docs.aws.amazon.com/cli/latest/reference/s3api/create-bucket.html do not list us-east-1 as a valid value for LocationConstraint.

snir911:
while here, i appended the bucket name to the role and policy

whitebrandy (approved):
lgtm

aws/raw-to-ami.sh
```diff
@@ -44,13 +48,13 @@ set_bucket_policies() {
 			"Effect":"Allow",
 			"Principal":{ "Service":"vmie.amazonaws.com" },
 			"Action": "sts:AssumeRole",
-			"Condition":{"StringEquals":{"sts:Externalid":"vmimport"}}
+			"Condition":{"StringEquals":{"sts:Externalid":"vmimport-${BUCKET_NAME}"}}
```
bpradipt, line 51: The policy name must be `vmimport` - https://docs.aws.amazon.com/vm-import/latest/userguide/required-permissions.html. Relevant text:

***
VM Import/Export requires a role to perform certain operations on your behalf. You must create a service role named vmimport with a trust relationship policy document that allows VM Import/Export to assume the role, and you must attach an IAM policy to the role. For more information, see [IAM Roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/roles-toplevel.html) in the IAM User Guide.
***

snir911: thanks, removed

bpradipt (approved):
/lgtm
Thanks @snir911

Issue #502, opened 2023-01-09
Install open-vm-tools when creating qcow2 image for vSphere


bpradipt:
open-vm-tools are required for vsphere pod VM images. This is handled when creating using vsphere packer provider but not when creating qcow2.

bpradipt:
@whitebrandy can you please take a look at it ?
I think one of the options is to add the install instructions in `misc-settings.sh`

PR #504, opened 2023-01-09
qcow2: template: attempt installing open-vm-tools for rhel and centos

whitebrandy:
This is useful if we are repurposing the qcow2 for a ovf.

Fixes: #502

### DIFF
--- a/podvm/qcow2/centos/qemu-centos.pkr.hcl
+++ b/podvm/qcow2/centos/qemu-centos.pkr.hcl
@@ -69,6 +69,7 @@ build {
   provisioner "shell" {
     remote_folder = "~"
     inline = [
+      "sudo dnf -y install open-vm-tools || exit 0",
       "sudo bash ~/misc-settings.sh"
     ]
   }
--- a/podvm/qcow2/rhel/qemu-rhel.pkr.hcl
+++ b/podvm/qcow2/rhel/qemu-rhel.pkr.hcl
@@ -69,6 +69,8 @@ build {
   provisioner "shell" {
     remote_folder = "~"
     inline = [
+    # requires subscription
+      "sudo dnf -y install open-vm-tools || exit 0",
       "sudo bash ~/selinux_relabel.sh"
     ]
   }


podvm/qcow2/centos/qemu-centos.pkr.hcl
```diff
@@ -69,6 +69,7 @@ build {
   provisioner "shell" {
     remote_folder = "~"
     inline = [
+      "sudo dnf -y install open-vm-tools || exit 0",
```
bpradipt, line 72: how about adding this to misc_settings.sh file ?
Also any thoughts on how this can be installed only for vsphere ?

whitebrandy: I used the script because misc_settings.sh is common for all including ubuntu. I think we should follow the ubuntu example and install open-vm-tools by default. That way, we don't have to worry about CLOUD_PROVIDER specific qcow2 images. What do you think ?

bpradipt: Got it. Can you also add it to ubuntu ?
Also later on it will be good to figure out a way to install specific packages for specific providers

whitebrandy: Actually, ubuntu already has open-vm-tools installed in the upstream cloud image. So, when converting the ubuntu qcow2, we don't need to install additional packages.

bpradipt: Thanks for clarifying @whitebrandy

snir911 (approved):
LGTM i guess changing it to be applied only for vsphere provider can be applied in follow up pr if wanted

caavery:
I was not able to create a peerpods vm with this commit due to vmware tools was not installed. This may be an error on my part or not having the permissions to download and install.

bpradipt:
> I was not able to create a peerpods vm with this commit due to vmware tools was not installed. This may be an error on my part or not having the permissions to download and install open-vm-tools. I did not receive any errors or warnings. Let me know what you think.

Can you check in the pod VM if open-vm-tools got installed or not ? Also for rhel you will need to build the image on a rhel system which has subscription enabled.

whitebrandy:
> I was not able to create a peerpods vm with this commit due to vmware tools was not installed. This may be an error on my part or not having the permissions to download and install open-vm-tools. I did not receive any errors or warnings. Let me know what you think.

Note that you need to setup subscription info to be able to successfully install. 
The build not failing is deliberate. We don't want to fail when someone wants to create an image but doesn't want to setup subscription.

caavery:
Yep I pulled down a registered image but neglected to subscribe.

caavery:
OK I'm withdrawing my concern.

Issue #506, opened 2023-01-11
vsphere: add a force option to vmdk-to-ovf.sh


whitebrandy:
Currently, the script works fine on the first run but will fail later on. Since the vm has a somewhat esoteric configuration, i.e. a vmdk that has been copied is used to create the VM and is in a different folder, removing the template is not a straightforward process.

PR #508, opened 2023-01-11
vsphere: vmdk-to-ovf.sh: add a force option

whitebrandy:
This option overwrites an existing template by making sure to delete template files, if they exist, before the current run.

The default is yes - always assume a force install.

Fixes: #506


caavery (approved):
LGTM

whitebrandy:
> /lgtm @whitebrandy assuming you have tested this :-)

Ofcourse!

whitebrandy:
> > /lgtm @whitebrandy assuming you have tested this :-)
> 
> Ofcourse!

I mean, not that I can guarantee that there are absolutely no bugs but I would have to be *really really* confident not to test my PowerCLI scripting!

Issue #515, opened 2023-01-12
controller: make install fails with required binary not found


Amulyam24:
```
peer-pod-controller % make install
/Users/amulyameka/go/src/github.com/caa/peer-pod-controller/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
bash: /Users/amulyameka/go/src/github.com/caa/peer-pod-controller/bin/controller-gen: No such file or directory
make: *** [manifests] Error 127
```

PR #516, opened 2023-01-12
controller: download binaries via go install instead of go get

Amulyam24:
Installing via `go get` is deprecated since 1.17. Instead, use `go install` to download the binaries.

Fixes: #515

Signed-off-by: Amulyam24 <amulmek1@in.ibm.com>

### DIFF
--- a/peer-pod-controller/Makefile
+++ b/peer-pod-controller/Makefile
@@ -51,6 +51,8 @@ IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
 
+CONTROLLER_TOOLS_VERSION = 0.10.0
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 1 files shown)


peer-pod-controller/Makefile
```diff
@@ -150,31 +150,17 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 .PHONY: controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
+	GOBIN=$(shell pwd)/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0
```
jensfr, line 153: Hi, thanks for this patch, it's definitely a valid change. Maybe we can test if the binary already exists locally and only call go install if it's not? 

I think when you create a new controller with a recent operator-sdk it creates something like this:

controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
$(CONTROLLER_GEN): $(LOCALBIN)
	test -s $(LOCALBIN)/controller-gen || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)

What do you think?

Amulyam24: @jensfr, yes. That would be better. Thanks for the suggestion. Addressed the change, PTAL.

snir911:
This will usually require `sudo` now to install it in LOCALBIN, IIUC earlier it was installed under `$PWD/bin`, i think it would be preferred if possible, not sure if it make sense or not.

snir911:
should it be modified manually? IIUC the project is scaffolded with the Makefile automatically, would it be possible to update it using the operator-sdk? it seems to [switch](https://github.com/kubernetes-sigs/kubebuilder/blob/156196cca2e726ba0085487d8b5b1a209f3da523/testdata/project-v4/Makefile#L152) to go install already

Amulyam24:
> should it be modified manually? IIUC the project is scaffolded with the Makefile automatically, would it be possible to update it using the operator-sdk? it seems to [switch](https://github.com/kubernetes-sigs/kubebuilder/blob/156196cca2e726ba0085487d8b5b1a209f3da523/testdata/project-v4/Makefile#L152) to go install already

@jensfr, can you confirm if this would work? I'm not sure about this.

jensfr:
> should it be modified manually? IIUC the project is scaffolded with the Makefile automatically, would it be possible to update it using the operator-sdk? it seems to [switch](https://github.com/kubernetes-sigs/kubebuilder/blob/156196cca2e726ba0085487d8b5b1a209f3da523/testdata/project-v4/Makefile#L152) to go install already

I think I missed [this step](https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.21.0/#for-golang-based-operators--change-makefile-targets-to-replace-go-get-with-go-install) when upgrading operator-sdk. Even upgrading operator-sdk would mean changing the Makefile manually. So I think it's okay to change it manually, as described in the upgrade instructions. Does that make sense?

Amulyam24:
@jensfr @snir911, addressed the changes, PTAL.

snir911 (approved):
LGTM, thanks!

snir911 (changes requested):
Hi, just noticed, the `$(LOCAL_BIN)` can't be the first one as it becomes the default make target instead of `all`

snir911 (approved):
LGTM, thanks!

bpradipt (approved):
/lgtm
Thanks @Amulyam24

Issue #512, opened 2023-01-11
qcow2: Provide public s390x image


jtumber-ibm:
Build upon #404 and #372 to create a s390x image in gh actions

PR #522, opened 2023-01-13
podvm/github-action: Support building public s390x qcow images 

jtumber-ibm:
The `cd-rom` option for qemu doesn't work for s390x as it can't auto-detect the layout. Uses `virtio-scsi` instead works for both `s390x` and `amd64` for the cloud-init data. Addtionally to make sure it boots from the `qcow2` drive instead of the cloud-init, need to explictly define the device and drive.

Fixes: #512
    
### DIFF
--- a/podvm/Dockerfile.podvm
+++ b/podvm/Dockerfile.podvm
@@ -11,10 +11,16 @@ FROM ${BUILDER_IMG} AS podvm_builder
 ARG CLOUD_PROVIDER=libvirt
 ARG PODVM_DISTRO=ubuntu
 ARG AA_KBC=offline_fs_kbc
+# If not provided, uses system architecture
+ARG ARCH
 
 ENV CLOUD_PROVIDER ${CLOUD_PROVIDER}
 ENV PODVM_DISTRO ${PODVM_DISTRO}
 ENV AA_KBC ${AA_KBC}
+ENV ARCH ${ARCH}
+
+# Installs add-ons for foreign target, if required
+RUN ./cloud-api-adaptor/podvm/hack/cross-build-extras.sh
 
 # Default to Ubuntu Focal amd64 release image. These variables can be overriden as needed
 ARG UBUNTU_IMAGE_URL=https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
--- a/podvm/Makefile.inc
+++ b/podvm/Makefile.inc
@@ -15,6 +15,8 @@ IMAGE_PREFIX := podvm
 
 HOST_ARCH   := $(shell uname -m)
 ARCH        ?= $(HOST_ARCH)
+# Normalise x86_64 / amd64 for input ARCH
+ARCH        := $(subst amd64,x86_64,$(ARCH))
 DEB_ARCH    := $(subst x86_64,amd64,$(ARCH))
 LIBC        ?= $(if $(findstring s390x,$(ARCH)),gnu,musl)
 RUST_TARGET := $(ARCH)-unknown-linux-$(LIBC)
@@ -121,6 +123,6 @@ $(ATTESTATION_AGENT_SRC):
 	git clone "$(ATTESTATION_AGENT_REPO)" "$(ATTESTATION_AGENT_SRC)"
 
 $(ATTESTATION_AGENT): $(ATTESTATION_AGENT_SRC)
-	cd "$(ATTESTATION_AGENT_SRC)" && $(RUST_FLAGS) $(MAKE) KBC="$(AA_KBC)" LIBC="$(LIBC)"
+	cd "$(ATTESTATION_AGENT_SRC)" && CC= ARCH=$(ARCH) $(MAKE) KBC="$(AA_KBC)" LIBC="$(LIBC)"
 	mkdir -p "$(@D)"
 	install --compare "$(ATTESTATION_AGENT_SRC)/app/target/$(RUST_TARGET)/release/attestation-agent" "$@"
--- a/podvm/qcow2/ubuntu/qemu-ubuntu.pkr.hcl
+++ b/podvm/qcow2/ubuntu/qemu-ubuntu.pkr.hcl
@@ -7,14 +7,17 @@ source "qemu" "ubuntu" {
   headless          = true
   iso_checksum      = "${var.cloud_image_checksum}"
   iso_url           = "${var.cloud_image_url}"
-  output_directory  = "output"
-  qemuargs          = [["-m", "${var.memory}"], ["-smp", "cpus=${var.cpus}"], ["-cdrom", "${var.cloud_init_image}"], ["-serial", "mon:stdio"]]
+  output_directory  = "${var.output_directory}"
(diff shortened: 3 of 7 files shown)


jtumber-ibm:
Can check `docker.io/jamesnelson/podvm-ibmcloud-ubuntu-amd64` and `docker.io/jamesnelson/podvm-ibmcloud-ubuntu-s390x` for test runs of this change

podvm/Dockerfile.podvm
```diff
@@ -11,10 +11,17 @@ FROM ${BUILDER_IMG} AS podvm_builder
...
+# If not provided, uses system architecture
+ARG ARCH
 
 ENV CLOUD_PROVIDER ${CLOUD_PROVIDER}
 ENV PODVM_DISTRO ${PODVM_DISTRO}
 ENV AA_KBC ${AA_KBC}
+ENV ARCH ${ARCH}
+
+# Installs add-ons for foreign target, if required
+COPY hack/cross-build-extras.sh cross-build-extras.sh
```
bpradipt, line 23: The podvm_builder should include all files under podvm which includes `hack/*`. This statement is not needed. Also we should avoid copying from local system unless absolutely needed to keep the container build process isolated from host.

podvm/Makefile.inc
```diff
@@ -15,6 +15,8 @@ IMAGE_PREFIX := podvm
 
 HOST_ARCH   := $(shell uname -m)
 ARCH        ?= $(HOST_ARCH)
+# Normalise x86_64 / amd64 for input ARCH
+ARCH        := $(subst amd64,x86_64,$(ARCH))
```
bpradipt, line 19: ARCH is also used as ARG in the Dockerfile. If it's passed as container build arg, then ARCH will be set first in L17 and then modified. But still want to re-confirm from you :-) .. Do you see any issues ?

podvm/qcow2/ubuntu/qemu-ubuntu.pkr.hcl
```diff
@@ -7,14 +7,18 @@ source "qemu" "ubuntu" {
...
+  output_directory  = "${var.output_directory}"
+  qemuargs          = [["-device", "virtio-blk,drive=virtio-drive,id=virtio-disk0,bootindex=1"], ["-drive", "file=${var.output_directory}/${var.qemu_image_name},if=none,cache=writeback,discard=ignore,format=qcow2,id=virtio-drive"], ["-device", "virtio-scsi"], ["-drive", "file=${var.cloud_init_image},format=raw,if=none,id=c1"], ["-device", "scsi-cd,drive=c1"], ["-m", "${var.memory}"], ["-smp", "cpus=${var.cpus}"], ["-serial", "mon:stdio"]]
   ssh_password      = "${var.ssh_password}"
   ssh_port          = 22
   ssh_username      = "${var.ssh_username}"
   ssh_wait_timeout  = "300s"
+  boot_wait         = "${var.boot_wait}"
   vm_name           = "${var.qemu_image_name}"
-  shutdown_command  = "sudo shutdown -h now" 
+  accelerator       = "tcg"
```
bpradipt, line 18: If I understand correctly, the above statement will force the build to always use tcg. 
```
The appropriate software must have already been installed on your build machine to use the accelerator you specified. 
When no accelerator is specified, Packer will try to use kvm if it is available but will default to tcg otherwise.
```
I think it's better to not set it so that `kvm` is used when available.

podvm/Makefile.inc
```diff
@@ -121,6 +123,6 @@ $(ATTESTATION_AGENT_SRC):
 	git clone "$(ATTESTATION_AGENT_REPO)" "$(ATTESTATION_AGENT_SRC)"
 
 $(ATTESTATION_AGENT): $(ATTESTATION_AGENT_SRC)
-	cd "$(ATTESTATION_AGENT_SRC)" && $(RUST_FLAGS) $(MAKE) KBC="$(AA_KBC)" LIBC="$(LIBC)"
+	cd "$(ATTESTATION_AGENT_SRC)" && CC= ARCH=$(ARCH) $(MAKE) KBC="$(AA_KBC)" LIBC="$(LIBC)"
```
bpradipt, line 126: is this `CC=` correct ?

podvm/Makefile.inc
jtumber-ibm: It matching with the ARG in the Dockerfile is on purpose. L17 won't set it again if it is set as a build-arg in the docker build. 
L19 means that we can specify `amd64` `x86_64` for the build-arg and it will still work as expected.

podvm/qcow2/ubuntu/qemu-ubuntu.pkr.hcl
jtumber-ibm: That is the correct understanding, however I found it still tried to use kvm on my machine even when it wasn't available. Happy to switch it back as I think in the Docker build process it will always use `tcg`.

podvm/Makefile.inc
jtumber-ibm: Yes, attestation agent includes `libgit2` library which doesn't build when specifying `CC`.

bpradipt (approved):
/lgtm
Thanks @jtumber-ibm

stevenhorsman:
@jtumber-ibm - Sorry to ask a simple question, but after this are we going to make https://quay.io/repository/confidential-containers/podvm-ibmcloud-ubuntu multiarch, or have two separate images for amd64 & s390x?

jtumber-ibm:
> @jtumber-ibm - Sorry to ask a simple question, but after this are we going to make https://quay.io/repository/confidential-containers/podvm-ibmcloud-ubuntu multiarch, or have two separate images for amd64 & s390x?

No worries it is a good question. I think not - as there is no intention that this is a runnable image so having the convenience of the manifest list might lead people to mistakenly use the wrong image.

stevenhorsman (approved):
I'm definitely not an expert, but LGTM.
I was a bit nosy, so I had a look at the test images created and they seem reasonably (without trying to run the qcow2 images, just looking at the naming):
```
$ ./bin/skopeo copy --insecure-policy docker://docker.io/jamesnelson/podvm-ibmcloud-ubuntu-amd64 oci:/tmp/test_amd64:latest
Getting image source signatures
Copying blob 3cee1b4adf11 skipped: already exists
Copying config fc02df85b6 done
Writing manifest to image destination
Storing signatures
$ sudo umoci unpack --image /tmp/test_amd64:latest /tmp/amd_bundle
$ ls /tmp/amd_bundle/rootfs/
podvm-e6102dc-dirty-amd64.qcow2
$ ./bin/skopeo copy --insecure-policy docker://docker.io/jamesnelson/podvm-ibmcloud-ubuntu-s390x oci:/tmp/test_s390x:latest
Getting image source signatures
Copying blob 52c32d2fa07a done
Copying config 3bb2b14095 done
Writing manifest to image destination
Storing signatures
$ sudo umoci unpack --image /tmp/test_s390x:latest /tmp/s390x_bundle
$ sudo ls /tmp/s390x_bundle/rootfs
podvm-e6102dc-dirty-s390x.qcow2
```

PR #525, opened 2023-01-15
reduce container builds layers size

snir911:


podvm/Dockerfile.podvm
```diff
@@ -8,7 +8,7 @@ ARG BUILDER_IMG="quay.io/confidential-containers/podvm-builder-ubuntu"
 
 FROM ${BUILDER_IMG} AS podvm_builder
 
-ARG CLOUD_PROVIDER=libvirt
+ARG CLOUD_PROVIDER
```
jtumber-ibm, line 11: Any particular reason for removing the default?

podvm/Makefile.inc
```diff
@@ -11,6 +11,12 @@ UBUNTU_RELEASE     = focal
 SKOPEO_VERSION     = 1.5.0
 UMOCI_VERSION      = 0.4.7
 
+ifeq ($(FORCE),false)
```
jtumber-ibm, line 14: could you explain this `force` change? I don't understand why it is required for this change, sorry if I am missing something

snir911: sure, AFAIU force is used to force rebuilding of the forwarder and agent, in the container builds this results in double build, once by `make binaries` and later by `make image` while nothing has changed and process is fixed and always running on clean environment. Using FORCE allows to "unforce" these targets by setting FORCE=false

podvm/Dockerfile.podvm
snir911: TBH it's not tightly coupled with this PR, removing the default makes the container runtime to print a warning if the variable is not set, just to avoid mistakes, nothing else

jtumber-ibm: I'd prefer to leave it set, but not a strong perference

podvm/Makefile.inc
jtumber-ibm: Would this PR have the same result without the force changes and just adding the `clean_sources` target?

In general I'd prefer the pattern of `make clean image` if an end user wanted to force a rebuild. For me doing `make binaries` followed by `make image` doesn't rebuild the binaries already.

snir911: The target are forced, thing is, if you don't do clean_sources target will be forced but the targets in the source folder makefiles will be skipped
I'll try to do something else

snir911: I removed the FORCE thing and changed it to a single stage build and added the clean_sources

wainersm:
hi @snir911, I appreciate you have worked on this as to build the podvm in containers it generates layers with gigabytes. I had to purge the images many times last week because my workstation was running out of space :)

While here, could you also reduce the size of the podvm_builder images?

When I remember I pass the hadolint (https://github.com/hadolint/hadolint) on the dockerfiles I create to catch problems. For the podvm_builder dockerfile, hadolint recommends (apart from other stuffs) to pass `--no-install-recommends` to apt-get as well as consolidate the many `RUN`; those changes should reduce the layers amount and size.

wainersm:
Before this PR:

```
$ docker images
REPOSITORY      TAG       IMAGE ID       CREATED          SIZE
podvm_libvirt   latest    d09643afc472   3 minutes ago    704MB
<none>          <none>    c6c79939f060   3 minutes ago    9.05GB
podvm_builder   latest    e93438e22e2d   33 minutes ago   2.83GB
ubuntu          20.04     d5447fc01ae6   6 weeks ago      72.8MB
```
After:
```
REPOSITORY      TAG       IMAGE ID       CREATED             SIZE
podvm_libvirt   latest    c67c99ccbd07   38 minutes ago      704MB
<none>          <none>    37f58ff90d2b   38 minutes ago      5.67GB
podvm_builder   latest    e3efdd254638   About an hour ago   2.83GB
ubuntu          20.04     d5447fc01ae6   6 weeks ago         72.8MB
```

jtumber-ibm (approved):
lgtm thanks @snir911

snir911:
* added go modcache cleanup and moved go cleaning to dockerfiles
* added --no-install-recommends as suggested by hadolint (most other warnings are not size related)
It saves a lot in the intermediate layers, hopefully it would help to reduce the resources (disk/mem) requirements of the building machine

podvm/Dockerfile.podvm
whitebrandy: Forcing the user to set CLOUD_PROVIDER will help with preventing bugs because someone forgot to set CLOUD_PROVIDER with the right target for the image. (However, as of today, I don't think there are any such dependenices for the image. )

wainersm:
@snir911 is it ready to merge? I mean, with my approval?

podvm/Dockerfile.podvm
```diff
@@ -31,12 +31,13 @@ ENV UBUNTU_IMAGE_CHECKSUM ${UBUNTU_IMAGE_CHECKSUM}
 
 
 # Building binaries separately allows for reuse in case of error
-# with image building step
-RUN cd cloud-api-adaptor/podvm && \
-     make binaries
+# with image building step OTOH it may increase layers size
```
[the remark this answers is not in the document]
snir911: I left it as it's helpful to uncomment this in case you want the binaries build to be cached

snir911:
yes, I wasn't sure if it's actually helpful in reducing the disk size required from the host, it does seem to help, ready on my side

bpradipt (approved):
/lgtm
I think having a build arg to decide whether to create a separate layer for the built binaries will be more helpful. But that can be done as a feature addition.

PR #526, opened 2023-01-16
cmd: Display proper error message when invalid provider is passed

Amulyam24:
When an invalid cloud provider is passed, it is currently not being handled by the adaptor. This PR adds a fix to display a message instead of failing silently.

Fixes: #523

Signed-off-by: Amulyam24 <amulmek1@in.ibm.com>

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/confidential-containers/cloud-api-adaptor/cmd"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/hypervisor"
@@ -24,6 +25,8 @@ import (
 )
 
 const programName = "cloud-api-adaptor"
+
+var supportedProviders = []string{"aws", "azure", "ibmcloud", "libvirt", "vsphere"}
 
 type daemonConfig struct {
 	helperDaemonRoot  string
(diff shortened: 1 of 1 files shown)


Amulyam24:
@surajssd, provided a fix for #523. PTAL.

cmd/cloud-api-adaptor/main.go
```diff
@@ -46,8 +48,13 @@ var hypcfg hypervisor.Config
 func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
 	if len(os.Args) < 2 {
-		fmt.Printf("%s aws|azure|ibmcloud|libvirt <options>\n", os.Args[0])
+		fmt.Printf("%s aws|azure|ibmcloud|libvirt|vsphere <options>\n", os.Args[0])
 		cmd.Exit(1)
+	} else if len(os.Args) == 2 {
```
bpradipt, line 53: I think a better option will be something like this:
```
if len(os.Args) < 2 || !contains(os.Args[1], supportedProviders)  {
...
}
```
wdyt ?

Amulyam24: @bpradipt, this is better. Changed it. PTAL

bpradipt (approved):
/lgtm
Thanks @Amulyam24

cmd/cloud-api-adaptor/main.go
```diff
@@ -45,8 +47,8 @@ var hypcfg hypervisor.Config
 
 func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
-	if len(os.Args) < 2 {
-		fmt.Printf("%s aws|azure|ibmcloud|libvirt <options>\n", os.Args[0])
+	if len(os.Args) < 2 || !contains(os.Args[1], supportedProviders) {
+		fmt.Printf("Invalid or no cloud provider passed.\n%s aws|azure|ibmcloud|libvirt|vsphere <options>\n", os.Args[0])
```
littlejawa, line 51: Can we keep a single list of strings to maintain?
Something like: 

```suggestion
		fmt.Printf("Invalid or no cloud provider passed.\n%s %s <options>\n", os.Args[0], strings.Join(supportedProviders, "|"))
```

Amulyam24: Changed it.

surajssd (approved):

Issue #548, opened 2023-01-19
open-vm-tools clean install fails during provisioning


whitebrandy:
If we install open-vm-tools via the provisioner, selinux blocks %post because of inconsistent labels.

PR #549, opened 2023-01-19
podvm: rhel: relabel before running misc-settings.sh

whitebrandy:
The relabeling in the provisioner is intended to fix labeling when packer copies over files to the image. It's run at the end during provisioning. However, misc-settings.sh may install packages and configure services to boot at startup. If the labels are not correct, this will fail.

Move labeling to right after we copy our target dir and run misc-settings in the end.

Fixes: #548

### DIFF
--- a/podvm/qcow2/rhel/qemu-rhel.pkr.hcl
+++ b/podvm/qcow2/rhel/qemu-rhel.pkr.hcl
@@ -49,6 +49,20 @@ build {
     ]
   }
 
+# relabel copied files right after copy-files.sh
+# to prevent other commands from failing
+  provisioner "file" {
+    source      = "qcow2/selinux_relabel.sh"
+    destination = "~/selinux_relabel.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    inline = [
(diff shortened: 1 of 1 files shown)


podvm/qcow2/rhel/qemu-rhel.pkr.hcl
```diff
@@ -50,30 +50,31 @@ build {
...
+    environment_vars = [
+        "CLOUD_PROVIDER=${var.cloud_provider}",
+        "PODVM_DISTRO=${var.podvm_distro}",
+	]
     inline = [
-      "sudo bash ~/selinux_relabel.sh"
+      "sudo -E bash ~/misc-settings.sh"
     ]
   }
+
```
snir911, line 79: extra line? also maybe add a comment in misc_settings that it must be after the relabeling for the future.. your call, LGTM overall, thanks!

whitebrandy: Thanks for the review, both fixed!

bpradipt (approved):
/lgtm
Thanks @whitebrandy 
I see that the same sequence is already there for CentOS

PR #555, opened 2023-01-23
vsphere: Log error return from Resource Pool

### LINKED ISSUE #553: vsphere: No error is logged when the Datacenter has more than one cluster
For now we do not support multiple datacenter clusters but there is no error printed out when more than one is configured.

caavery:
Fixes #553


wainersm (approved):
LGTM. Thanks @caavery !

littlejawa (approved):

Issue #558, opened 2023-01-23
vsphere make image does not create a latest tag in quay.io


caavery:
ARCHES=linux/amd64 registry=quay.io/cavery/new http://quay.io/cavery/new CLOUD_PROVIDER=vsphere make image

So a few things that are different and I'm not sure if they are intentional or not. 

*There is no latest tag

*The make image creates a new directory cloud-api-adaptor-vsphere where it did not before.
old command:
ARCHES=linux/amd64 registry=quay.io/cavery/new/cloud-api-adaptor-vsphere CLOUD_PROVIDER=vsphere make image

Now I have to run this
ARCHES=linux/amd64 registry=quay.io/cavery/new CLOUD_PROVIDER=vsphere make image
or I get the image in cavery /new/cloud-api-adaptor-vsphere/cloud-api-adaptor-vsphere instead of cavery /new/cloud-api-adaptor-vsphere

*And could we default to ARCHES=linux/amd64 ?

jtumber-ibm:
Hi @caavery seperating out the registry and the image name was on purpose to help consistancy of naming across hosting providers. i.e. if we host it in `uk.icr.io/james-tumber` or `quay.io/cavery/new` we don't also have to specify the image name `cloud-api-adaptor-vsphere`.

I'll make a PR for the other changes, to include the latest tag and default to single arch

PR #562, opened 2023-01-25
image: include latest tag and default to single arch

jtumber-ibm:
Fixes: #558

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -10,7 +10,7 @@ registry="${registry:-quay.io/confidential-containers}"
 name="cloud-api-adaptor-${CLOUD_PROVIDER}"
 tag=$(date +%Y%m%d%H%M%s)
 
-supported_arches=${ARCHES:-"linux/amd64,linux/s390x"}
+supported_arches=${ARCHES:-"linux/amd64"}
 
 function build_caa_payload() {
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -10,7 +10,7 @@ registry="${registry:-quay.io/confidential-containers}"
 name="cloud-api-adaptor-${CLOUD_PROVIDER}"
 tag=$(date +%Y%m%d%H%M%s)
 
-supported_arches=${ARCHES:-"linux/amd64,linux/s390x"}
+supported_arches=${ARCHES:-"linux/amd64"}
```
stevenhorsman, line 13: Sorry to ask a basic question, but I assume the ibm cloud version will set `ARCHES=linux/amd64,linux/s390x`. I think that's the case in the GHA workflow, but do we need to support this in any scripts for 'manually' (ie not GHA) creating the CAA ibm cloud payload, or would we just doc this in the ibm cloud section?

jtumber-ibm: I think a doc in ibmcloud section is most appropriate

stevenhorsman (approved):
LGTM. Thanks!

caavery (approved):
LGTM

PR #567, opened 2023-01-25
WIP: allow to set qemu_path from makefile

snir911:


podvm/Dockerfile.podvm.centos
```diff
@@ -11,10 +11,12 @@ FROM ${BUILDER_IMG} AS podvm_builder
 ARG CLOUD_PROVIDER=libvirt
 ARG PODVM_DISTRO=centos
 ARG AA_KBC=offline_fs_kbc
+ARG QEMU_BINARY="/usr/libexec/qemu-kvm"
```
whitebrandy, line 14: Without having to add an extra variable to pass, you could probably just include the logic in the Makefile to first search /usr/libexec/qemu-kvm followed by /usr/bin/qemu-system-x86_64 for the the qemu binary.

podvm/Makefile
```diff
@@ -29,11 +29,13 @@ ifeq ($(PODVM_DISTRO),ubuntu)
 	-var cloud_image_url=${UBUNTU_IMAGE_URL} \
 	-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM})
 ifneq ($(HOST_ARCH),$(ARCH))
-	$(eval OPTS += -var qemu_binary=qemu-system-${ARCH} \
-	-var machine_type=${QEMU_MACHINE_TYPE_${ARCH}} \
+	$(eval OPTS += -var machine_type=${QEMU_MACHINE_TYPE_${ARCH}} \
 	-var boot_wait=300s)
+ifndef QEMU_BINARY
+	$(eval OPTS += -var qemu_binary=qemu-system-${ARCH})
```
whitebrandy, line 35: I know this WIP but does this work ? I recall the right way is to create an user variable and then do something like qemu_binary = S{var.qemu_image_binary} in the packer script.

snir911: It does work (it  was here before), i think the  reason is because in the packer files we cannot applying this according to some logic (identify arch in this case)

podvm/Dockerfile.podvm.centos
snir911: right, thing is we are now supports different architectures also, I'll need to think how to handle this

podvm/Makefile
whitebrandy: Looking here - https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/qcow2/ubuntu/variables.pkr.hcl, qemu_binary is already defined as a user variable and that's why this works. Although not related to this MR and is mostly a nit: please consider renaming the user variable qemu_binary to distinguish it from the packer variable qemu_binary.

snir911: TBH i'm not sure why we set it as it's the [default of packer anyhow](https://developer.hashicorp.com/packer/plugins/builders/qemu#qemu_binary)

whitebrandy: Yeah, that's what I was referring to in the first comment. The "-var ..." syntax is only for user defined variables. You cannot set a predefined variable that way - that has to go in the main script. That's why the main script has `qemu_binary=$(var.qemu_binary}`

whitebrandy: It's easier to read the template if I know which of them are user defined variables and which of them predefined. It's actually confusing to someone new to packer trying to do something similar.

jtumber-ibm: Maybe I'm confused, what would be the alternative? Is there a better way to update that value?

I don't see the difference between `var.qemu_binary` and `var.qemu_image_binary` to tell me if it is user defined or not.

whitebrandy: Ofcourse there is :) qemu_image_binary is not mentioned in https://developer.hashicorp.com/packer/plugins/builders/qemu. So, that's a cue for someone to go look where it came from. Anyway, as I  mentioned in my previous comment, this was a nit but generally, I believe mixing user defined variables and predefined variables makes the code confusing to read.

Issue #566, opened 2023-01-25
podvm: qemu_binary is fixed for centos and rhel


snir911:
as a result build will fail systems which qemu-system-* is used

PR #569, opened 2023-01-26
podvm: set qemu-system-* link to qemu-kvm

snir911:
in the builder containers that installs qemu-kvm instead of setting qemu_binary in hcl files

Fixes: #566

### DIFF
--- a/podvm/Dockerfile.podvm_builder.rhel
+++ b/podvm/Dockerfile.podvm_builder.rhel
@@ -25,6 +25,8 @@ RUN dnf groupinstall -y 'Development Tools' && \
     rm -rf /usr/local/go && tar -C /usr/local -xzf go${GO_VERSION}.linux-amd64.tar.gz && \
     rm -f go${GO_VERSION}.linux-amd64.tar.gz
 
+# set a correspond qemu-system-* named link to qemu-kvm
+RUN ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-$(uname -m)
 
 # cloud-utils package is not available for rhel.
 RUN git clone https://github.com/canonical/cloud-utils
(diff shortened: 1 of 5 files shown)


snir911:
@whitebrandy @jtumber-ibm this is another suggestion following to https://github.com/confidential-containers/cloud-api-adaptor/pull/567 which i somehow closed

jtumber-ibm:
@snir911 this wouldn't allow us to specify a different arch than the host arch. I think the approach is clearer in the method you had before

snir911:
@jtumber-ibm it's only a link, if you pass some qemu_binary to any other available binary it will work.
BTW, AFAIU qemu-kvm is basically qemu-system-[architecture of the host], so you won't be able to cross build on OSs where only qemu-kvm is available (centos8 & rhel 8) unless you install it explicitly from somewhere (the other arch binaries are not available on cnetos/rhel 8 default repos)

snir911:
@jtumber-ibm, @whitebrandy  WDYT?

podvm/Dockerfile.podvm_builder.rhel
```diff
@@ -24,6 +24,8 @@ RUN dnf groupinstall -y 'Development Tools' && \
     rm -rf /usr/local/go && tar -C /usr/local -xzf go${GO_VERSION}.linux-amd64.tar.gz && \
     rm -f go${GO_VERSION}.linux-amd64.tar.gz
 
+# set a correspond qemu-system-* named link to qemu-kvm
+RUN ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-$(uname -m)
```
whitebrandy, line 28: IIUC, this will only work with qemu-system-x86_64, for other arches, you have to specify a qemu_binary path as well in your packer script, right ? I agree with @jtumber-ibm, I think the previous version was cleaner :)

snir911:
@whitebrandy @jtumber-ibm does it make sense now? i left the link in addition to the option to set QEMU_BINARY if needed

whitebrandy (approved):
LGTM, thanks Snir!

jtumber-ibm (approved):
lgtm

Issue #530, opened 2023-01-18
Operator: can't connect to cri-runtime.sock in cloud-api-adaptor-con container


genjuro214:
The problem is found when fixing #491. Pls refer to comment: https://github.com/confidential-containers/cloud-api-adaptor/issues/491#issuecomment-1371880366

The problem seems not always reproduced, pls refer to comments in PR https://github.com/confidential-containers/cloud-api-adaptor/pull/494

Just create this issue, and monitor if the problem really exists.

jtumber-ibm:
I've also found switching cri_runtime_endpoint.yaml to

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: cloud-api-adaptor-daemonset
  namespace: confidential-containers-system
  labels:
    app: cloud-api-adaptor
spec:
  template:
    spec:
      containers:
      - name: cloud-api-adaptor-con
        env:
        - name: CRI_RUNTIME_ENDPOINT
          value: /run/containerd/containerd.sock
        volumeMounts:
        - mountPath: /run/containerd # in-container CRI_RUNTIME_ENDPOINT default
          name: cri-runtime-endpoint
      volumes:
      - name: cri-runtime-endpoint
        hostPath:
          path: /run/containerd # SET! (crio's default: /var/run/crio/crio.sock)
          type: ""
```

Fixes the problem

mattarnoatibm:
I also reproduced this problem using the default cri_runtime_endpoint.yaml and fixed it by changing it to

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: cloud-api-adaptor-daemonset
  namespace: confidential-containers-system
  labels:
    app: cloud-api-adaptor
spec:
  template:
    spec:
      containers:
      - name: cloud-api-adaptor-con
        env:
        - name: CRI_RUNTIME_ENDPOINT
          value: /run/containerd/containerd.sock
        volumeMounts:
        - mountPath: /run/containerd # in-container CRI_RUNTIME_ENDPOINT default
          name: cri-runtime-endpoint
      volumes:
      - name: cri-runtime-endpoint
        hostPath:
          path: /run/containerd # SET! (crio's default: /var/run/crio/crio.sock)
          type: ""
```

PR #575, opened 2023-02-01
operator: fix cri runtime endpoint

mattarnoatibm:
The operator install for a kubernetes cluster using IBM Cloud VPC Infrastructure requires the following cri_runtime_endpoint.yaml to work.

Fixes: #530

Signed-off-by: Matthew Arnold <mattarno@uk.ibm.com>

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 CLOUD_PROVIDER=${1:-$CLOUD_PROVIDER}
-CRI_RUNTIME_ENDPOINT=${CRI_RUNTIME_ENDPOINT:-/run/peerpod/cri-runtime.sock}
+CRI_RUNTIME_ENDPOINT=${CRI_RUNTIME_ENDPOINT:-/run/cri-runtime.sock}
 optionals+=""
 
 # Ensure you add a space before the closing quote (") when updating the optionals
--- a/install/overlays/aws/cri_runtime_endpoint.yaml
+++ b/install/overlays/aws/cri_runtime_endpoint.yaml
@@ -11,7 +11,7 @@ spec:
       containers:
       - name: cloud-api-adaptor-con
         volumeMounts:
-        - mountPath: /run/peerpod/cri-runtime.sock # in-container CRI_RUNTIME_ENDPOINT default
+        - mountPath: /run/cri-runtime.sock # in-container CRI_RUNTIME_ENDPOINT default
           name: cri-runtime-endpoint
       volumes:
       - name: cri-runtime-endpoint
--- a/install/overlays/azure/cri_runtime_endpoint.yaml
+++ b/install/overlays/azure/cri_runtime_endpoint.yaml
@@ -11,7 +11,7 @@ spec:
       containers:
       - name: cloud-api-adaptor-con
         volumeMounts:
-        - mountPath: /run/peerpod/cri-runtime.sock # in-container CRI_RUNTIME_ENDPOINT default
+        - mountPath: /run/cri-runtime.sock # in-container CRI_RUNTIME_ENDPOINT default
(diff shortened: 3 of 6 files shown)


mattarnoatibm:
I'm not aware of any install of the peer pods operator on IBM Cloud that has worked with the original cri_runtime_endpoint.yaml since October. @jtumber-ibm and @genjuro214 have done more testing with the peer pods operator install on IBM Cloud and seen the same issue.

I've not tested the peer pods operator install on other cloud providers but I would be interested to know if the peer pods operator install using the original cri_runtime_endpoint.yaml used by the AWS, Azure, Libvirt and VSphere cloud providers. The cri_runtime_endpoint.yaml files for AWS, Azure, Libvirt and VSphere are have the same content, which is the same as the cri_runtime_endpoint.yaml for IBM Cloud that I would like this PR to replace.

snir911:
Hi, AFAIR i tested with libvirt, if there's an issue i wouldn't expect it to be ibmcloud specific, I'll give it a try on other providers
when tested you set `CRI_RUNTIME_ENDPOINT=/run/containerd/containerd.sock`?

snir911:
Similar is happening for AWS, reason seem to be lack of permissions, setting the CAA pod to privileged seem to help, that's odd, it requires additional investigation. i think it can be set in this file so that it will be applied only when cri-sock is used

mattarnoatibm:
> Hi, AFAIR i tested with libvirt, if there's an issue i wouldn't expect it to be ibmcloud specific, I'll give it a try on other providers when tested you set `CRI_RUNTIME_ENDPOINT=/run/containerd/containerd.sock`?

Hi @snir911, yes, I think the `env:` in the container spec ensures `CRI_RUNTIME_ENDPOINT` is set to `/run/containerd/containerd.sock` in the Pods in the `cloud-api-adaptor-daemonset` Daemonset.

snir911:
@mattarnoatibm i think there's a conflict because we mount both `/run/peerpod/cri-runtime.sock` socket and the `/run/peerpod/` folder for the caa pod.
If it's not a burden i'd suggest to simply switch `/run/peerpod/cri-runtime.sock` with `/run/cri-runtime.sock` or some other socket path for all providers and entrypoint.sh instead of the current commit

mattarnoatibm:
Hi @snir911. Thanks for your feedback. I made the change you suggested for all cloud providers.

I tested the install CAA using the peer pods operator and create a peer pod on IBM Cloud. It worked and a peer pod could be created. In the CAA pod `cloud-api-adaptor-con` container

```
    Mounts:
...
      /run/cri-runtime.sock from cri-runtime-endpoint (rw)
```

```
 Volumes:
...
   pods-dir:
    Type:          HostPath (bare host directory volume)
    Path:          /run/peerpod
    HostPathType:  
```

It would be great to get some help to test the other cloud providers.

snir911:
Thanks @mattarnoatibm, looks good, could you please also just update the default path [here](https://github.com/confidential-containers/cloud-api-adaptor/blob/9ec0295b657258a03a66fba5b9cab425d35f501a/entrypoint.sh#L4)
I'll give it a test on other provider
(BTW checks failure seems unrelated to this patch)

mattarnoatibm:
> Thanks @mattarnoatibm, looks good, could you please also just update the default path [here](https://github.com/confidential-containers/cloud-api-adaptor/blob/9ec0295b657258a03a66fba5b9cab425d35f501a/entrypoint.sh#L4) I'll give it a test on other provider (BTW checks failure seems unrelated to this patch)

Thanks @snir911. I've made the update to entrypoint.sh.

snir911 (approved):
LGTM, thanks @mattarnoatibm 
I can confirm it works also with AWS

jtumber-ibm (approved):
lgtm, thanks @mattarnoatibm

PR #583, opened 2023-02-03
Azure: Add support for user defined environment variables

kartikjoshi21:
Add change to support user defined subnet, NSG and image ids to improve flexibility.

Signed-off-by: Kartik Joshi <kartikjoshi@microsoft.com>

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -5,7 +5,7 @@ CRI_RUNTIME_ENDPOINT=${CRI_RUNTIME_ENDPOINT:-/run/peerpod/cri-runtime.sock}
 optionals+=""
 
 # Ensure you add a space before the closing quote (") when updating the optionals
-# example: 
+# example:
 # following is the correct method: optionals+="-option val "
 # following is the incorrect method: optionals+="-option val"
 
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -40,7 +40,7 @@ exec cloud-api-adaptor-aws aws \
 }
 
 azure() {
-test_vars AZURE_CLIENT_ID AZURE_CLIENT_SECRET AZURE_TENANT_ID AZURE_VNET_NAME AZURE_SUBNET_NAME AZURE_NSG_NAME
+test_vars AZURE_CLIENT_ID AZURE_CLIENT_SECRET AZURE_TENANT_ID AZURE_SUBSCRIPTION_ID AZURE_SUBNET_ID AZURE_IMAGE_ID
```
jepio, line 43: You're missing AZURE_NSG_ID here

kartikjoshi21: We can create azure vm without  NSG ID as well. I think we can keep it as optional param here.

jepio (approved):

surajssd (approved):

knrt10 (approved):

Issue #587, opened 2023-02-07
aws: issue in uploading raw image to be converted to ami


snir911:
in the helper script when running
`aws/raw-to-ami.sh <raw img> peer-pods us-east-2`
i hit:
`An error occurred (InvalidParameter) when calling the ImportSnapshot operation: The given S3 object is not local to the region. The object must be located in us-east-1.`

PR #588, opened 2023-02-07
aws: explicitly specify region for aws commands

snir911:
in aws/raw-to-ami.sh

Fixed: #587

### DIFF
--- a/aws/raw-to-ami.sh
+++ b/aws/raw-to-ami.sh
@@ -3,13 +3,13 @@
 [ "$DEBUG" == 'true' ] && set -x
 
 ARGC=$#
-if [ $ARGC -ne 3 ]; then
-    echo "USAGE: $(basename $0) <Image Path> <S3 Bucket Name> <Region>"
+if [ $ARGC -ne 3 ] && [ $ARGC -ne 2 ]; then
+    echo "USAGE: $(basename $0) <Image Path> <S3 Bucket Name> [Region]"
     exit 1
 fi
 IMAGE_FILE_PATH=$1
 BUCKET_NAME=$2
-REGION=$3
+REGION=${3:-$AWS_REGION}
 
 IMAGE_NAME="$(basename -- ${IMAGE_FILE_PATH})"
 AMI_NAME="${IMAGE_NAME%.*}"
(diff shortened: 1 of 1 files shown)


aws/raw-to-ami.sh
```diff
@@ -54,7 +54,7 @@ set_bucket_policies() {
 }
 EOF
 
-	aws iam create-role --role-name vmimport --assume-role-policy-document "file://${TRUST_POLICY_JSON_FILE}"
+	aws iam create-role --role-name vmimport --assume-role-policy-document "file://${TRUST_POLICY_JSON_FILE}" --region ${REGION}
```
bpradipt, line 57: May be a better approach is to take the region specified in the argument of the raw-to-ami.sh and use it for the aws command ?
Also can we use the REGION from the `AWS_REGION` env variable (which is also a common approach) and if it's not set, then throw an error to explicitly provide the REGION option ?

snir911: > May be a better approach is to take the region specified in the argument of the raw-to-ami.sh and use it for the aws command ?

this is what i did (or i didn't understand what you mean)

about the AWS_REGION, it's a bit tricky as it could also come from AWS_DEFAULT_REGION or from the profile configuration, but indeed maybe i can just take into account only AWS_REGION only

snir911: sent another version

aws/raw-to-ami.sh
```diff
@@ -23,6 +23,7 @@ ROLE_POLICY_JSON_FILE="${TMPDIR}/role-policy.json"
 BUCKET_POLICY_JSON_FILE="${TMPDIR}/bucket-policy.json"
 
 pre_checks() {
+	[[ $REGION ]] || { echo "Region not fount, either add as argument to the command or set AWS_REGION " 1>&2 ; exit 1; }
```
bpradipt, line 26: nit: fount -> found

bpradipt (approved):
/lgtm
Just a nit and this should be good to go.
Thanks @snir911

littlejawa (approved):
lgtm
Thanks @snir911

PR #594, opened 2023-02-09
pkg/adaptor: Add Teardown() to cloud.Provider

### LINKED ISSUE #593: New CAA framework: No DeleteProvider hook.
There does not seem to be a hook in the new CAA frame work so I can logout of vcenter when the CAA terminates.

type Cloud interface {
    ParseCmd(flags *flag.FlagSet)
    LoadEnv()
    NewProvider() (cloud.Provider, error)
} 

I just realized it was missing . I used to call it in Shutdown.

 func (s *server) Shutdown() error {
    DeleteGovmomiClient(s.gclient)
    s.stopOnce.Do(func() {
        close(s.stopCh)
    })
    return nil
}
@yoheiueda @bpradipt

yoheiueda:
This patch adds a hook to clean up a cloud provider.

Fixes #593

### DIFF
--- a/pkg/adaptor/cloud/aws/provider.go
+++ b/pkg/adaptor/cloud/aws/provider.go
@@ -169,3 +169,7 @@ func (p *awsProvider) DeleteInstance(ctx context.Context, instanceID string) err
 	return nil
 
 }
+
+func (p *awsProvider) Teardown() error {
+	return nil
+}
--- a/pkg/adaptor/cloud/azure/provider.go
+++ b/pkg/adaptor/cloud/azure/provider.go
@@ -264,3 +264,7 @@ func (p *azureProvider) DeleteInstance(ctx context.Context, instanceID string) e
 
 	return nil
 }
+
+func (p *azureProvider) Teardown() error {
+	return nil
+}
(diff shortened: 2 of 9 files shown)


yoheiueda:
@caavery  It looks like the error message does not come from CAA. Probably, it comes from cri-o, and kubelet reports the error?

https://github.com/containerd/containerd/blob/a6f3cbe0ada87187b7110a7dbed20d519f067417/pkg/truncindex/truncindex.go#L41
https://github.com/cri-o/cri-o/blob/6f7b7c27f389be058190bda7f27aaf27b3e516b8/internal/lib/container.go#L39
https://github.com/cri-o/cri-o/blob/a5e79c2cf8737c096909cb2af6b2713c41a6dddf/server/container_status.go#L31
https://github.com/grpc/grpc-go/blob/55dfae6e5becc928314500087c01690708deef72/internal/status/status.go#L140

Can you identify what the container 2d116a actually is? Did you delete the CAA pod while you were running a peer pod?

caavery:
@yoheiueda There were no peer pods running. I'll try it again later today.

caavery (approved):
LGTM

Thanks 
Cathy

bpradipt (approved):
/lgtm
Thanks @yoheiueda

PR #608, opened 2023-02-11
podvm: allow the builder to checkout a refspec

### LINKED ISSUE #607: podvm: allow the builder to checkout a pull request ref
On CI it will be needed to build the podvm from the sources in the pull request. Currently the podvm_builder dockerfile gets the `CAA_SRC` and `CAA_SRC_BRANCH` arguments to clone and checkout the repository like:

```
$ git clone "$CAA_SRC" -b $CAA_SRC_BRANCH cloud-api-adaptor
```

Which fails if `CAA_SRC_BRANCH` is a pull request refspec:

```
$ CAA_SRC_BRANCH="refs/pull/606/merge"
$ git clone "$CAA_SRC" -b $CAA_SRC_BRANCH cloud-api-adaptor
Cloning into 'cloud-api-adaptor'...
fatal: Remote branch refs/pull/606/merge not found in upstream origin

```

wainersm:
The podvm builder dockerfile uses the CAA_SRC and CAA_SRC_BRANCH build arguments to, respectively, get the repository and branch from where the CAA sources should be cloned. However, it may be needed to checkout from a git refspec instead of the branch. For example, on CI it should build the sources from the pull request refspec.

This introduced the CAA_SRC_REF build argument that instructs the dockerfile to fetch and checkout on the refspec. Thus, that variable takes precedence over CAA_SRC_BRANCH when set.

Fixes #607


snir911:
For CI and development purposes, wouldn't it make more sense to allow checking out on a ref (and also allow to fetch different repos/branches) on the actual building Dockerfiles? it will allow to reuse the builder containers

wainersm:
> For CI and development purposes, wouldn't it make more sense to allow checking out on a ref (and also allow to fetch different repos/branches) on the actual building Dockerfiles? it will allow to reuse the builder containers

Hi @snir911 !

I'm not sure I got your comment. With this changes it will be possible to checkout any repo and ref. For example:
```
$ docker build -t podvm_builder --build-arg CAA_SRC=https://github.com/wainersm/cc-cloud-api-adaptor --build-arg CAA_SRC_REF="refs/pull/23/merge" -f Dockerfile.podvm_builder .
```

I kept the `CAA_SRC_BRANCH` because checking a single branch will produce a smaller tree, thus a smaller image. Whereas to checkout a ref you need to clone the entire repo, fetch & then checkout the ref.

Do you think it can be done differently?

snir911:
Hi, @wainersm, I meant allowing overriding/fetching/checking-out different repos/branches/refs in the building container i.e. in the Dockerfile.podvm* files instead of the Dockerfile.podvm_builder* files.

bpradipt:
> Hi, @wainersm, I meant allowing overriding/fetching/checking-out different repos/branches/refs in the building container i.e. in the Dockerfile.podvm* files instead of the Dockerfile.podvm_builder* files.

Good idea. This will allow reuse of the builder container and save time. 
@wainersm wdyt ?

wainersm:
> > Hi, @wainersm, I meant allowing overriding/fetching/checking-out different repos/branches/refs in the building container i.e. in the Dockerfile.podvm* files instead of the Dockerfile.podvm_builder* files.
> 
> Good idea. This will allow reuse of the builder container and save time. @wainersm wdyt ?

ah, got it @snir911 . It is a good idea to me either, @bpradipt 

One question though: it seems we don't need the sources on the builder image, rather completely move the cloning/checking-out logic to the building container?

bpradipt:
> > > Hi, @wainersm, I meant allowing overriding/fetching/checking-out different repos/branches/refs in the building container i.e. in the Dockerfile.podvm* files instead of the Dockerfile.podvm_builder* files.
> > 
> > 
> > Good idea. This will allow reuse of the builder container and save time. @wainersm wdyt ?
> 
> ah, got it @snir911 . It is a good idea to me either, @bpradipt
> 
> One question though: it seems we don't need the sources on the builder image, rather completely move the cloning/checking-out logic to the building container?

For released builds we might need it. For example builder image for release vX.Y.Z having all the pieces to reproduce the build.

wainersm:
@bpradipt , @snir911 ,

double checking I understood it: we want to allow developers to overwrite the caa repo and refspec on both builder and building containers? i.e. add the following change to all the Dockerfile.podvm.* files:

```
diff --git a/podvm/Dockerfile.podvm b/podvm/Dockerfile.podvm
index 5bca50f..4301f32 100644
--- a/podvm/Dockerfile.podvm
+++ b/podvm/Dockerfile.podvm
@@ -13,12 +13,23 @@ ARG PODVM_DISTRO=ubuntu
 ARG AA_KBC=offline_fs_kbc
 # If not provided, uses system architecture
 ARG ARCH=x86_64
+ARG CAA_SRC=""
+ARG CAA_SRC_REF=""
 
 ENV CLOUD_PROVIDER ${CLOUD_PROVIDER}
 ENV PODVM_DISTRO ${PODVM_DISTRO}
 ENV AA_KBC ${AA_KBC}
 ENV ARCH ${ARCH}
 
+RUN if [ -n "${CAA_SRC}" ]; then \
+      rm -rf cloud-api-adaptor && \
+      git clone ${CAA_SRC} cloud-api-adaptor;\
+    fi && \
+    if [ -n "${CAA_SRC_REF}" ]; then \
+      cd cloud-api-adaptor && \
+      git fetch origin ${CAA_SRC_REF} && \
+      git checkout FETCH_HEAD -b ${CAA_SRC_REF} ;\
+    fi
 # Installs add-ons for foreign target, if required
 RUN ./cloud-api-adaptor/podvm/hack/cross-build-extras.sh
 
```

bpradipt:
> @bpradipt , @snir911 ,
> 
> double checking I understood it: we want to allow developers to overwrite the caa repo and refspec on both builder and building containers? i.e. add the following change to all the Dockerfile.podvm.* files:

I got confused with builder and building containers :-). But the example that you quoted helps clear my confusion.
So we want to allow developers to overwrite CAA repo when building the Pod VM image, which is modifying all the Dockerfile.podvm.* files (and not the Dockerfile.podvm_builder.*)
 
> 
> ```
> diff --git a/podvm/Dockerfile.podvm b/podvm/Dockerfile.podvm
> index 5bca50f..4301f32 100644
> --- a/podvm/Dockerfile.podvm
> +++ b/podvm/Dockerfile.podvm
> @@ -13,12 +13,23 @@ ARG PODVM_DISTRO=ubuntu
>  ARG AA_KBC=offline_fs_kbc
>  # If not provided, uses system architecture
>  ARG ARCH=x86_64
> +ARG CAA_SRC=""
> +ARG CAA_SRC_REF=""
>  
>  ENV CLOUD_PROVIDER ${CLOUD_PROVIDER}
>  ENV PODVM_DISTRO ${PODVM_DISTRO}
>  ENV AA_KBC ${AA_KBC}
>  ENV ARCH ${ARCH}
>  
> +RUN if [ -n "${CAA_SRC}" ]; then \
> +      rm -rf cloud-api-adaptor && \
> +      git clone ${CAA_SRC} cloud-api-adaptor;\
> +    fi && \
> +    if [ -n "${CAA_SRC_REF}" ]; then \
> +      cd cloud-api-adaptor && \
> +      git fetch origin ${CAA_SRC_REF} && \
> +      git checkout FETCH_HEAD -b ${CAA_SRC_REF} ;\
> +    fi
>  # Installs add-ons for foreign target, if required
>  RUN ./cloud-api-adaptor/podvm/hack/cross-build-extras.sh
>  
> ```

+1

snir911:
@bpradipt would it make sense to completely remove the repo handling from the podvm_builder containers?
Once we'll have releases we can either modify the Dockerfiles to clone the latest release tag by default (unless CAA_REF,CAA_REPO... are specified) or to build the containers/image and tag it with the released version.

bpradipt:
> @bpradipt would it make sense to completely remove the repo handling from the podvm_builder containers? Once we'll have releases we can either modify the Dockerfiles to clone the latest release tag by default (unless CAA_REF,CAA_REPO... are specified) or to build the containers/image and tag it with the released version.

@snir911 Usually, the code will be tied to the toolchain - golang version, operator sdk version, rust version etc. The purpose of the builder container is to have the required toolchain that can build the specific code version.
If we decouple the code from the toolchain, then the purpose won't be met and we run the risk of failed or incorrect builds due to the incompatibility between the toolchain and the code.
Having said that it's not a hard problem to solve. For example we can include a toolchain version file with the code and ensure it's tracked with the code. @stevenhorsman already created an issue which should address this requirement.

Further we add a Makefile target to first check for compatibility with the toolchain when building the code. Once we have this support, we can remove the code from the builder container and only provide required toolchains in the builder container to speed up the different image creation process.

Does this make sense ?

snir911:
I see now, thanks @bpradipt 
BTW different problem might be there's no 1:1 mapping between the container image used for building and the podVM image used for running, as it may be slightly different (maybe irrelevant once everything is statically built)

wainersm:
@bpradipt @snir911 I updated this PR so that now the sources can only be overridden on the podvm build (i.e. the Dockerfile.podvm.* files). I also added a fix to CentOS, apparently they changed the base image recently.

podvm/Dockerfile.podvm
```diff
@@ -13,12 +13,23 @@ ARG PODVM_DISTRO=ubuntu
 ARG AA_KBC=offline_fs_kbc
 # If not provided, uses system architecture
 ARG ARCH=x86_64
+ARG CAA_SRC=""
+ARG CAA_SRC_REF=""
```
bpradipt, line 17: You want CAA_SRC CAA_SRC_REF to be available only as build time options right, and hence you don't add it to ENV ?

bpradipt (approved):
/lgtm
Thanks @wainersm 
Add a minor clarification and assuming this was tested.
I didn't get time to test the changes.

podvm/Dockerfile.podvm
wainersm: Yest @bpradipt , they are build time options, I didn't see any reason to make them exported to runtime environment.

wainersm:
> /lgtm Thanks @wainersm Add a minor clarification and assuming this was tested. I didn't get time to test the changes.

Yep, it tested it. That's how I hit the problem I fixed for CentOS :)

PR #609, opened 2023-02-13
Stop execution of escapes check in CI

### LINKED ISSUE #604: CI always reports escape analysis results
Currently, CI always report results of escape analysis for `main.go` as shown in the attached image.

These escaping variables do not mean bugs or memory leaks. These information are just hints for performance optimizations.  In my opinion, reporting these results in CI seem confusing, since they are shown as failures. I think it is safe for us to disable escape checks in CI.

warnings

yoheiueda:
Escape analysis checks results are useful for performance optimization, but not useful for finding bugs and memory leaks. Running this test in CI shows confusing errors.

Fixes #604

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -56,5 +56,4 @@ jobs:
           sudo apt-get install -y libvirt-dev
       - name: check
         run: make CLOUD_PROVIDER=${{ matrix.provider }} check
-      - name: escapes
-        run: make CLOUD_PROVIDER=${{ matrix.provider }} escapes
+
--- a/Makefile
+++ b/Makefile
@@ -57,11 +57,9 @@ endif
 .PHONY: escapes
 escapes: ## golang memeory escapes check
 ifeq ($(CLOUD_PROVIDER),libvirt)
-	go build $(GOFLAGS) -gcflags="-m -l" -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go | grep "escapes to heap" || true
-	go build $(GOFLAGS) -gcflags="-m -l" -o agent-protocol-forwarder cmd/agent-protocol-forwarder/main.go | grep "escapes to heap" || true
+	go build $(GOFLAGS) -gcflags="-m -l" ./... 2>&1 | grep "escapes to heap" 1>&2 || true
 else
-	CGO_ENABLED=0 go build -gcflags="-m -l" $(GOFLAGS) -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go | grep "escapes to heap" || true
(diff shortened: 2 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -56,5 +56,4 @@ jobs:
           sudo apt-get install -y libvirt-dev
       - name: check
         run: make CLOUD_PROVIDER=${{ matrix.provider }} check
-      - name: escapes
```
SamYuan1990, line 59: in discussion phase, I suppose we can enable this job to see different between others.

Makefile
```diff
@@ -57,11 +57,11 @@ endif
 .PHONY: escapes
 escapes: ## golang memeory escapes check
 ifeq ($(CLOUD_PROVIDER),libvirt)
-	go build $(GOFLAGS) -gcflags="-m -l" -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go | grep "escapes to heap" || true
-	go build $(GOFLAGS) -gcflags="-m -l" -o agent-protocol-forwarder cmd/agent-protocol-forwarder/main.go | grep "escapes to heap" || true
+	go build $(GOFLAGS) -gcflags="-m -l" -o cloud-api-adaptor cmd/cloud-api-adaptor/main.go 2>&1 | grep "escapes to heap" || true
```
SamYuan1990, line 60: for the problem here in fact it's related to/limited by our project structure.
```
yuanyi@yuanyideMacBook-Pro cloud-api-adaptor % go build -gcflags="-m -l" -o cloud-api-adaptor ./cmd/cloud-api-adaptor ./pkg
directory cmd/cloud-api-adaptor is contained in a module that is not one of the workspace modules listed in go.work. You can add the module to the workspace using:
        go work use .
no Go files in /Users/yuanyi/go/src/github.com/SamYuan1990/cloud-api-adaptor/pkg
```
Let's find some way to make it better. as for now, we have latest golang as 1.20 for now. (doesn't mean we need to bump up version to 1.20 for now, just figure out solution.)

SamYuan1990: if we make it just to package/folder, there will be nothing detected. for example: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/4163646155/jobs/7204298902#step:6:8

yoheiueda: Thank you for finding out how to check all files other than `main.go`

I post some comments at https://github.com/confidential-containers/cloud-api-adaptor/issues/604

huoqifeng (approved):
/lgtm, thank you! @yoheiueda

bpradipt (approved):
/lgtm
Thanks @yoheiueda

PR #611, opened 2023-02-13
[RFC] github-actions: add common e2e workflow

wainersm:
**Description**

As we already have a simple but yet working [e2e testing framework](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/test/e2e), I began to work on the other pieces to implement the CI for libvirt.

This PR implements a workflow that can be common to all cloud e2e jobs. It does build the components that will be used to deploy peer pods:
  - build the podvm builder
  - build the podvm image
  - extract the qcow2 & save in a cache
  - build the cloud-api-adaptor image

As is, the implemented jobs on this PR are useful to check the deployment components can be built. But I envision each cloud should implement at least one additional job to effectively execute the tests on their testing environments.  

**Implementation**

* This depends on PR 608: [podvm: allow the builder to checkout a refspec](https://github.com/confidential-containers/cloud-api-adaptor/pull/608). That PR has got some interesting idea that will likely cascade in changes to this one
* Images are pushed to Github's images registry. I tried to use a local started registry but `docker buildx` fails to connect. In the end I think that using github's registry is better than a local registry  because it is public and so allows fetching the cloud-api-adaptor image by the operator from within any cloud provider. Besides the fact that devels will be able to fetch the images to debug failed e2e jobs
   * Because images are pushed for each PR it might hit the storage limit at some point in time. We could have an automation to delete the images when the PR is closed
 * The `docker buildx` action allow for caching components. I didn't explore that, it can be done on a follow up
 * So far the jobs build an Ubuntu podvm for libvirt to ease my life. I should enable it to more os/providers combinations as I get the first reviews for this PR.

### DIFF
--- a/.github/workflows/test_e2e.yaml
+++ b/.github/workflows/test_e2e.yaml
@@ -1,0 +1,147 @@
+name: test-e2e
+on: [pull_request]
+
+jobs:
+  # Build and push the podvm builder image for each OS.
+  #
+  podvm_builder:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          # Please keep this list in alphabetical order.
+          #- centos
+          - ubuntu
+        include:
+          #- os: centos
+          #  dockerfile: Dockerfile.podvm_builder.centos
+          - os: ubuntu
+            dockerfile: Dockerfile.podvm_builder
+    steps:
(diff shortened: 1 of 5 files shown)


wainersm:
A variant of this implementation is now running the libvirt e2e CI.

Issue #617, opened 2023-02-19
libvirt: remove non-essential libvirt build tags


snir911:
as it will ease import `"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/cloud/libvirt"` from different modules if needed

PR #618, opened 2023-02-19
libvirt: remove non-essential libvirt build tags

snir911:
to allow easy importing of the pkg

Fixes: #617

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,22 +45,17 @@ help: ## Display this help.
 	@printf "$$(git rev-parse HEAD 2>/dev/null)" >$@
 	@test -n "$$(git status --porcelain --untracked-files=no)" && echo -dirty >>$@ || true
 
-$(BINARIES): .git-commit $(SOURCES)
-ifeq ($(CLOUD_PROVIDER),libvirt)
-	$(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
-else
+agent-protocol-forwarder: .git-commit $(SOURCES)
 	$(GOOPTIONS) CGO_ENABLED=0 go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
-endif
+
+cloud-api-adaptor: .git-commit $(SOURCES)
+	$(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
 
 ##@ Development
 
 .PHONY: escapes
 escapes: ## golang memeory escapes check
-ifeq ($(CLOUD_PROVIDER),libvirt)
 	go build $(GOFLAGS) -gcflags="-m -l" ./... 2>&1 | grep "escapes to heap" 1>&2 || true
-else
-	CGO_ENABLED=0 go build $(GOFLAGS) -gcflags="-m -l" ./... 2>&1 | grep "escapes to heap" 1>&2 || true
-endif
 
 .PHONY: test
 test: ## Run tests.
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,7 +30,6 @@ jobs:
         with:
           go-version: ${{ matrix.go_version }}
       - name: Install build dependencies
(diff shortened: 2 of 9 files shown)


whitebrandy (commented):
```
and disable it for the +agent-protocol-forwarder
```
Since this is a change in how agent-protocol-forwarder is being built for libvirt, I would suggest that you add a few lines in the commit message that agent-protocol-forwarder, even for the libvirt build, does not use any of the libvirt go bindings and hence, it's safe to compile it with CGO=1. The advantage is that we get a static self-contained binary for agent-protocol-forwarder.

yoheiueda:
I think it is a good idea to enable CGO by default, and remove non-essential build contraint for `libvirt`.

This build constraint is an essential one. Right? Is there any other essential use of build tag for `libvirt`?

https://github.com/confidential-containers/cloud-api-adaptor/blob/5d18b67acdb693439b4aaddd149a1dc52fda582b/cmd/cloud-api-adaptor/cloudmgr/libvirt.go#L1

yoheiueda:
I agree to remove the `libvirt` build constraint from files under `pkg/adaptor/cloud/libvirt`, but I am wondering whether it makes sense to use a build constraint for `cgo` in the files like this.

```
//go:build cgo
```

If someone wants to cross-build the `cloud-api-adaptor` binary by setting the `GO_ARCH` environment variable, CGO needs to be disabled. If files for libvirt has such build contraints, I think we can easily cross build CAA.

Another benefit may be development with IDE on non-linux environment. I am using VS Code on Mac, and have difficulty to enable CGO to work with libvirt related code.  When I open CAA with VS Code, I got the following error in `pkg/adaptor/cloud/libvirt/libvirt.go`.

```
cannot use libvirt.DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE (constant unknown with invalid type) as "libvirt.org/go/libvirt".DomainInterfaceAddressesSource value in argument to dom.ListAllInterfaceAddresses
```

If `//go:build cgo` is specified for the files related to libvirt, I think CGO-related errors will disappear with VS Code on Mac.

snir911:
@yoheiueda sounds like a good option, I'll try it

snir911:
modified, i didn't touch the `escapes` target , i think it should be fine as is

cmd/cloud-api-adaptor/cloudmgr/manager.go
```diff
@@ -1,3 +1,5 @@
+//go:build libvirt || aws || azure || ibmcloud || vsphere
```
yoheiueda, line 1: Why do we need this build constraint?

When someone forks this repo and wants to implement a new cloud provider, this line is also needed to be updated. If this line does not exist, I think adding a new cloud provider becomes simpler.

Makefile
```diff
@@ -45,12 +45,11 @@ help: ## Display this help.
...
-	$(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
-else
+agent-protocol-forwarder: .git-commit $(SOURCES)
 	$(GOOPTIONS) CGO_ENABLED=0 go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
-endif
+
+cloud-api-adaptor: .git-commit $(SOURCES)
+	$(GOOPTIONS) CGO_ENABLED=1 go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
 
 ##@ Development
```
yoheiueda, line 55: could you replace the ifeq-else-endif lines for the `escapes` check with the single line below?

```
	go build $(GOFLAGS) -gcflags="-m -l" ./... 2>&1 | grep "escapes to heap" 1>&2 || true
```
We no more need to check whether CLOUD_PROVIDER is libvirt or not here.

yoheiueda:
>  i didn't touch the escapes target , i think it should be fine as is

https://github.com/confidential-containers/cloud-api-adaptor/blob/378101f7768b122bba110a40b54b89f2694bfff2/Makefile#L56-L62

could you replace the ifeq-else-endif lines for the `escapes` check with the single line below?

```
	go build $(GOFLAGS) -gcflags="-m -l" ./... 2>&1 | grep "escapes to heap" 1>&2 || true
```
We no more need to check whether CLOUD_PROVIDER is libvirt or not here.

cmd/cloud-api-adaptor/cloudmgr/manager.go
snir911: not having this allows you to compile cloudmgr without setting a cloud-provider tag, which worth nothing, IIUC it is expected to be removed soon once we will switch to single binary and compile all providers
However, I don't mind removing it already

snir911:
updated according to comments, removed WIP

yoheiueda (approved):
LGTM. Thank you very much!

bpradipt (approved):
/lgtm
Thanks @snir911

bpradipt:
There are few additional github action workflows which will require changes - like image.yaml, podvm.yaml . 
It's ok to handle it in a subsequent PR.

Issue #612, opened 2023-02-14
azure: NIC and disk not getting deleted on cleanup


knrt10:
Fixing this.

PR #619, opened 2023-02-20
azure: fix cleaning up resources

knrt10:
VM, nic and disk were not getting cleaned up when instance deletion request was made.

Fixes: #612

### DIFF
--- a/pkg/adaptor/cloud/azure/provider.go
+++ b/pkg/adaptor/cloud/azure/provider.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"regexp"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
@@ -24,6 +25,7 @@ import (
 
 var logger = log.New(log.Writer(), "[adaptor/cloud/azure] ", log.LstdFlags|log.Lmsgprefix)
 var errNotReady = errors.New("address not ready")
+var errNotFound = errors.New("VM name not found")
 
 const (
 	maxInstanceNameLen = 63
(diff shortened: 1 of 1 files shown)


pkg/adaptor/cloud/azure/provider.go
```diff
@@ -251,7 +252,10 @@ func (p *azureProvider) DeleteInstance(ctx context.Context, instanceID string) e
 		return fmt.Errorf("creating VM client: %w", err)
 	}
 
-	pollerResponse, err := vmClient.BeginDelete(ctx, p.serviceConfig.ResourceGroupName, instanceID, nil)
+	// instanceID in the form of /subscriptions/<subID>/resourceGroups/<resource_name>/providers/Microsoft.Compute/virtualMachines/<VM_Name>.
+	vmName := strings.SplitAfter(instanceID, "virtualMachines/")[1]
```
mkulke, line 256: It feels a bit unfortunate that we have to mangle strings like this. If the azure api doesn't accept an id, but a name as an input for delete operations. we maybe should treat that name as an identifier for the azure code (afaik within a sub + resource-group a name would be a unique identifier). 

In any case I think we might want to be extra diligent with this operation, afaik delete operation will not return fail on non-existing operations (at least that's the behaviour for `az vm delete`. Some ideas:

- cover a failure of string extraction + return an error, so there's at least a log
- use a regex to extract it to be sure we are mangling a proper string:
`^/subscriptions/$sub/resourceGroups/$rg/providers/Microsoft.Compute/virtualMachines/(.*)$`

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -260,7 +264,63 @@ func (p *azureProvider) DeleteInstance(ctx context.Context, instanceID string) e
...
+
+	logger.Printf("deleted disk successfully: %s", diskName)
+
+	return nil
+}
+
+func (p *azureProvider) deleteNetworkInterface(ctx context.Context, nicName string) error {
+	nicClient, err := armnetwork.NewInterfacesClient(p.serviceConfig.SubscriptionId, p.azureClient, nil)
+	if err != nil {
+		return fmt.Errorf("creating network interfaces client: %w", err)
```
mkulke, line 310:
```suggestion
		return fmt.Errorf("creating network interface client: %w", err)
```

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -260,7 +264,63 @@ func (p *azureProvider) DeleteInstance(ctx context.Context, instanceID string) e
...
+	}
+
+	pollerResponse, err := nicClient.BeginDelete(ctx, p.serviceConfig.ResourceGroupName, nicName, nil)
+	if err != nil {
+		return fmt.Errorf("beginning deletion of network interface: %w", err)
```
mkulke, line 315:
```suggestion
		return fmt.Errorf("beginning network interface deletion: %w", err)
```

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -260,7 +264,63 @@ func (p *azureProvider) DeleteInstance(ctx context.Context, instanceID string) e
...
+	}
+
+	_, err = pollerResponse.PollUntilDone(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("polling network interface deletion: %w", err)
```
mkulke, line 320:
```suggestion
		return fmt.Errorf("waiting for network interface deletion: %w", err)
```

pkg/adaptor/cloud/azure/provider.go
knrt10: Make sense, updated

mkulke (approved):
thx. lgtm

bpradipt (approved):
/lgtm
Thanks @knrt10

PR #621, opened 2023-02-21
all: migrate from k8s.gcr.io to registry.k8s.io

CIPHERTron:
This fix is a part of an umbrella issue https://github.com/kubernetes/k8s.io/issues/4780


snir911 (approved):
Thanks for the patch, LGTM overall
nit, if you don't mind please add `Fixes: kubernetes/k8s.io#4780` to the commit msg for reference (also very nit-i think `all:` would fit better than `fix:`)

CIPHERTron:
> Thanks for the patch, LGTM overall nit, if you don't mind please add `Fixes: kubernetes/k8s.io#4780` to the commit msg for reference (also very nit-i think `all:` would fit better than `fix:`)

@snir911 I've addressed your comments. PTAL. Thanks!

bpradipt (approved):
/lgtm
Thanks @CIPHERTron

bpradipt:
Force merging this.

PR #627, opened 2023-02-22
Align the struct fields for optimal memory use

bpradipt:
It would be good to add `fieldalignment` to the linter or `make check` option.  However couldn't find an easy way to do it.
Hope to implement it as a separate issue.


snir911 (approved):

yoheiueda (approved):
LGTM

PR #628, opened 2023-02-22
Switch to single binary

bpradipt:


bpradipt:
I haven't changed the github actions in this PR. Will be handling it in a follow-on PR.

bpradipt:
Updated the build.yaml to use the new var and also added a new action to ensure building binary with all providers succeeds

Makefile
```diff
@@ -93,9 +98,17 @@ image: ## Build and push docker image to $registry
 
 .PHONY: deploy 
 deploy: ## Deploy cloud-api-adaptor using the operator, according to install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml file.
+ifneq ($(CLOUD_PROVIDER),)
```
snir911, line 101: what about having something like:
```
%:
        kubectl apply -f install/yamls/deploy.yaml
        kubectl apply -k install/overlays/$@

deploy: $(filter-out deploy, $(MAKECMDGOALS))
        @for i in $^; do ./$$i; done

```
this will also allow to use CLOUD_PROVIDERS instead of BUILD_PROVIDERS if proffered 
disclaimer: I'm not a makefile expert but IIUC it would allow to do: `make deploy aws `

Makefile
```diff
@@ -93,9 +98,17 @@ image: ## Build and push docker image to $registry
...
 	kubectl apply -f install/yamls/deploy.yaml
 	kubectl apply -k install/overlays/$(CLOUD_PROVIDER)
+else
+	$(error CLOUD_PROVIDER is not set)
+endif
 
 .PHONY: delete
 delete: ## Delete cloud-api-adaptor using the operator, according to install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml file.
```
snir911, line 109: same, maybe also worth changing to undeploy?

.github/workflows/build-all.yaml
```diff
@@ -0,0 +1,42 @@
...
+            providers: "aws,azure,ibmcloud,vsphere"
+
+    steps:
+      - name: Checkout the pull request code
+        uses: actions/checkout@v3
+      - name: Setup Golang version ${{ matrix.go_version }}
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go_version }}
+      - name: Install build dependencies
```
snir911, line 35: nit: we can add a condition for only when libvirt is included (dynamic) only no?

Makefile
```diff
@@ -1,21 +1,26 @@
...
 BINARIES    := cloud-api-adaptor agent-protocol-forwarder
 SOURCEDIRS  := ./cmd ./pkg
 PACKAGES    := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
 SOURCES     := $(shell find $(SOURCEDIRS) -name '*.go' -print)
 
+ifeq (,$(findstring libvirt,$(BUILD_PROVIDERS)))
+	# Force static build since libvirt provider is not part of the providers to be built. 
+	override GOOPTIONS += CGO_ENABLED=0
+endif
+
```
snir911, line 23: I think this will allow to revert the caa, forwader targets below back to:
```
$(BINARIES): .git-commit $(SOURCES)
        $(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
```

snir911: i just noticed it will not take into account in some of the targets below such as vet, escapse if the default is without cgo should they also used without? IDK

bpradipt: I ran the following targets without any issues. 
```
make vet
make escapes
BUILD_PROVIDERS="aws,azure,ibmcloud,vsphere,libvirt" make vet
BUILD_PROVIDERS="aws,azure,ibmcloud,vsphere,libvirt" make escapes
```

Makefile
bpradipt: I'm not aware of the above syntax. Will give it a try.
Just to understand, you are suggesting using CLOUD_PROVIDERS instead of BUILD_PROVIDERS, which makes sense. 
However I don't understand how the user will provide a single provider just for deployment ?

jtumber-ibm:
Could it be possible to just have a "dev" and "release" list (names not important) where "dev" is all providers and "release" is all providers minus libvirt? Just thinking of reducing the number of updates when adding new providers

Makefile
bpradipt: @snir911 are you suggesting running `deploy` target for all the providers that are part of CLOUD_PROVIDERS ?

Makefile
snir911: notice that it will run with CGO regardless to if libivirt i specified or not, I'm not sure if it's a problem or not, good that it passes..

Makefile
bpradipt: I'll probably address it in a subsequent PR as there are places like e2e where it's used.

.github/workflows/build-all.yaml
bpradipt: yes. Fixed in new update

jtumber-ibm:
Will also require updates to https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/.github/workflows/image.yaml

Makefile
bpradipt: I think it shouldn't be a problem since these are static checks. We can revisit if we see any issues.

bpradipt:
> Will also require updates to https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/.github/workflows/image.yaml

Yeah, thanks..

bpradipt:
I'm yet to test the multi-arch builds. Will test and update this issue.

jtumber-ibm (approved):
assuming multiarch builds are still ok, which I think they should be from the changes. lgtm thanks @bpradipt

snir911 (approved):
LGTM assuming tested, thanks!

Issue #623, opened 2023-02-21
vsphere: Increase types of vCenter configs that can be used for template cloning


caavery:
The CAA will now allow a greater variety of vCenter configurations. This includes data centers with multiple clusters, hosts, or data stores. DRS  ( dynamic resource scheduler ) Automation of the type Manual is also supported.  This is where only a data center and a cluster name are required inputs for the indicated vCenter. The peer pod is located according to the DRS suggestion.

Issue #635, opened 2023-02-23
vsphere: Add additional startup option checking to the CAA


caavery:
With supporting more vcenter configs ( issue #623 ) we need to add  more  initial checking of the optional inputs to the CAA during startup in NewProvider instead of doing the work in the bash script entrypoint.sh or later in the CAA when the user goes to create a pod. 

example: If the user selects DRS then a cluster name is required

@bpradipt  @snir911

PR #641, opened 2023-02-27
vsphere: Support more types of vcenter configs

caavery:
Fixes: #623

This patch also includes a fix in DeleteInstance when the instanceID is null.

### DIFF
--- a/cmd/cloud-api-adaptor/cloudmgr/vsphere.go
+++ b/cmd/cloud-api-adaptor/cloudmgr/vsphere.go
@@ -23,16 +23,16 @@ type vsphereMgr struct{}
 func (_ *vsphereMgr) ParseCmd(flags *flag.FlagSet) {
 
 	flags.StringVar(&vspherecfg.VcenterURL, "vcenter-url", "", "URL of vCenter instance to connect to")
-	flags.StringVar(&vspherecfg.UserName, "user-name", "", "Username, defaults to `GOVC_USERNAME`")
-	flags.StringVar(&vspherecfg.Password, "password", "", "Password, defaults to `GOVC_PASSWORD`")
+	flags.StringVar(&vspherecfg.UserName, "user-name", "", "vCenter Username")
+	flags.StringVar(&vspherecfg.Password, "password", "", "vCenter Password")
 	flags.StringVar(&vspherecfg.Thumbprint, "thumbprint", "", "SHA1 thumbprint of the vcenter certificate. Enable verification of certificate chain and host name.")
 	flags.StringVar(&vspherecfg.Template, "template", "podvm-template", "vCenter template to deploy")
-	flags.StringVar(&vspherecfg.Datacenter, "data-center", "", "vCenter desination datacenter name")
-	flags.StringVar(&vspherecfg.Vcluster, "vcluster-name", "", "vCenter desination cluster name for DRS placement")
+	flags.StringVar(&vspherecfg.Datacenter, "data-center", "", "vCenter destination datacenter name")
 	flags.StringVar(&vspherecfg.Datastore, "data-store", "", "vCenter datastore")
-	flags.StringVar(&vspherecfg.Resourcepool, "resource-pool", "", "vCenter desination resource pool")
-	flags.StringVar(&vspherecfg.Deployfolder, "deploy-folder", "", "vCenter vm desintation folder relative to the vm inventory path (your-data-center/vm). \nExample '-deploy-folder peerods' will create or use the existing folder peerpods as the \ndeploy-folder in /datacenter/vm/peerpods")
-
+	flags.StringVar(&vspherecfg.Deployfolder, "deploy-folder", "", "vCenter vm destination folder relative to the vm inventory path (your-data-center/vm). \nExample '-deploy-folder peerods' will create or use the existing folder peerpods as the \ndeploy-folder in /datacenter/vm/peerpods")
+	flags.StringVar(&vspherecfg.Cluster, "cluster", "", "vCenter destination cluster name ")
+	flags.StringVar(&vspherecfg.DRS, "drs", "false", "Use DRS for clone placement in destination Vcenter cluster")
+	flags.StringVar(&vspherecfg.Host, "host", "", "vCenter host name of resource pool destination")
 }
 
 func (_ *vsphereMgr) LoadEnv() {
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -90,18 +90,19 @@ exec cloud-api-adaptor-libvirt libvirt \
 }
 
 vsphere() {
-test_vars GOVC_USERNAME GOVC_PASSWORD GOVC_URL GOVC_DATACENTER GOVC_DATASTORE
+test_vars GOVC_USERNAME GOVC_PASSWORD GOVC_URL GOVC_DATACENTER
 
 [[ "${GOVC_TEMPLATE}" ]] && optionals+="-template ${GOVC_TEMPLATE} "
-[[ "${GOVC_VCLUSTER}" ]] && optionals+="-vcluster ${GOVC_VCLUSTER} "
-[[ "${GOVC_RESOURCE_POOL}" ]] && optionals+="-resource-pool ${GOVC_RESOURCE_POOL} "
+[[ "${GOVC_VCLUSTER}" ]] && optionals+="-cluster ${GOVC_VCLUSTER} "
 [[ "${GOVC_FOLDER}" ]] && optionals+="-deploy-folder ${GOVC_FOLDER} "
+[[ "${GOVC_HOST}" ]] && optionals+="-host ${GOVC_HOST} "
+[[ "${GOVC_DRS}" ]] && optionals+="-drs ${GOVC_DRS} "
+[[ "${GOVC_DATASTORE}" ]] && optionals+="-data-store ${GOVC_DATASTORE} "
 
 set -x
 exec cloud-api-adaptor-vsphere vsphere \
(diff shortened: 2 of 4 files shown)


caavery:
There is a file missing from my commit. Fixing it.

caavery:
OK I added entrypoint.sh to the commit

bpradipt:
It will be helpful to add the commented out parameters with a brief description to https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/install/overlays/vsphere/kustomization.yaml 
Similar to  https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/install/overlays/aws/kustomization.yaml

caavery:
Yes that's a good idea, however I'd like to do that in the next patch for issue #635 which does more initial checking and I have ready to go.

bpradipt:
> Yes that's a good idea, however I'd like to do that in the next patch for issue #635 which does more initial checking and I have ready to go.

It should be fine to address it as part of #635

cmd/cloud-api-adaptor/cloudmgr/vsphere.go
```diff
@@ -23,16 +23,17 @@ type vsphereMgr struct{}
...
+	flags.StringVar(&vspherecfg.Password, "password", "", "vCenter Password")
 	flags.StringVar(&vspherecfg.Thumbprint, "thumbprint", "", "SHA1 thumbprint of the vcenter certificate. Enable verification of certificate chain and host name.")
 	flags.StringVar(&vspherecfg.Template, "template", "podvm-template", "vCenter template to deploy")
 	flags.StringVar(&vspherecfg.Datacenter, "data-center", "", "vCenter desination datacenter name")
-	flags.StringVar(&vspherecfg.Vcluster, "vcluster-name", "", "vCenter desination cluster name for DRS placement")
 	flags.StringVar(&vspherecfg.Datastore, "data-store", "", "vCenter datastore")
-	flags.StringVar(&vspherecfg.Resourcepool, "resource-pool", "", "vCenter desination resource pool")
 	flags.StringVar(&vspherecfg.Deployfolder, "deploy-folder", "", "vCenter vm desintation folder relative to the vm inventory path (your-data-center/vm). \nExample '-deploy-folder peerods' will create or use the existing folder peerpods as the \ndeploy-folder in /datacenter/vm/peerpods")
-
+	// flags.StringVar(&vspherecfg.Resourcepool, "resource-pool", "", "vCenter desination resource pool")
```
bpradipt, line 33: Is the commented code intentional ?

pkg/adaptor/cloud/vsphere/provider.go
```diff
@@ -122,9 +124,114 @@ func (p *vsphereProvider) CreateInstance(ctx context.Context, podName, sandboxID
...
+
+	var poolref *types.ManagedObjectReference
+
+	if strings.EqualFold(p.serviceConfig.DRS, "true") {
+
+		// If the user has selected DRS we check that there is a cluster name,
+		// ignore any non relevant inputs, and the rest is up to DRS
+
+		if p.serviceConfig.Cluster == "" {
+			err = fmt.Errorf("DRS requires a cluster name")
```
bpradipt, line 144: You can simply return `nil, fmt.Errorf("DRS requires a cluster name")`

pkg/adaptor/cloud/vsphere/provider.go
```diff
@@ -122,9 +124,114 @@ func (p *vsphereProvider) CreateInstance(ctx context.Context, podName, sandboxID
...
+		}
+
+		rspec := *recs[0].Action[0].(*types.PlacementAction).RelocateSpec
+		relocateSpec.Datastore = rspec.Datastore
+		relocateSpec.Host = rspec.Host
+		relocateSpec.Pool = rspec.Pool
+
+	} else if p.serviceConfig.Host != "" {
+
+		// Here a cluster name is required if there is a cluster and no DRS. The host hangs off the cluster
```
bpradipt, line 183: Can you please add a reference doc link which describes the DRS flow or just describe the algo ? This will help to understand the entire logical flow for anyone looking at this later on

cmd/cloud-api-adaptor/cloudmgr/vsphere.go
caavery: Yes. I was thinking of supporting resource by name at some point. I can take it out.

pkg/adaptor/cloud/vsphere/provider.go
caavery: I actually have this in the second patch but I can include it here in this one.

caavery: I can describe it better.

caavery: @bpradipt OK I think I addressed your concerns. I also decided to remove the call to  DefaultResoucePool. This was intended to be a last ditch effort to identify a resource pool when the user had not provided enough information. Unfortunately it only works in very simple config e.g. a datacenter with only one cluster with one host and one datastore. So it basically allows the user to skirt following the config rules which can cause confusion if any additions to the datacenter are made and the call no longer works.

cmd/cloud-api-adaptor/cloudmgr/vsphere.go
```diff
@@ -23,16 +23,16 @@ type vsphereMgr struct{}
...
 	flags.StringVar(&vspherecfg.Template, "template", "podvm-template", "vCenter template to deploy")
 	flags.StringVar(&vspherecfg.Datacenter, "data-center", "", "vCenter desination datacenter name")
-	flags.StringVar(&vspherecfg.Vcluster, "vcluster-name", "", "vCenter desination cluster name for DRS placement")
 	flags.StringVar(&vspherecfg.Datastore, "data-store", "", "vCenter datastore")
-	flags.StringVar(&vspherecfg.Resourcepool, "resource-pool", "", "vCenter desination resource pool")
 	flags.StringVar(&vspherecfg.Deployfolder, "deploy-folder", "", "vCenter vm desintation folder relative to the vm inventory path (your-data-center/vm). \nExample '-deploy-folder peerods' will create or use the existing folder peerpods as the \ndeploy-folder in /datacenter/vm/peerpods")
-
+	flags.StringVar(&vspherecfg.Cluster, "cluster", "", "vCenter desination cluster name ")
+	flags.StringVar(&vspherecfg.DRS, "drs", "false", "Use DRS for clone placement in desination Vcenter cluster")
+	flags.StringVar(&vspherecfg.Host, "host", "", "vCenter host name of resource pool desination")
```
bpradipt, line 35: nit: desination -> destination for all the above 3 statements

cmd/cloud-api-adaptor/cloudmgr/vsphere.go
```diff
@@ -23,16 +23,16 @@ type vsphereMgr struct{}
...
 	flags.StringVar(&vspherecfg.Thumbprint, "thumbprint", "", "SHA1 thumbprint of the vcenter certificate. Enable verification of certificate chain and host name.")
 	flags.StringVar(&vspherecfg.Template, "template", "podvm-template", "vCenter template to deploy")
 	flags.StringVar(&vspherecfg.Datacenter, "data-center", "", "vCenter desination datacenter name")
-	flags.StringVar(&vspherecfg.Vcluster, "vcluster-name", "", "vCenter desination cluster name for DRS placement")
 	flags.StringVar(&vspherecfg.Datastore, "data-store", "", "vCenter datastore")
-	flags.StringVar(&vspherecfg.Resourcepool, "resource-pool", "", "vCenter desination resource pool")
 	flags.StringVar(&vspherecfg.Deployfolder, "deploy-folder", "", "vCenter vm desintation folder relative to the vm inventory path (your-data-center/vm). \nExample '-deploy-folder peerods' will create or use the existing folder peerpods as the \ndeploy-folder in /datacenter/vm/peerpods")
-
+	flags.StringVar(&vspherecfg.Cluster, "cluster", "", "vCenter desination cluster name ")
+	flags.StringVar(&vspherecfg.DRS, "drs", "false", "Use DRS for clone placement in desination Vcenter cluster")
```
bpradipt, line 34: You can also use a BoolVar. If the option `-drs` is missing it's `false`, if the option is present then it's `true`.

bpradipt (approved):
/lgtm
@caavery tried switching to `BoolVar` but it turned out to be  non-intuitive when compared to using `true` and `false` string. We will look at BoolVar option handling in a follow-on PR.

entrypoint.sh
```diff
@@ -90,18 +90,19 @@ exec cloud-api-adaptor-libvirt libvirt \
 }
 
 vsphere() {
-test_vars GOVC_USERNAME GOVC_PASSWORD GOVC_URL GOVC_DATACENTER GOVC_DATASTORE
+test_vars GOVC_USERNAME GOVC_PASSWORD GOVC_URL GOVC_DATACENTER
```
snir911, line 93: I was about to suggest removing the GOVC_DATASTORE install/overlays/vsphere/kustomization.yaml but i see we missing it there, however, while here I'd suggest to add the GOVC_DATACENTER there as it's also missing and it is mandatory

snir911: however, generally LGTM :)

caavery: The next patch for issue #635 contains an updated kustomization.yaml as per my conversation with Pradipta https://github.com/confidential-containers/cloud-api-adaptor/pull/641#issuecomment-1448131203

 GOVC_DATACENTER is no longer mandatory if you request DRS. The next patch checks for that in NewProvider.

snir911 (approved):
thanks!

Issue #637, opened 2023-02-24
cloud-api-adaptor container image usage


bpradipt:
We have two types of builds:
- dev: includes all the provider including libvirt. 
- release: includes all the provider excluding libvirt. 

We need to decide on the following:
1. Publish both dev and release builds simultaneously for every PR merge ?
2. Publish only dev builds for every PR merge and release builds per release ?
3. Publish release builds for every PR merge?

bpradipt:
@jtumber-ibm @snir911 @wainersm @stevenhorsman

stevenhorsman:
My initial reaction is number 2, if the other difference is libvirt them I don't see much harm in publishing that on each PR merge. I don't have strong opinions on this though, so happy to go with the other suggestions.

jtumber-ibm:
It would be nice to have the s390x images as up to date as the x86 for testing

bpradipt:
@jtumber-ibm in that case we need to build both release and dev builds or figure out a way to create dev builds (no static builds) for s390x. Is it possible? Based on my previous experience with cross-compiling on ppc64le there were issues when using cgo so had to build natively. Hence the question.

jtumber-ibm:
> 

If we need to wait for #119 to be able to have a s390x dev image then we can try to priorities that? Happy to have just dev (without s390x) for each PR at the moment.

bpradipt:
@jtumber-ibm we won't be able to have a pure go implementation for libvirt in near future :-( . The official libvirt-go bindings need cgo and for long term maintainability it's better we stick to the official libvirt-go bindings. 
I see no harm in creating different builds for each PR.  We will need to figure out a way to tag or name it properly.

snir911:
2 sounds like a good option (i assume runtime issues are less likely to appear with the non-cgo (release) build, although i'm not completely sure if it's actually true)

bpradipt:
> @jtumber-ibm we won't be able to have a pure go implementation for libvirt in near future :-( . The official libvirt-go bindings need cgo and for long term maintainability it's better we stick to the official libvirt-go bindings. I see no harm in creating different builds for each PR. We will need to figure out a way to tag or name it properly.

By different builds, I meant dev and release builds.

PR #642, opened 2023-02-27
gh-action: build dev and release images for every PR merge

bpradipt:
Fixes: #637

@jtumber-ibm @stevenhorsman @snir911 PTAL.

The PR will ensure both dev and release builds for every PR merged and use `dev-$timestamp` and `release-$timestamp` as tags. Also the `latest` tag will be only for dev release.

Hope this approach will be able to meet the requirements for multi-arch as well.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -8,21 +8,26 @@ script_dir=$(dirname "$(readlink -f "$0")")
 
 registry="${registry:-quay.io/confidential-containers}"
 name="cloud-api-adaptor"
-tag=$(date +%Y%m%d%H%M%s)
+commit_id=${1:-$(git rev-parse HEAD)}
 
 supported_arches=${ARCHES:-"linux/amd64"}
 
 function build_caa_payload() {
 	pushd "${script_dir}/.."
 
+	local tag_string="-t ${registry}/${name}:latest -t ${registry}/${name}:dev-${commit_id}"
+
+	if [[ "$RELEASE_BUILD" == "true" ]]; then
+		tag_string="-t ${registry}/${name}:${commit_id}"
+	fi
+
+
 	docker buildx build --platform "${supported_arches}" \
 		--build-arg RELEASE_BUILD="${RELEASE_BUILD}" \
 		-f Dockerfile \
-		-t "${registry}/${name}:${tag}" \
-		-t "${registry}/${name}:latest" \
+		${tag_string} \
 		--push \
 		.
-
 	popd
 }
 
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ clean: ## Remove binaries.
 
 .PHONY: image
 image: ## Build and push docker image to $registry
-	hack/build.sh
+	hack/build.sh $(shell cat .git-commit)
 
 ##@ Deployment
 


jtumber-ibm (approved):

hack/build.sh
```diff
@@ -15,13 +15,23 @@ supported_arches=${ARCHES:-"linux/amd64"}
...
 	pushd "${script_dir}/.."
 
-	docker buildx build --platform "${supported_arches}" \
-		--build-arg RELEASE_BUILD="${RELEASE_BUILD}" \
-		-f Dockerfile \
-		-t "${registry}/${name}:${tag}" \
-		-t "${registry}/${name}:latest" \
-		--push \
-		.
+	if [[ "$RELEASE_BUILD" == "true" ]]; then
```
wainersm, line 18: Hi @bpradipt , one suggestion:

```
local image="-t ${registry}/${name}:latest -t ${registry}/${name}:dev-${tag}"
if [[ "$RELEASE_BUILD" == "true" ]]; then
     image="-t ${registry}/${name}:release-${tag}"
fi

docker buildx build --platform "${supported_arches}" \
	--build-arg RELEASE_BUILD="${RELEASE_BUILD}" \
	-f Dockerfile \
	${image} \
	--push \
	.

```

wainersm:
@bpradipt also, wouldn't be better to use the commit SHA1 rather than the timestamp? So changes can be easily track-able

hack/build.sh
bpradipt: sure @wainersm

bpradipt:
> @bpradipt also, wouldn't be better to use the commit SHA1 rather than the timestamp? So changes can be easily track-able

Yeah, good suggestion.

bpradipt:
@jtumber-ibm @wainersm @stevenhorsman ptal. Have addressed all the previous comments.

snir911 (approved):
@bpradipt  thanks, LGTM overall, one nit point - would it make sense to use dev tags as is and use just hash (without `release-`) for the releases? (also i assume once we'll have releases we'll switch to version tags)

stevenhorsman (approved):
LGTM. I'm happy with Snir's suggestion to remove the release prefix, but it's not a dealbreaker for me given that we'll likely update it once we get proper releases out.

bpradipt:
> @bpradipt thanks, LGTM overall, one nit point - would it make sense to use dev tags as is and use just commit-id (without `release-`) for the releases? (also i assume once we'll have releases we'll switch to version tags)

Yeah I can do the change.

stevenhorsman:
Not a blocker, but just FYI - in the other confidential-container container images, we use the full commit hash, not just the short version.

bpradipt:
> Not a blocker, but just FYI - in the other confidential-container container images, we use the full commit hash, not just the short version.

Hmm didn't realise this. I'm taking the commit id from the Makefile itself. 
Probably we can decide whether to use the long or short commit hash and make it uniform across the project ?

stevenhorsman:
> > Not a blocker, but just FYI - in the other confidential-container container images, we use the full commit hash, not just the short version.
> 
> Hmm didn't realise this. I'm taking the commit id from the Makefile itself. Probably we can decide whether to use the long or short commit hash and make it uniform across the project ?

Ah ok - I think the idea is, if we take the commit from the overall project, then we can easily go back an identify what code we were running in a specific version, which isn't going to be as easy as if we just use the `Makefile`. It looks like we are using the long commit id in this way in the [coco runtime payload](https://quay.io/repository/confidential-containers/runtime-payload-ci?tab=tags) at least.

bpradipt:
Ah I see now. The one that I'm setting up in build.sh is a short id. I'll fix it.

hack/build.sh
```diff
@@ -8,21 +8,26 @@ script_dir=$(dirname "$(readlink -f "$0")")
 
 registry="${registry:-quay.io/confidential-containers}"
 name="cloud-api-adaptor"
-tag=$(date +%Y%m%d%H%M%s)
+commit_id=${1:-$(git rev-parse HEAD)}
```
bpradipt, line 11: @stevenhorsman ^^

stevenhorsman: Thanks!

stevenhorsman:
> @stevenhorsman please check if it's fine now..

That looks good to me, so I'm happy to merge it and then double check it's working as expected on future PRs!

Issue #643, opened 2023-02-27
all: Update cloud provider kustomatization.yaml to use the new single image binary


stevenhorsman:
We used to have specific CAA images for each cloud provider and rely on kustomization.yaml to set them. After https://github.com/confidential-containers/cloud-api-adaptor/pull/628 there is just a single image, so we can set this in `caa-pod.yaml`.

The next question is what changes to remove in the cloud provider specific kustomization.yaml files. Given that we still use the `CLOUD_PROVIDER` value in `entrypoint.sh` to specify a particular cloud provider to use I think this means that we don't currently have the ability to support mixed cloud providers in a single operator deployment at the moment? Due to this I think keeping the name suffix makes sense.

There is then the question about the image kustomize. There are scenarios that you might want to test an image other than the latest (e.g. when testing a specific release, and particularly for ibmcloud where the latest doesn't support s390x), so I'm not sure if it is worth leaving the section in but commented out, or just removing altogether?

PR #644, opened 2023-02-27
all: Update Cloud provider kustomization

stevenhorsman:
- Update the kustomization files for the cloud providers now we have a single binary, so we point to the latest versions

Fixes: #643

### DIFF
--- a/install/overlays/libvirt/kustomization.yaml
+++ b/install/overlays/libvirt/kustomization.yaml
@@ -3,8 +3,8 @@ bases:
 nameSuffix: -libvirt
 
 images:
-- name: kustomize.this/cloud-api-adaptor/image:url
-  newName: quay.io/confidential-containers/cloud-api-adaptor-libvirt # change image if needed
+- name: cloud-api-adaptor
+  newName: quay.io/confidential-containers/cloud-api-adaptor # change image if needed
   newTag: latest
 
 generatorOptions:
--- a/install/yamls/caa-pod.yaml
+++ b/install/yamls/caa-pod.yaml
@@ -22,7 +22,7 @@ spec:
             name: peer-pods-secret
         - configMapRef:
             name: peer-pods-cm
-        image: kustomize.this/cloud-api-adaptor/image:url
+        image: cloud-api-adaptor
         imagePullPolicy: Always
         name: cloud-api-adaptor-con
         securityContext:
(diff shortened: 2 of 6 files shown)


stevenhorsman:
@bpradipt - based on our quick slack conversation earlier I've created my suggestion as this PR. I'm not sure if this is on the right lines though, so have left it as a draft PR for now. If you find that I'm way off base then please feel free to update the issue with a better proposal. Thanks!

install/overlays/libvirt/kustomization.yaml
```diff
@@ -2,10 +2,11 @@ bases:
...
 
-images:
-- name: kustomize.this/cloud-api-adaptor/image:url
-  newName: quay.io/confidential-containers/cloud-api-adaptor-libvirt # change image if needed
-  newTag: latest
+# change image if needed
+# images:
+# - name: quay.io/confidential-containers/cloud-api-adaptor:latest
+#   newName: quay.io/confidential-containers/cloud-api-adaptor
+#   newTag: release-<timestamp>
```
bpradipt, line 9: nit: libvirt will not be there in release builds. So the newTag can refer to `dev-<timestamp>`

snir911: also i think it was switched to release-<commit-id>

install/yamls/caa-pod.yaml
```diff
@@ -22,7 +22,7 @@ spec:
             name: peer-pods-secret
         - configMapRef:
             name: peer-pods-cm
-        image: kustomize.this/cloud-api-adaptor/image:url
+        image: quay.io/confidential-containers/cloud-api-adaptor:latest
```
snir911, line 25: how about changing this to:
```
image: cloud-api-adaptor
```
and then uncomment it in all kustomization.yaml files to:
```
images:
- name: cloud-api-adaptor
  newName: quay.io/confidential-containers/cloud-api-adaptor
  newTag: latest
```

stevenhorsman: Thanks for the suggestion - I've pushed a change based on this suggestion. Once people are is general agreement I'll squash the commits and move this PR out of draft.

install/overlays/libvirt/kustomization.yaml
bpradipt: `dev-<commitid>` as per latest changes for the dev builds which includes libvirt :-)

bpradipt (approved):
/lgtm
Thanks @stevenhorsman

snir911 (approved):
LGTM, thanks!

Issue #646, opened 2023-02-27
Remove deprecated io/ioutil package


wainersm:
According to golangci-lint the`io/ioutil` package is deprecated. We may want to remove its dependency from our code base.

```
pkg/podnetwork/tuntest/util.go:10:2: SA1019: package io/ioutil is deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
pkg/adaptor/server_test.go:9:2: SA1019: package io/ioutil is deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
pkg/adaptor/cloud/azure/provider.go:11:2: SA1019: package io/ioutil is deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
```

PR #647, opened 2023-02-28
pkg: use io or os package instead of deprecated io/ioutil

Amulyam24:
Some functions in package `io/ioutil` are deprecated as of go 1.16 and 1.17. Move to using functions from package `io` or `os`.

Fixes:#646


yoheiueda (approved):
LGTM. Thank you!

snir911 (approved):

Issue #639, opened 2023-02-27
Update operator-sdk and golang version


bpradipt:
CCv0 kata-agent is using golang-1.19 now. Also the latest version of operator-sdk uses golang-1.19.
We will need to update the projects to use golang-1.19 and latest operator-sdk

- [webhook](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/webhook)
- [peer-pods-controller](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/peer-pods-controller)
- ci workflows
- Dockerfiles

PR #648, opened 2023-02-28
Bump golang version to 1.19

Amulyam24:
Fixes: #639


peer-pod-controller/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/peer-pod-controller
 
-go 1.18
+go 1.19
```
bpradipt, line 3: Afaik, this will also need changes to dependencies like controller-mgr etc as the version of golang is tied to operator-sdk version.

Amulyam24: @bpradipt, added changes. PTAL.
There is currently no CI check for validating the manifests or related code in peer pod controller. It would be good to add a simple check, wdyt?

yoheiueda (approved):
LGTM

peer-pod-controller/go.mod
bpradipt: yeah, it'll be a good addition to the CI.

Amulyam24: Okay, I'll create a follow up PR shortly.

peer-pod-controller/controllers/peerpodconfig_controller.go
```diff
@@ -326,9 +325,9 @@ func (r *PeerPodConfigReconciler) getNodesWithLabels(nodeLabels map[string]strin
 
 	if err := r.Client.List(context.TODO(), nodes, listOpts...); err != nil {
 		r.Log.Error(err, "Getting list of nodes having specified labels failed")
-		return err, &corev1.NodeList{}
+		return &corev1.NodeList{}, err
 	}
-	return nil, nodes
+	return nodes, nil
```
bpradipt, line 330: Since the method signature is changed, subsequent changes need to be done where it's called.

bpradipt: I'm wondering why no error is thrown during compilation
```
err, nodesList := r.getNodesWithLabels(map[string]string{"node-role.kubernetes.io/worker": ""})
```

Amulyam24: Missed it due to some local issue on my side, fixed it. PTAL!

bpradipt (approved):
/lgtm
Thanks @Amulyam24

PR #613, opened 2023-02-14
ibmcloud: add e2e test ibmcloud provisioner

huoqifeng:
Fixes: #490

Signed-off-by: Qi Feng Huo <huoqif@cn.ibm.com>

- [x] Provisioner https://github.com/confidential-containers/cloud-api-adaptor/issues/490
  - [x] VPC & Subnet 
  - [x] CoS and PodVM
  - [x] IKS Cluster 
  - [ ] ~~Operator~~ 
  - [x] Use prop files rather than system variables
  - [x] Expose the ability to developer
  - [x] Test

Self managed cluster will be covered by story #620

PR #650, opened 2023-02-28
test/e2e: read provision properties from file

### LINKED ISSUE #592: Implement an e2e test framework
The e2e test framework should implement a common pipeline for all the cloud providers to ensure they are all testing the same scenarios. At the same time it should be extensible to allow specific implementation and test cases.

This framework is meant to be used on CI as well as on developer's workstation.

An PoC was sent in https://github.com/confidential-containers/cloud-api-adaptor/pull/362. The community discussed and decided on have an Go implementation based on the https://github.com/kubernetes-sigs/e2e-framework

wainersm:
This allow passing properties to the provisioner via a file which path should be specified in the TEST_E2E_PROVISION_FILE variable.

Fixes #592

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v66.0.0+incompatible
 	github.com/Azure/go-autorest/autorest v0.11.27
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.11
+	github.com/BurntSushi/toml v1.2.0
 	github.com/avast/retry-go/v4 v4.3.2
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.6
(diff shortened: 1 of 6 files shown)


huoqifeng (approved):
/lgtm, thank you! @wainersm

huoqifeng:
> Hi @huoqifeng ! I think this implements the "Use prop files rather than system variables" feature in your PR #613 . Let me know whether this is what you had in mind or not, otherwise we can work on a better solution.
> 
> BTW, I am working on a way of allow each provider to edit the Kustomize overlay files in `install/overlays/<cloud>` before the operator gets installed. That's why I need to read the properties from somewhere, I chose from a file for security reasons.

I was thinking add `DoKustomize` in interface `CloudProvision` and call it before `kustomizeHelper.Apply`. It's nice if you can add similar implementation, thanks!

stevenhorsman (approved):
LGTM. Thanks @wainersm!

Issue #458, opened 2022-12-22
Refactor cloud provider code


yoheiueda:
Cloud API Adaptor supports multiple cloud providers, but the current implementation contains a lot of duplicated code among multiple providers.

We need to refactor cloud provider code to reduce such duplication as well as addressing the following issues.

### Modification of the duplicated code

If we need to implement a new feature in the duplicated code, we need to write the same code to all of the duplicated code.
Example: https://github.com/confidential-containers/cloud-api-adaptor/pull/381

### Build constrains

We currently rely on build constraints to select a specific cloud provider. We can enable only a single cloud provider in a single binary, and we cannot enable multiple cloud providers in a single binary, even though we need to specify a cloud provider name as a command line argument. This design prevents us to build a common binary that supports all cloud providers. 

https://github.com/confidential-containers/cloud-api-adaptor/blob/2c01e3445274dc250e92c1686e53fbf078d5f458/pkg/adaptor/hypervisor/registry/ibmcloud.go#L1-L13

### Unit tests in CI

Currently, unit tests are executed for all cloud providers. This is redundant, since the same tests are executed multiple times. We can execute unit tests using a single run if we can fix the build constraint problem.

https://github.com/confidential-containers/cloud-api-adaptor/blob/2c01e3445274dc250e92c1686e53fbf078d5f458/.github/workflows/build.yaml#L44

yoheiueda:
We first introduce a common package to refactor cloud provider code, and then refactor each cloud provider code one by one.

- [ ] Introduce common package for the shared code among multiple cloud providers
- [ ] Refactor ibmcloud provider
- [ ] Refactor aws provider
- [ ] Refactor azure provider
- [ ] Refactor vsphere provider
- [ ] Refactor libvirt provider
- [ ] Remove dead code
- [ ] Update CI to avoid redundant unit test execution

yoheiueda:
@bpradipt  refactoring ibmcloud code is done, and we need to refactor the other cloud providers one by one.

The new code layout is described here https://github.com/confidential-containers/cloud-api-adaptor/pull/329#issuecomment-1376843922.

The ibmcode code is https://github.com/confidential-containers/cloud-api-adaptor/pull/329/commits/c89daf702de4e342df8b9b52ec1c19fdec4d59ae for a reference.

knrt10:
I can work on the Azure part if it's okay @yoheiueda ?

bpradipt:
[...]

PR #654, opened 2023-03-01
makefile: Parameterize built-in cloud providers

### LINKED ISSUE #653: Makefile: Parameterize built-in cloud providers
Currently, build-in cloud providers are specified as `-tags` option in `GOFLAGS`.

https://github.com/confidential-containers/cloud-api-adaptor/blob/8f9cf9c3c95fd87106f67ceca87af1abd6030636/Makefile#L21

An explicit variable name `BUILTIN_CLOUD_PROVIDERS` is more comprehensive for users, and will also simplifies the Makefile rules related to `CGO_ENALED`.

yoheiueda:
This PR introduces the BUILTIN_CLOUD_PROVIDERS variable in Makefile to specify cloud providers that to be built into the cloud-api-adaptor binary.

This patch also refactors Makefile to simplify the rules to derive build tags and CGO_ENABLED.

Fixes #653

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         run: |
           # Build each provider separately
-          make GOFLAGS="-tags=${{ matrix.provider }}" build
+          make BUILTIN_CLOUD_PROVIDERS="${{ matrix.provider }}" build
       - name: Test
         run: |
           go install github.com/jstemmer/go-junit-report@v1.0.0
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ ARCH        ?= $(subst x86_64,amd64,$(shell uname -m))
 # Default is dev build. To create release build set RELEASE_BUILD=true
 RELEASE_BUILD ?= false
 CLOUD_PROVIDER ?=
-GOOPTIONS   ?= GOOS=linux GOARCH=$(ARCH)
+GOOPTIONS   ?= GOOS=linux GOARCH=$(ARCH) CGO_ENABLED=0
+GOFLAGS     ?=
 BINARIES    := cloud-api-adaptor agent-protocol-forwarder
 SOURCEDIRS  := ./cmd ./pkg
 PACKAGES    := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
@@ -17,13 +18,9 @@ SOURCES     := $(shell find $(SOURCEDIRS) -name '*.go' -print)
 TEST_E2E_TIMEOUT ?= 20m
 
 ifeq ($(RELEASE_BUILD),true)
-	# Force static build since libvirt provider is not part of the providers to be built. 
-	GOFLAGS ?= -tags=aws,azure,ibmcloud,vsphere
-	# FIXME: this results in agent-protocol-forwarder having <... CGO_ENABLED=0 CGO_ENABLED=0 ...>
-	# Harmless but still needs a better fix
-	GOOPTIONS += CGO_ENABLED=0
+	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere
 else
-	GOFLAGS ?= -tags=aws,azure,ibmcloud,vsphere,libvirt
+	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere libvirt
 endif
 
 all: build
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -55,11 +52,21 @@ help: ## Display this help.
...
 	@test -n "$$(git status --porcelain --untracked-files=no)" && echo -dirty >>$@ || true
 
-agent-protocol-forwarder: .git-commit $(SOURCES)
-	$(GOOPTIONS) CGO_ENABLED=0 go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
+GOFLAGS += -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' \
+                     -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'"
 
-cloud-api-adaptor: .git-commit $(SOURCES)
-	$(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
+# Build tags required to build cloud-api-adaptor are deribed from BUILTIN_CLOUD_PROVIDERS.
```
bpradipt, line 58: nit: deribed -> derived

Makefile
```diff
@@ -55,11 +52,21 @@ help: ## Display this help.
...
+# When libvirt is specified, CGO_ENABLED is set to 1.
+space := $() $()
```
bpradipt, line 60: Can you please explain this ? Is it any number of spaces between the provider strings ?

Makefile
yoheiueda: Fixed. Thank you.

Makefile
yoheiueda: This is the technique described in the GNU Make manual. 
https://www.gnu.org/software/make/manual/html_node/Syntax-of-Functions.html

The number of spaces between provider strings should be one. I don't have an idea to handle consecutive spaces in Makefile. 

If BUILTIN_CLOUD_PROVIDERS contains two consecutive spaces,
```
BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud  vsphere
```

the tag option looks like this.
```
-tags aws,azure,ibmcloud,,vsphere
```

I confirm that `go build` does not complain about this empty build tag, so there is no real harm here.

Makefile
```diff
@@ -55,11 +52,21 @@ help: ## Display this help.
...
+# Build tags required to build cloud-api-adaptor are derived from BUILTIN_CLOUD_PROVIDERS.
+# When libvirt is specified, CGO_ENABLED is set to 1.
+space := $() $()
+comma := ,
+cloud-api-adaptor: GOFLAGS += -tags $(subst $(space),$(comma),$(BUILTIN_CLOUD_PROVIDERS))
+
+ifneq (,$(filter libvirt,$(BUILTIN_CLOUD_PROVIDERS)))
+cloud-api-adaptor: GOOPTIONS := $(subst CGO_ENABLED=0,CGO_ENABLED=1,$(GOOPTIONS))
+endif
+
```
yoheiueda, line 67: The `filter` function accepts whitespace-separated words as an input.
https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

Makefile
bpradipt: Would it be better to have the BUILTIN_CLOUD_PROVIDERS as a comma separated list ?
I have no preferences. Trying to figure out least error prone approach when users want to set custom `BUILTIN_CLOUD_PROVIDERS` before build.

yoheiueda: I noticed that the `scrip` function replaces each consecutive spaces with a single space. I'll use this function to fix the problem.
https://www.gnu.org/software/make/manual/html_node/Text-Functions.html#index-stripping-whitespace

yoheiueda: @bpradipt updated

bpradipt:
@yoheiueda this looks good. Can you please also update docs/DEVELOPMENT.md with instructions to use `BUILTIN_CLOUD_PROVIDERS`

Makefile
```diff
@@ -55,11 +52,21 @@ help: ## Display this help.
...
+space := $() $()
+comma := ,
+cloud-api-adaptor: GOFLAGS += -tags $(subst $(space),$(comma),$(strip $(BUILTIN_CLOUD_PROVIDERS)))
+
+ifneq (,$(filter libvirt,$(BUILTIN_CLOUD_PROVIDERS)))
+cloud-api-adaptor: GOOPTIONS := $(subst CGO_ENABLED=0,CGO_ENABLED=1,$(GOOPTIONS))
+endif
+
+$(BINARIES): .git-commit $(SOURCES)
+	$(GOOPTIONS) go build $(GOFLAGS) -o "$@" "cmd/$@/main.go"
```
snir911, line 69: When the cloud-api-adaptor target is called the GOFLAGS and GOOPTION are overridden only for the cloud-api-adatptor target? i.e. it won't take into account for the forwarder?

Makefile
snir911: `findstring` would fit better here no?

yoheiueda: `filter` recognizes whitespace-separated words, so it does exact matches. `findstring` just searches a substring, so incorrect matches may occur.

Makefile
yoheiueda: Only `cloud-api-adaptor` have build constrains for cloud providers. `agent-protocol-forwarder` does not contains code that are related to cloud providers. So, we only need to specify build tags for `cloud-api-adaptor`.

.github/workflows/build.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         run: |
           # Build each provider separately
-          make GOFLAGS="-tags=${{ matrix.provider }}" build
+          make BUILTIN_CLOUD_PROVIDERS="${{ matrix.provider }}" build
```
yoheiueda, line 40: I also needed to update CI.

I think we can also stop running unti tests per cloud provider, and run unit tests for all providers at once. I will raise another PR to do that.
https://github.com/confidential-containers/cloud-api-adaptor/issues/458#issuecomment-1430780931

yoheiueda:
> @yoheiueda this looks good. Can you please also update docs/DEVELOPMENT.md with instructions to use `BUILTIN_CLOUD_PROVIDERS`

I updated the doc.

bpradipt (approved):
/lgtm
Thanks @yoheiueda

Makefile
yoheiueda: I notice that build tags are also used in another target such as `test-e2e`. I'll update Makefile.

https://github.com/confidential-containers/cloud-api-adaptor/blob/8b0a7a8de5cc458ae6b8255ef8ef476614cc7bca/Makefile#L77-L78

Makefile
```diff
@@ -55,11 +52,21 @@ help: ## Display this help.
...
+GOFLAGS += -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' \
+                     -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'"
 
-cloud-api-adaptor: .git-commit $(SOURCES)
-	$(GOOPTIONS) go build $(GOFLAGS) -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'" -o "$@" "cmd/$@/main.go"
+# Build tags required to build cloud-api-adaptor are derived from BUILTIN_CLOUD_PROVIDERS.
+# When libvirt is specified, CGO_ENABLED is set to 1.
+space := $() $()
+comma := ,
+GOFLAGS += -tags=$(subst $(space),$(comma),$(strip $(BUILTIN_CLOUD_PROVIDERS)))
```
yoheiueda, line 62: build tags are now set for all targets.

snir911: as you mentioned it's not needed for forwarder, but i guess it won't hurt..

snir911 (approved):
LGTM, thanks!

snir911:
@yoheiueda #659 will allow to set the tags as for the caa only as it was before no?

yoheiueda:
> @yoheiueda #659 will allow to set the tags for the caa only as it was before no?

Yes, I am aware of #659. Currently, all cloud providers are liked in `/test/provisioner/provision.go`, but 
I am thinking of applying `BUILTIN_CLOUD_PROVIDERS` to e2e tests as well.

https://github.com/confidential-containers/cloud-api-adaptor/blob/8573d8e55a5fd9e89ceb07fc4b5bddcd30659b2a/test/provisioner/provision.go#L78-L90

PR #659, opened 2023-03-02
e2e: remove build tags in provision_*.go files

### LINKED ISSUE #658: e2e test: compile error if only the tag "ibmcloud" is specified
When attempting to run e2e tests for the ibmcloud provider only, I set tag as ibmcloud only:
```
CLOUD_PROVIDER=ibmcloud GOFLAGS="-tags=ibmcloud" make test-e2e
```

However, I got following compile error:
```
test/e2e/provision.go:80:22: undefined: NewAzureCloudProvisioner
test/e2e/provision.go:82:22: undefined: NewLibvirtProvisioner
note: module requires Go 1.18
FAIL	github.com/confidential-containers/cloud-api-adaptor/test/e2e [build failed]
FAIL
```

genjuro214:
Avoid compile error when using tag to filter tests

Fixes #658

Signed-off-by: Lei Li <genjuro214@gmail.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         run: |
           # Build each provider separately
-          make GOFLAGS="-tags=${{ matrix.provider }}" build
+          make BUILTIN_CLOUD_PROVIDERS="${{ matrix.provider }}" build
       - name: Test
         run: |
           go install github.com/jstemmer/go-junit-report@v1.0.0
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ ARCH        ?= $(subst x86_64,amd64,$(shell uname -m))
 # Default is dev build. To create release build set RELEASE_BUILD=true
 RELEASE_BUILD ?= false
 CLOUD_PROVIDER ?=
-GOOPTIONS   ?= GOOS=linux GOARCH=$(ARCH)
+GOOPTIONS   ?= GOOS=linux GOARCH=$(ARCH) CGO_ENABLED=0
+GOFLAGS     ?=
 BINARIES    := cloud-api-adaptor agent-protocol-forwarder
 SOURCEDIRS  := ./cmd ./pkg
 PACKAGES    := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
@@ -17,13 +18,9 @@ SOURCES     := $(shell find $(SOURCEDIRS) -name '*.go' -print)
 TEST_E2E_TIMEOUT ?= 20m
 
 ifeq ($(RELEASE_BUILD),true)
-	# Force static build since libvirt provider is not part of the providers to be built. 
-	GOFLAGS ?= -tags=aws,azure,ibmcloud,vsphere
-	# FIXME: this results in agent-protocol-forwarder having <... CGO_ENABLED=0 CGO_ENABLED=0 ...>
-	# Harmless but still needs a better fix
-	GOOPTIONS += CGO_ENABLED=0
+	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere
 else
-	GOFLAGS ?= -tags=aws,azure,ibmcloud,vsphere,libvirt
+	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere libvirt
 endif
 
 all: build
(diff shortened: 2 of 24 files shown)


yoheiueda:
Could you add the following build constraint in `test/e2e/provision_libvirt.go`?  

```
//go:build cgo
```

The build tag `cgo` is automatically set only when CGO is enabled. This will be convenient for developers in environments where libvirt library is not available as discussed at https://github.com/confidential-containers/cloud-api-adaptor/pull/618#issuecomment-1437738050

We probably need a new file something like `test/e2e/provision_libvirt_nocgo.go` with the following content.

```
//go:build !cgo

package e2e

func NewLibvirtProvisioner(properties map[string]string) (CloudProvision, error) {
	panic("CGO is not enabled")
}
```

This file will suppress a build error at the `provision.go` when CGO is not enabled. 
https://github.com/confidential-containers/cloud-api-adaptor/blob/8573d8e55a5fd9e89ceb07fc4b5bddcd30659b2a/test/provisioner/provision.go#L82

genjuro214:
@yoheiueda , I have made changes according to your comments above. However, when running e2e with CGO disabled, I encountered the following error:
```
# CGO_ENABLED=0 CLOUD_PROVIDER=ibmcloud make test-e2e
go test -v -tags=aws,azure,ibmcloud,vsphere,libvirt -timeout 20m -count=1 ./test/e2e
# github.com/confidential-containers/cloud-api-adaptor/test/e2e [github.com/confidential-containers/cloud-api-adaptor/test/e2e.test]
test/e2e/libvirt_test.go:23:15: undefined: libvirt.NewConnect
test/e2e/libvirt_test.go:28:38: undefined: libvirt.CONNECT_LIST_DOMAINS_ACTIVE
note: module requires Go 1.18
FAIL	github.com/confidential-containers/cloud-api-adaptor/test/e2e [build failed]
FAIL
```

Do we also need to update the build constraint in `test/e2e/libvirt_test.go`:
```
//go:build libvirt && cgo
```

So that the libvirt tests will not be run with CGO disabled? Is this expected?

yoheiueda:
Oh, yes.

You need to add the same build constraint to `libvirt_test.go`.


```
//go:build cgo
```

If you remove `libvirt` tag from `provision_libvirt.go`, I don't think `libvirt` is necessary in `libvirt_test.go`

genjuro214:
I think `libvirt` is necessary in `libvirt_test.go`, so that we can run libvirt tests only with CGO enabled?
```
CGO_ENABLED=1 GOFLAGS="-tags=libvirt" CLOUD_PROVIDER=libvirt make test-e2e
```

huoqifeng:
@genjuro214 I'm not sure whether we want to remove the build flag because usually the test will be run against a single provider. Maybe an alternative approach is like this https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/cloudmgr/manager.go#L19

huoqifeng:
If the purpose of GetCloudProvisioner factory is intend to remove the build flag, it's OK for me :-) @wainersm

yoheiueda:
> I think `libvirt` is necessary in `libvirt_test.go`, so that we can run libvirt tests only with CGO enabled?
> 
> ```
> CGO_ENABLED=1 GOFLAGS="-tags=libvirt" CLOUD_PROVIDER=libvirt make test-e2e
> ```

Ah, yes. That is correct. Please keep the `libvirt` tag in the file.

BTW, built-in cloud providers are now specified by `BUILTIN_CLOUD_PROVIDERS`. `CLOUD_PROVIDER` does not control the `-tags` option, although directly specifying `GOFLAGS` still works.

https://github.com/confidential-containers/cloud-api-adaptor/pull/654

yoheiueda:
> @genjuro214 I'm not sure whether we want to remove the build flag because usually the test will be run against a single provider.

The current code structure causes a build error when a single cloud provider tag is specified, so I think this PR is valid as a fix for the problem.

>  Maybe an alternative approach is like this https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/cmd/cloud-api-adaptor/cloudmgr/manager.go#L19

I also agree that we should refactor the related code in a similar way, so that a single or subset of cloud providers can be enabled. I think it is OK to refactor the code after this PR is merged.

yoheiueda (approved):
LGTM

huoqifeng (approved):
/lgtm, I'm OK for the fix and let's refactor in later PR, thanks @genjuro214

genjuro214:
Can't rebase and merge due to conflicts. I tried to create a new PR: https://github.com/confidential-containers/cloud-api-adaptor/pull/669
and close this one.

PR #673, opened 2023-03-03
Add timeout to CRI endpoint connection

jtumber-ibm:


stevenhorsman (approved):
This looks reasonable to me and hopefully fixes the issues we've see with hanging on ibm cloud. Thanks @jtumber-ibm!

pkg/adaptor/proxy/proxy.go
```diff
@@ -105,26 +105,27 @@ func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error)
...
+		timeout, cancel := context.WithTimeout(ctx, p.retryInterval)
+		defer cancel()
+		conn, err := grpc.DialContext(timeout, p.criSocketPath, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock(),
 			grpc.WithContextDialer(func(ctx context.Context, target string) (net.Conn, error) {
 				return (&net.Dialer{}).DialContext(ctx, "unix", target)
 			}),
 		)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to established cri uds connection to %s", p.criSocketPath)
```
wainersm, line 118: Could you also print `retryInterval`?

wainersm: hmm... ok, it depends on the error type, it might not fail due to the timeout.

wainersm:
@jtumber-ibm `retryInterval` has been used in `AgentProxy.dial()` and now `initCriClient()`. Wouldn't have sense to have a timeout specific to cri client connection? I suppose it would be a smaller timeout because the connection is local, or I got it wrong?

pkg/adaptor/proxy/proxy.go
jtumber-ibm: This doesn't have any retry logic, as the socket should either be there or not at this point. Unlike the agent-protocol-forwarder connection.

jtumber-ibm:
> @jtumber-ibm `retryInterval` has been used in `AgentProxy.dial()` and now `initCriClient()`. Wouldn't have sense to have a timeout specific to cri client connection? I suppose it would be a smaller timeout because the connection is local, or I got it wrong?

We could reduce the timeout for the cri cilent connection, but I don't see the benefits to be honest. The next step after here is the agent-protocol-forwarder connection which normally takes 20+ seconds for the podvm to come online so the overall performance doesn't get degraded. But I'm not opposed to adding in a different timeout just for cri client.

wainersm (approved):
LGTM

PR #681, opened 2023-03-08
Enable tls communication between agent proxy and agent-protocol-forwarder

bpradipt:


entrypoint.sh
```diff
@@ -12,6 +12,10 @@ optionals+=""
 [[ -S ${CRI_RUNTIME_ENDPOINT} ]] && optionals+="-cri-runtime-endpoint ${CRI_RUNTIME_ENDPOINT} "
 [[ "${PAUSE_IMAGE}" ]] && optionals+="-pause-image ${PAUSE_IMAGE} "
 [[ "${VXLAN_PORT}" ]] && optionals+="-vxlan-port ${VXLAN_PORT} "
+[[ "${CACERT_FILE}" ]] && optionals+="-ca-cert-file ${CACERT_FILE} "
+[[ "${CERT_FILE}" ]] && optionals+="-cert-file ${CERT_FILE} "
+[[ "${CERT_KEY}" ]] && optionals+="-cert-key ${CERT_KEY} "
```
snir911, line 17: the three of these must be set together no? maybe worth checking it

bpradipt: Good point. The cert_file and cert_key are must. And if self-signed certificate is used then cacert_file is also needed. I have added this in doc. I'll update the entrypoint.sh to reflect this.

podvm/qcow2/misc-settings.sh
```diff
@@ -38,4 +38,14 @@ then
...
 fi
+
+
+if [ -e /etc/certificates/tls.crt ] && [ -e /etc/certificates/tls.key ] && [ -e /etc/certificates/ca.crt ]; then
+    # Update systemd service file to add additional options
+    sed -i "/ExecStart/s/$/ \-cert\-file \/etc\/certificates\/tls.crt \-cert\-key \/etc\/certificates\/tls.key \-ca\-cert\-file \/etc\/certificates\/ca.crt/" /etc/systemd/system/agent-protocol-forwarder.service
+elif [ -e /etc/certificates/tls.crt ] && [ -e /etc/certificates/tls.key ] && [ ! -e /etc/certificates/ca.crt ]; then
+    # Update systemd service file to add additional options
+    sed -i "/ExecStart/s/$/ \-cert\-file \/etc\/certificates\/tls.crt \-cert\-key \/etc\/certificates\/tls.key/" /etc/systemd/system/agent-protocol-forwarder.service
+fi
```
yoheiueda, line 49: How about using the `EnvironmentFile` directive in `agent-protocol-forwarder.service`, instead of directly modifying it?

`agent-protocol-forwarder.service` will looks like this.
```
[Service]
ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -kata-agent-socket /run/kata-containers/agent.sock $TLS_OPTIONS
EnvironmentFile=-/etc/default/agent-protocol-forwarder
```

`misc-settings.sh` will look like this.
```
if [ -e /etc/certificates/tls.crt ] && [ -e /etc/certificates/tls.key ] && [ -e /etc/certificates/ca.crt ]; then
   cat <<END >> /etc/default/agent-protocol-forwarder
TLS_OPTIONS=-cert-file /etc/certificates/tls.crt -cert-key /etc/certificates/tls.key -ca-cert-file /etc/certificates/ca.crt
END
elif
...
```

pkg/adaptor/proxy/proxy.go
```diff
@@ -62,12 +66,33 @@ func NewAgentProxy(socketPath, criSocketPath string, pauseImage string) AgentPro
...
+
+		logger.Printf("TLS is configured. Configure TLS dialer")
+
+		// Create a TLS configuration object
+		config, err := tlsutil.GetTLSConfigFor(p.tlsConfig)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create tls config: %v", err)
+		}
+		dialer.Config = config
+	}
```
yoheiueda, line 95: Is it possible to simply this code like this?

```
	var dialer interface {
		DialContext(ctx context.Context, network, address string) (net.Conn, error)
	}

	if p.tlsConfig != nil && !reflect.DeepEqual(tlsutil.TLSConfig{}, *p.tlsConfig) {
		logger.Printf("TLS is configured. Configure TLS dialer")

		// Create a TLS configuration object
		config, err := tlsutil.GetTLSConfigFor(p.tlsConfig)
		if err != nil {
			return nil, fmt.Errorf("Failed to create tls config: %v", err)
		}
		dialer = &tls.Dialer{
			Config: config,
		}
	} else {
		dialer = &net.Dialer{}
	}
```

pkg/adaptor/proxy/proxy.go
```diff
@@ -78,8 +103,13 @@ func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error)
...
 
-			// TODO: Support TLS
-			conn, err = (&net.Dialer{}).DialContext(ctx, "tcp", address)
+			if dialer.Config == nil {
+				// Use generic tcp dialer
+				conn, err = dialer.NetDialer.DialContext(ctx, "tcp", address)
+			} else {
+				// Use tls dialer
+				conn, err = dialer.DialContext(ctx, "tcp", address)
+			}
```
yoheiueda, line 112: Then, these line will be like this.
```
conn, err = dialer.DialContext(ctx, "tcp", address)
```

pkg/adaptor/proxy/proxy.go
```diff
@@ -62,12 +66,33 @@ func NewAgentProxy(socketPath, criSocketPath string, pauseImage string) AgentPro
...
 }
 
 func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error) {
 
 	var conn net.Conn
+	var config *tls.Config
+
+	dialer := &tls.Dialer{
+		NetDialer: &net.Dialer{
+			Timeout: p.retryInterval,
```
yoheiueda, line 80: Isn't this `Timeout` redundant, since `ctx, cancel := context.WithTimeout(ctx, p.retryInterval)` is in line 103?

pkg/util/tls/tls.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to load root certificates: %w", err)
+		}
+		tlsConfig.RootCAs = rootCAs
+		// This is important otherwise you'll hit the following error
+		// cannot validate certificate for <IP> because it doesn't contain any IP SAN
+		// Since it's not possible to know the IP address of the Kata VM apriori, using a well-defined
+		// hostname here. Other option is to create certificates with IP SAN having all the
+		// IPs in the network range
+		tlsConfig.ServerName = podvmServername
```
yoheiueda, line 147: Is this also set in pod VM side? Probably it is harmless, but may be confusing.

I think this is important point, so we should explicitly set this filed in the dialer code of proxy in the cloud-api-adaptor side.

docs/tls-proxy-forwarder.md
```diff
@@ -0,0 +1,73 @@
...
+plane communication between `cloud-api-adaptor` that is running in the K8s worker
+node and the `agent-protocol-forwarder` that is running in the Pod VM. More
+specifically the communication between the proxy, that is run via
+`cloud-api-adaptor` and the `agent-protocol-forwarder` that is running inside the
+Pod VM will be using TLS.
+
+### Add the server certificate to the Pod VM image
+
+- Copy the server certificate, key and (optionally) the CA certificate to `~/cloud-api-adaptor/podvm/files/etc/certificates`.
+You will need CA certificate if using self-signed certificates.
```
yoheiueda, line 26: Do we need `.gitignore` for these files?

pkg/util/tls/tls.go
bpradipt: This is not required in the pod VM side and it's harmless.

pkg/adaptor/proxy/proxy.go
bpradipt: Right. It's not needed.

pkg/adaptor/proxy/proxy.go
bpradipt: Thanks for the suggestion. I'll try this and let you know.

bpradipt: I tested the changes on libvirt using both tls and plain connection

docs/tls-proxy-forwarder.md
bpradipt: Yeah.. better to ignore the certificate files inside the directory

podvm/qcow2/misc-settings.sh
bpradipt: I have used the above approach in the updated PR

pkg/adaptor/proxy/proxy.go
bpradipt: Done

podvm/qcow2/misc-settings.sh
```diff
@@ -38,4 +38,17 @@ then
 # else if...
 #
 fi
+
+if [ -e /etc/certificates/tls.crt ] && [ -e /etc/certificates/tls.key ] && [ -e /etc/certificates/ca.crt ]; then
+    # Update systemd service file to add additional options
+    cat <<END >> /etc/default/agent-protocol-forwarder
+    TLS_OPTIONS=-cert-file /etc/certificates/tls.crt -cert-key /etc/certificates/tls.key -ca-cert-file /etc/certificates/ca.crt
```
yoheiueda, line 45: The leading spaces before `TLS_OPTIONS` are not necessary.

podvm/qcow2/misc-settings.sh
```diff
@@ -38,4 +38,17 @@ then
...
+END
+elif [ -e /etc/certificates/tls.crt ] && [ -e /etc/certificates/tls.key ] && [ ! -e /etc/certificates/ca.crt ]; then
+    # Update systemd service file to add additional options
+    cat <<END >> /etc/default/agent-protocol-forwarder
+    TLS_OPTIONS=-cert-file /etc/certificates/tls.crt -cert-key /etc/certificates/tls.key
```
yoheiueda, line 50: same as above

yoheiueda (approved):
LGTM other than the minor comments.

entrypoint.sh
```diff
@@ -12,6 +12,10 @@ optionals+=""
 [[ -S ${CRI_RUNTIME_ENDPOINT} ]] && optionals+="-cri-runtime-endpoint ${CRI_RUNTIME_ENDPOINT} "
 [[ "${PAUSE_IMAGE}" ]] && optionals+="-pause-image ${PAUSE_IMAGE} "
 [[ "${VXLAN_PORT}" ]] && optionals+="-vxlan-port ${VXLAN_PORT} "
+[[ "${CACERT_FILE}" ]] && optionals+="-ca-cert-file ${CACERT_FILE} "
+[[ "${CERT_FILE}" ]] && [[ "${CERT_KEY}" ]] && optionals+="-cert-file ${CERT_FILE} -cert-key ${CERT_KEY} "
+[[ "${CERT_KEY}" ]] && optionals+="-cert-key ${CERT_KEY} "
```
snir911, line 17: Is it needed here? now if you set CERT_FILE and CERT_KEY the cmdline will look as follow:
`-cert-file ${CERT_FILE} -cert-key ${CERT_KEY} -cert-key ${CERT_KEY} `

install/overlays/aws/tls_certs_volume_mount.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - name: cloud-api-adaptor-con
+        volumeMounts:
+        - mountPath: /etc/certificates
+          name: certs
+      volumes:
+      - name: certs
+        secret:
+          secretName: certs-for-tls
+
+# to apply this uncomment the patchesStrategicMerge of this file in kustomization.yaml
```
snir911, line 21: Just a thought for the future, maybe we can avoid duplicating these sort of files and make kustomize to pick them from the same directory for all providers

snir911 (approved):
LGTM, thanks, pls check the comment on entrypoint.sh

install/overlays/aws/tls_certs_volume_mount.yaml
bpradipt: Yeah. It'll be much simpler.

entrypoint.sh
bpradipt: Good catch. Updated it

Issue #689, opened 2023-03-09
Make e2e ibmcloud provisioner for managed cluster robust and support overlay operator


huoqifeng:
- [ ] Generate `COS_SERVICE_URL` and `VPC_SERVICE_URL` based on region
- [x] Use VPC and Subnet `name` only without ID in properties and retrieve the ID if it's existing, create it otherwise.
- [ ] Look up `COS_INSTANCE_ID` by `COS_INSTANCE_NAME`??
- [x] Look up `SSH_KEY_ID` by `SSH_KEY_NAME`?? or accept a public key and upload it??
- [x] Check IKS cluster created completely before install operator
- [x] Check IKS cluster deleted completely before delete subnet and vpc
- ??

huoqifeng:
@stevenhorsman @liudalibj please add more comments here for any enhancements you have thought about.

huoqifeng:
Looks this is not reasonable always: `Generate COS_SERVICE_URL and VPC_SERVICE_URL based on region`.
Because the URL might not be related with region only.

PR #703, opened 2023-03-13
ibmcloud: e2e prov remove iam and vpc url in props

huoqifeng:
Fixes: #689


yoheiueda (approved):
LGTM

stevenhorsman (approved):
This looks great in allows overriding of the `IAM_SERVICE_URL` and `VPC_SERVICE_URL` values if required (e.g. in staging), but not having them in the default properties file which makes it easer for 'normal' users. Thanks Qifeng!

PR #700, opened 2023-03-09
vsphere: Add additional startup option checking

caavery:
Fixes: #635

This patch also updates kustomization.yaml with the new options and provides better descriptions.

PR #704, opened 2023-03-13
vsphere: Add additional startup option checking

caavery:
Fixes: #635

This patch also updates kustomization.yaml with the new options and provides better descriptions.


caavery:
New pull request replacing  PR #700

pkg/adaptor/cloud/vsphere/provider.go
```diff
@@ -49,6 +54,27 @@ func NewProvider(config *Config) (cloud.Provider, error) {
...
 }
 
+func checkConfig(config *Config) error {
+
+	// Do some initial checks of the optional input values
+
+	if config.DRS == "true" {
+		if config.Cluster == "" {
+			return fmt.Errorf("Error: A cluster name is required with DRS")
+		}
```
snir911, line 64: nit: i wonder if it make sense to have fallback try to use Host and Datastore in cases where DRS=true but cluster is not set

install/overlays/vsphere/kustomization.yaml
```diff
@@ -18,9 +18,32 @@ configMapGenerator:
...
+                       # If not using DRS a datastore name is required.
+                       # If using DRS the datastore name is ignored.
+
+  #- GOVC_FOLDER=""     # Uncomment to place your peerpod VM in an existing desintation folder
+                        # or create a new one if it does not exist in the VM inventory path
+                        # (GOVC_DATACENTER/vm/GOVC_FOLDER).
+
   #- PAUSE_IMAGE="" # Uncomment and set if you want to use a specific pause image
-  #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
+  - VXLAN_PORT="" # set to use "9000" or change if you want to use a specific vxlan port. Defaults to 4789.
```
snir911, line 46: better to keep commented if not mandatory

snir911 (commented):
Thanks @caavery, LGTM overall, i added minor comments

pkg/adaptor/cloud/vsphere/provider.go
caavery: The problem with that is the user has misconfigured the CAA and we need to point that out. They are asking for DRS and if we are not doing what they ask to try to help them it may sound good but it also leads to confusion. Meaning DRS requires a cluster if they have omitted a cluster name they have ignored the most crucial part of the config.

install/overlays/vsphere/kustomization.yaml
caavery: K I'll fix this.

install/overlays/vsphere/kustomization.yaml
```diff
@@ -18,9 +18,32 @@ configMapGenerator:
...
+                       # If not using DRS a datastore name is required.
+                       # If using DRS the datastore name is ignored.
+
+  #- GOVC_FOLDER=""     # Uncomment to place your peerpod VM in an existing desintation folder
+                        # or create a new one if it does not exist in the VM inventory path
+                        # (GOVC_DATACENTER/vm/GOVC_FOLDER).
+
   #- PAUSE_IMAGE="" # Uncomment and set if you want to use a specific pause image
-  #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
+  # - VXLAN_PORT="" # Uncomment and set to use "9000" or change if you want to use a specific vxlan port. Defaults to 4789.
```
snir911, line 46: very nit: seems like there's additional space here

snir911 (approved):
Thanks @caavery !

install/overlays/vsphere/kustomization.yaml
```diff
@@ -36,8 +59,12 @@ secretGenerator:
 - name: peer-pods-secret
   namespace: confidential-containers-system
   literals:
-  - GOVC_USERNAME="" # set
-  - GOVC_PASSWORD="" # set
+  - GOVC_USERNAME=""   # set of the vCenter username is required.
```
bpradipt, line 62: nit: `set of the vCenter` -> `set if the vCenter`

install/overlays/vsphere/kustomization.yaml
```diff
@@ -36,8 +59,12 @@ secretGenerator:
...
+
+  - GOVC_PASSWORD=""   # set of the vCenter password is required.
+
```
bpradipt, line 65: nit: `set of the vCenter` -> `set if the vCenter`

pkg/adaptor/cloud/vsphere/provider.go
```diff
@@ -49,6 +54,27 @@ func NewProvider(config *Config) (cloud.Provider, error) {
...
+		return nil
+	}
+
+	// Hosts and Datastores are required when not using DRS
+	// We cannot check for a cluster as hosts do not have to be part of a cluster.
+
+	if config.Host == "" || config.Datastore == "" {
+		return fmt.Errorf("Error: A host and datastore name are required")
```
bpradipt, line 72: Are both host and datastore needed or either of them?

pkg/adaptor/cloud/vsphere/provider.go
```diff
@@ -35,6 +35,11 @@ func NewProvider(config *Config) (cloud.Provider, error) {
 
 	logger.Printf("vsphere config: %#v", config.Redact())
 
+	err := checkConfig(config)
+	if err != nil {
```
bpradipt, line 39: Should the error be logged here ?

install/overlays/vsphere/kustomization.yaml
caavery: The set is required. There is no if.  I could say # The setting of the vCenter blah is required.

pkg/adaptor/cloud/vsphere/provider.go
caavery: Both are required so if either of them are empty it is an error.

install/overlays/vsphere/kustomization.yaml
caavery: Or  'Setting the vCenter blah is required'

pkg/adaptor/cloud/vsphere/provider.go
caavery: All errors from NewProvider are logged in cmd/cloud-api-adaptor/main.go
func main() {

	starter, err := config.Setup()
	if err != nil {
		fmt.Fprintf(os.Stderr, "%s: %s\n", os.Args[0], err)
		cmd.Exit(1)
	}

So if I wanted to be consistent I should remove this logging in NewProvider and return the formatted error
govmomiClient, err := NewGovmomiClient(*config)
        if err != nil {
                logger.Printf("Error creating vcenter session for cloud provider:  %s", err)
                return nil, err
        }

install/overlays/vsphere/kustomization.yaml
caavery: Yeah same as below. There is no if. They must be set.

pkg/adaptor/cloud/vsphere/provider.go
bpradipt: thanks for the clarification

bpradipt (approved):
/lgtm
Thanks @caavery

Issue #708, opened 2023-03-14
remove the -<cloudprovider> suffix from all k8s objects


snir911:
as we moved to single binary it would be more suitable to have a generic name than a  cloud-provider specific naming

PR #710, opened 2023-03-14
controller: remove cloud-provider suffix from k8s objects

snir911:
deployments, cm, secrets..

Fixes: #708

### DIFF
--- a/install/overlays/aws/kustomization.yaml
+++ b/install/overlays/aws/kustomization.yaml
@@ -3,7 +3,6 @@ kind: Kustomization
 
 bases:
 - ../../yamls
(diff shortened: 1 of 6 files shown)


snir911:
@wainersm no, currently we supports only single cloud provider at a time, the same provider that host the cluster nodes, I'm not aware to any plans to change it
The two reasons i find to remove the suffix are that we moved to not cloud specific binary and that it will simplify secret/cm handling from the controllers

yoheiueda (approved):
LGTM

bpradipt (approved):
/lgtm
Thanks @snir911

Issue #375, opened 2022-11-26
Handle dangling cloud resources


bpradipt:
In the following issue - https://github.com/confidential-containers/cloud-api-adaptor/issues/328 we discussed about the dangling cloud resource problem. Currently cloud-api-adaptor (CAA) is stateless and this results in cases where the pod VM instances created by the caa is not deleted if the worker node reboots (on error, node maintenance etc).

To handle such cases, we felt the need to have a CRD for pod VM to maintain the state and perform required cleanups.

cc @yoheiueda @jensfr @surajssd

yoheiueda:
Currently, `cloud-api-adaptor` uses a on-memory map to manage mapping between VM instances and pod VMs. `cloud-api-adaptor` runs on a worker node, so a worker node crash results in a loss of the mapping information. This leads to dangling pod VMs.

To solve this problem, we can use a CRD object to represent a pod VM.  CRD objects are resilient to worker node crashes, and we can implement a controller to monitor CRD objects to clean dangling pod VMs.

This is the (incomplete) controller code that implements a similar mechanism. This code comes from  my project project, so cannot directly be applied to cloud-api-adaptor, but we can use it as a reference.

https://gist.github.com/yoheiueda/122619c2b12457edd46105741a14bf09

PeerPod is a CRD that represents a peer pod VM. There is a 1-1 mapping between Pod and PeerPod objects. A PeerPod objects contains VM instance information such as ID and IP address.

A PeerPod has a owner reference to the Pod object. This reference manages mapping between PeerPod and Pod, and guarantees the PeerPod object is automatically deleted, when the corresponding Pod object is deleted.
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

When the controller detects a new Pod creation with a specific runtimeClass, it creates a PeerPod object with a owner reference to the original Pod object.

The controller monitors PeerPod object creation, and when it detects a new PeerPod object, it talks with a cloud API endpoint to create a new VM instance. When it obtains VM information such as IP address of the newly created VM instance, it updates the PeerPod object to include such VM information.

When CAA receives a CreateSandboxRequest from Kata shim, it talks with Kubernetes API server to wait for the corresponding PeerPod object becoming ready. Then, CAA gets the IP address information from the updated PeerPod object, and establishes a tunnel for agent protocol communication between kata shim and kata agent.

When the controller detects deletion of a PeerPod objects, it talks with the cloud API endpoint to delete the VM instance.

[...]

PR #701, opened 2023-03-10
ibmcloud: e2e test harden cluster subnet and vpc

huoqifeng:
Fixes: #689

PR #711, opened 2023-03-15
PeerPods CR & controller for handling dangling cloud resources

snir911:
Fixes: #375 
Here's a suggestion for monitoring and deleting dangling resources, this is RFC as I'd like to get feedback on the general idea and implementation but it's also WIP as there're still things to fix and improve

**Architecture & Workflow:**

Creation time:
With every successful peer-pod VM creation cloud-api-adaptor will create also a PeePod CR (predefined by the operator) which contains the cloud provider and its VM instance id.

Owner references:
The PeerPod CR is owned by the original Pod object that represents the peer-pod, upon pod deletion background cascading deletion getting into action which means the Pod will be deleted first and the GC will handle the owned PeerPod CR later.

Deletion time:
(Normal case) When remote hypervisor will get the stopVM request (upon pod deletion) it will delete the podVM instance and if successful it will remove the finalizer attached to the owned PeerPod object so it can then be cleaned by the GC
(Failure case) In case for some reason cloud-api-adaptor doesn’t honor the delete request of it fails to delete it, the finalizer is not removed and then when peer-pods controller will recognize deletion that the GC is trying to perform for the owned PeerPod object and that it still has the finalizer it will comprehend it needs to preform the deletion of podVM resource based on the PeerPod CR fields

* Currently supports and tested only with libvirt and aws

 TODOs:
- [ ] Docs
- [x] Move to a separate directory?
- [x] Adapt to #710
- [ ] Tests
- [ ] support more providers
- general fixes &  improvements

Depends-on: #755 (caa side)
to test use https://github.com/snir911/cloud-api-adaptor/tree/375merged
with the mentioned images


peer-pod-controller/controllers/peerpod_controller.go
```diff
@@ -0,0 +1,148 @@
...
+		libvirtcfg.PoolName = getEnvOrDefault("LIBVIRT_POOL", "default")
+		libvirtcfg.NetworkName = getEnvOrDefault("LIBVIRT_NET", "default")
+		provider, _ = libvirt.NewProvider(&libvirtcfg)
+	default:
+		return fmt.Errorf("unknown provider: %s", cloudName)
+	}
+
+	if cloud := cloudmgr.Get(cloudName); cloud != nil {
+		cloud.LoadEnv() // this doesn't work ATM as it requires providers build tags for cloudmgr
+	}
```
yoheiueda, line 138: If we can specify build tags here, I think we can simplify these lines to something like this.

```
	cloud := cloudmgr.Get(cloudName)
	if cloud == nil {
		return fmt.Errorf("unknown provider: %s", cloudName)
	}
	provider, err := cloud.NewProvider()
	if err != nil {
		return err
	}
```

peer-pod-controller/Dockerfile
```diff
@@ -15,13 +16,11 @@ COPY api/ api/
 COPY controllers/ controllers/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
```
yoheiueda, line 19: Can we specify build tags for cloud providers here in the same way of other binariers like `cloud-api-adaptor` and `cluster-provisioner`?

Build tags are derived from `BUILTIN_CLOUD_PROVIDERS` in the main Makefile.
https://github.com/confidential-containers/cloud-api-adaptor/blob/900b330e4502ba219abcd900bdf7c541539e2a1c/Makefile#L21-L26

peer-pod-controller/controllers/peerpod_controller.go
```diff
@@ -0,0 +1,148 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if pp.ObjectMeta.GetDeletionTimestamp() == nil { // TODO: consider filter events
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.setProvider(ctx, pp.Spec.CloudProvider); err != nil { // TODO: set only once?
+		return ctrl.Result{}, err
+	}
```
yoheiueda, line 70: If we only need to support a single cloud provider in a Kubernetes cluster, we don't need to set a cloud provider per pod here.

We can specify a cloud provider name as an argument to the controller binary, and create a Provider obect in `mainl.go` like [`cmd/cloud-api-adaptor/main.go`](https://github.com/confidential-containers/cloud-api-adaptor/blob/900b330e4502ba219abcd900bdf7c541539e2a1c/cmd/cloud-api-adaptor/main.go). Then we can pass the Provider object to this controller.

peer-pod-controller/main.go
```diff
@@ -84,6 +85,13 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "PeerPodConfig")
 		os.Exit(1)
 	}
+	if err = (&controllers.PeerPodReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "PeerPod")
+		os.Exit(1)
+	}
```
yoheiueda, line 94: As mentioned above, if we can specify a cloud provider name as a command line argument, we can create a Provider object here, and pass it in this struct.

pkg/util/k8sops/peerpod.go
yoheiueda, line 1: I think this package is only used in `"pkg/adaptor/cloud"`, so we can put this package under `"pkg/adaptor"`

peer-pod-controller/controllers/peerpod_controller.go
```diff
@@ -0,0 +1,148 @@
...
+
+	logger.Info("deleting instance", "InstanceID", pp.Spec.InstanceID, "CloudProvider", pp.Spec.CloudProvider)
+	if err := provider.DeleteInstance(ctx, pp.Spec.InstanceID); err != nil {
+		return ctrl.Result{}, err
+	}
```
yoheiueda, line 75: Is `DeleteInstance` called by both controller and `cloud-api-adaptor`? If so, I think DeleteInstance will return an error like `instance not found`. Such an error probably means that the instance has been already deleted. Is it harmless to return such error here?

pkg/util/k8sops/peerpod.go
```diff
@@ -0,0 +1,125 @@
...
+	client        *kubernetes.Clientset
+	uclient       *rest.RESTClient // use generated client instaed
+	cloudProvider string
+}
+
+func NewPeerPodService() *PeerPodService {
+	cloudProvider := os.Getenv("CLOUD_PROVIDER") // TODO: don't get from env var directly
+	if cloudProvider == "" {
+		logger.Printf("NewPeerPodService: failed to get cloudProvider")
+		return nil
```
yoheiueda, line 36: Can we return an error explicitly like this?

```
		return nil, errors.New("NewPeerPodService: failed to get cloudProvider")
```

pkg/adaptor/cloud/cloud.go
```diff
@@ -124,6 +126,9 @@ func NewService(provider Provider, proxyFactory proxy.Factory, workerNode podnet
 		workerNode:   workerNode,
 	}
 	s.cond = sync.NewCond(&s.mutex)
+	if s.ppService = k8sops.NewPeerPodService(); s.ppService == nil {
+		logger.Printf("failed to create PeerPodService, runtime failure may result in dangling resources")
+	}
```
yoheiueda, line 131: Can we handle errors explicitly like this?

```
if ppService, err  := k8sops.NewPeerPodService(); err != nil {
	logger.Printf("failed to create PeerPodService, runtime failure may result in dangling resources: %v", err)
} else {
	s.ppService = ppService
}
```

yoheiueda (commented):
Thank you very much for proposing this great enhancement! This looks a right approach to me.

peer-pod-controller/api/v1alpha1/peerpod_types.go
```diff
@@ -0,0 +1,67 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// PeerPodSpec defines the desired state of PeerPod
+type PeerPodSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Foo is an example field of PeerPod. Edit peerpod_types.go to remove/update
+	CloudProvider string `json:"cloudProvider,omitempty"`
```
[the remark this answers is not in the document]
snir911: Not really, at first i used it as it was not very simple to pass parameters at build time using the operator-sdk.
I'm still not sure how to do it but it indeed will be useful if we'll support multi cloud, and generally it just make sense to have context about the which provider the InstanceID refers to

peer-pod-controller/controllers/peerpod_controller.go
snir911: I couldn't find how to do it, IIRC adding tags caused the controller-gen to fail
I thought we may be able to go the other way around, i.e. to build all unless there's a tag

peer-pod-controller/Dockerfile
snir911: here's tags might work I'll have a look

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork v1.1.0
 	github.com/IBM-Cloud/bluemix-go v0.0.0-20230120122421-afb48116b8f1
 	github.com/IBM/go-sdk-core/v5 v5.10.2
-	github.com/IBM/vpc-go-sdk v0.32.0
+	github.com/IBM/vpc-go-sdk v1.0.1
```
stevenhorsman, line 12: I think QiFeng purposely downgraded the vpc-go-sdk to the latest maintained one: https://github.com/confidential-containers/cloud-api-adaptor/pull/701#issuecomment-1463371224

peer-pod-controller/controllers/peerpod_controller.go
snir911: It seems the returned error comes from the provider's api, I'l check for the existence of the finalizer before deleting so that also if the deletion will fail due to `instance not found`, the next try will skip it and complete with success

peer-pod-controller/controllers/peerpod_controller.go
snir911: I used CLOUD_PROVIDER instead of setting as parameter to the binary

peer-pod-controller/controllers/peerpod_controller.go
snir911: done

snir911:
I addresses all the issues commented, i hit some module issues due to the pkgs movements, hence, it blocks me from testing ATM but i don't expect code to have additional dramatic changes.
(also need to rename peer-pod-ctlr to peer-pod-ctrl)

peer-pod-controller/api/v1alpha1/peerpod_types.go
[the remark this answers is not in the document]
yoheiueda: OK. If we want to support multiple cloud providers, there is also possibility to support multiple accounts of a single cloud provider. So, we need a label to represents a pair of cloud provider name and account identifier.

A possible simple way to support such situation is to run multiple controllers, each of which is responsible only for a single cloud provider and a single account.  A controller should have a startup option to specify a value to be used for the label. The default value can be the same string of a cloud provider, since a single account is usually used for a single cloud provider.

Each controller determines which PeerPod objects it should handle by filtering the objects with the label.

snir911:
removed the WIP & RFC, please review
**NOTE:** it still requires to remove some hacks that was done for testing before merging
all other changes can be done as follow up also supporting other providers

peer-pod-ctrl/controllers/peerpod_controller.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2023.
```
bpradipt, line 2: nit: Better to use the confidential containers copyright

yoheiueda:
> Please remove any commits that are really not required for peerpod CR. For example I think this commit is not needed - all: mv cloudmgr pkg under pkg/

I think this specific change is based on my previous comment.

It is OK with me to keep using the original package location of `cmd/cloud-api-adaptor/cloudmgr`. I'll submit another PR to reorganize the `cloudmgr` functionality into each cloud provider code in `pkg/cloud/*`.

bpradipt:
> > Please remove any commits that are really not required for peerpod CR. For example I think this commit is not needed - all: mv cloudmgr pkg under pkg/
> 
> I think this specific change is based on my previous comment.
> 
> It is OK with me to keep using the original package location of `cmd/cloud-api-adaptor/cloudmgr`. I'll submit another PR to reorganize the `cloudmgr` functionality into each cloud provider code in `pkg/cloud/*`.

@yoheiueda thanks for clarifying. I think since this PR is quite big, keeping it specific to the controller required changes will be better from review and test standpoint. wdyt ?

snir911:
> @snir911 thanks for implementing this important feature. Overall it looks good. I need to do a second pass but before that few generic comments to help move forward.
> 
> 1. Please squash few commits. For example the reconcile loop function and the api etc creation can be in one PR.

you mean to split the api calles from the CAA to a different PR?

> 2. Please remove any commits that are really not required for peerpod CR. For example I think this commit is not needed - `all: mv cloudmgr pkg under pkg/`

As @yoheiueda mentioned i moved it as he requested, however, i think reverting it and reorganize later also make sense, also i think it would help to avoid the need to point go.mod to the root dir
 
> 3. Add proper copyright statements for CoCo

ok
> 4. Please rename the folder to `peerpod-controller` to match the CR - `peerpod`.

The reason i used "ctrl" and not "controller" is because when i tried to stick with this naming in the other operator ([peer-pod-config-controller](https://github.com/confidential-containers/cloud-api-adaptor/pull/714) it resulted in failure with some of the objects definitions due to a too long name (default names the operator-sdk is using)
Of course it can be modified, but i thought it would be better to be unified with the naming conventions in both operator

bpradipt:
> > @snir911 thanks for implementing this important feature. Overall it looks good. I need to do a second pass but before that few generic comments to help move forward.
> > 
> > 1. Please squash few commits. For example the reconcile loop function and the api etc creation can be in one PR.
> 
> you mean to split the api calles from the CAA to a different PR?

These two commits
- 8e0206b11a9e3d73a9090ea08592e379a7e59553
- b2b243461f9ebc4851cdf9855c3e0f4d48ab801e
- 
> 
> > 4. Please rename the folder to `peerpod-controller` to match the CR - `peerpod`.
> 
> The reason i used "ctrl" and not "controller" is because when i tried to stick with this naming in the other operator ([peer-pod-config-controller](https://github.com/confidential-containers/cloud-api-adaptor/pull/714) it resulted in failure with some of the objects definitions due to a too long name (default names the operator-sdk is using) Of course it can be modified, but i thought it would be better to be unified with the naming conventions in both operator

:-( ..Let's use `peerpod-ctrl` (peerpod as a single word to denote the CR)

bpradipt (approved):
/lgtm
Thanks @snir911. The best part is that the actual controller code is pretty less :)

peer-pod-controller/api/v1alpha1/peerpod_types.go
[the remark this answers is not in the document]
yoheiueda: I think is OK to keep this field. We will come back when we support multiple cloud providers.

yoheiueda (approved):
LGTM

Issue #705, opened 2023-03-14
Automatic generation of TLS certificates


yoheiueda:
The agent protocol communication between `cloud-api-adaptor` and `agent-protocol-forwarder` can now be encrypted with TLS when certificates and keys are specified in the command line argument. #681

I think we can automate this TLS configuration by generating necessary certificates in cloud-api-adaptor in most of use cases.

From the perspective of `agent-protocol-forwarder`, network traffics coming from outside of a peer pod VM are not trusted at all. `cloud-api-adaptor` is not a trusted entity for `agent-protocol-forwarder`, so it can't trust the incoming network traffic even if the connection is encrypted. Ultimately, `agent-protocol-forwarder` should validate the incoming agent API requests by other means.

From the perspective of `cloud-api-adaptor`, TLS still has some meanings to reduce security risks as a defense-in-depth approach.

Since administrators of `cloud-api-adaptor` need to trust cloud environments such as the worker node VM instance and its background cloud infrastructure, `cloud-api-adaptor` can generate a pair of server cert/key for a peer pod VM, and then send them to the peer pod VM via cloud-init data.

I don't think sending a server private key to a peer pod VM via cloud-init data degrades security. From the perspective of `cloud-api-adaptor`, cloud-init data is considered as a secure communication channel. From the perspective of `agent-protocol-forwarder`, the TLS connection is not trusted anyway.

The steps to generate certificates and keys will be something like this. We can reuse the TLS mechanism introduced in #681.

1. At start up, `cloud-api-adaptor` generates the following two sets of certificate/key pairs
   * Self-signed client certificate and its private key
   * Self-signed server CA certificate and its private key
2. Before creating a peer pod VM, `cloud-api-adaptor` generates a pair of server certificate and private key using the server CA certificate.
3.  Before creating a peer pod VM, `cloud-api-adaptor` sends the generated server certificate/key as well as client certificate to a newly created pod VM via cloud-init data
4. `agent-protocol-adaptor` starts TLS listener using the server cert/key
5. `cloud-api-adaptor` initiates TLS connection to agent-protocol-forwarder using the client cert/key
6. `agent-protocol-adaptor` validates incoming TLS connection using the client certificate
7. `cloud-api-adaptor` validates the server certificate sent from agent-protocol-forwarder using the server CA certificate

Note that we can still use manual TLS configuration for use cases that the above assumptions are not met.

bpradipt:
@yoheiueda few questions
[...]

PR #720, opened 2023-03-17
TLS: Automatic mTLS configuration

yoheiueda:
This PR introduces automatic mTLS configuration option. This is a quick convenient way to set up mTLS without manually preparing server/client TLS certificates.

The motivation is described in #705.

The mechanism is described in the documentation, and I copy and paste it here.

> `cloud-api-adaptor` running in a worker node communicates with `agent-protocol-forwarder` running in a peer pod VM using a mutual TLS connection. This mTLS connection is used to transfer TTRPC requests and responses of [agent protocol](https://github.com/kata-containers/kata-containers/blob/CCv0/src/libs/protocols/protos/agent.proto) between `kata-shim` and `kata-agent`.
> 
> There are two options to configure mTLS connections.
> 
> * Automatic configuration - `cloud-api-adaptor` automatically generates and and necessary keys and certificates
> * Manual configuration - Users need to prepare necessary keys and certificates, and correctly deploy them.
> 
> Automatic configuration is a quick convenient option, but has some consideration points. If you want to fully control deployment of certificates and private keys, you can adopt the manual configuration option. You can also completely disable TLS encryption for testing purpose.
> 
> ## Automatic configuration
> 
> Automatic TLS configuration is enabled by default. When command line options for TLS manual configuration are NOT specified, TLS certificates and keys are automatically generated and deployed appropriately.
> 
> Mutual TLS configuration consists of server and client authentications.
> 
> When the `-ca-cert-file` option of `cloud-api-adaptor` is NOT specified, a Certificate Authority (CA) service is enabled. The CA service issues a server certificate when `cloud-api-adaptor` creates a new peer pod VM. The generated certificate and private key are passed to the new pod VM as cloud-init data in a CreateInstance API call of cloud provider.
> 
> When the `-cert-file` and `-cert-key` options of `cloud-api-adaptor` are NOT specified, `cloud-api-adaptor` generates a self-signed client certificate and its private key, and passes the client certificate to the new peer pod VM as cloud-init data in a CreateInstance API call of cloud provider.
> 
> ### Security consideration points
> 
> Note that a server private key is passed to a peer pod VM as cloud-init data in an API call of cloud provider. This seems that there is a security risk here, but the security risk is considered small in practice. TLS session keys reside in memory of a worker node. This means that cloud administrators can access the session keys and possibly decrypt TLS traffics, unless the worker node is in a secure enclave. While cloud administrators can access TLS session keys, passing private keys via cloud API does not significantly increase security risks. One possible attack scenario is that a malicious cloud administrator injects a malformed private key, and the golang standard crypto library has a vulnerability when parsing such malformed key.
> 
> In fact, this automatic TLS configuration increases attack surfaces. If you need automation of TLS certificate management while you want to minimize attack surface,  another possible option is to construct your own system based on the Kubernetes [cert-manager](https://cert-manager.io/) with the manual TLS configuration.


pkg/util/tlsutil/certgen.go
```diff
@@ -0,0 +1,236 @@
...
+
+// Certificate generation is based on https://github.com/golang/go/blob/master/src/crypto/tls/generate_cert.go
+
+// Automatic mutual TLS configuration works as follows
+//
+// 1. At start up, cloud-api-adaptor generates the following two sets of certificate/key pairs
+//    * Self-signed client certificate and its private key
+//    * Self-signed server CA certificate and its private key
+// 2. Before creating a peer pod VM, cloud-api-adaptor generates a pair of server certificate and private key using the server CA certificate.
+// 3. When creating a peer pod VM, cloud-api-adaptor sends the generated server certificate/key as well as client certificate to a newly created pod VM via cloud-init data
```
bpradipt, line 28: Shouldn't the pod VM use server certificate/key and ca certificate ?

pkg/forwarder/forwarder.go
```diff
@@ -39,6 +39,10 @@ type Config struct {
 	PodNetwork   *tunneler.Config `json:"pod-network"`
 	PodNamespace string           `json:"pod-namespace"`
 	PodName      string           `json:"pod-name"`
+
+	TLSServerKey  string `json:"tls-server-key,omitempty"`
+	TLSServerCert string `json:"tls-server-cert,omitempty"`
+	TLSClientCA   string `json:"tls-client-ca,omitempty"`
```
bpradipt, line 44: nit: gofmt ?

pkg/adaptor/proxy/proxy.go
```diff
@@ -97,7 +103,13 @@ func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error)
...
 		// we are using a well-defined hostname here. Other option is to create
 		// certificates with IP SAN having all the IPs in the network range
-		config.ServerName = podvmServername
+		// When CA service is enabled, a server certificate is automatically generated for
+		// the instance VM name.
+		if p.caService != nil {
+			config.ServerName = p.serverName
+		} else {
+			config.ServerName = podvmServername
+		}
```
bpradipt, line 112: With automatic mTLS option available, do we really need to keep non-TLS option ?

pkg/util/tlsutil/certgen.go
yoheiueda: For server authentication, since there are multiple servers (peer pod VMs) per worker node, cloud-api-adaptor generates a single server CA certificate, and then issues a server certificate from it for each peer pod VM.

For client authentication, since there is only a single client (cloud-api-adaptor process) per worker node, I don't think cloud-api-adaptor needs to generate a client CA certificate. Instead, cloud-api-adaptor generates a self-signed client certificate, and agent-protocol-forwarder uses the client certificate as a root certificate for client authentication.

pkg/forwarder/forwarder.go
yoheiueda: Could you  explain in a little more detail? `gofmt` doesn't show anything in my environment.

```
gofmt -s -d pkg/forwarder/forwarder.go
```

pkg/adaptor/proxy/proxy.go
yoheiueda: A possible use case is debugging the proxy communication to by analyzing packet dumps captured tcpdump. I agree that non-TLS option is not necessary in most cases.

huoqifeng:
@yoheiueda thanks for doing this, I'm still reading the code but have a more higher level question, is mTLS specific for PeerPod? Is it also a problem for normal CoCo pod and kata-shim? Is this a potential feaature we can introduce to normal CoCo pod?

pkg/adaptor/cloud/cloud.go
```diff
@@ -204,28 +204,48 @@ func (s *cloudService) CreateVM(ctx context.Context, req *pb.CreateVMRequest) (r
...
+	agentProxy := s.proxyFactory.New(serverName, socketPath)
+
 	daemonConfig := forwarder.Config{
 		PodNamespace: namespace,
 		PodName:      pod,
 		PodNetwork:   podNetworkConfig,
+		TLSClientCA:  string(agentProxy.ClientCA()),
+	}
+
+	if caService := agentProxy.CAService(); caService != nil {
```
huoqifeng, line 232: what does else mean?

pkg/adaptor/proxy/factory.go
```diff
@@ -3,27 +3,52 @@
...
+	pauseImage    string
+	criSocketPath string
+	tlsConfig     *tlsutil.TLSConfig
+	caService     tlsutil.CAService
 }
 
-func NewFactory(pauseImage, criSocketPath string, tlsConfig *tls.TLSConfig) Factory {
+func NewFactory(pauseImage, criSocketPath string, tlsConfig *tlsutil.TLSConfig) Factory {
+
+	if tlsConfig != nil && !tlsConfig.HasCertAuth() {
```
huoqifeng, line 21: what does else mean?

pkg/adaptor/proxy/proxy.go
```diff
@@ -256,3 +267,20 @@ func (p *agentProxy) Shutdown() error {
...
+	return p.caService
+}
+
+func (p *agentProxy) ClientCA() (certPEM []byte) {
+
+	if p.tlsConfig == nil {
+		return nil
+	}
+	if p.tlsConfig.CAFile != "" {
+		// When a client CA file is explicitly specified, we don't need to put it in cloud-init data
```
huoqifeng, line 281: How does the CAFile contents get used then?

pkg/util/tlsutil/certgen.go
bpradipt: Hmm, I was thinking the flow like this. 
1. Create a self-signed root CA (ca.crt)
2. Create client (client.crt, client.key) and server (server.crt, server.key) certificates, signed by root CA
3. CAA sends the server.crt, server.key and ca.crt to the pod VM via cloud-init. 
Instead of  CAA sending the server.crt, server.key and client.crt which is what I understood from your earlier comments.

pkg/adaptor/proxy/proxy.go
bpradipt: Right, for debugging we might need plaintext. Thanks.

pkg/adaptor/proxy/proxy.go
huoqifeng: I seems don't see the file contents get used?

huoqifeng: Also where were the `CertFile` and `KeyFile` contents used?

huoqifeng: oh, I see, this is additional change for `Automatic mTLS configuration`.

pkg/adaptor/proxy/factory.go
yoheiueda: `tlsConfig == nil` means that TLS encryption is disabled.
`tlsConfig.HasCertAuth()` means that CA certificate is manually specified for server authentication.

So, the `else` clause here means either of the above cases.

pkg/util/tlsutil/certgen.go
yoheiueda: The flow In my current implementation is like this.
1. Create a self-signed root CA (ca.crt, ca.key) **only for server certificates to be issued**
2. Create a self-signed **client** certificate (client.crt, client.key)
3. Issue a server certificate (server.crt, server.key) signed by the root server  CA (ca.key)
4. CAA sends server.crt, server.key (as server-side key) and client.crt (as client root certificate) to the pod VM via cloud-init

In general, we can use difference trusted root certificate pools for server and client authentications in mTLS, although it is completely valid to use the single same root certificate for both client and server authentications.

In my current implementation, I don't use the same certificate as client and server root certificates, just because we don't have to do so. I just use separated trust chains in client and server authentications. Of course, we can use the same root certificate for both server and client authentications, but I don't come up with any strong reason to do so.

If we use the same CA to issue a client certificate, both server and client certificates are validated by the same CA certificate. It means that a peer pod VM can use the server certificate to connect to another peer pod VM by mimicking a client certificate. This is very factitious scenario, and I am not sure what does this actually mean, but separation of trust chain in client and server authentications will prevent this kind of misuse.

@bpradipt do you have any idea regarding the use of the same CA for both client and server authentications?

yoheiueda:
> @yoheiueda thanks for doing this, I'm still reading the code but have a more higher level question, is mTLS specific for PeerPod? Is it also a problem for normal CoCo pod and kata-shim? Is this a potential feaature we can introduce to normal CoCo pod?

This TLS configuration is for the communication between kata-shim and kata-agent. In the peer pod VM case, this communication is possibly exposed to the public network. The intention is to protect agent protocol communication from third parties' random attacks. Please note that the TLS encryption of agent protocol communication does not prevent attacks by malicious cloud administrators.

If the communication between kata-shim and kata-agent does not go out of a host machine, I don't think encryption of the agent protocol communication with TLS has some benefits.

pkg/adaptor/proxy/factory.go
huoqifeng: Thanks Yohei, I just realized `manually specified` is existing code.

pkg/util/tlsutil/certgen.go
bpradipt: In all the self-signed certificate scenarios that I have seen, I saw single root CA being used to sign all the certificates. Kind of mimicking trusted root CA used within an organisation. Hence the flow of not using the root `ca.crt` in the cert chain felt like an anti-pattern to me. But honestly I'm no expert in this, and what you mentioned above also makes sense. So I'll leave it to your discretion :-)

pkg/forwarder/forwarder.go
bpradipt: My bad. Please ignore

huoqifeng (approved):
/lgtm, thank you! @yoheiueda

docs/tls-proxy-forwarder.md
```diff
@@ -1,6 +1,33 @@
-## Enable TLS communication between proxy and agent-protocol-forwarder
+# TLS communication between `cloud-api-adaptor` and `agent-protocol-forwarder`
 
-To enable TLS, you'll need the following:
+`cloud-api-adaptor` running in a worker node communicates with `agent-protocol-forwarder` running in a peer pod VM using a mutual TLS connection. This mTLS connection is used to transfer TTRPC requests and responses of [agent protocol](https://github.com/kata-containers/kata-containers/blob/CCv0/src/libs/protocols/protos/agent.proto) between `kata-shim` and `kata-agent`.
+
+There are two options to configure mTLS connections.
+
+* Automatic configuration - `cloud-api-adaptor` automatically generates and and necessary keys and certificates
```
bpradipt, line 7: nit: generates and and => generates

yoheiueda: fixed. thank you!

docs/tls-proxy-forwarder.md
```diff
@@ -1,6 +1,33 @@
-## Enable TLS communication between proxy and agent-protocol-forwarder
+# TLS communication between `cloud-api-adaptor` and `agent-protocol-forwarder`
 
-To enable TLS, you'll need the following:
+`cloud-api-adaptor` running in a worker node communicates with `agent-protocol-forwarder` running in a peer pod VM using a mutual TLS connection. This mTLS connection is used to transfer TTRPC requests and responses of [agent protocol](https://github.com/kata-containers/kata-containers/blob/CCv0/src/libs/protocols/protos/agent.proto) between `kata-shim` and `kata-agent`.
+
+There are two options to configure mTLS connections.
+
+* Automatic configuration - `cloud-api-adaptor` automatically generates and necessary keys and certificates
```
bpradipt, line 7:
```suggestion
Automatic configuration - `cloud-api-adaptor` automatically generates necessary keys and certificates
```

bpradipt (approved):
/lgtm
Just a minor nit which I added as a suggestion

PR #725, opened 2023-03-20
aws: podvm creation based on RHEL AMI

snir911:
using packer's AWS plugin

Fixes: #724


aws/image/Makefile
```diff
@@ -4,20 +4,22 @@
...
-		-var account_id=${AWS_ACCOUNT_ID} \
-		-var region=${AWS_REGION} \
+	packer build -var region=${AWS_REGION} \
 		-var instance_type=${INSTANCE_TYPE} \
-		-var ami_name=${IMAGE_NAME} .
+		-var ami_name=${IMAGE_NAME} ./${PODVM_DISTRO}/
 	rm -fr toupload
 
+default-vpc:
+	aws ec2 create-default-vpc --region ${AWS_REGION} || true
```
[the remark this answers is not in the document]
snir911: if it's already there it prints the error and continues.
as this vpc is used only for creating a temporary VM for installing all bits and that's it i thought it wouldn't be required for majority of the use cases, do you think it's needed?

aws/image/rhel/aws-rhel.pkr.hcl
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+source "amazon-ebs" "rhel" {
+  ami_name      = "${var.ami_name}"
+  instance_type = "${var.instance_type}"
+  region        = "${var.region}"
+  source_ami_filter {
+    filters = {
+      name                = "RHEL-8.7.0_HVM-*"
```
whitebrandy, line 16: May the filter could also be a user variable so that RHEL9 could also use this script ?

aws/image/rhel/aws-rhel.pkr.hcl
```diff
@@ -0,0 +1,87 @@
...
+  provisioner "file" {
+    source      = "misc-settings.sh"
+    destination = "~/misc-settings.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    inline = [
+      "sudo -E bash ~/misc-settings.sh"
+    ]
```
whitebrandy, line 85: Maybe we should run selinux_relabel on misc_settings before executing. Most likely, it's fine either way since root will have an unconfined label by default.

aws/image/copy-files.sh
```diff
@@ -6,12 +9,18 @@ if [ -e /tmp/files/etc/aa-offline_fs_kbc-resources.json ]; then
...
+        sudo mkdir -p /etc/certificates
+	sudo cp -a /tmp/files/etc/certificates/tls.crt /etc/certificates/
+	sudo cp -a /tmp/files/etc/certificates/tls.key /etc/certificates/
+fi
+
+# If self-signed certificates, then ca certificate needs to be provided
+if [ -e /tmp/files/etc/certificates/ca.crt ]; then
+	sudo cp -a /tmp/files/etc/certificates/ca.crt /etc/certificates/
+fi
+
```
whitebrandy, line 22: Just confirming since this part is new to me. What happens if there are no certificates at all ? Everything will continue to work but unencrypted ?

aws/image/Makefile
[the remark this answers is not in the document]
bpradipt: Ability to use an existing VPC will be needed. Otherwise it'll be severely limited.

aws/image/rhel/aws-rhel.pkr.hcl
snir911: I think once will RHEL 9 will work let's just switch to it, the filter naming has very unique pattern sometimes (or we can have the two options)

aws/image/rhel/aws-rhel.pkr.hcl
snir911: I didn't get you, 
before executing of what?

aws/image/copy-files.sh
snir911: @bpradipt to the call (i just copied it :P )

aws/image/Makefile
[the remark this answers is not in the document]
snir911: I'll modify it to also allow mentioning VPC

aws/image/rhel/aws-rhel.pkr.hcl
whitebrandy: before executing the script. As I mentioned, most likely you won't have a problem because you are executing this from the root directory and the default label there has higher privileges.

wainersm:
Hi @snir911,

The e2e tests for AWS will have code to create the VPC, push the qcow2, create the AMI, create the k8s cluster, and so on, leveraging the AWS SDK for Go. That seems to have many overlaps with the Terraform scripts you are going to create/improve on this PR. So I was wondering how we could conciliate the two efforts and end up with a single code base.

IIUC  IBM folks have invested on the provision library (https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/test/provisioner) of the e2e framework and implemented a command-line tool  (https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/cmd/cluster-provisioner) to be used stand-along (i.e. outside of the e2e tests). Perhaps we should follow that same path? @huoqifeng , did I understand correctly your plans?

@snir911 , ah, I've already a code to push the qcow2 EBS and creates the AMI using the AWS SDK for Go. It probably is broken because we changed the base e2e framework a lot recently but I will open an PR as WIP so that you can have a look.

aws/image/copy-files.sh
bpradipt: Yeah, everything will continue to work as-is

bpradipt:
@wainersm this PR is only about creating AWS pod VM AMI using publicly available AMIs. 
I don't see any overlap other than creating VPC if it doesn't exist. 
IMHO vpc creation can be removed and this PR should only take VPC as an input. How the VPC gets created should be left to any preferred tool of the user. 

Same applies to K8s cluster creation. There are numerous ways to create a K8s cluster and we should just have an opinionated way for cluster creation in CI and leave it there.

aws/image/rhel/aws-rhel.pkr.hcl
snir911: IIUC selinux runs before as you meant

aws/image/Makefile
```diff
@@ -4,20 +4,37 @@
...
+$(IMAGE_FILE): setopts $(BINARIES) $(FILES)
 	mkdir -p toupload
-	packer build -var vpc_id=${VPC_ID} \
-		-var subnet_id=${SUBNET_ID} \
-		-var account_id=${AWS_ACCOUNT_ID} \
+	packer build ${OPTS} \
 		-var region=${AWS_REGION} \
 		-var instance_type=${INSTANCE_TYPE} \
-		-var ami_name=${IMAGE_NAME} .
+		-var ami_name=${IMAGE_NAME} ./${PODVM_DISTRO}/
```
bpradipt, line 18: How PODVM_DISTRO is being used ?

wainersm:
> @wainersm this PR is only about creating AWS pod VM AMI using publicly available AMIs. I don't see any overlap other than creating VPC if it doesn't exist. IMHO vpc creation can be removed and this PR should only take VPC as an input. How the VPC gets created should be left to any preferred tool of the user.
> 

@snir911 explained me the goal. So never mind :)

> Same applies to K8s cluster creation. There are numerous ways to create a K8s cluster and we should just have an opinionated way for cluster creation in CI and leave it there.

aws/image/Makefile
snir911: just to point to the disto file

bpradipt: I see. Thanks

bpradipt: Do you plan to cover CentOS in a separate PR ?

snir911: I haven't planned to add CentOS as part of this PR, this can be done in a following PR

aws/README.md
```diff
@@ -10,22 +10,19 @@
 
 - Set environment variables
 ```
-export AWS_ACCOUNT_ID="REPLACE_ME"
-export AWS_REGION="REPLACE_ME"
+export AWS_REGION="us-east-1" # mandatory
+export PODVM_DISTRO=rhel # mandatory
+export INSTANCE_TYPE=t3.small # optional, default is t3.small
+export IMAGE_NAME=peer-pod-ami-image # optional
+export VPC_ID=vpc-01234567890abcdef # optional, otherwise, it sets and use the default vpc
```
bpradipt, line 17:
```suggestion
export VPC_ID=vpc-01234567890abcdef # optional, otherwise, it creates and uses the default vpc
```

bpradipt (approved):
/lgtm
Thanks @snir911

whitebrandy (approved):
lgtm, thanks Snir

PR #728, opened 2023-03-21
test: use fatal for program exit errors in e2e test

huoqifeng:
Fixes: #726

### DIFF
--- a/cmd/cluster-provisioner/main.go
+++ b/cmd/cluster-provisioner/main.go
@@ -6,13 +6,30 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
 	"os"
 
 	pv "github.com/confidential-containers/cloud-api-adaptor/test/provisioner"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


stevenhorsman (approved):
LGTM. Thanks Qifeng & Yohei for the suggestion

wainersm (approved):
With the @yoheiueda 's observation to remove the `os.exit(1)` calls.

Hey @huoqifeng  thanks! I saw you guys using the logger on the code for ibmcloud, I thought it would be a good improvement to the framework too but I forgot to say :(

huoqifeng:
> `log.Fatal` internally calls `os.Exit(1)`, so I don't think you need to call `os.Exit(1)` by yourself here.
> 
> https://github.com/sirupsen/logrus/blob/v1.9.0/logger.go#L189-L191

Right, it will call os.Exit eventually https://github.com/sirupsen/logrus/blob/f8bf7650dccb756cea26edaf9217aab85500fe07/logger.go#L342, thanks @yoheiueda

yoheiueda (approved):
LGTM

PR #738, opened 2023-03-23
cmd: Notify systemd about service state changes

### LINKED ISSUE #735: cmd: Notify systemd when process becomes ready
Currently, no `Type` option is specified in `agent-protocol-forwarder.service`, so the start-up type of the `agent-protocol-forwarder` systemd service is set to the default value `simple`. This means that systemd will consider the `agent-protocol-forwader` service started immediately after the process has been forked.

This is not a convenient setting when we write another systemd service that depends on `agent-protocol-forwader`. Systemd will start the  service immediately after `agent-protocol-forwader` is forked, even though it is not ready to serving requests.

The start-up type `notify` lets systemd to wait until the `agent-protocol-forwader` service becomes ready.
https://www.freedesktop.org/software/systemd/man/sd_notify.html#Description

https://github.com/confidential-containers/cloud-api-adaptor/blob/a89512575ba52df9ac21ff4fb58639ae73f5c80f/podvm/files/etc/systemd/system/agent-protocol-forwarder.service#L8-L12

yoheiueda:
This patch enables notification to systemd using the sd_notify protocol.  When the program is running as a systemd service with type=notify, and becomes ready for serving requests, systemd gets notified.

Fixes #735

### DIFF
--- a/cmd/starter.go
+++ b/cmd/starter.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 
+	"github.com/coreos/go-systemd/daemon"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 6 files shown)


bpradipt (approved):
/lgtm
Thanks @yoheiueda 
I tested this successfully on libvirt with ubuntu podvm as well.

huoqifeng (approved):
/lgtm, thank you! @yoheiueda

Issue #747, opened 2023-03-27
go: Enabling go.work causes error due to invalid go.mod files


yoheiueda:
This is a sub task of #475.

When I enabled Go workspace with the procedure showen below, I noticed that some go commands and make for `peer-pod-controller` fails due to invalid `go.mod` file in the the `podvm` directory.

```
$ go list ./cmd
github.com/confidential-containers/cloud-api-adaptor/cmd
$ go work init
$ go work use -r .
$ cat go.work
go 1.19

use (
	.
	./azure
	./peer-pod-controller
	./podvm
	./volumes/csi-wrapper
	./webhook
)
$ go list ./cmd
go: no module declaration in go.mod. To specify the module path:
	go mod edit -module=example.com/mod
$ cd peer-pod-controller
$ make
︙
bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
Error: err: exit status 1: stderr: go: no module declaration in go.mod. To specify the module path:
	go mod edit -module=example.com/mod
```

PR #538, opened 2023-01-18
azure: use common makefile for generating images

knrt10:
fixes: #535
Signed-off-by: Kautilya Tripathi <ktripathi@microsoft.com>

PR #500, opened 2023-01-06
build: exclude systemd/system dirs from CAA go module

mattarnoatibm:
When cloud-api-adaptor is required by another Go project as a Go module, ignore the systemd/system directories as they contain a file run-kata\x2dcontainers.mount that breaks Go module imports.

Fixes #483

Signed-off-by: Matthew Arnold <mattarno@uk.ibm.com>

PR #748, opened 2023-03-27
go: Use valid go.mod format in the podvm directory

yoheiueda:
The `go.mod` file in the `podvm` directory is empty to void the issue addressed by #500. But, empty `go.mod` file causes errors when we run some go commands as described in #747.

This PR keeps the go.mod file, but changes the file to use a valid go.mod file format. I confirmed that referencing `cloud-api-adaptor` repo as a go module from another go program still works after this change is introduced.

This PR also removes `azure/go.sum`, since this file is no longer necessary, as #538 has removed the files that caused the problem.


mattarnoatibm (approved):
LGTM

bpradipt (approved):
/lgtm

PR #750, opened 2023-03-27
cloud: add new provider - IBM Cloud PowerVS

Amulyam24:
This PR adds the initial implementation of running peerpods on PowerVS. 

More info on Power Virtual Server(PowerVS) can be found [here](https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-about-virtual-server). 

Fixes: #749

cc @mkumatag


bpradipt:
@Amulyam24 thanks for the PR. May I please request to create a separate PR to parameterise the number of proxy connection retries ?

cc @yoheiueda

pkg/adaptor/cloud/ibmcloud/powervs_provider.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		if *in.Status == "ERROR" {
+			return nil, fmt.Errorf("instance is in error state")
+		}
+
+		if *in.Status == "ACTIVE" {
+			pvslogger.Printf("instance is in desired state: %s", *in.Status)
+			break
+		}
```
yoheiueda, line 102: In the case of IBM Cloud VPC, `CreateInstance` does not wait until the newly created VM insstance becomes ACTIVE state. Instead, `CreateInstance` returns when the API call for `CreateInstance` returns and IP address is assigned.

After `CreateInstance` is called, `cloud-api-adaptor` waits until it can establish a network connection with `agent-protocol-forwarder` running in a new peer pod VM instance.

In the case of PowerVS, are IP addresses assigned to a VM instance before it becomes active? If so, you don't need to wait here.

https://github.com/confidential-containers/cloud-api-adaptor/blob/0935bab6f61988ede05655b9638c3e4371fef847/pkg/adaptor/cloud/ibmcloud/provider.go#L145-L197

https://github.com/confidential-containers/cloud-api-adaptor/blob/0935bab6f61988ede05655b9638c3e4371fef847/pkg/adaptor/cloud/cloud.go#L312-L350

https://github.com/confidential-containers/cloud-api-adaptor/blob/0935bab6f61988ede05655b9638c3e4371fef847/pkg/adaptor/proxy/proxy.go#L214-L216

stevenhorsman:
> Thank you very much for contributing this great enhancement!
> 
> I think separating packages with IBM Cloud VPC is better than using a single package with it.
> 
> I think we should have chosen `ibmcloudvpc` or something for the provider of IBM Cloud VPC. The cloud provider name and build tag name `ibmcloud` is already referenced in various locations, so changing the name `ibmcloud` for IBM Cloud VPC for something else from now may causes compatibility issues.
> 
> @stevenhorsman @huoqifeng do you have any comments?

Yes - I think that we probably need to leave the current VPC implementation as is with `ibmcloud`, and then add `ibmcloud_powervs` or similar for the new cloud provider.

pkg/adaptor/cloud/ibmcloud/powervs_provider.go
Amulyam24: > In the case of PowerVS, are IP addresses assigned to a VM instance before it becomes active? If so, you don't need to wait here.

@yoheiueda, in PowerVS, the IP is assigned to the VM once it is in `ACTIVE` state after the `BUILD` stage(during which it is `nil`) unlike VPC where `0.0.0.0` is assigned at first. Hence the check is needed for fetching the IP and then proceeding with establishing a connection with `agent-protocol-forwarder`.

Amulyam24: > I think we should have chosen ibmcloudvpc or something for the provider of IBM Cloud VPC. The cloud provider name and build tag name ibmcloud is already referenced in various locations, so changing the name ibmcloud for IBM Cloud VPC for something else from now may causes compatibility issues.

Addressed it, PTAL!

cmd/cloud-api-adaptor/cloudmgr/ibmcloud_vpc.go
yoheiueda, line 1: Could you keep the original file name for the original IBM VPC provider? I think this filename, package name, and build tag need to be consistent

yoheiueda (changes requested):

pkg/adaptor/cloud/ibmcloud/powervs_provider.go
yoheiueda: > @yoheiueda, in PowerVS, the IP is assigned to the VM once it is in `ACTIVE` state after the `BUILD` stage(during which IP is `nil`) unlike VPC where `0.0.0.0` is assigned at first. Hence the check is needed for fetching the IP and then proceeding with establishing a connection with `agent-protocol-forwarder`.

I see. So, the code looks OK as it is.

cmd/cloud-api-adaptor/cloudmgr/ibmcloud_vpc.go
Amulyam24: Addressed it. PTAL

yoheiueda (approved):
LGTM

Makefile
```diff
@@ -22,7 +22,7 @@ TEST_E2E_TIMEOUT ?= 20m
 ifeq ($(RELEASE_BUILD),true)
 	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere
 else
-	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere libvirt
+	BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud ibmcloud_powervs vsphere libvirt
```
yoheiueda, line 25: One additional comment.

Currently, there is no documentation about how to enable PowerVS in cloud-api-adaptor yet. How about adding `ibmcloud_powervs` in builtin provider list, when the documentation is ready?

bpradipt: I second that. Let's add it to builtin list after documentation and may be few test CI runs

bpradipt: Also required will be the basic podvm image creation steps for ppc64le.

Amulyam24: Sure, makes sense. I will submit a follow-up PR regarding the documentation and podvm image for PowerVS.

cmd/cloud-api-adaptor/main.go
```diff
@@ -37,7 +37,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 
 	if len(os.Args) < 2 {
 		fmt.Printf("%s <provider_name> [options] | version\n", os.Args[0])
-		fmt.Printf("   The <provider_name> should be one of aws|azure|ibmcloud|libvirt|vsphere>\n")
+		fmt.Printf("   The <provider_name> should be one of aws|azure|ibmcloud|ibmcloud-powervs|libvirt|vsphere>\n")
```
yoheiueda, line 40: Could you rebase this PR?

As https://github.com/confidential-containers/cloud-api-adaptor/pull/788 has been merged, we no longer need to hard-code cloud provider names in `main.go`.

Amulyam24: Addressed, PTAL.

Amulyam24:
Hi, can this PR be merged?

huoqifeng:
> Hi, can this PR be merged?

It could be merged after at least 2 reviewers agreed with the change, In this case I think @stevenhorsman has not admitted the further change yet.  @Amulyam24

stevenhorsman:
> > Hi, can this PR be merged?
> 
> It could be merged after at least 2 reviewers agreed with the change, In this case I think @stevenhorsman has left comments and not admitted the further change yet. @Amulyam24

Sorry, With the 0.5 release I've not had a chance to look at this yet, but I'll try and get to it in the next couple of days

stevenhorsman:
@Amulyam24 - I've not been able to test the power code, but looking through it, it seems ok. Can you rebase your branch on the latest staging and then I can try running the ibmcloud (VPC) e2e tests. Thanks

pkg/adaptor/cloud/ibmcloud-powervs/manager.go
```diff
@@ -0,0 +1,37 @@
...
+func (_ *Manager) ParseCmd(flags *flag.FlagSet) {
+
+	flags.StringVar(&ibmcloudPowerVSConfig.ApiKey, "api-key", "", "IBM Cloud API key, defaults to `IBMCLOUD_API_KEY`")
+	flags.StringVar(&ibmcloudPowerVSConfig.Zone, "zone", "", "PowerVS zone name")
+	flags.StringVar(&ibmcloudPowerVSConfig.ServiceInstanceID, "service-instance-id", "", "ID of the PowerVS Service Instance")
+	flags.StringVar(&ibmcloudPowerVSConfig.NetworkID, "network-id", "", "ID of the network instance")
+	flags.StringVar(&ibmcloudPowerVSConfig.ImageID, "image-id", "", "ID of the boot image")
+	flags.StringVar(&ibmcloudPowerVSConfig.SSHKey, "ssh-key", "", "Name of the SSH Key")
+	flags.Float64Var(&ibmcloudPowerVSConfig.Memory, "memory", 2, "Amount of memory in GB")
+	flags.Float64Var(&ibmcloudPowerVSConfig.Processors, "cpu", 0.5, "Number of processors allocated")
```
bpradipt, line 25: Is there any instance size or flavour concept yet in powervs?

Amulyam24:
> @Amulyam24 - I've not been able to test the power code, but looking through it, it seems ok. Can you rebase your branch on the latest staging and then I can try running the ibmcloud (VPC) e2e tests. Thanks

Sure @stevenhorsman, Rebased it.

pkg/adaptor/cloud/ibmcloud-powervs/manager.go
Amulyam24: @bpradipt, not yet.

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	if len(*pvsInstances) <= 0 {
+		return nil, fmt.Errorf("there are no instances created")
+	}
+
+	ins := (*pvsInstances)[0]
+	instanceID := *ins.PvmInstanceID
+
+	num := 1
```
bpradipt, line 89: Instead of the for loop can you use retry.Do here with a timeout?

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+
+	if len(ips) > 0 {
+		return ips, nil
+	}
+
+	// If IP is not assigned to the instance, fetch it from DHCP server
+	var ip *string
+	for i := 0; i < maxRetries; i++ {
```
bpradipt, line 166: Same comment here. Can you use retry.Do with a timeout?

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,256 @@
...
+	)
+
+	if err != nil {
+		err = fmt.Errorf("failed to reach the desired instance state: %s", err)
+		logger.Print(err)
+		return nil, err
+
+	}
+
+	ips, err := p.getVMIPs(ins)
```
yoheiueda, line 115: Could you pass `ctx` to `getVMIPs`?

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if len(ips) > 0 {
+		return ips, nil
+	}
+
+	// If IP is not assigned to the instance, fetch it from DHCP server
+	var ip *string
+
+	ctx, cancel := context.WithTimeout(context.Background(), 750*time.Second)
```
yoheiueda, line 169: Then, please create a new context using `ctx` from the caller.

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,256 @@
...
+
+	logger.Printf("deleted an instance %s", instanceID)
+	return nil
+}
+
+func (p *ibmcloudPowerVSProvider) Teardown() error {
+	return nil
+}
+
+func (p *ibmcloudPowerVSProvider) getVMIPs(instance *models.PVMInstance) ([]net.IP, error) {
```
yoheiueda, line 143: Add `ctx` parameter here.

pkg/adaptor/cloud/ibmcloud-powervs/powervs.go
```diff
@@ -0,0 +1,71 @@
...
+	session, err := ibmpisession.NewIBMPISession(options)
+	if err != nil {
+		return nil, err
+	}
+
+	return &powervsService{
+		instanceClient: instance.NewIBMPIInstanceClient(context.Background(), session, serviceinstanceID),
+		networkClient:  instance.NewIBMPINetworkClient(context.Background(), session, serviceinstanceID),
+		dhcpClient:     instance.NewIBMPIDhcpClient(context.Background(), session, serviceinstanceID),
+	}, nil
```
yoheiueda, line 47: Passing new contexts in this new function looks strange to me.  A context should be created per request. Right?

Probably, we should only keep `session` and `serviceinstanceID` in `powervsService`. Then, prepare functions like below.
```
func (s *powervsService) newInstanceClient(ctx context.Context)  *instance.IBMPIInstanceClient {
    return instance.NewIBMPIInstanceClient(ctx, s.session, s.serviceinstanceID)
}

func (s *powervsService) newNetworkClient(ctx context.Context)  *instance.IBMPINetworkClient {
    return instance.NewIBMPINetworkClient(ctx, s.session, s.serviceinstanceID)
}

func (s *powervsService) newIDhcpClient(ctx context.Context)  *instance.IBMPIDhcpClient {
    return instance.NewIBMPIDhcpClient(ctx, s.session, s.serviceinstanceID)
}
```

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
Amulyam24: Hi @bpradipt, I tried using `retry.Do` as below
```
        ctx, cancel := context.WithTimeout(ctx, 150*time.Second)
	defer cancel()

	logger.Printf("Waiting for instance to reach state: ACTIVE")
	err = retry.Do(
		func() error {
			in, err := p.powervsService.instanceClient.Get(*ins.PvmInstanceID)
			if err != nil {
				return fmt.Errorf("failed to get the instance: %v", err)
			}

			if *in.Status == "ERROR" {
				return fmt.Errorf("instance is in error state")
			}

			if *in.Status == "ACTIVE" {
				logger.Printf("instance is in desired state: %s", *in.Status)
				return nil
			}

			return fmt.Errorf("failed to reach desired instance state")
		},
		retry.Context(ctx),
		retry.Attempts(0),
	)
```
I'm seeing a delay of around 60s before the function returns even though the instance enters the `ACTIVE` state. Not sure if there's an internal delay causing it, I've tried a couple of retry options with no luck. Any idea?

yoheiueda: @Amulyam24 did you try `retry.MaxDelay(time.Second)`?

`BackoffDelay` is used by default, so delay time increases exponentially like 0.1s, 0.2s, 0.4s, 0.8s, 1.6s, 3.2s, 6.4s, 12.8s, 25.6s, 51.2s, ... `MaxDelay` sets an upper limit for delay time.

Amulyam24: That worked, Thanks @yoheiueda!

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
Amulyam24: Addressed it, PTAL.

pkg/adaptor/cloud/ibmcloud-powervs/powervs.go
Amulyam24: @yoheiueda, added `ctx` and modified it accordingly, PTAL.

yoheiueda: @Amulyam24 thank you for the udpates.


It looks like your change is not what I meant. `newPowervsClient` is called per provider creation, not per request. Right? So, per-request `ctx` must be used for processing a request. My proposed code uses per-request `ctx`.

pkg/adaptor/cloud/ibmcloud-powervs/powervs.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type powervsService struct {
+	instanceClient *instance.IBMPIInstanceClient
+	networkClient  *instance.IBMPINetworkClient
+	dhcpClient     *instance.IBMPIDhcpClient
+}
+
+func newPowervsClient(apikey, serviceinstanceID, zone string) (*powervsService, error) {
+	ctx := context.Background()
```
yoheiueda, line 22: this `ctx` is not per request.

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -0,0 +1,247 @@
...
+		Memory:     core.Float64Ptr(p.serviceConfig.Memory),
+		Processors: core.Float64Ptr(p.serviceConfig.Processors),
+		ProcType:   core.StringPtr(p.serviceConfig.ProcessorType),
+		SysType:    p.serviceConfig.SystemType,
+		UserData:   base64.StdEncoding.EncodeToString([]byte(userData)),
+	}
+
+	logger.Printf("CreateInstance: name: %q", instanceName)
+
+	pvsInstances, err := p.powervsService.instanceClient.Create(body)
```
yoheiueda, line 72: We can use per-request `ctx` like this
```
	pvsInstances, err := p.powervsService.instanceClient(ctx).Create(body)
```

pkg/adaptor/cloud/ibmcloud-powervs/powervs.go
Amulyam24: @yoheiueda, yes, that's right. Made the changes, PTAL.

bpradipt (approved):
/lgtm
Thanks @Amulyam24

PR #752, opened 2023-03-27
proxy: add new flag for number of proxy retries

Amulyam24:
A new flag proxy-retries has been introduced to customize the number of retries needed for establishing agent proxy connection.

Fixes: #751


yoheiueda (commented):
The logic of the max retry count is my corner-cutting implementation when I initially implemented the proxy code.

I think it is more appropriate to use timeout than max count. Can you use an option to specify a timeout value instead of specifying retry count?

bpradipt:
@yoheiueda should we use https://github.com/avast/retry-go just like in adaptor/proxy/service.go

yoheiueda:
> @yoheiueda should we use https://github.com/avast/retry-go just like in adaptor/proxy/service.go

I agree that it is a good idea to use such library. I think either is OK to implement it with the library in this PR, or create another PR to adopt the library.

Amulyam24:
Thanks for the suggestion @bpradipt and @yoheiueda. It makes sense to create a new PR for adopting the library wherever needed.

PR #756, opened 2023-03-29
ibmcloud: fix cluster bug for x86 architecture based cluster

liudalibj:
- Fix binariers playbook for create a x86 architecture based cluster in ibmcloud


huoqifeng (commented):
I did a quick search, looks like variable mapping is a better solution? like: https://stackoverflow.com/questions/58169348/how-is-the-architecture-fact-called-in-ansible/73096785#73096785

liudalibj:
> I did a quick search, looks like variable mapping is a better solution? like: https://stackoverflow.com/questions/58169348/how-is-the-architecture-fact-called-in-ansible/73096785#73096785

Thanks, let me try to use this better way.

stevenhorsman (approved):
LGTM. Thanks @liudalibj and @huoqifeng for the suggestion!

jtumber-ibm (approved):

Issue #758, opened 2023-03-29
Pod vm image creation with latest Ubuntu on x86_64


bpradipt:
This happens in some scenarios due to ssh auth timing before VM boot

PR #759, opened 2023-03-29
podvm: Increase boot_wait time for all distros

bpradipt:
The default boot_wait of 10s for packer qemu builder is not sufficient in some environments. Set it to a higher value of 5 min to give enough time for the VM to boot for customisations.

Fixes: #758


huoqifeng (approved):
/lgtm, thanks @bpradipt

yoheiueda (approved):
LGTM

Issue #760, opened 2023-03-29
Provide flexibility to override rebuilding of the binaries when creating podvm images


bpradipt:
The podvm image creation process is time consuming (easily running into upwards of 30 min). 
The primary reason is building of all the required binaries before image creation. Sometimes we have a requirement to generate a new podvm image using only the updated configs. Rebuilding the binaries is not really needed. 
It will be useful to have the flexibility to choose when to rebuild the binaries and disable it if reusing pre-built binaries.

PR #761, opened 2023-03-29
podvm: Add the ability to disable BINARIES target

bpradipt:
Running "make BINARIES= image" will not rebuild the BINARIES target and just create the image with what is already available under podvm/files.
This helps with quick podvm image customisations.

Fixes: #760


yoheiueda (approved):
LGTM

huoqifeng (approved):
/lgtm

PR #766, opened 2023-03-29
Patch series to improve podvm image generation process

bpradipt:
This series adds support for creating a separate container image for podvm binaries and uses the same for creating podvm image.


bpradipt:
The gh-action needs to be updated to create all the images. I plan to handle this in a separate PR if it's fine.

podvm/Dockerfile.podvm
```diff
@@ -33,22 +33,23 @@ RUN if [ -n "${CAA_SRC}" ]; then \
...
 ENV UBUNTU_IMAGE_CHECKSUM ${UBUNTU_IMAGE_CHECKSUM}
 
+# Copy the binaries to podvm/files folder
+COPY --from=podvm_binaries /podvm-binaries.tar.gz /src/cloud-api-adaptor/podvm/files
+RUN tar xvf /src/cloud-api-adaptor/podvm/files/podvm-binaries.tar.gz -C /src/cloud-api-adaptor/podvm/files
 
-# Building binaries separately allows for reuse in case of error
-# with image building step OTOH it may increase layers size
-#RUN cd cloud-api-adaptor/podvm && \
-#     make binaries
```
snir911, line 47: I'd prefer to leave it but not a deal but fine with removing as well

snir911 (approved):
Thanks, tested rhel on aws, worked well

podvm/Dockerfile.podvm
bpradipt: Removed it to avoid any confusion. But I can add commented out code for anyone wanting to rebuilding the binaries when creating the image

stevenhorsman (commented):
I've run this locally to create an s390x ibmcloud based image (full disclosure - I've not tried running a peer pod from it though), so it is general good. I have a question in the code and then
I wonder if we should update the doc in https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/README.md as part of this to specify the extra step:
```
docker build -t podvm_binaries_s390x --build-arg ARCH=s390x --build-arg BUILDER_IMG=podvm_builder:latest --build-arg CLOUD_PROVIDER=ibmcloud -f Dockerfile.podvm_binaries .
```
and updated the podvm build to pass in `BINARIES_IMG` as well?

podvm/Dockerfile.podvm
```diff
@@ -5,20 +5,20 @@
 # Builds pod vm image inside container
 #
 ARG BUILDER_IMG="quay.io/confidential-containers/podvm-builder-ubuntu"
+ARG BINARIES_IMG="quay.io/confidential-containers/podvm-binaries-centos"
```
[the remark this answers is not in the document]
bpradipt: Thanks for the catch. The above line is incorrect.
The binaries and  builder should be using the same distro although the builds are static unless it's rhel. But still

bpradipt:
> I've run this locally to create an s390x ibmcloud based image (full disclosure - I've not tried running a peer pod from it though), so it is general good. I have a question in the code and then I wonder if we should update the doc in https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/README.md as part of this to specify the extra step:
> 
> ```
> docker build -t podvm_binaries_s390x --build-arg ARCH=s390x --build-arg BUILDER_IMG=podvm_builder:latest --build-arg CLOUD_PROVIDER=ibmcloud -f Dockerfile.podvm_binaries .
> ```
> 
> and updated the podvm build to pass in `BINARIES_IMG` as well?

Thanks for mentioning this. I updated the README.

podvm/README.md
```diff
@@ -73,48 +74,82 @@ $ docker build -t podvm_builder \
...
-The builder image has to be indicated via `BUILDER_IMG` build argument to docker. Apart from that, the `CLOUD_PROVIDER` argument is mandatory and specifies the cloud provider that binaries should be built for. Once again, one builder image can be used to build for multiple
-cloud providers and architectures.
+```
+$ docker build -t podvm_binaries \
+         --build-arg BUILDER_IMG=podvm_builder \
+	 -f Dockerfile.podvm_binaries .
 
-Below is shown how to build an Ubuntu image for libvirt:
+```
+The build process can take significant time.
```
stevenhorsman, line 93: Sorry to nitpick, but I think this build is sensitive to architecture too? If so it might be worth adding a sentence about the option to setting `--build-arg ARCH`?

podvm/README.md
```diff
@@ -73,48 +74,82 @@ $ docker build -t podvm_builder \
...
 
-The process to build the podvm container image will effectively compile binaries and create the qcow2 file.
+In order to build the binaries you should be using the corresponding dockerfile
+of the Linux distro for which the builder image was built.
+For example, if the builder image was built with *Dockerfile.podvm_builder.DISTRO* then
+you should use the *Dockerfile.podvm_binaries.DISTRO* to build the image with podvm binaries.
 
-In order to build the podvm image you should be using the corresponding dockerfile of the Linux distro for which the builder image was built. For example, if the builder image was built with *Dockerfile.podvm_builder.DISTRO* then you should use the *Dockerfile.podvm.DISTRO* to build the podvm image.
+Assuming you have built the podvm_builder image for Ubuntu as explained in the previous step,
+running the following command will build the image with the podvm binaries.
```
stevenhorsman, line 85: Sorry to nitpick, but I think this build is sensitive to architecture too? If so it might be worth adding a sentence about the option to setting `--build-arg ARCH`?

podvm/Dockerfile.podvm_binaries
```diff
@@ -0,0 +1,44 @@
...
+# Installs add-ons for foreign target, if required
+RUN ./cloud-api-adaptor/podvm/hack/cross-build-extras.sh
+
+RUN cd cloud-api-adaptor/podvm && \
+     make binaries
+
+RUN tar czvf /podvm-binaries.tar.gz -C /src/cloud-api-adaptor/podvm/files usr/
+RUN tar czvf /pause-bundle.tar.gz -C /src/cloud-api-adaptor/podvm/files pause_bundle/
+
+FROM ubuntu:20.04
```
whitebrandy, line 42: I am wondering if the distro specific build stage is necessary ? Can we use the minimal `FROM scratch` instead since we are only copying ?

podvm/README.md
```diff
@@ -73,48 +74,82 @@ $ docker build -t podvm_builder \
...
+	 --build-arg BINARIES_IMG=podvm_binaries \
+	 -f Dockerfile.podvm .
 ```
 
-This step will take several minutes to complete, mostly because `packer` will use the QEMU builder in emulation mode when running within container. If you are using podman then you can speed up QEMU by enabling native virtualization, i.e., passing the `--device=/dev/kvm` argument to enable KVM inside the container.
+This step will take several minutes to complete, mostly because `packer` will
+use the QEMU builder in emulation mode when running within container. If you
+are using podman then you can speed up QEMU by enabling native virtualization,
+i.e., passing the `--device=/dev/kvm` argument to enable KVM inside the
+container.
```
stevenhorsman, line 120: Nit: consider using the special GFM for Tips here?
```suggestion
use the QEMU builder in emulation mode when running within container.
> **Tip:** If you are using podman then you can speed up QEMU by enabling native
> virtualization, by passing the `--device=/dev/kvm` argument to enable KVM inside
> the container.
```
(I hope I got the formatted correct, but have no preview in the PR comments)

podvm/README.md
```diff
@@ -73,48 +74,82 @@ $ docker build -t podvm_builder \
...
 
-Also beware that the process consume a bunch of memory and disk from the host. As a tip, if the build fail at the point QEMU was launched but packer couldn't connect via ssh then it might indicate lack of memory (try to increase the amount of memory if running on VM).
+Also beware that the process consume a bunch of memory and disk from the host.
+As a tip, if the build fail at the point QEMU was launched but packer couldn't
+connect via ssh then it might indicate lack of memory (try to increase the
+amount of memory if running on VM).
```
stevenhorsman, line 125: Nit: formatting and adding example error
```suggestion
> **Note:** Beware that the process consume a bunch of memory and disk from the host.
If the build fails at the point QEMU was launched but packer couldn't
connect via ssh, with an error similar to:
> ```
> Build 'qemu.ubuntu' errored after 5 minutes 57 seconds: Timeout waiting for SSH.
> ```
> then it might indicate lack of memory, so try to increase the amount of memory if running on VM.
```

stevenhorsman (approved):
Thanks very much for adding the updated documentation, it's generally really good, so I feel quite guilty, but I've made a few suggestions. Nothing major though.

podvm/Dockerfile.podvm_binaries
bpradipt: I explicitly added it to make it easier for running commands using the same image if needed. Let's keep it for sometime and we can change it going forward.

bpradipt:
> Thanks very much for adding the updated documentation, it's generally really good, so I feel quite guilty, but I've made a few suggestions. Nothing major though.

Thanks for your suggestions. I have updated the README

stevenhorsman (approved):
Looks good to me. Thanks for the updates!

podvm/Dockerfile.podvm_binaries
whitebrandy: Sounds good. The advantage of scratch is it's a no-op. So, you actually save some cycles by avoiding creating a layer.

Issue #754, opened 2023-03-28
Have a better mechanism to handle retrying agent proxy connection


Amulyam24:
Originally discussed on #752

> @yoheiueda should we use https://github.com/avast/retry-go just like in adaptor/proxy/service.go
  > I agree that it is a good idea to use such library. I think either is OK to implement it with the library in this PR, or create another PR to adopt the library.

PR #771, opened 2023-03-31
pkg: enhance retry logic for establishing connections

Amulyam24:
This PR introduces a new flag `proxy-timeout` which represents the maximum time in minutes for establishing agent proxy connection. It also refactors the retry logic for establshing connection between agent proxy and `agent-protocol-forwarder` on peer pod VM and between `agent-protocol-forwarder` and kata agent.

Fixes: #754

### DIFF
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/confidential-containers/cloud-api-adaptor/cmd"
 	"github.com/confidential-containers/cloud-api-adaptor/cmd/cloud-api-adaptor/cloudmgr"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor"
+	"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/proxy"
 	daemon "github.com/confidential-containers/cloud-api-adaptor/pkg/forwarder"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/podnetwork/tunneler/vxlan"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/util/tlsutil"
@@ -71,6 +72,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&tlsConfig.KeyFile, "cert-key", "", "cert key")
 			flags.BoolVar(&tlsConfig.SkipVerify, "tls-skip-verify", false, "Skip TLS certificate verification - use it only for testing")
 			flags.BoolVar(&disableTLS, "disable-tls", false, "Disable TLS encryption - use it only for testing")
+			flags.DurationVar(&cfg.serverConfig.ProxyTimeout, "proxy-timeout", proxy.DefaultProxyTimeout, "Maximum timeout in minutes for establishing agent proxy connection")
 
 			flags.StringVar(&cfg.networkConfig.TunnelType, "tunnel-type", podnetwork.DefaultTunnelType, "Tunnel provider")
 			flags.StringVar(&cfg.networkConfig.HostInterface, "host-interface", "", "Host Interface")
--- a/pkg/adaptor/proxy/factory.go
+++ b/pkg/adaptor/proxy/factory.go
@@ -3,7 +3,11 @@
 
 package proxy
 
-import "github.com/confidential-containers/cloud-api-adaptor/pkg/util/tlsutil"
+import (
+	"time"
+
+	"github.com/confidential-containers/cloud-api-adaptor/pkg/util/tlsutil"
+)
 
 type Factory interface {
 	New(serverName, socketPath string) AgentProxy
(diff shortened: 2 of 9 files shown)


cmd/cloud-api-adaptor/main.go
```diff
@@ -71,6 +71,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 			flags.StringVar(&tlsConfig.KeyFile, "cert-key", "", "cert key")
 			flags.BoolVar(&tlsConfig.SkipVerify, "tls-skip-verify", false, "Skip TLS certificate verification - use it only for testing")
 			flags.BoolVar(&disableTLS, "disable-tls", false, "Disable TLS encryption - use it only for testing")
+			flags.IntVar(&cfg.serverConfig.ProxyTimeout, "proxy-timeout", 3, "Maximum timeout in minutes for establishing agent proxy connection")
```
yoheiueda, line 74: How about using the GO's duration string here, instead of specifying a value in minutes?
https://pkg.go.dev/time#ParseDuration

pkg/adaptor/proxy/factory.go
```diff
@@ -14,9 +14,10 @@ type factory struct {
 	criSocketPath string
 	tlsConfig     *tlsutil.TLSConfig
 	caService     tlsutil.CAService
+	proxyTimeout  int
```
yoheiueda, line 17: We can use `time.Duration` here.
```
	proxyTimeout  time.Duration
```

pkg/adaptor/proxy/factory.go
```diff
@@ -14,9 +14,10 @@ type factory struct {
...
 }
 
-func NewFactory(pauseImage, criSocketPath string, tlsConfig *tlsutil.TLSConfig) Factory {
+func NewFactory(pauseImage, criSocketPath string, tlsConfig *tlsutil.TLSConfig, proxyTimeout int) Factory {
```
yoheiueda, line 20: We can use `time.Duration` here, too.

pkg/adaptor/proxy/proxy.go
```diff
@@ -117,38 +113,28 @@ func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error)
...
 			conn, err = dialer.DialContext(ctx, "tcp", address)
-
-			if err == nil || count == maxRetries {
-				return
+			if err != nil {
+				return err
 			}
-			<-ctx.Done()
-		}()
+			return nil
```
yoheiueda, line 127: I think you can just return `err` without examining `err`.
```
return err
```

pkg/adaptor/proxy/proxy.go
```diff
@@ -117,38 +113,28 @@ func (p *agentProxy) dial(ctx context.Context, address string) (net.Conn, error)
...
-			err := fmt.Errorf("reaches max retry count. gave up establishing agent proxy connection to %s: %w", address, err)
-			logger.Print(err)
-			return nil, err
-		}
-		logger.Printf("failed to establish agent proxy connection to %s: %v. (retrying... %d/%d)", address, err, count, p.maxRetries)
-
-		count++
+	if err != nil {
+		logger.Printf("failed to establish agent proxy connection to %s", address)
+		return nil, err
```
yoheiueda, line 134: Could you wrap the original error like this?
```
err = fmt.Errorf("failed to establish agent proxy connection to %s: %w", address, err)
logger.Print(err)
return nil, err
```

If the error message becomes too long, you can specify the `LastErrorOnly` option.
https://pkg.go.dev/github.com/avast/retry-go#LastErrorOnly

yoheiueda (commented):
Could you wrap your Git commit message?

https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/4575067357/jobs/8077453701?pr=771

cmd/cloud-api-adaptor/main.go
snir911, line 74: Would it make sense to have a variable in entrypoint.sh for modifying it (similarly to pause image: https://github.com/confidential-containers/cloud-api-adaptor/blob/e12e76fd6cce33524c52cb93365634a862c63632/entrypoint.sh#L13)

pkg/adaptor/proxy/proxy.go
```diff
@@ -25,18 +26,18 @@ import (
...
+	defaultCriTimeout = 1 * time.Second
 
 	// The server TLS certificate must have this as SAN
 	// TODO: Avoid hard coding of server name
 	podvmServername = "podvm-server"
 )
 
-var logger = log.New(log.Writer(), "[adaptor/proxy] ", log.LstdFlags|log.Lmsgprefix)
+var (
+	DefaultProxyTimeout, _ = time.ParseDuration("3m")
```
yoheiueda, line 38: You don't need to define a variable here to parse "3m". See the above comment.

cmd/cloud-api-adaptor/main.go
yoheiueda: I agree.

@Amulyam24 could you also update `entrypoint.sh` as well?

Amulyam24: Sure @yoheiueda. Made the changes, PTAL.

yoheiueda (approved):
LGTM

snir911 (approved):
LGTM, thanks!

Issue #774, opened 2023-03-31
ibmcloud: Support staging for image import


stevenhorsman:
Sometimes it's helpful to test on staging, so adding the ability to support this in import.sh, the push.sh does would be helpful

PR #775, opened 2023-03-31
ibmcloud: Add the ability to import to staging

stevenhorsman:
- Add the ability to create the VPC image to staging, like push.sh does, by allowing the api endpoint to be overridden

Fixes: #774


jtumber-ibm:
I think this needs doc updates as you will also need to update the cos endpoint to use staging too which might not be obviously

jtumber-ibm (approved):

stevenhorsman:
@jtumber-ibm - I've added some documentation now, so take a look and see if you think it is enough. I was assuming that if you were using staging you'd likely be 'in the know', but thanks for challenging me that that was probably not a good assumption.

Jordan9500 (approved):
LGTM! 

nit: I dont know if we should limit it to them adding if it is dev/staging or production?

Issue #794, opened 2023-04-05
proxytimeout is not working as intended


bpradipt:
The proxytimeout is not working as intended due to default retry iterations set to 10.
The default iterations need to be set to `0` to allow cancellation via context.

PR #795, opened 2023-04-05
proxy: Fix proxytimeout handling

bpradipt:
Also increase the default timeout to account for different environments

Fixes: #794


yoheiueda (approved):
LGTM. Thank you!

jensfr (approved):
lgtm, thanks @bpradipt

Issue #573, opened 2023-02-01
Support multiple cloud providers in a single binary


yoheiueda:
The current implementation uses build tag to select a single cloud provider code to be built into a single binary. This design can reduce the binary size, since unnecessary cloud provider code are not build into a binary.

For a case where binary size does not matter, we can enhance `cloud-api-adaptor` to support multiple or all of cloud providers into a single binary.  

Currently, we need to specify a cloud provider name in the first argument to the command line, so a `cloud-api-adaptor` process can dynamically select an appropriate cloud provider based on the argument.

https://github.com/confidential-containers/cloud-api-adaptor/blob/e09a1ea0c8a22d8d192ef5815eb585a48b629fc9/cmd/cloud-api-adaptor/main.go#L58-L59

While supporting multiple cloud providers does not complicate the code, it will have some benefits.

- We can create a single universal binary that supports all cloud providers
- CI tests are needed to be executed once ([currently CI tests needed to be executed per cloud provider](https://github.com/confidential-containers/cloud-api-adaptor/blob/e09a1ea0c8a22d8d192ef5815eb585a48b629fc9/.github/workflows/build.yaml#L15-L20))
- Friendly to go lint (currently we need to suppress golint errors by editing [.golangci.yaml](https://github.com/confidential-containers/cloud-api-adaptor/blob/e09a1ea0c8a22d8d192ef5815eb585a48b629fc9/.golangci.yaml#L281-L287))
-  Friendly to IDEs like VS Code (currently we need to specify a cloud provider like `GOFLAGS=tags=ibmcloud`, and language support does not work in the code for the other cloud providers)

The current libvirt support uses CGO, so it is probably convenient to keep using a build tag for the libvirt provider code, until swiching to pure go libvirt implementation is done. https://github.com/confidential-containers/cloud-api-adaptor/issues/119

bpradipt:
@yoheiueda switching to pure go implementation for libvirt will be risky going forward. The pure go implementation is available from - https://github.com/digitalocean/go-libvirt whereas the one we use in CAA is the `official` libvirt go module available from - https://gitlab.com/libvirt/libvirt-go-module
Imho let's continue using the official libvirt go module which requires dynamic linking. 

Given that we create a container image for deployment of CAA, I think it should be ok to switch to a single dynamically linked binary for cloud-api-adaptor and agent-protocol-forwarder.

Wdyt ?

yoheiueda:
[...]

PR #841, opened 2023-04-19
ci: Unify redundant unit test execution

### LINKED ISSUE #840: Some CI tests are redandant
Previously, single binary supports cloud provider specified with with a build tag. This has been changed to support multiple providers in a single binary #573.

Become of this change, some CI jobs execute the identical sets of tests regardless of specified build tags.

We should unify such redundant CI jobs to reduce resource usage as well as waiting time.

yoheiueda:
This PR reduces redundant execution of unit tests in CI. `make test` now runs all unit tests including ones for multiple cloud providers since #573, so we don't need to execute unit tests for the build tag for each cloud provider.

This PR also adds execution of golangci-lint for the sub modules in addition to the main module. `volumes/csi-wrapper` has some lint errors, so it will be added in a follow-on PR.

Fixes #840


huoqifeng (approved):
/lgtm, I can see `-tags=aws,azure,ibmcloud,vsphere,libvirt` is used in `make test` in https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/4745602657/jobs/8428150684, thanks @yoheiueda

bpradipt (approved):
/lgtm
Thanks @yoheiueda

PR #849, opened 2023-04-20
csi-wrapper: Fix lint errors and enable lint checks in CI

yoheiueda:
This PR fixes the errors reported by golangci-lint in the csi-wrapper module.

I only refactored the code to supress lint errors, and I only added TODO comments at the locations where
errors are reported like this.
```
Error return value of `func` is not checked (errcheck)
```

This PR also bumps the version of golangci-lint to v1.52.2, since  the old version has a bug in processing the --prefix-path option, which is necessary to run the lint tool in sub modules.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,5 +1,5 @@
-# (C) Copyright Red Hat 2022.
-# SPDX-License-Identifier: Apache-2.0
+# (C) Copyright Confidential Containers Contributors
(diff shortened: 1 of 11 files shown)


liudalibj (approved):
LGTM, thanks @yoheiueda

huoqifeng (approved):
/lgtm, thanks @yoheiueda help fixed the lint errors in csi-wrapper!

Issue #598, opened 2023-02-09
Add versioning system for dependencies


stevenhorsman:
At the moment of the components we pull in (most notably to the podvm image) are either not versioned (e.g. kata-agent, attestation-agent)- so we pull in the latest code, or versions in particular files (skopeo, umoci, pause image, ubuntu version). In order to ensure our 'supply chain' is reliable and trusted and in future to enable us to have reproducible builds, we should consider adding a centralised place to control the versions of files, maybe similar to [versions.yaml](https://github.com/kata-containers/kata-containers/blob/CCv0/versions.yaml) that kata-containers has.

jtumber-ibm:
Example

```yaml
---
cloud-img:
  - distro: ubuntu
    architecture: amd64
    source:
      url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
      checksum: 3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
  - distro: ubuntu
    architecture: s390x
    source:
      url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-s390x.img
      checksum: 24673aa86785573d3a92e15166ff81beff88cbb0abc01938f156eb1332e87cd3
```

wainersm:
Taking the `Dockerfile.podvm` as an example, there are `UBUNTU_IMAGE_URL` and `UBUNTU_IMAGE_CHECKSUM` build arguments:

```
ARG UBUNTU_IMAGE_URL=https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
ARG UBUNTU_IMAGE_CHECKSUM=3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
```

How to feed the information from version.yaml (suppose we will use that file) as the example @jtumber-ibm provided on the comment above? Currently the podvm (and podvm_builder) build process is manual. Does it mean we will need a script or Make targets to automate the build process so that it reads from the version.yaml to pass the correct versions as arguments to the podvm build?

jtumber-ibm:
>
[...]

PR #851, opened 2023-04-21
cmd: Show version information

### LINKED ISSUE #850: Support showing version information in commands
Since a first release is available, we can add ability to show version information in commands.

yoheiueda:
This patch updates Makefile to derive version information from the latest git release tag, and embed it in binaries.

The version information is also shown in the begging of log messages from the commands.

Fixes #850 

A version information is shown like this
```
cloud-api-adaptor version v0.5.0-dev
  commit: f3f604e9cd8be01a1922b8c84e95bd4afd4b0bad-dirty
  go: go1.19.4
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,16 +46,18 @@ help: ## Display this help.
 # Targets that depend on .gits-commit can use $(shell cat .git-commit) to get a
 # git revision string.  They will only be rebuilt if the revision string
 # actually changes.
-# TODO When a release is created change these steps to use: git describe --abbrev=0 --tags to pull the latest release tag on a specific branch.
 .PHONY: .git-commit.tmp
 .git-commit: .git-commit.tmp
 	@cmp $< $@ >/dev/null 2>&1 || cp $< $@
 .git-commit.tmp:
-	@printf "$$(git rev-parse HEAD 2>/dev/null)" >$@
+	@printf "$$(git rev-parse HEAD 2>/dev/null || echo unknown)" >$@
 	@test -n "$$(git status --porcelain --untracked-files=no)" && echo -dirty >>$@ || true
 
-GOFLAGS += -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' \
-                     -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'"
+version = $(shell git describe --match "v[0-9]*" --abbrev=0 --tags --dirty=-dev || echo unknown)
+commit  = $(shell cat .git-commit)
+
+GOFLAGS += -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(version)' \
+                     -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(commit)'"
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -46,16 +46,18 @@ help: ## Display this help.
...
 .git-commit: .git-commit.tmp
 	@cmp $< $@ >/dev/null 2>&1 || cp $< $@
 .git-commit.tmp:
-	@printf "$$(git rev-parse HEAD 2>/dev/null)" >$@
+	@printf "$$(git rev-parse HEAD 2>/dev/null || echo unknown)" >$@
 	@test -n "$$(git status --porcelain --untracked-files=no)" && echo -dirty >>$@ || true
 
-GOFLAGS += -ldflags="-X 'github.com/confidential-containers/cloud-api-adaptor/cmd.VERSION=$(shell echo "unknown")' \
-                     -X 'github.com/confidential-containers/cloud-api-adaptor/cmd.COMMIT=$(shell cat .git-commit)'"
+version = $(shell git describe --match "v[0-9]*" --abbrev=0 --tags --dirty=-dev || echo unknown)
```
Jordan9500, line 56: Do we want the response to be unknown against a branch? 


For example, the version branch produces this: 
`git describe --match "v[0-9]*" --abbrev=0 --tags --dirty=-dev`
```
fatal: No names found, cannot describe anything.
```

Jordan9500 (approved):
Other than the question this LGTM! Thanks for the update @yoheiueda

cmd/agent-protocol-forwarder/main.go
```diff
@@ -67,6 +69,12 @@ func (cfg *Config) Setup() (cmd.Starter, error) {
 		cfg.tlsConfig = &tlsConfig
 	}
 
+	cmd.ShowVersion(programName)
+
+	if showVersion {
+		cmd.Exit(0)
```
bpradipt, line 75: nit: a question for my understanding. I see that Exit is mapped to os.Exit. Is there any benefit of doing this ?

bpradipt (approved):
/lgtm
thanks @yoheiueda

Makefile
yoheiueda: @Jordan9500 Thank you very much for pointing out the problem. I was able to reproduce the problem by directly cloning my personal repo. I didn't noticed the problem, since I am usually working in a local repository that is cloned from this upstream repo.

I don't have a good idea to work around the problem when tags are not available. Possibly, we can do something when versioning system for dependencies is introduced #598. Please let me proceed with the current approach for now.

cmd/agent-protocol-forwarder/main.go
yoheiueda: It is for unit tests. We can replace `Exit` in unit tests, and can check whether an exit code is expected one or not.

https://github.com/confidential-containers/cloud-api-adaptor/blob/f3f604e9cd8be01a1922b8c84e95bd4afd4b0bad/cmd/parse_test.go#L19-L23

Go does not provide a simple solution for testing `os.Exit` as described here.
https://stackoverflow.com/q/26225513

One possible solution was to use [Go monkeypatching](https://github.com/bouk/monkey), but it does not support s390x, and now it is archived.

It originally comes from my internal project. I don't think we need it unless we really need full code coverage in unit tests.

PR #856, opened 2023-04-21
Use correct podvm binaries image as default

bpradipt:


stevenhorsman (approved):
LGTM. It's a good job we set the BINARIES_IMG explicitly in the release workflow :)

snir911 (approved):

Issue #820, opened 2023-04-17
Parameterise the CAA image in peerpodconfig controller


bpradipt:
https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/peerpodconfig-ctrl/controllers/peerpodconfig_controller.go#L223

PR #862, opened 2023-04-24
peerpodconfig-ctrl: remove provider fetching

snir911:
as it's not needed anymore.
if validator for the peerpod ConfigMap/Secret fields is needed can be implemented in separate

Fixes: #820


bpradipt (approved):
/lgtm
Thanks @snir911

yoheiueda (approved):
LGTM. Thank you!

PR #874, opened 2023-04-27
Refactor the build procedure of the cloud-api-adaptor container image

### LINKED ISSUE #863: Limit files to be sent as container image build context
The current build process of the `cloud-api-adaptor` container image sends all files in the entire repository directory as follows.

https://github.com/confidential-containers/cloud-api-adaptor/blob/fb83fdb36d1899f822b7423729357a9298775968/hack/build.sh#L26-L31

https://github.com/confidential-containers/cloud-api-adaptor/blob/fb83fdb36d1899f822b7423729357a9298775968/Dockerfile#L5

This causes problems when the directory contains built binary files. For example, `build.sh` may fail to cross-compile a binary when the directory already contains a built binary for the host machine, since `make` does not rebuild a new binary for a target architecture.

Another problem is that container image build may takes long time when the directory contains large files such as QCOW2 files.

We can limit files to be sent by preparing a `.dockerignore` file. We can also reduce build time by selectively copying the files with the `COPY` statement in `Dockerfile`.

https://docs.docker.com/engine/reference/builder/#dockerignore-file

Note that the logic to derive version and commit strings in ` Makefile` may not work if we selectively send files to docker daemon, so we need to explicitly pass version and commit strings to `build.sh`.
 
https://github.com/confidential-containers/cloud-api-adaptor/blob/fb83fdb36d1899f822b7423729357a9298775968/Makefile#L56-L57

yoheiueda:
This PR optimizes how to build `cloud-api-adaptor` container images with Docker.

- Reduce files to be sent to a Docker daemon as build context
- Reorganize Dockerfile so that build cache is efficiently used
- Add .dockerignore to avoid accidental addition of unnecessary files

The logic of the version string generation in Makefile is also fixed in this PR. A version string should have suffix "-dev" for post-release commits, but the current Makefile always generates a version string without the suffix. This PR contains a patch that fixes the problem.

Fixes #863


mattarnoatibm (approved):
LGTM

snir911 (approved):
LGTM, thanks!

Issue #881, opened 2023-04-30
peerpod-ctrl: avoid the requirement for CM/Secret existence before controller start


snir911:
For peerpod-ctrl to be working properly it requires all of the cloud-provider configurations to be present in advance, usually they are being set as environment variables using secretRef and configMapRef.
Normally this is not an issue, however, sometimes Secret and ConfigMap are not available at the time the controller starts (e.g. when peerpod-ctrl is integrated as part of a set of controllers with a single manager)
One solution may be fetching the Secret/ConfigMap at runtime.

PR #882, opened 2023-04-30
WIP: peerpod-ctrl: dynamically fetch cloud provider config

snir911:
at runtime, when Provider is not passed

Fixes: #881

### DIFF
--- a/peerpod-ctrl/controllers/peerpod_controller.go
+++ b/peerpod-ctrl/controllers/peerpod_controller.go
@@ -18,13 +18,19 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/confidential-containers/cloud-api-adaptor/cmd/cloud-api-adaptor/cloudmgr"
 	confidentialcontainersorgv1alpha1 "github.com/confidential-containers/cloud-api-adaptor/peerpod-ctrl/api/v1alpha1"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/cloud"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -47,6 +57,20 @@ func (r *PeerPodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	logger := log.FromContext(ctx)
 	pp := confidentialcontainersorgv1alpha1.PeerPod{}
 
+	// cloud provider was not set, try to fetch cloud provider and its configs dyanmically from configmap or secret
```
bpradipt, line 60:
```suggestion
cloud provider was not set, try to fetch cloud provider and its configs dynamically from configmap or secret
```

peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -83,3 +107,44 @@ func (r *PeerPodReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	peerpodscm := corev1.ConfigMap{}
+	peerpodssecret := corev1.Secret{}
+	ns := os.Getenv("PEERPODS_NAMESPACE")
+	if ns == "" {
+		return fmt.Errorf("PEERPODS_NAMESPACE is not set")
+	}
+
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ppConfigMap, Namespace: ns}, &peerpodscm); err == nil {
+		for k, v := range peerpodscm.Data {
+			os.Setenv(k, v)
```
bpradipt, line 121: Do you need to set all the k,v pairs or just need the CLOUD_PROVIDER name ?

snir911: I need to set all of those required for cloud provider authentication, later, we may be able to change the way we are getting the the configuration in the cloud pkg and also use it here instead of using env vars

bpradipt: Ok, can you please add the above as a comment in the code.

peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -47,6 +59,20 @@ func (r *PeerPodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			return ctrl.Result{RequeueAfter: (1 * time.Minute)}, nil
+		}
+
+		var pErr error
+		r.Provider, pErr = SetProvider()
+		if pErr != nil {
+			return ctrl.Result{}, pErr
+		}
+	}
+
```
yoheiueda, line 75: I suppose we don't need mutex here, since this method is not called concurrently. Right?

yoheiueda (approved):
LGTM. Thank you very much!

bpradipt (approved):
/lgtm
Thanks @snir911 

The WIP needs to be removed

peerpod-ctrl/controllers/peerpod_controller.go
snir911: setProvider? no

snir911:
made some changes and removed WIP

peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -83,3 +109,47 @@ func (r *PeerPodReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return fmt.Errorf("ConfigMap Error: %v, Secret Error: %v", cmErr, secretErr)
+	}
+
+	return nil
+}
+
+func SetProvider() (cloud.Provider, error) {
+	cloudName := os.Getenv("CLOUD_PROVIDER")
+	if cloud := cloudmgr.Get(cloudName); cloud != nil {
+		cloud.LoadEnv() // we assume LoadEnv knows to load all nessecry configs
```
bpradipt, line 146: Nit: nessecry -> necessary

PR #893, opened 2023-05-03
Add versions.yaml file

jtumber-ibm:
Super simple yaml file to document what the default versions currently are for our third party dependencies.

Also included is an example of how this would be used in the Makefile to get the ubuntu image URL and checksum.

Welcome any comments. Thanks.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,0 +1,31 @@
+---
+# Sources for a base cloud os images
+cloudimg:
+  ubuntu:
+    focal:
+      amd64:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
+        checksum: "sha256:3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280"
+      s390x:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-s390x.img
+        checksum: "sha256:24673aa86785573d3a92e15166ff81beff88cbb0abc01938f156eb1332e87cd3"
+  centos:
+    stream-8:
+      amd64:
+        url: https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2
+        checksum: "sha256:b80ca9ccad8715aa1aeee582906096d3190b91cebe05fcb3ac7cb197fd68a7fe"
+tools:
+  golang: 1.19.6
+  rust: 1.66.0
+  protoc: 3.11.4
+# Referenced Git repositories
+git:
+  attestation-agent:
+    url: https://github.com/confidential-containers/attestation-agent
+    reference: v0.5.0
+  kata-containers:
+    url: https://github.com/kata-containers/kata-containers
+    reference: CCv0
+  cloud-api-adaptor:
+    url: https://github.com/confidential-containers/cloud-api-adaptor
+    reference: staging


stevenhorsman:
@jtumber-ibm - It looks pretty good to me so far, albeit not the most user friendly thing, but `make` to shell normally isn't!. I think we also want the attestation-agent repo in to the versions file as well.

Something that might be worth considering if whether we want to try and follow the template of the kata-containers versions.yaml file https://github.com/kata-containers/kata-containers/blob/CCv0/versions.yaml which has a slightly different format. It's not mandatory, but might be worth aligning with them on it.

I guess the next steps (once everyone is happy with the versions.yaml format) is to start adopting it where we pull in versions at the moment. Is your plan to do that under a separate PR?

Thanks!

versions.yaml
```diff
@@ -0,0 +1,34 @@
...
+      amd64:
+        url: https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2
+        checksum: b80ca9ccad8715aa1aeee582906096d3190b91cebe05fcb3ac7cb197fd68a7fe
+  redhat:
+    rhel8:
+      amd64:
+        url:
+        checksum:
+# Fixed version references
+version:
```
bpradipt, line 23: how about using `tools` or `build-tools` instead of `version` for golang, rust, protoc, packer, docker/buildx ?

bpradipt:
@jtumber-ibm the approach looks good. When do you plan to move it out of RFC ?

Eventually the gh-action, Dockerfiles etc can be updated to refer to the versions.yaml to pickup the right versions

podvm/Makefile
```diff
@@ -28,9 +22,14 @@ PACKER_DEFAULT_OPTS ?= -var boot_wait=300s
...
 
-setopts:
+setopts: tools
 ifeq ($(PODVM_DISTRO),ubuntu)
 	@echo defined
+
+	# Check for set but empty or unset URL and CHECKSUM
+	$(eval UBUNTU_IMAGE_URL ?= $(shell yq '.cloudimg.ubuntu.$(UBUNTU_RELEASE).$(DEB_ARCH).url' < ../versions.yaml))
+	$(eval UBUNTU_IMAGE_CHECKSUM ?= $(shell yq '.cloudimg.ubuntu.$(UBUNTU_RELEASE).$(DEB_ARCH).checksum' < ../versions.yaml))
+
```
yoheiueda, line 32: Can we refactor Makefile not to use the `setopts` target and the `eval` function  like this?  Variables defined with `=` are evaluated lazily. I don't think such refactoring is not the scope of this PR, but I think using `yq` for many variables makes Makefile more complicated, so I think it might be worth refactoring.

```

UBUNTU_IMAGE_URL = $(shell yq '.cloudimg.ubuntu.$(UBUNTU_RELEASE).$(DEB_ARCH).url' < ./versions.yaml)
UBUNTU_IMAGE_CHECKSUM = $(shell yq '.cloudimg.ubuntu.$(UBUNTU_RELEASE).$(DEB_ARCH).checksum' < ./versions.yaml)

ifeq ($(PODVM_DISTRO),ubuntu)

OPTS := -var qemu_image_name=$(IMAGE_FILE) \
	-var cloud_image_url=$(UBUNTU_IMAGE_URL) \
	-var cloud_image_checksum=$(UBUNTU_IMAGE_CHECKSUM)

ifneq ($(HOST_ARCH),$(ARCH))
OPTS += -var machine_type=$(QEMU_MACHINE_TYPE_$(ARCH))
ifndef QEMU_BINARY
OPTS += -var qemu_binary=qemu-system-$(ARCH)
endif
endif
endif

︙
$(IMAGE_FILE): $(BINARIES) $(FILES) tools
```

yoheiueda (commented):
Overall it looks good to me. I have some minor comments.

podvm/Makefile
jtumber-ibm: I agree it is/could get messy, it is a larger task to refactor that which can be looked at in a different issue.

jtumber-ibm:
> @jtumber-ibm the approach looks good. When do you plan to move it out of RFC ?
> 
> Eventually the gh-action, Dockerfiles etc can be updated to refer to the versions.yaml to pickup the right versions

Moved out of RFC, if people are happy to stay with this format then we can merge next week?

@bpradipt @yoheiueda @stevenhorsman

stevenhorsman:
Are you planning to pull out the ubuntu/yq usage from this PR altogether and do it separately, so this just has the version.yaml format?

Is there issue(s) that enumerate the follow on tasks after this merged as well?

Thanks!

jtumber-ibm:
> Are you planning to pull out the ubuntu/yq usage from this PR altogether and do it separately, so this just has the version.yaml format?
> 
> Is there issue(s) that enumerate the follow on tasks after this merged as well?
> 
> Thanks!

I think that is probably wise

versions.yaml
```diff
@@ -0,0 +1,34 @@
+---
+# Sources for a base cloud os images
+cloudimg:
+  ubuntu:
+    focal:
+      amd64:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
+        checksum: 3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
```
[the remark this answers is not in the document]
bpradipt: @jtumber-ibm looks like the earlier change got reverted

versions.yaml
bpradipt: @jtumber-ibm looks like the earlier change got reverted

versions.yaml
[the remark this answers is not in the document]
jtumber-ibm: Thanks for noticing, fixed

jtumber-ibm:
@bpradipt do you have values that can be used for the rhel section?

All, any version changes you'd like to see before being merged?

versions.yaml
```diff
@@ -0,0 +1,33 @@
...
+    focal:
+      amd64:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
+        sha256sum: 3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
+      s390x:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-s390x.img
+        sha256sum: 24673aa86785573d3a92e15166ff81beff88cbb0abc01938f156eb1332e87cd3
+  centos:
+    stream-8:
+      amd64:
```
wainersm, line 14: Hi @jtumber-ibm , first of all, thanks for working on that! I honestly don't like much the versions.yaml even on kata containers but I don't have a better solution :(

Regarding this line in specific, could you change from amd64 to x86_64? Usually using "amd64" implies that we will need a function/code to translate arch (as in `uname -m`).

versions.yaml
```diff
@@ -0,0 +1,33 @@
...
+        url: https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2
+        sha256sum: b80ca9ccad8715aa1aeee582906096d3190b91cebe05fcb3ac7cb197fd68a7fe
+  redhat:
+    rhel8:
+      amd64:
+        url:
```
bpradipt, line 20: @snir911 is there a rhel8 or rhel9 url link and sha256sum that we can use here ?

versions.yaml
```diff
@@ -0,0 +1,33 @@
...
+        sha256sum:
+tools:
+  golang: 1.19.6
+  rust: 1.66.0
+  protoc: 3.11.4
+# Referenced Git repositories
+git:
+  kata-containers:
+    url: https://github.com/kata-containers/kata-containers
```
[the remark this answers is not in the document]
jtumber-ibm: I believe you could do this without any changes to this file's structure, i.e just specify a path instead and git will treat it as a "remote", even if it is technically local (when this file gets used in the build process).

versions.yaml
jtumber-ibm: Hi @wainersm, I wish everyone would just pick one and stick with it, i.e. in Go-land it's `amd64` and rust went with `x86_64` and you rightly say that the kernel reports `x86_64`. 
As this is primarily a Go project I'd sway towards using `amd64` and we already have logic to switch from `x86_64` in place. 
Happy to switch if it would make other parts of the project easier

versions.yaml
```diff
@@ -0,0 +1,33 @@
+---
+# Sources for a base cloud os images
+cloudimg:
+  ubuntu:
+    focal:
+      amd64:
+        url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
+        sha256sum: 3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
```
yoheiueda, line 8: Previously, I commented that we should use `sha256sum` instead of `checksum` here to avoid ambiguity, but I'd like to take back my comment. Sorry.

After that, I noticed that Packer's option `iso_checksum` accepts a form of `"sha256:{$checksum}"`.

https://github.com/hashicorp/packer-plugin-sdk/blob/8331daf9a68be731b67b59621179cd3909458986/multistep/commonsteps/iso_config.go#L98-L118

So, I think it is better to specify a hash type in a value like below.

```
url: https://cloud-images.ubuntu.com/releases/focal/release-20230107/ubuntu-20.04-server-cloudimg-amd64.img
checksum: sha256:3895e38566e5c2c019f5c6f825ab7570ee34dac6b9142fab0c7e5a78084c4280
```

versions.yaml
snir911: no ATM

stevenhorsman (approved):
I think that most of the comments have been address on this, so I'm happy for it to get merged, ~once the RHEL link is added~ (it looks like this isn't available now?), so we can start the implementation and begin using it.

One possible addition - given that most of our other dependencies are in here I wonder if it is easier to add the attestation-agent now rather than wait for the implementation?

snir911:
yes, the rhel image link is dynamically created and there's no static one available ATM AFIAK, let's skip it for now

bpradipt (approved):
/lgtm
Thanks @jtumber-ibm

PR #894, opened 2023-05-03
azure: resource group is essential for deletion

snir911:
fetch it from also environment for peerpod-ctrl usage

Fixes: #881


littlejawa (approved):
lgtm

Issue #897, opened 2023-05-04
k8sops: Failing to release PeerPod object


snir911:
cloud-api-adaptor fails to release PeerPod object due to:
`[adaptor/cloud] failed to release PeerPod peerpods.confidentialcontainers.org "4221d360-8e45-4481-becc-9490d4be3019" not found `
This does not affect the peerpod-ctrl operation which deletes them

PR #898, opened 2023-05-04
k8sops: map Pods to their owned PeerPod objects

snir911:
so their name can be fetched correctly at release

Fixes: #897

<del>added hold until verified </del>


bpradipt (approved):
/lgtm

Issue #871, opened 2023-04-26
Remove old peer pods deployment fork


stevenhorsman:
Currently our installation of peer pods is based on having our own version of the kata-runtime payload that we deploy with a combination of
```
kubectl apply -f install/yamls/deploy.yaml
```
to in the peer pods fork of the controller manager and then our fork of the ccruntime through `ccruntime-peer-pods.yaml`

In https://github.com/confidential-containers/cloud-api-adaptor/issues/870 we should have updated to switch away from this process. Once this is stable and everyone is happy we should clean up the old code. I think this involves us deleting
- pre-install and runtime payload directories
- yamls/deploy.yaml and yamls/ccruntime-peer-pods.yaml
- Any other files?

Issue #870, opened 2023-04-26
Update install docs/Makefile/provisioning/e2e test to use 'upstream' CoCo payload


stevenhorsman:
Currently our installation of peer pods is based on having our own version of the kata-runtime payload that we deploy with a combination of
```
kubectl apply -f install/yamls/deploy.yaml
```
to in the peer pods fork of the controller manager and then our fork of the ccruntime through `ccruntime-peer-pods.yaml`

Now we've updated the coco operator to have the `kata-remote` in https://github.com/confidential-containers/operator/tree/main/config/samples/ccruntime/peer-pods we want to switch away from this duplication and move towards:
- Using CoCo operator to deploy the base stack
- Deploy the CAA pod and config (but not the ccruntime)

*Notes*

- This is a big change and probably involves updating a far few docs, the deploy makefile, any tools/e2e processes we have that deploy/install peer pods
- To make things easier to transition we should do https://github.com/confidential-containers/cloud-api-adaptor/issues/869 first so that we
- I'd like to find the least destructive way possible of doing this, and delete the yaml files as a separate step once we've stabilized this, which probably means excluding `ccruntime-peer-pods.yaml` from the kustomize rather than deleting it right now.

If anyone has a better approach, then I'd be really happy to hear it

stevenhorsman:
I've been having a look at this and made some progress with the upstream controller/payload, but hit an issue where the runtime-payload that we link to in the [peer-pods ccruntime](https://github.com/confidential-containers/operator/blob/main/config/samples/ccruntime/peer-pods/kustomization.yaml) is x86 only. I have a [branch](https://github.com/confidential-containers/operator/compare/main...stevenhorsman:peer-pods-s390x?expand=1#diff-a45d8628589061742bbaedf9450d611107524fba3375c23cb61aa85a2cbb72bf) with an s390x version, but that feels like it doesn't scale nicely to me, so I'm going to get Fabiano's opinion on whether I can try and make a multi-arch version of the runtime-payloads to avoid this...

stevenhorsman:
We are still blocked on this until the multi-arch image build of the CC runtime is created and the operator updated to use it in https://github.com/confidential-containers/operator/pull/169, but I've done some speculative work on it on my [branch](https://github.com/stevenhorsman/cloud-api-adaptor/tree/operator-switch) I'm sure there is a better way to update the provisioner, but once we have some upstream builds to test I'll check if this solution works.

PR #903, opened 2023-05-05
Update runtime payload runtimeclassname

stevenhorsman:
This PR updates the peer pod fork of the runtime payload to install to the `kata-remote` runtime class.
This is in prep for us switching over to the upstream coco operator rather than using our own fork.

I don't think there is any automation for building the payload image, so this is fairly safe to merge, but then we will need to build the runtime-payload manually and follow up with the doc, webhook and e2e compatible fixes in #904

### DIFF
--- a/install/runtime-payload/scripts/kata-deploy.sh
+++ b/install/runtime-payload/scripts/kata-deploy.sh
@@ -131,7 +131,7 @@ function cleanup_different_shims_base() {
 }
 
 function configure_crio_runtime() {
-	local runtime="kata"
+	local runtime="kata-remote"
 	local configuration="configuration"
 	if [ -n "${1-}" ]; then
 		runtime+="-$1"
@@ -172,7 +172,7 @@ function configure_crio() {
 }
 
 function configure_containerd_runtime() {
-	local runtime="kata"
+	local runtime="kata-remote"
 	local configuration="configuration"
 	if [ -n "${1-}" ]; then
 		runtime+="-$1"
--- a/install/yamls/ccruntime-peer-pods.yaml
+++ b/install/yamls/ccruntime-peer-pods.yaml
@@ -23,9 +23,9 @@ spec:
         name: containerd-conf
       - mountPath: /opt/confidential-containers/
         name: kata-artifacts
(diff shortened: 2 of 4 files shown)


stevenhorsman:
To limit the chances of having broken changes I suggest we review this and #904 at the same time then the flow is:
- Merge #903 
- Manually rebuild `quay.io/confidential-containers/peer-pods-runtime-payload`
- Merge #904 

The clock for breaking changes exists once the peer-pods-runtime-payload is pushed

bpradipt:
@stevenhorsman weren't we suppose to remove the custom payload from this repo and solely rely on the kata-containers payload that is used with operator and have the remote hypervisor config ?

stevenhorsman:
> @stevenhorsman weren't we suppose to remove the custom payload from this repo and solely rely on the kata-containers payload that is used with operator and have the remote hypervisor config ?

Yes, but we seemed to all agree that doing it in stages was the best way to limit the breaking changes, so switching to the coco operator is in #870 and the removal of the custom payload is part of #871. If you want to merge all of these changes into a single PR we can, but I think it will be very disruptive to upstream and any downstream projects to switch it all over in a single go?

bpradipt:
> Yes, but we seemed to all agree that doing it in stages was the best way to limit the breaking changes, so switching to the coco operator is in #870 and the removal of the custom payload is part of #871. If you want to merge all of these changes into a single PR we can, but I think it will be very disruptive to upstream and any downstream projects to switch it all over in a single go?

I see.  I don't have any concern with the approach. Just wanted to ensure that there is no duplication.

stevenhorsman:
> I see. I don't have any concern with the approach. Just wanted to ensure that there is no duplication.

Thanks, it is a little bit wasteful fixing up (and updating) the old peer-pods-runtime-payload before we remove it, but I hope it de-risks the switch, so it's a trade off that I'm happy to make. Looking forward to getting rid of the duplicated code soon though :)

bpradipt (approved):
/lgtm
Thanks @stevenhorsman

snir911 (approved):
LGTM thanks

PR #905, opened 2023-05-05
Use same pod vm image settings for different providers

bpradipt:


jensfr (approved):
lgtm, thanks @bpradipt

PR #910, opened 2023-05-09
peerpodconfig: Handle empty nodesList

bpradipt:
The controller retrieves the list of nodes based on the peerpodconfig nodeselector. The number of nodes in the list is used to proportionally advertise the extended resources among the nodes. However the code didn't handle the case where the list can be empty. This PR fixes the same and doesn't advertise extended resources for empty nodesList

Fixes: #909


yoheiueda (approved):
LGTM

snir911 (approved):

PR #955, opened 2023-05-12
ci: add shellcheck to lint workflow and fix issues

katexochen:
For now, this fixes findings with severity "warning" and also uses this severity level for the workflow.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -4,53 +4,65 @@
 # Run linting tools on the sources of the project.
 ---
 name: lint
-on: [pull_request]
+on: [workflow_dispatch, pull_request]
+env:
+  GO_VERSION: "1.19"
 jobs:
-  golangci-lint_job:
-    name: golangci-lint
+  vet-and-fmt:
+    name: vet and fmt
     runs-on: ubuntu-latest
     permissions:
       contents: read
-    strategy:
-      fail-fast: false
-      matrix:
-        go_version:
-          - 1.19
-        workdir:
-          - .
-          - peerpod-ctrl
-          - peerpodconfig-ctrl
-          - volumes/csi-wrapper
-          - webhook
     steps:
       - name: Checkout the pull request code
         uses: actions/checkout@v3
-      - name: Setup Golang version ${{ matrix.go_version }}
+      - name: Setup Golang version ${{ env.GO_VERSION }}
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ matrix.go_version }}
+          go-version: ${{ env.GO_VERSION }}
       - name: Install dependencies
         run: |
           sudo apt-get update -y
           sudo apt-get install -y libvirt-dev
-      - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v3
-        with:
-          working-directory: ${{ matrix.workdir }}
-          args: --path-prefix=${{ matrix.workdir }}
-          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.52.2
+      - name: check
+        run: |
+          make fmt
+          make vet
 
-  go_check:
+  golangci-lint:
+    name: golangci-lint
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
-      - uses: actions/checkout@main
-      - uses: actions/setup-go@main
+      - name: Checkout the pull request code
+        uses: actions/checkout@v3
+      - name: Setup Golang version ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v3
         with:
-          go-version: 1.19
-      - name: Install build dependencies
+          go-version: ${{ env.GO_VERSION }}
+      - name: Install dependencies
         run: |
           sudo apt-get update -y
           sudo apt-get install -y libvirt-dev
+      - name: Install golangci-lint
+        env:
+          GOLANGCI_LINT_VERSION: "1.52.2"
+        working-directory: /tmp
+        run: |
+          curl -sSfLO "https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz"
+          tar -xzf "golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz"
+          sudo mv "golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64/golangci-lint" /usr/local/bin
       - name: check
-        run: make check
+        run: make golangci-lint
+
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout the pull request code
+        uses: actions/checkout@v3
+      - name: Run shellcheck
+        run: make shellcheck
--- a/aws/image/misc-settings.sh
+++ b/aws/image/misc-settings.sh
@@ -24,13 +24,13 @@ then
 	    #fallthrough
 	    ;&
 	rhel)
-	    (! dnf list --installed | grep open-vm-tools 2>&1 >/dev/null) && \
+	    (! dnf list --installed | grep open-vm-tools >/dev/null 2>&1) && \
 		(! dnf -y install open-vm-tools) && \
 		     echo "$PODVM_DISTRO: Error installing package required for cloud provider: $CLOUD_PROVIDER" 1>&2 && exit 1
 	    ;;
 	ubuntu)
-	    (! dpkg -l | grep open-vm-tools 2>&1 > /dev/null) && apt-get update && \
-		(! apt-get -y install open-vm-tools 2>&1 > /dev/null) && \
+	    (! dpkg -l | grep open-vm-tools > /dev/null 2>&1) && apt-get update && \
+		(! apt-get -y install open-vm-tools > /dev/null 2>&1) && \
 		     echo "$PODVM_DISTRO: Error installing package required for cloud provider: $CLOUD_PROVIDER" 1>&2  && exit 1
 	    ;;
 	*)
--- a/aws/image/misc-settings.sh
+++ b/aws/image/misc-settings.sh
@@ -1,4 +1,4 @@
-# Uncomment the awk statement if you want to use mac as 
+# Uncomment the awk statement if you want to use mac as
 # dhcp-identifier in Ubuntu
 #awk '/^[[:space:]-]*dhcp4/{ split($0,arr,/dhcp4.*/)
 #                           gsub(/-/," ", arr[1])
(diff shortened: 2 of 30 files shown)


yoheiueda (commented):
Thank you very much for this great work!

I found some false-positives. Could you check my comments?

I also noticed that the script below is also a shell script, while it does not have the `.sh` extension. Does the shellcheck action detect this file as well, or do we need to change the script filename to have `.sh`?

[podvm/files/usr/local/bin/kata-agent-clean](https://github.com/confidential-containers/cloud-api-adaptor/blob/a9941bba4692c22fada21b5b9f70a1986c74690d/podvm/files/usr/local/bin/kata-agent-clean)

ibmcloud/terraform/cluster/scripts/setup.sh
```diff
@@ -48,7 +46,7 @@ if ! [[ -e /var/lib/kubelet/kubeadm-flags.env ]]; then
     # https://cloud.ibm.com/docs/vpc?topic=vpc-configuring-address-prefixes&locale=en
 
     curl -sL -o /tmp/kube-flannel.yml https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
-    sed -i 's|"10.244.0.0/16"|"172.20.0.0/16"|' /tmp/kube-flannel.yml
+    sed -i 's|10.244.0.0/16|172.20.0.0/16|' /tmp/kube-flannel.yml
```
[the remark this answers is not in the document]
katexochen: Sure, do we have to escape the double quotes int the sed statement to not break the outer double-quoted string (bash -x -c "....", line 39)?

katexochen:
Thanks for the review @yoheiueda

> Does the shellcheck action detect this file as well, or do we need to change the script filename to have .sh?

This file should be checked. Beside file extension, the action [looks for executable files with a shebang](https://github.com/ludeeus/action-shellcheck/blob/00cae500b08a931fb5698e11e79bfbd38e612a38/action.yaml#L198-L204).

ibmcloud/terraform/cluster/scripts/setup.sh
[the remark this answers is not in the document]
yoheiueda: Oh, I didn't notice the outer double quotes, although I originally wrote this script 😅 

Yes, the line will look like this?
```
sed -i 's|\"10.244.0.0/16\"|\"172.20.0.0/16\"|' /tmp/kube-flannel.yml
```

ibmcloud/image/build.sh
```diff
@@ -201,7 +201,8 @@ if [ "${SE_BOOT}" = "1" ]; then
     done
     printf "\nFormatting boot-se partition\n"
     mke2fs -t ext4 -L boot-se ${tmp_nbd}p1
-    export boot_uuid=$(blkid ${tmp_nbd}p1 -s PARTUUID -o value)
```
wainersm, line 204: It will be so hard for me to make this old habit die! :)

ibmcloud/image/push.sh
```diff
@@ -82,7 +82,7 @@ while image_id=$(ibmcloud is images --output=json --visibility=private | jq -r "
     fi
 done
 
-echo -e "\nUploading $image_path to cloud object stroage at "$cos_bucket" with key $object_key\n"
```
wainersm, line 84: maybe stroage is a typo?

wainersm: I suspect here the original author wanted $cos_bucket enclosed by double-quotes.

volumes/csi-wrapper/hack/ibm/cleanandreinstall.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/bash
-# this script is for develop test only, user must prepared env via wrap-ibm-vpc-block-csi-driver.md 
+# this script is for develop test only, user must prepared env via wrap-ibm-vpc-block-csi-driver.md
 kubectl delete po nginx
 kubectl delete pvc my-pvc
 kubectl -n kube-system delete po nginx
 kubectl -n kube-system delete pvc my-pvc
 pv_name=$(kubectl get pv |grep ibmc-vpc-block-5iops-tier|awk '{print $1}')
 kubectl delete pv $pv_name
-cd ~/ibm-vpc-block-csi-driver
```
wainersm, line 9: what's the problem here?

aws/image/misc-settings.sh
```diff
@@ -24,13 +24,13 @@ then
 	    #fallthrough
 	    ;&
 	rhel)
-	    (! dnf list --installed | grep open-vm-tools 2>&1 >/dev/null) && \
+	    (! dnf list --installed | grep open-vm-tools >/dev/null 2>&1) && \
```
wainersm, line 27: yet another old habit hard to die.

wainersm:
Great work @katexochen !

One question: is it possible to have a Makefile target for running shellcheck locally as part of `check`?  So people don't get used to run the CI just for the sake of running shellcheker as it waste resources.

volumes/csi-wrapper/hack/ibm/cleanandreinstall.sh
wainersm: I mean, I wonder if exiting the script at first directory failure is the right thing. As this seems a cleanup script, perhaps it should move to the next  block if a given directory doesn't exist, to continue the clean up process.

wainersm:
> > is it possible to have a Makefile target for running shellcheck locally as part of `check`?
> 
> Why isn't golangci-lint part of `make check`?

Good question. When I introduced golangci-lint I forgot to add on `make check` :(

katexochen:
There seems to be some weird issue with golangci-lint and the `--path-prefix` flag. When not using the flag, [there are findings in the `volumes/csi-wrapper` package](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/5002275289/jobs/8962021507?pr=955#step:6:114). If I set the flag, these findings disappear (and the weren't found by the [lint workflow before, which used the flag](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/.github/workflows/lint.yaml#L40)). I didn't manage to reproduce this behavior in another directory. The version without the flag is currently on this branch, you can check the one with [this debug branch](https://github.com/katexochen/cloud-api-adaptor/tree/test/golangci-pathprefix).

volumes/csi-wrapper/hack/ibm/cleanandreinstall.sh
katexochen: I added your suggestion.

yoheiueda:
@katexochen The findings are intentionally excluded by the settings in `.golangci.yaml`. This YAML file is for all modules in this repository, so a path prefix is added to each package path. When `--path-prefix` is not specified, paths do not include the prefix, and this exclude rule does not match.

https://github.com/confidential-containers/cloud-api-adaptor/blob/f8a9de44e384fa42dd8e35a86ced526750ff34b8/.golangci.yaml#L299-L303

yoheiueda (commented):
Why custom scripts (shellcheck.sh and golangci-lint.sh) are used, instead of actions?

I guess the intention is to run all tests sequentially, not in parallel.

If there are both golang and script errors, I think only the go error is reported, since `make test` stops executing remaining tests when the golang test fails.

I think it is convenient to run the remaining tests even when a preceding test fails.

katexochen:
> Why custom scripts (shellcheck.sh and golangci-lint.sh) are used, instead of actions?

@yoheiueda My intention was to execute the exact same checks in CI as in local dev environment. The shellcheck action is essentially a script around shellcheck, like the one I wrote. I think duplicating things like exclude and include list will let them diverge at some point. Also the call of `make check` was already there, I just deleted the steps that would now be duplicated.

But I agree that it would be nice to not terminate if a step fails to also get the results of the other linters.
I see the following two ways we could enable this:

1. Don't run the `check` target but the targets underneath, execute in different jobs.
2. Add a meta script that sequentially executes the different checks and returns the result afterwards.

hack/shellcheck.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# Configuration
+scanDir="." # Directory to scan
+shellcheckOpts=( # Options to pass to ShellCheck
+    "--severity=warning"
+)
+ignorePaths=( # Paths to ignore when running ShellCheck
+    "*./.git/*"
+    "*.go"
+)
```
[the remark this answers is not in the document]
katexochen: Sure, I can add this. I guess in consequence newly created, local scripts that are not yet added to Git won't be linted, but I guess that is a downside we can accept.

hack/golangci-lint.sh
```diff
@@ -0,0 +1,89 @@
...
+if ! command -v golangci-lint &> /dev/null; then
+    echoerr "golangci-lint is not installed, please install it."
+    echoerr "You can find installation instructions at https://golangci-lint.run/usage/install/."
+    exit 1
+fi
+
+# Configuration
+excludeModules=(
+    "./podvm" # see the comment in podvm/go.mod
+)
```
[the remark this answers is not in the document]
katexochen: Would you prefer to just add these paths to the list, or check the files at runtime for their Git status?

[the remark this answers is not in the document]
katexochen: I've added a check using git.

hack/shellcheck.sh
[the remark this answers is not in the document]
katexochen: Never mind, this shouldn't be a problem. I've added a check using git.

katexochen:
> Don't run the check target but the targets underneath, execute in different jobs.

I've implemented 1. for now.

.github/workflows/lint.yaml
```diff
@@ -4,53 +4,69 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout the pull request code
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libvirt-dev shellcheck
```
yoheiueda, line 70: Looks like the Ubuntu runner image already contains the `shellcheck` command. I guess we don't need to install it by ourselves.

https://github.com/actions/runner-images/blob/9d9b3a110dfc98100cdd09cb2c957b9a768e2979/images/linux/Ubuntu2204-Readme.md?plain=1#L363

hack/golangci-lint.sh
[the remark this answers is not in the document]
yoheiueda: Thank you very much!

hack/shellcheck.sh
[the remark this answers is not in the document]
yoheiueda: Thank you very much!

.github/workflows/lint.yaml
yoheiueda: I don't think `libvirt-dev` is necessary just to run `shellcheck`

yoheiueda (approved):
LGTM. Thank you very much!

(Please don't forget to squash the fixup commits.)

katexochen:
Ping @snir911 @jtumber-ibm @wainersm

bpradipt:
This looks good to be merged.
Thanks @katexochen

Issue #959, opened 2023-05-16
ibmcloud: support to upload se-podvm image and update SE document


liudalibj:
In the provision code when we `UploadPodvm` we always use `ubuntu-20-04-s390x` as the `osName` for s390x arch https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/test/provisioner/provision_ibmcloud.go#L886, we should use `hyper-protect-1-0-s390x` as the osName for the se-podvm image.

And the verify SE enabled steps are missed.

PR #960, opened 2023-05-16
ibmcloud: upload se-podvm image with correct osName

liudalibj:
- How to verify one container is SE enabled
- Use correct osName when create image base on se-podvm image

fixes: #959


stevenhorsman (approved):
LGTM. Thanks for adding the verification information and the uploadImage instance profile check!

huoqifeng (approved):
/lgtm
Thank you @liudalibj

Issue #962, opened 2023-05-16
peerpodconfig-ctrl: set CAA_IMAGE as relatedImage in CSV of operator


jensfr:
> .spec.relatedImages: list any related images, or other container images that your Operator might require to perform their functions. This list can include operand images as well. All image references should be specified by digest (SHA) and not by tag.

CAA_IMAGE is the name of our env variable for the pull spec of the cloud-api-adaptor image. If we add the prefix RELATED_IMAGE_ and add it to config/manager/manager.yaml it will be automatically added to the CSV when the bundle is generated and the env var will be available in the controller code. 

It is also helpful for digest pinning by build tools like [OSBS](https://osbs.readthedocs.io/en/latest/index.html)

PR #963, opened 2023-05-16
peerpodconfig-ctrl: rename CAA_IMAGE env var

jensfr:
rename it to RELATED_IMAGE_ENV_VAR so that we can include it as a related image in the CSV and by that enable automatic digest pinning in tools like OSBS
(https://osbs.readthedocs.io/en/latest/users.html#creating-the-relatedimages-section)

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/962


bpradipt (approved):
/lgtm
Thanks @jensfr

snir911 (approved):

Issue #979, opened 2023-05-19
s390x podvm image build, the qemu instance boot time need be increased


liudalibj:
By default we will waiting 5 mins for the qemu instance to bootstrap, but for s390x podvm image build it seems that we will need more time.

from https://github.com/liudalibj/cloud-api-adaptor/actions/runs/5011343888/jobs/8982069481
```
...
[2023-05-18T00:40:52.648Z] ==> qemu.ubuntu: Typing the boot commands over VNC...
[2023-05-18T00:40:52.648Z]     qemu.ubuntu: Not using a NetBridge -- skipping StepWaitGuestAddress
[2023-05-18T00:40:52.648Z] ==> qemu.ubuntu: Using SSH communicator to connect: 127.0.0.1
[2023-05-18T00:40:52.648Z] ==> qemu.ubuntu: Waiting for SSH to become available...
[2023-05-18T00:42:00.386Z] ==> qemu.ubuntu: Error waiting for SSH: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
[2023-05-18T00:42:00.386Z] ==> qemu.ubuntu: Deleting output directory...
[2023-05-18T00:42:00.386Z] Build 'qemu.ubuntu' errored after 7 minutes 13 seconds: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
[2023-05-18T00:42:00.386Z] 
[2023-05-18T00:42:00.386Z] ==> Wait completed after 7 minutes 13 seconds
[2023-05-18T00:42:00.386Z] 
...
```

PR #980, opened 2023-05-19
podvm: increase s390x podvm build boot_wait

liudalibj:
- increase s390x podvm build boot_wait to 600s to avoid qemu ssh connection timeout 

fixes: #979
Signed-off-by: Da Li Liu <liudali@cn.ibm.com>

### DIFF
--- a/podvm/Makefile
+++ b/podvm/Makefile
@@ -37,8 +37,8 @@ ifeq ($(PODVM_DISTRO),ubuntu)
 	-var cloud_image_url=${UBUNTU_IMAGE_URL} \
 	-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM} \
 	-var se_boot=${SE_BOOT})
-ifeq ($(SE_BOOT),1)
-	$(eval PACKER_DEFAULT_OPTS = -var boot_wait=480s )
+ifeq ($(ARCH),s390x)
+	$(eval PACKER_DEFAULT_OPTS = -var boot_wait=600s )
 endif
 ifneq ($(HOST_ARCH),$(ARCH))
 	$(eval OPTS += -var machine_type=${QEMU_MACHINE_TYPE_${ARCH}})


podvm/Makefile
```diff
@@ -37,8 +37,8 @@ ifeq ($(PODVM_DISTRO),ubuntu)
 	-var cloud_image_url=${UBUNTU_IMAGE_URL} \
 	-var cloud_image_checksum=${UBUNTU_IMAGE_CHECKSUM} \
 	-var se_boot=${SE_BOOT})
-ifeq ($(SE_BOOT),1)
-	$(eval PACKER_DEFAULT_OPTS = -var boot_wait=480s )
+ifeq ($(ARCH),s390x)
```
huoqifeng, line 40: My understanding is to increase the timeout to 10 minutes both for SE and non-SE image for s390x? Previous timeout for non-SE image was 5 minutes and SE image was 8 minutes?
Is it good enough to increase SE image from 8 minutes to 10 minutes and non-SE image from 5 minutes to 10 minutes? Shall we increase timeout more larger?

huoqifeng: Given the errors like below `errored after 7 minutes 13 seconds`, looks 10 minutes is not safe?
```
Build 'qemu.ubuntu' errored after 7 minutes 13 seconds: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain
```

liudalibj: > My understanding is to increase the timeout to 10 minutes both for SE and non-SE image for s390x? Previous timeout for non-SE image was 5 minutes and SE image was 8 minutes? Is it good enough to increase SE image from 8 minutes to 10 minutes and non-SE image from 5 minutes to 10 minutes? Shall we increase timeout more larger?

yeah, it's both se and noe-se image, I go through most of the failed job 10 minutes for boot the qemu s390x instance should be enough.

huoqifeng (approved):
/lgtm
Thanks @liudalibj

yoheiueda (approved):
LGTM.

I wondering why we need to rely on `boot_wait`, since necessary boot_time may be different depending on environment. I'll check whether there is another way to wait for VM startup.

Issue #1001, opened 2023-05-24
ibmcloud: e2e Test cases for ibmcloud to check Peer Pod Logs User from container


sudharshanibm3:
Need to write e2e-test cases for ibmcloud to check peer-pod Logs User from container

Steps to test peer pod with Logs of User from container :

- Build and push a customised multi-arch image from following Dockerfile
```
FROM --platform="${TARGETPLATFORM}" alpine:latest AS testuser
RUN addgroup -S othergroup && adduser -S otheruser -G othergroup
USER otheruser
ENTRYPOINT [ "/bin/sh", "-c", "whoami" ]
```
- Create a peer-pod with the customized image
- Let the pod to reach `Completed` status
- Run commands to get logs from the pod `kubectl logs <podname>`
- The Expected output of the log must be `otheruser`

PASS cases - If the log matches the expected output `otheruser`
FAIL cases 
    - If the Pod Fails to start
    - If the Pod Logs different output
    - If the Pod fails to reach `Completed` Status

PR #942, opened 2023-05-11
ibmcloud: Added test case to check peer pod created with job

sudharshanibm3:
Added test cases to check peer pod created via job in ibmcloud cloud-provider

Resolves: #930, Resolves: #1004

PR #1002, opened 2023-05-24
ibmcloud: e2e-test cases for ibmcloud to check peer-pod Logs User from container

sudharshanibm3:
Added e2e-test cases for ibmcloud to check peer-pod Logs User from container

Fixes: #1001 

Note: Current PR is dependent on #942

### DIFF
--- a/.github/workflows/test-images.yaml
+++ b/.github/workflows/test-images.yaml
@@ -1,0 +1,68 @@
+name: Create Test Images
+on:
+  push:
+    branches:
+      - "staging"
+    paths:
+      - 'test/e2e/fixtures/Dockerfile.*'
+
+jobs:
+  list-dockerfiles:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ env.MATRIX }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "MATRIX=$(ls test/e2e/fixtures/Dockerfile.* | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_ENV
+  build:
+    name: Create Test Images
+    needs: list-dockerfiles
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        targets: ${{ fromJson(needs.list-dockerfiles.outputs.matrix) }}
+        arch: [amd64, s390x]
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
+    - name: Login to Docker container Registry
+      uses: docker/login-action@v2
+      with:
+        registry: docker.io
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - name: Login to Quay container Registry
+      uses: docker/login-action@v2
+      with:
+        registry: quay.io
+        username: ${{ secrets.QUAY_USERNAME }}
+        password: ${{ secrets.QUAY_PASSWORD }}
+    - name: determine docker tag by splitting branch name on slash
+      id: docker-tag
+      shell: bash
+      run: |
+          for path in "${{matrix.targets}}"; do
+            file_name=$(basename "$path")
+            name="${file_name##*.}"
+            output_paths+=("${name,,}")
+          done
+          echo "DOCKER_TAG=${output_paths[*]}" >> $GITHUB_ENV
+    - name: Build and push
+      uses: docker/build-push-action@v3
+      with:
+        tags: |
+          quay.io/${{ secrets.QUAY_USERNAME }}/test-images:${{env.DOCKER_TAG}}
+        push: true
+        context: .
+        platforms: linux/s390x,linux/amd64
+        file: ${{matrix.targets}}
+        build-args: |
+          "ARCH=${{ matrix.arch }}"
--- a/test/e2e/common_suite_test.go
+++ b/test/e2e/common_suite_test.go
@@ -6,6 +6,7 @@ package e2e
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
@@ -50,6 +51,7 @@ type testCase struct {
 	job                  *batchv1.Job
 	testCommands         []testCommand
 	expectedPodLogString string
+	podState             v1.PodPhase
 }
 
 func (tc *testCase) withConfigMap(configMap *v1.ConfigMap) *testCase {
@@ -79,6 +81,11 @@ func (tc *testCase) withTestCommands(testCommands []testCommand) *testCase {
 
 func (tc *testCase) withExpectedPodLogString(expectedPodLogString string) *testCase {
 	tc.expectedPodLogString = expectedPodLogString
+	return tc
+}
+
+func (tc *testCase) withCustomPodState(customPodState v1.PodPhase) *testCase {
+	tc.podState = customPodState
 	return tc
 }
 
@@ -114,9 +121,11 @@ func (tc *testCase) run() {
 				if err = client.Resources().Create(ctx, tc.pod); err != nil {
 					t.Fatal(err)
 				}
-				if err = wait.For(conditions.New(client.Resources()).PodRunning(tc.pod), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
+
+				if err = wait.For(conditions.New(client.Resources()).PodPhaseMatch(tc.pod, tc.podState), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
 					t.Fatal(err)
 				}
+
 			}
 			return ctx
 		}).
@@ -145,31 +154,40 @@ func (tc *testCase) run() {
 						t.Errorf("Job Created pod with Invalid log")
 					}
 				}
-
 			}
+
 			if tc.pod != nil {
 				if err := client.Resources(tc.pod.Namespace).List(ctx, &podlist); err != nil {
 					t.Fatal(err)
 				}
-				tc.assert.HasPodVM(t, tc.pod.Name)
-				for _, testCommand := range tc.testCommands {
-					var stdout, stderr bytes.Buffer
+				if tc.expectedPodLogString != "" {
+					LogString, err := comparePodLogString(ctx, client, *tc.pod, tc.expectedPodLogString)
+					if err != nil {
+						t.Fatal(err)
+					}
+					t.Logf("Log output of peer pod:%s", LogString)
+				}
+				if tc.podState == v1.PodRunning && tc.testCommands != nil {
+					for _, testCommand := range tc.testCommands {
+						var stdout, stderr bytes.Buffer
 
-					for _, podItem := range podlist.Items {
-						if podItem.ObjectMeta.Name == tc.pod.Name {
-							if err := cfg.Client().Resources(tc.pod.Namespace).ExecInPod(ctx, tc.pod.Namespace, tc.pod.Name, testCommand.containerName, testCommand.command, &stdout, &stderr); err != nil {
-								t.Log(stderr.String())
-								t.Fatal(err)
-							}
+						for _, podItem := range podlist.Items {
+							if podItem.ObjectMeta.Name == tc.pod.Name {
+								if err := cfg.Client().Resources(tc.pod.Namespace).ExecInPod(ctx, tc.pod.Namespace, tc.pod.Name, testCommand.containerName, testCommand.command, &stdout, &stderr); err != nil {
+									t.Log(stderr.String())
+									t.Fatal(err)
+								}
 
-							if !testCommand.testCommandStdoutFn(stdout) {
-								t.Fatal(fmt.Errorf("Command %v running in container %s produced unexpected output on stdout: %s", testCommand.command, testCommand.containerName, stdout.String()))
+								if !testCommand.testCommandStdoutFn(stdout) {
+									t.Fatal(fmt.Errorf("Command %v running in container %s produced unexpected output on stdout: %s", testCommand.command, testCommand.containerName, stdout.String()))
+								}
 							}
 						}
 					}
 				}
-			}
+				tc.assert.HasPodVM(t, tc.pod.Name)
 
+			}
 			return ctx
 		}).
 		Teardown(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
@@ -232,9 +250,50 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
 		testName:      testName,
 		assert:        assert,
 		assessMessage: assessMessage,
+		podState:      v1.PodRunning,
 	}
 
 	return testCase
+}
+
+func comparePodLogString(ctx context.Context, client klient.Client, customPod v1.Pod, expectedPodlogString string) (string, error) {
+	podLogString := ""
+	var podlist v1.PodList
+	clientset, err := kubernetes.NewForConfig(client.RESTConfig())
+	if err != nil {
+		return podLogString, err
+	}
+	if err := client.Resources(customPod.Namespace).List(ctx, &podlist); err != nil {
+		return podLogString, err
+	}
+	for _, pod := range podlist.Items {
+		if pod.ObjectMeta.Name == customPod.Name {
+			func() {
+				req := clientset.CoreV1().Pods(customPod.Namespace).GetLogs(pod.ObjectMeta.Name, &v1.PodLogOptions{})
+				podLogs, err := req.Stream(ctx)
+				if err != nil {
+					return
+				}
+				defer podLogs.Close()
+				buf := new(bytes.Buffer)
+				_, err = io.Copy(buf, podLogs)
+				if err != nil {
+					return
+				}
+				podLogString = strings.TrimSpace(buf.String())
+			}()
+		}
+	}
+
+	if err != nil {
+		return podLogString, err
+	}
+
+	if podLogString != expectedPodlogString {
+		return podLogString, errors.New("Error: Pod Log doesn't match with Expected String")
+	}
+
+	return podLogString, nil
 }
 
 func getSuccessfulAndErroredPods(ctx context.Context, t *testing.T, client klient.Client, job batchv1.Job) (int, int, string, error) {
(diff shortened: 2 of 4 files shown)


stevenhorsman (commented):
Hey Sudharshan, can you remove the jobs test commit from this PR please as it isn't related to this issue? Also I think you need a commit description and DCO sign off in the workflow commit. Thanks!

stevenhorsman (commented):
On first glance it looks okay, but I think there is a lot of overlap in code with #942, so I think we need to get that merged first and then rebase on it otherwise we'll likely have a lot of conflicts

sudharshanibm3:
Hi @stevenhorsman ,
Rebased the code with latest staging branch

.github/workflows/test-images.yaml
```diff
@@ -0,0 +1,71 @@
...
+jobs:
+  list-dockerfiles:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' }}
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "::set-output name=matrix::$(ls test/e2e/fixtures/Dockerfile.* | jq -R -s -c 'split("\n")[:-1]')"
```
stevenhorsman, line 18: Just to check - is `jq` installed on `ubuntu-latest`?

stevenhorsman: Any chance you can kick this off manually in your fork to check this? (though I appreciate that you might need to update the registry for that.

stevenhorsman: Also - it looks like `set-output` is deprecated now: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

sudharshanibm3:
Hi @stevenhorsman ,
Here I ran the [workflow](https://github.com/sudharshanibm3/cloud-api-adaptor/actions/runs/5179007906) in my forked repo with my [registry](https://quay.io/repository/sudharshanibm3/test-images?tab=tags)
Screenshot 2023-06-05 at 9 02 37 PM

.github/workflows/test-images.yaml
stevenhorsman: Thanks for the fixes!

.github/workflows/test-images.yaml
```diff
@@ -0,0 +1,71 @@
...
+  build:
+    name: Create Test Images
+    needs: list-dockerfiles
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        targets: ${{ fromJson(needs.list-dockerfiles.outputs.matrix) }}
```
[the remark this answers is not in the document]
sudharshanibm3: I think it's due to the inability in unmarshalling property of the matrix
here I found something related to that, [solution](https://github.com/nektos/act/issues/927#issuecomment-1035013048)

.github/workflows/test-images.yaml
```diff
@@ -0,0 +1,70 @@
...
+        tags: |
+          quay.io/${{ secrets.QUAY_USERNAME }}/test-images:${{env.DOCKER_TAG}}
+        push: true
+        context: .
+        platforms: linux/s390x,linux/amd64
+        file: ${{matrix.targets}}
+        build-args: |
+          "ARCH=${{ matrix.arch }}"
+          "UBUNTU_IMAGE_URL="
+          "UBUNTU_IMAGE_CHECKSUM="
```
stevenhorsman, line 70: Why are these arguments needed?

sudharshanibm3: Sorry, I think these arguments are not required. Removed the arguments in updated code

test/e2e/common_suite_test.go
```diff
@@ -236,6 +254,39 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
 
 	return testCase
 }
+func ComparePodLogString(ctx context.Context, t *testing.T, client klient.Client, Pod v1.Pod, ExpectedPodlogString string) (bool, string, error) {
```
stevenhorsman, line 257: A new line between functions makes it easier to read IMHO

test/e2e/common_suite_test.go
```diff
@@ -236,6 +254,39 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
 
 	return testCase
 }
```
stevenhorsman, line 256: A new line between functions makes it easier to read IMHO

test/e2e/common_suite_test.go
stevenhorsman, line 257: Convention is for go variables to start with lower case letters
```suggestion
func ComparePodLogString(ctx context.Context, t *testing.T, client klient.Client, pod v1.Pod, expectedPodlogString string) (bool, string, error) {
```

stevenhorsman, line 257: Is this function intended to be exported? otherwise I think it should start with a lower case too?

test/e2e/common_suite_test.go
```diff
@@ -114,15 +112,58 @@ func (tc *testCase) run() {
...
 					t.Fatal(err)
 				}
-				if err = wait.For(conditions.New(client.Resources()).PodRunning(tc.pod), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
-					t.Fatal(err)
+				if tc.expectedPodLogString != "" {
+					if err = wait.For(conditions.New(client.Resources()).PodPhaseMatch(tc.pod, v1.PodSucceeded), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
+						t.Fatal(err)
+					}
+				} else {
+					if err = wait.For(conditions.New(client.Resources()).PodRunning(tc.pod), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
```
stevenhorsman, line 120: Quick question - do the custom test images go into/end on a different state to the others e.g. nginx and that's why we need to do this branch?

test/e2e/common_suite_test.go
```diff
@@ -114,15 +112,58 @@ func (tc *testCase) run() {
...
+					}
 				}
+
 			}
 			return ctx
 		}).
 		Assess(tc.assessMessage, func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			client := cfg.Client()
 			var podlist v1.PodList
+			if tc.pod != nil {
```
stevenhorsman, line 131: Any particularly reason that you moved the `if tc.pod != nil` logic up above the `if tc.job != nil` branch in this PR?

sudharshanibm3: No It wont affect the logic of the code.  Placement can be changed if required

test/e2e/common_suite_test.go
sudharshanibm3: Yeah 
nginx image will be in `Running` state
our custom image will go to `Completed` state 
so made a separate branch for tracking pods with `Completed` state

sudharshanibm3:
Hi @stevenhorsman ,
Updated the requested changes

test/e2e/common_suite_test.go
stevenhorsman: I guess my concern here is that this branch isn't really based on whether we have expected logging, but rather what the image is, so we could get into a situation where we want to check the logs of an image that stays in running. I wonder if instead of branching on this if we could update the wait to look for either running or completed and then it should be valid for all cases?

test/e2e/common_suite_test.go
stevenhorsman: Yes - I'd rather keep the original placement as it makes the diff in the PR smaller, thanks

stevenhorsman:
> Hi @stevenhorsman , Updated the requested changes

Thanks, I've added comments on a couple of them.

BTW @mattarnoatibm - if you have any cycles to review this before your vacation then I'd definitely value your opinion!

test/e2e/common_suite_test.go
sudharshanibm3: Let me know if this solution works

- Decalred a `podState` in testCase struct
- If any pod requires different state to assess rather than `Running`, they can declare via `withCustomPodState`
- So users can declare different podState like `Succeeded`, `Pending`,`Unknown`,`Failed` and podState will dynamically add those states while setup part

Implemented above approach in updated code, please let me know your views on this approach

test/e2e/common_suite_test.go
```diff
@@ -50,6 +50,7 @@ type testCase struct {
 	job                  *batchv1.Job
 	testCommands         []testCommand
 	expectedPodLogString string
+	podState             string
```
mattarnoatibm, line 53: Can we make the type of `podState` a `v1.podPhase`?

test/e2e/common_suite_test.go
```diff
@@ -114,15 +120,58 @@ func (tc *testCase) run() {
 				if err = client.Resources().Create(ctx, tc.pod); err != nil {
 					t.Fatal(err)
 				}
-				if err = wait.For(conditions.New(client.Resources()).PodRunning(tc.pod), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT)); err != nil {
-					t.Fatal(err)
+				if tc.podState != "" {
```
mattarnoatibm, line 123: `wait.For(conditions.New(client.Resources()).PodRunning(tc.pod), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT))` should be equivalent to `wait.For(conditions.New(client.Resources()).PodPhaseMatch(tc.pod, v1.PodRunning)`, so I think we can remove the need to have this `if` `else` branch by

- Setting a default `podState` of `v1.PodRunning` in `newTestCase`
- Change the `wait.For` code to `wait.For(conditions.New(client.Resources()).PodPhaseMatch(tc.pod, tc.podState), wait.WithTimeout(WAIT_POD_RUNNING_TIMEOUT))`, which covers test cases where the test expects a pod to reach the running phase and where the test expects a pod to reach the Succeeded phase.

test/e2e/common_suite_test.go
```diff
@@ -114,15 +120,58 @@ func (tc *testCase) run() {
...
+					isLogStringValid, LogString, err := comparePodLogString(ctx, t, client, *tc.pod, tc.expectedPodLogString)
+					if err != nil {
+						t.Fatal(err)
+					}
+					if isLogStringValid {
+						t.Logf("Log output of peer pod:%s", LogString)
+					} else {
+						t.Errorf("Invalid Log output from peer pod:%s", LogString)
+					}
+				} else if tc.testCommands != nil {
```
mattarnoatibm, line 153: I don't think having checking logs and running a command inside a pod container in exclusive branches is going to be valid for all test cases we could write, because you could write a test case where we want to check the logs in a pod that reaches the Running phase, and then run a command inside a container in that running pod.

I suggest checking if `tc.podState` is `v1.PodRunning` and running the test commands if that condition is true, and making that conditional separate from the `tc.expectedPodLogString != ""` check.

test/e2e/common_suite_test.go
```diff
@@ -237,6 +263,40 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
 	return testCase
 }
 
+func comparePodLogString(ctx context.Context, t *testing.T, client klient.Client, customPod v1.Pod, expectedPodlogString string) (bool, string, error) {
```
mattarnoatibm, line 266: I don't see `t *testing.T` used anywhere in this function.

test/e2e/common_suite_test.go
```diff
@@ -237,6 +263,40 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
...
+				_, err = io.Copy(buf, podLogs)
+				if err != nil {
+					return
+				}
+				podLogString = strings.TrimSpace(buf.String())
+			}()
+		}
+	}
+	if podLogString != expectedPodlogString {
+		return false, podLogString, nil
```
mattarnoatibm, line 295: Instead of returning `false` and a `nil` you can return a  new error (e.g., from `errors.New()`) here.

This also lets you simplify the types returned by  `comparePodLogString` to just `(string, error)`.

test/e2e/fixtures/Dockerfile.testuser
```diff
@@ -0,0 +1,4 @@
+FROM --platform="${TARGETPLATFORM}" alpine:latest
+RUN addgroup -S othergroup && adduser -S otheruser -G othergroup
+USER otheruser
+ENTRYPOINT [ "/bin/sh", "-c", "whoami" ]
```
mattarnoatibm, line 4: NP: Missing final newline.

test/e2e/ibmcloud_test.go
```diff
@@ -92,6 +92,13 @@ func TestCreatePeerPodWithJob(t *testing.T) {
 	doTestCreatePeerPodWithJob(t, assert)
 }
 
+func TestCreatePeerPodWithUserImage(t *testing.T) {
```
mattarnoatibm, line 95: I think the name of this test case and the name of the `doTestCreatePeerPodWithUserImage` function should be different. I think the purpose of this test case it to test fetching logs from a pod container. The fact the pod container image is a user-built (or test case built) image is, I believe, circumstantial. So I think a name like `TestCreatePeerPodAndCheckLogs`.

mattarnoatibm (changes requested):
Thanks for contributing and refining these tests.

I've added some comments and suggestions.

test/e2e/ibmcloud_test.go
sudharshanibm3: Hi @mattarnoatibm ,
There are test case which  need to be merged after this PR, which checks the logs of workdir, envvariable, etc
So can we keep this name as `TestCreatePeerPodAndCheckUserLogs`, `TestCreatePeerPodAndCheckWorkDirectoryLogs`...  ?

mattarnoatibm: Suggestion - since this test case checks the output of a container image entrypoint command, rename the test case to `TestCreatePeerPodAndCheckOutputOfEntrypointCmd`?

sudharshanibm3:
Hi @mattarnoatibm ,
Updated code with requested changes

test/e2e/common_suite_test.go
stevenhorsman: I think we are still waiting for this fix?

test/e2e/ibmcloud_test.go
stevenhorsman: So I'm not sure that this works fully like this Sudharshan? We have a number of tests that do this behaviour, but they will use different images and expect different logged output, so I think whilst we could have a basic test called `TestCreatePeerPodAndCheckOutputOfEntrypointCmd`, we would need to pass in the pod name, image, expected log, testcase name, description etc to have it unique for the 3/4 different test cases and at that point I'm not sure if it's messier to have a shared function with ~7/8 parameters rather than individual ones?

test/e2e/common_suite_test.go
sudharshanibm3: Exchanged `job` and `pod` statements in updated code

test/e2e/ibmcloud_test.go
stevenhorsman: I'm not a huge fan of using the map to get around there being a lot of parameters. I guess maybe the best thing to do it go back to the `TestCreatePeerPodAndCheckUserLogs` name and then look at refactoring it once we add more tests with common code, rather than attempting to prematurely refactor for the future?

sudharshanibm3: Hi @stevenhorsman ,
Reverted the test name to `TestCreatePeerPodAndCheckUserLogs`

stevenhorsman: It still looks to be named `doTestCreatePeerPodAndCheckOutputOfEntrypointCmd`? Anyway, the main thing is to remove the map/extra parameters and review that once we add the env/workdir test

stevenhorsman (approved):
This looks good to me - as mentioned we can work on further refactoring once the subsequent related changes are ready

test/e2e/common_suite_test.go
```diff
@@ -237,6 +255,40 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
...
+				_, err = io.Copy(buf, podLogs)
+				if err != nil {
+					return
+				}
+				podLogString = strings.TrimSpace(buf.String())
+			}()
+		}
+	}
+	if podLogString != expectedPodlogString {
+		return podLogString, errors.New(err.Error())
```
mattarnoatibm, line 287: This will throw a nil pointer dereference if nothing in the inner function returns an error but the pod container logs from the pod are not equal to the pod container logs the test case expected.

I'd suggest creating a new error object with a new custom error message string.

I think you need a separate check before this check to return `podLogString, err` if `err` is not nil.

test/e2e/common_suite_test.go
```diff
@@ -292,7 +344,7 @@ func getSuccessfulAndErroredPods(ctx context.Context, t *testing.T, client klien
 func doTestCreateSimplePod(t *testing.T, assert CloudAssert) {
 	namespace := envconf.RandomName("default", 7)
 	pod := newNginxPod(namespace)
-	newTestCase(t, "SimplePeerPod", assert, "PodVM is created").withPod(pod).run()
+	newTestCase(t, "SimplePeerPod", assert, "PodVM is created").withPod(pod).withCustomPodState(v1.PodRunning).run()
```
mattarnoatibm, line 347: I'd make the `newTestCase` function add a default pod state `v1.PodRunning` to the test case so you don't have to use `withCustomPodState(v1.PodRunning)` when creating every test case.

test/e2e/common_suite_test.go
```diff
@@ -237,6 +255,40 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
 	return testCase
 }
 
+func comparePodLogString(ctx context.Context, client klient.Client, customPod v1.Pod, expectedPodlogString string) (string, error) {
+	podLogString := ""
+	var podlist v1.PodList
+	clientset, err := kubernetes.NewForConfig(client.RESTConfig())
+	if err != nil {
+		return podLogString, errors.New(err.Error())
```
mattarnoatibm, line 263:
```suggestion
		return podLogString, err
```

test/e2e/common_suite_test.go
```diff
@@ -237,6 +255,40 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
...
+	}
+	if err := client.Resources(customPod.Namespace).List(ctx, &podlist); err != nil {
+		return podLogString, errors.New(err.Error())
```
mattarnoatibm, line 266:
```suggestion
		return podLogString, err

**Review by mattarnoatibm — CHANGES_REQUESTED** [review]
Thanks for the changes.

I've got a few more changes to request that fall outside of general refactoring.

**Review by mattarnoatibm — CHANGES_REQUESTED** [review]

File: test/e2e/common_suite_test.go [hunk 19]
```diff
@@ -232,11 +250,48 @@ func newTestCase(t *testing.T, testName string, assert CloudAssert, assessMessag
...
+				_, err = io.Copy(buf, podLogs)
+				if err != nil {
+					return
+				}
+				podLogString = strings.TrimSpace(buf.String())
+			}()
+		}
+	}
+	if podLogString != expectedPodlogString {
+		if err != nil {
```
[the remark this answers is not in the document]
sudharshanibm3: Hi @mattarnoatibm ,
I think updated code resolves this error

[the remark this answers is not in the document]
mattarnoatibm: I think there's a corner case where `expectedPodlogString` is `""` and `err` is not nil where the `comparePodLogString` function will return `podLogString, nil` when it should return `podLogString, err`. So you could catch that by taking the `if err != nil {` case outside of the `if podLogString != expectedPodlogString {` case.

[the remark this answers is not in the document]
sudharshanibm3: ok, then added an extra condition in updated code. Could you please check if that's ok?

mattarnoatibm (approved):
LGTM. Thanks for the contribution!

Issue #1011, opened 2023-05-25
The target path for latest AA has changed


bpradipt:
Existing path in Makefile.inc `app/target/$(RUST_TARGET)/release/attestation-agent`
`app` subfolder no longer exists when building latest AA. The updated path is
`target/$(RUST_TARGET)/release/attestation-agent`

PR #1032, opened 2023-06-01
podvm: Update AA install path

bpradipt:
Fixes: #1011


snir911 (approved):

stevenhorsman (approved):
LGTM. I re-built the podvm using this source ref and it ran successfully. I then used that image and all the e2e tests passed. Thanks for the fix!

PR #1057, opened 2023-06-07
[WIP] Load defaults from versions.yaml for podvm build

jtumber-ibm:
Built off of https://github.com/confidential-containers/cloud-api-adaptor/pull/1018 

- Renamed existing Makefile to Makefile.image
- Refactored existing Makefile to target Docker image build process
- Created Makefile.defaults to load the required values from versions.yaml

Files Updated (Excluding changes from 1018)

- hack/yq.sh
- podvm/Makefile
- podvm/Makefile.image
- podvm/Makefile.default


jtumber-ibm:
@yoheiueda could you take a look at the Makefile changes (excluding the Makefile.defaults for now) and see if it is worth adding into your #1018 ?

yoheiueda:
@jtumber-ibm Your proposed change looks good to me.

With your change, we can build a builder image from the remote repo based on centos by this.
```
make docker-builder-remote-centos
```

And, we can build a builder image from the remote repo based on centos by this.
```
make docker-podvm-local-ubuntu
```

One thing I noticed is that my proposed Dockefile refactoring requires `docker buildx`, so I think it is better to use `docker buildx build` explicitly.

Another point is that how we can support `podman build` in addition to `docker`.

So, is it possible to parameterize the container image build command like this?

```
BUILD_CMD = docker buildx build
```

Then, we can change the build command like this.
```
make BUILD_CMD='podman build' build-podvm-local-ubuntu
```

@bpradipt any comments?

yoheiueda:
I think the Makefile.defaults part is work-in-progress, and you are possibly working on a similar idea, but I think we can define a fuction like below to reduce redundancy. (I don't come up with a good function name though.)

```
define query
	$(or $(shell $(YQ_PATH) '.$(1) | select(. != null)' $(VERSIONS_SRC)), \
	     $(error $(1) is not set, and could not be automatically set from $(VERSIONS_SRC)))
endef

ubuntu_amd64_IMAGE_URL ?= $(call query,cloudimg.ubuntu.focal.amd64.url)
```

jtumber-ibm:
@yoheiueda from a quick test it seems that `podman` doesn't support only building the required stages

yoheiueda:
Thank you for the investigation! Probably, we can support docker buildx only for the time being.

PR #1059, opened 2023-06-07
0.6.0 release versions

stevenhorsman:
Update kata-containers and attestation-agent versions to use 0.6.0 release/tags

### DIFF
--- a/pkg/adaptor/proxy/proxy_test.go
+++ b/pkg/adaptor/proxy/proxy_test.go
@@ -283,6 +283,9 @@ func (m *agentMock) WaitProcess(ctx context.Context, req *pb.WaitProcessRequest)
 func (m *agentMock) UpdateContainer(ctx context.Context, req *pb.UpdateContainerRequest) (*types.Empty, error) {
 	return &types.Empty{}, nil
 }
+func (m *agentMock) UpdateEphemeralMounts(ctx context.Context, req *pb.UpdateEphemeralMountsRequest) (*types.Empty, error) {
+	return &types.Empty{}, nil
+}
 func (m *agentMock) StatsContainer(ctx context.Context, req *pb.StatsContainerRequest) (*pb.StatsContainerResponse, error) {
 	return &pb.StatsContainerResponse{}, nil
 }
@@ -365,6 +368,9 @@ func (m *agentMock) GetVolumeStats(ctx context.Context, req *pb.VolumeStatsReque
 	return &pb.VolumeStatsResponse{}, nil
 }
 func (m *agentMock) ResizeVolume(ctx context.Context, req *pb.ResizeVolumeRequest) (*types.Empty, error) {
+	return &types.Empty{}, nil
+}
+func (p *agentMock) RemoveStaleVirtiofsShareMounts(ctx context.Context, req *pb.RemoveStaleVirtiofsShareMountsRequest) (*types.Empty, error) {
 	return &types.Empty{}, nil
 }
 func (m *agentMock) PullImage(ctx context.Context, req *pb.PullImageRequest) (*pb.PullImageResponse, error) {
--- a/pkg/adaptor/shim_test.go
+++ b/pkg/adaptor/shim_test.go
@@ -364,6 +364,10 @@ func (s *agentService) UpdateContainer(ctx context.Context, req *agent.UpdateCon
 	log.Printf("agent call: %T %#v", req, req)
 	return &types.Empty{}, nil
 }
+func (s *agentService) UpdateEphemeralMounts(ctx context.Context, req *agent.UpdateEphemeralMountsRequest) (*types.Empty, error) {
+	log.Printf("agent call: %T %#v", req, req)
+	return &types.Empty{}, nil
+}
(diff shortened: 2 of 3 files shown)


stevenhorsman:
So it looks like we aren't compatible with the latest version of kata runtime, so we can either rollback, or try and fix it up. I'll look at the latter to see how doable it is. I guess a lesson for next release cycle that we need to try and stay more up to date.

stevenhorsman:
So the remaining error we are seeing is in peerpod-ctrl:
```
golangci-lint run --path-prefix=./peerpod-ctrl
WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package agentproto: could not load export data: no export data for "github.com/confidential-containers/cloud-api-adaptor/pkg/util/agentproto"
ERRO Running error: 1 error occurred:
	* can't run linter goanalysis_metalinter: buildir: failed to load package agentproto: could not load export data: no export data for "github.com/confidential-containers/cloud-api-adaptor/pkg/util/agentproto"
```

I'm not sure if this is because there is some issue with peerpod-ctrl not being able to reference/use the agent protos package as it is referencing github.com/confidential-containers/cloud-api-adaptor v0.5.1-0.20230504043629-580abeb128aa so there is a package mismatch?

stevenhorsman:
> I could try doing `go mod edit -replace github.com/confidential-containers/cloud-api-adaptor=../` to see if that helps?

I have tried this and pushed to the branch, but github is not updating the PR apparently 🤷

stevenhorsman:
So `github.com/confidential-containers/cloud-api-adaptor=../` seems to have fixed the error, so I guess we merge this, do the release and then update the go module to point to the release. Does that sound okay to you @bpradipt?
We might have to think about it a bit more in future if we want to avoid this though?

bpradipt:
Hmm, this will be a problem going forward as well. We might have to version the CAA first and then update the rest of them. 

@yoheiueda any thoughts ?

bpradipt:
Btw, we should be good for now and address it post the release

.github/workflows/podvm_builder.yaml
```diff
@@ -63,4 +63,4 @@ jobs:
           "CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor"
           "CAA_SRC_REF=staging"
           "KATA_SRC=https://github.com/kata-containers/kata-containers"
-          "KATA_SRC_BRANCH=CCv0"
+          "KATA_SRC_BRANCH=CC-0.6.0"
```
bpradipt, line 66: nit: the commit message will require a change to mention 0.6.0 release

bpradipt (approved):
/lgtm
The commit message will require changes.

peerpod-ctrl/go.mod
```diff
@@ -77,7 +77,7 @@ require (
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20220913141151-9b49a6ddc6fd // indirect
+	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20230606170044-338e18e4fd46 // indirect
```
yoheiueda, line 80: Could you revert this change, and run `go mod tidy` again?

We need a separated PR to update this line, since this `go.mod` file contains a reference to the `cloud-api-adaptor` main module as follows.

```
github.com/confidential-containers/cloud-api-adaptor v0.5.1-0.20230504043629-580abeb128aa
```

We need to update these lines at the same time in a separated PR after this PR gets merged.

peerpod-ctrl/go.mod
```diff
@@ -131,3 +131,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/confidential-containers/cloud-api-adaptor => ../
```
yoheiueda, line 135: We can remove this `replace` directive if we apply the change I mentioned above.

yoheiueda (approved):
LGTM.

I think it is OK for now to include the `replace` directive. If we want remove the `replace` directive, we need an additional separated PR as described below.

I will prepare some documentation how to update go.mod for a release next time.

peerpod-ctrl/go.mod
stevenhorsman: I'm slightly concerned that this means we will be building the release with some references that are 9 months old, but peerpod-ctrl isn't part of the release at the moment, so I'm happy to give this a go

stevenhorsman: @yoheiueda - the linting tests still failed in the same way they did before I added the replace, so I think I need to revert your suggestion to get them passing in the short-term?

peerpod-ctrl/go.mod
stevenhorsman: As mentioned above this still results in the linting tools failing: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/5209437464/jobs/9399283534?pr=1059 :(

PR #1066, opened 2023-06-08
Go mod caa 0.6

stevenhorsman:
Post 0.6 release, bump the cloud-api-adaptor go mods to reference the released version


bpradipt (approved):
/lgtm
Thanks @stevenhorsman

stevenhorsman:
@liudalibj @genjuro214 - With the cloud-api-adaptor bump, which also brings some kubernetes module version bumps, we are getting some errors in the csi-wrapper:
```
volumes/csi-wrapper/pkg/peerpodvolume/controller.go:50:37: Error return value of `(k8s.io/client-go/tools/cache.SharedInformer).AddEventHandler` is not checked (errcheck)
	informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
I wondered if you had any time to look into those and we can get them into into this PR?

yoheiueda:
It looks like this commit changes the return value of the function.
https://github.com/kubernetes/client-go/commit/ecdc8bf7296ca836be500025e53a1b27506e3a6c

We just ignore the lint error in the CSI wrapper for now. #849 

So, I think we can just ignore return value like this with a TODO comment.

```
// TODO: error check
_, _ = informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```

stevenhorsman:
> So, I think we can just ignore return value like this with a TODO comment.
> 
> // TODO: error check
> _, _ = informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{

Thanks very much for the info, I'll give this a try and see if any other issue crop up like this

yoheiueda (approved):
LGTM. Thank you very much!

PR #1069, opened 2023-06-09
volume/csi-wrapper: make sure CSI target path has been mounted before…

### LINKED ISSUE #1068: volume/csi-wrapper: make sure CSI target path has been mounted before creating workload container
In https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/volumes/csi-wrapper/hack/ibm/nginx-kata-with-my-pvc-and-csi-wrapper.yaml, we can see `ibm-vpc-block-podvm-node-driver`, `csi-podvm-wrapper` and `nginx` are started in sequence, so that the `nginx` can use the mount path `/mount-path` which was mount/config by `ibm-vpc-block-podvm-node-driver` and `csi-podvm-wrapper`. However, although the start times are in sequence but the end times are not. For example, start sequence is:
`ibm-vpc-block-podvm-node-driver` -> `csi-podvm-wrapper` -> `nginx`
But the `nginx` could be started while `csi-podvm-wrapper` just started but not config the storage yet, considering the future encryption work takes longer time, it becomes a critical problem.

----

For this problem, we have a draft solution like below:

## Draft solution

1. In [Node wrapper](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/volumes/csi-wrapper/pkg/wrapper/nodeservice.go#L115), we invoke kata **direct-volume** API to add target path from **NodePublishVolumeRequest** as direct volume.

The target path usually looks like: `/var/lib/kubelet/pods/<POD_ID>/volumes/kubernetes.io~csi/<PVC_ID>/mount`

**direct-volume** API will create a directory on worker node. It looks like: `/run/kata-containers/shared/direct-volumes/<base64 encode of volume path>`

2. The target path will be a mount source of the OCI object in **CreateContainerRequest** for the workload container.

For example:
```
{
  "destination": "/mount-path",
  "type": "bind",
  "source": "/var/lib/kubelet/pods/11820868-f8ad-4622-bc7e-f251781ac2ee/volumes/kubernetes.io~csi/pvc-c1cc455b-bc53-469f-8bd0-5a55909fe52c/mount",
  "options": [
    "rbind",
    "rslave",
    "rw"
  ]
},
```

In adaptor proxy [CreateContainer](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/adaptor/proxy/service.go#L94) func, we can go through OCI mount sources, and compare with the volume paths from `/run/kata-containers/shared/direct-volumes`

If matched, add a new **annotation** to OCI, for example:
```
"annotations": {
    "io.confidentialcontainers.org.peerpodvolumes.target_path": "/var/lib/kubelet/pods/11820868-f8ad-4622-bc7e-f251781ac2ee/volumes/kubernetes.io~csi/pvc-c1cc455b-bc53-469f-8bd0-5a55909fe52c/mount",
    ...
}
```

3. In agent-protocol-forwarder interceptor [CreateContainer](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/pkg/forwarder/interceptor/interceptor.go#L75) func, we can handle the **CreateContainerRequest** before redirecting it to **kata-agent**

Get the target path from OCI annotations with key "io.confidentialcontainers.org.peerpodvolumes.target_path"

Go through OCI mount source and compare with target path. 

If matched, check if the target path has been mounted using command `findmnt -n -o TARGET --target <target_path>`

If yes, continue to redirect **CreateContainerRequest** to **kata-agent**

If no, start a timer to query every 5 seconds, until the target path is mounted or time out

genjuro214:
… creating workload container

- in adaptor proxy, identify the target path and add it to OCI annotations
- In forworder interceptor, get the target path from OCI annotations
- wait for device is mounted to the target path, and then redirect CreateContainerRequest to kata-agent

fixes #1068

### DIFF
--- a/install/overlays/ibmcloud/kata_direct_volumes_mount.yaml
+++ b/install/overlays/ibmcloud/kata_direct_volumes_mount.yaml
@@ -1,0 +1,21 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: cloud-api-adaptor-daemonset
+  namespace: confidential-containers-system
+  labels:
+    app: cloud-api-adaptor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cloud-api-adaptor-con
+        volumeMounts:
+        - mountPath: /run/kata-containers/shared/direct-volumes
+          name: kata-direct-volumes-dir
+      volumes:
+      - name: kata-direct-volumes-dir
+        hostPath:
+          path: /run/kata-containers/shared/direct-volumes
+          type: DirectoryOrCreate
+
--- a/pkg/adaptor/proxy/service.go
+++ b/pkg/adaptor/proxy/service.go
@@ -5,8 +5,11 @@ package proxy
 
 import (
 	"context"
+	b64 "encoding/base64"
 	"fmt"
 	"net"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/avast/retry-go/v4"
@@ -24,7 +27,12 @@ type proxyService struct {
 	pauseImage string
 }
 
-const defaultPauseImage = "registry.k8s.io/pause:3.7"
+const (
+	defaultPauseImage            = "registry.k8s.io/pause:3.7"
+	kataDirectVolumesDir         = "/run/kata-containers/shared/direct-volumes"
+	volumeTargetPathKey          = "io.confidentialcontainers.org.peerpodvolumes.target_path"
+	csiPluginEscapeQualifiedName = "kubernetes.io~csi"
+)
 
 func newProxyService(dialer func(context.Context) (net.Conn, error), criClient *criClient, pauseImage string) *proxyService {
 
@@ -94,16 +102,20 @@ func (s *proxyService) getImageName(annotations map[string]string) (string, erro
 func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (*types.Empty, error) {
 
 	logger.Printf("CreateContainer: containerID:%s", req.ContainerId)
-	if len(req.OCI.Annotations) > 0 {
-		logger.Print("    annotations:")
-		for k, v := range req.OCI.Annotations {
-			logger.Printf("        %s: %s", k, v)
-		}
-	}
 	if len(req.OCI.Mounts) > 0 {
 		logger.Print("    mounts:")
 		for _, m := range req.OCI.Mounts {
 			logger.Printf("        destination:%s source:%s type:%s", m.Destination, m.Source, m.Type)
+
+			if isNodePublishVolumeTargetPath(m.Source, kataDirectVolumesDir) {
+				req.OCI.Annotations[volumeTargetPathKey] = m.Source
+			}
+		}
+	}
+	if len(req.OCI.Annotations) > 0 {
+		logger.Print("    annotations:")
+		for k, v := range req.OCI.Annotations {
+			logger.Printf("        %s: %s", k, v)
 		}
 	}
 	if len(req.Storages) > 0 {
@@ -167,6 +179,17 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 	}
 
 	return res, err
+}
+
+func isNodePublishVolumeTargetPath(volumePath, directVolumesDir string) bool {
+	if !strings.Contains(filepath.Clean(volumePath), "/volumes/"+csiPluginEscapeQualifiedName+"/") {
+		return false
+	}
+
+	volumeDir := filepath.Join(directVolumesDir, b64.URLEncoding.EncodeToString([]byte(volumePath)))
+	_, err := os.Stat(volumeDir)
+
+	return err == nil
 }
 
 func (s *proxyService) StartContainer(ctx context.Context, req *pb.StartContainerRequest) (*types.Empty, error) {
--- a/pkg/forwarder/interceptor/interceptor.go
+++ b/pkg/forwarder/interceptor/interceptor.go
@@ -14,9 +14,16 @@ import (
 	"github.com/avast/retry-go/v4"
 	"github.com/gogo/protobuf/types"
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
+	"github.com/moby/sys/mountinfo"
 	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/util/agentproto"
+)
+
+const (
+	volumeTargetPathKey = "io.confidentialcontainers.org.peerpodvolumes.target_path"
+	volumeCheckInterval = 5 * time.Second
+	volumeCheckTimeout  = 3 * time.Minute
 )
 
 var logger = log.New(log.Writer(), "[forwarder/interceptor] ", log.LstdFlags|log.Lmsgprefix)
@@ -87,12 +94,21 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 		logger.Printf("    %s: %q", ns.Type, ns.Path)
 	}
 
+	volumeTargetPath := req.OCI.Annotations[volumeTargetPathKey]
 	if len(req.OCI.Mounts) > 0 {
 		for _, m := range req.OCI.Mounts {
 			if _, err := os.Stat(m.Source); os.IsNotExist(err) && m.Type == "bind" {
 				logger.Printf("mount source %s doesn't exist, try to create", m.Source)
 				if err = os.MkdirAll(m.Source, os.ModePerm); err != nil {
 					logger.Printf("Failed to create dir: %v", err)
+				}
+			}
+
+			if isTargetPath(m.Source, volumeTargetPath) {
+				logger.Printf("Waiting for device mounted to: %s", m.Source)
+				err := waitForDeviceMounted(ctx, m.Source)
+				if err != nil {
+					return nil, err
 				}
 			}
 		}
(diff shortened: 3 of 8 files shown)


pkg/adaptor/proxy/service.go
```diff
@@ -94,16 +102,20 @@ func (s *proxyService) getImageName(annotations map[string]string) (string, erro
...
-			logger.Printf("        %s: %s", k, v)
-		}
-	}
 	if len(req.OCI.Mounts) > 0 {
 		logger.Print("    mounts:")
 		for _, m := range req.OCI.Mounts {
 			logger.Printf("        destination:%s source:%s type:%s", m.Destination, m.Source, m.Type)
+
+			if isNodePublishVolumeTargetPath(m.Source, kataDirectVolumesDir) {
+				req.OCI.Annotations[volumeTargetPathKey] = m.Source
```
huoqifeng, line 111: What if multiple mount path in a single container?

huoqifeng (commented):
Thanks @genjuro214 figured out a solution for the sequence issue! Added a comments in the PR.

pkg/adaptor/proxy/service.go
genjuro214: Multiple mount paths are not considered in this PR. We can figure out how to support multiple mount paths in new issues, I think.

huoqifeng: Looks not too complex to support multiple mounts? How about set `req.OCI.Annotations[volumeTargetPathKey]` to a strings array or a string separated by `,` to represents a string array?

huoqifeng: And I agree we can take one mount path in one containers as a start if it's too complex to support multiple mounts.

install/overlays/ibmcloud/kata_direct_volumes_mount.yaml
```diff
@@ -0,0 +1,23 @@
...
+        volumeMounts:
+        - mountPath: /run/kata-containers/shared/direct-volumes
+          name: kata-direct-volumes-dir
+      volumes:
+      - name: kata-direct-volumes-dir
+        hostPath:
+          path: /run/kata-containers/shared/direct-volumes
+          type: DirectoryOrCreate
+
+# to apply this uncomment the patchesStrategicMerge of this file in kustomization.yaml
```
liudalibj, line 22: In kustomization.yaml the `- kata_direct_volumes_mount.yaml` already uncommented.
So is this line still need?

liudalibj: maybe comment it in kustomization.yaml is better.

pkg/forwarder/interceptor/interceptor.go
```diff
@@ -107,6 +120,54 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 	return res, err
 }
 
+func isTargetPath(path, targetPath string) bool {
+	return targetPath != "" && targetPath == path
+}
+
+func waitForDeviceMounted(path string) error {
+	checkInterval := 5 * time.Second
```
liudalibj, line 128: make `checkInterval` and `timeout` as const? after L24, maybe
```
const (
	volumeTargetPathKey = "io.confidentialcontainers.org.peerpodvolumes.target_path"
	volumeCheckInterval := 5 * time.Second
	volumeTimeout := 2 * time.Minute
)
```

genjuro214:
Addressed @liudalibj 's comments above

huoqifeng:
@bpradipt @yoheiueda do you have any comments on this?

liudalibj (approved):
LGTM thanks @genjuro214

genjuro214:
@huoqifeng , created a new issue #1077 to track the multiple pod volumes support.

pkg/adaptor/proxy/service.go
```diff
@@ -94,16 +102,20 @@ func (s *proxyService) getImageName(annotations map[string]string) (string, erro
 func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (*types.Empty, error) {
 
 	logger.Printf("CreateContainer: containerID:%s", req.ContainerId)
-	if len(req.OCI.Annotations) > 0 {
-		logger.Print("    annotations:")
-		for k, v := range req.OCI.Annotations {
-			logger.Printf("        %s: %s", k, v)
-		}
-	}
```
bpradipt, line 102: This change looks unrelated to the PR.

bpradipt:
I have a generic question. Can the changes to common code (interceptor, service etc) for csi-wrapper be included under a feature flag? This will make new changes easier for alternative approaches.

pkg/adaptor/proxy/service.go
genjuro214: Just adjust the print order. Because we may add new annotation when going through OCI Mounts, it's better print annotations after that

pkg/forwarder/interceptor/interceptor.go
```diff
@@ -95,6 +104,14 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 					logger.Printf("Failed to create dir: %v", err)
 				}
 			}
+
+			if isTargetPath(m.Source, volumeTargetPath) {
+				logger.Printf("Waiting for device mounted to: %s", m.Source)
+				err := waitForDeviceMounted(m.Source)
```
yoheiueda, line 110: I think `waitForDeviceMounted` should accept `ctx` as an argument, and should cancel waiting when `ctx` is cancelled.

pkg/forwarder/interceptor/interceptor.go
```diff
@@ -107,6 +124,51 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
...
 
+func isTargetPath(path, targetPath string) bool {
+	return targetPath != "" && targetPath == path
+}
+
+func waitForDeviceMounted(path string) error {
+	timer := time.NewTimer(volumeTimeout)
+
+	for {
+		isMounted := isMounted(execFindmnt, path)
```
yoheiueda, line 135: How about utilizing this existing library?

https://pkg.go.dev/github.com/moby/sys/mountinfo

pkg/forwarder/interceptor/interceptor.go
```diff
@@ -107,6 +124,51 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 	return res, err
 }
 
+func isTargetPath(path, targetPath string) bool {
+	return targetPath != "" && targetPath == path
+}
+
+func waitForDeviceMounted(path string) error {
+	timer := time.NewTimer(volumeTimeout)
```
yoheiueda, line 132: We are using the `retry` package to simplify this kind of retry logic in other places.

https://github.com/confidential-containers/cloud-api-adaptor/blob/a771f96274c8b98f9f450b9477ad0572902ae94e/pkg/adaptor/proxy/proxy.go#L121-L130

pkg/adaptor/proxy/service.go
bpradipt: Ok

pkg/forwarder/interceptor/interceptor.go
genjuro214: Great, I'll have a try.

genjuro214:
Updated the PR according to @yoheiueda 's comments above.

genjuro214:
@yoheiueda Do you have any further comments regarding the new changes?

pkg/adaptor/proxy/service.go
```diff
@@ -169,6 +181,17 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 	return res, err
 }
 
+func isNodePublishVolumeTargetPath(volumePath, directVolumesDir string) bool {
+	if !strings.Contains(volumePath, csiPluginEscapeQualifiedName) {
```
yoheiueda, line 185: Can the following code work?

```
if filepath.Base(volumepath) != csiPluginEscapeQualifiedName {
```

yoheiueda:
@liudalibj @huoqifeng @genjuro214 

I guess this PR is necessary, since CSI wrapper component runs as a sidecar container in a pod. Can we remove this kind of workaround, if we run the CSI wrapper componenet as a systemd service?

huoqifeng:
> @liudalibj @huoqifeng @genjuro214
> 
> I guess this PR is necessary, since CSI wrapper component runs as a sidecar container in a pod. Can we remove this kind of workaround, if we run the CSI wrapper componenet as a systemd service?

There is other problem when run it as systemd service, e.g can not talk to api-server because there is no token available if the code not running in a container. @yoheiueda

yoheiueda:
@huoqifeng 

I see. I think it is OK for now. I think we chose the design as the first prototype. In general, volumes should be ready before containers in a pod start running. I think we may need to rethink the design sometime later.

pkg/adaptor/proxy/service.go
genjuro214: No. volumepath value may looks like: `/var/lib/kubelet/pods/cc962e00-5575-4ca4-a651-afb5e1ef5d90/volumes/kubernetes.io~csi/pvc-c07c4faf-790f-4c1b-b626-6202937967f0/mount`, csiPluginEscapeQualifiedName value is `kubernetes.io~csi`

huoqifeng:
@yoheiueda are you happy with the approach and the changes made against your comments?

pkg/adaptor/proxy/service.go
yoheiueda: Thank you for the explanation.

So, how about this one?  This causes less false-positives.

```
if !strings.Contains(filepath.Clean(path), "/volumes/"+csiPluginEscapeQualifiedName+"/") {
```

genjuro214: @yoheiueda , sure. Made the changes and add test code

yoheiueda (approved):
LGTM. Thank you very much!

PR #1074, opened 2023-06-12
ibmcloud: Automatic vpc parameters

jtumber-ibm:
By only setting the following values I could deploy a peer pod on an ibmcloud managed cluster (IKS).

```yaml
configMapGenerator:
- name: peer-pods-cm
  literals:
  - CLOUD_PROVIDER="ibmcloud"
  - IBMCLOUD_PODVM_IMAGE_ID="r018-943c68bb-0741-485c-8256-78299593b9eb" #set
  - IBMCLOUD_PODVM_INSTANCE_PROFILE_NAME="bz2-2x8" #set
```
&
```yaml
secretGenerator:
- name: peer-pods-secret
  literals:
  - IBMCLOUD_IAM_PROFILE_ID="<profile_id>" # set
```

### DIFF
--- a/pkg/adaptor/cloud/ibmcloud/provider.go
+++ b/pkg/adaptor/cloud/ibmcloud/provider.go
@@ -85,16 +85,18 @@ func (p *ibmcloudVPCProvider) getInstancePrototype(instanceName, userData string
 		UserData: &userData,
 		Profile:  &vpcv1.InstanceProfileIdentity{Name: &p.serviceConfig.ProfileName},
 		Zone:     &vpcv1.ZoneIdentity{Name: &p.serviceConfig.ZoneName},
-		Keys: []vpcv1.KeyIdentityIntf{
-			&vpcv1.KeyIdentity{ID: &p.serviceConfig.KeyID},
-		},
-		VPC: &vpcv1.VPCIdentity{ID: &p.serviceConfig.VpcID},
+		Keys:     []vpcv1.KeyIdentityIntf{},
+		VPC:      &vpcv1.VPCIdentity{ID: &p.serviceConfig.VpcID},
 		PrimaryNetworkInterface: &vpcv1.NetworkInterfacePrototype{
 			Subnet: &vpcv1.SubnetIdentity{ID: &p.serviceConfig.PrimarySubnetID},
 			SecurityGroups: []vpcv1.SecurityGroupIdentityIntf{
 				&vpcv1.SecurityGroupIdentityByID{ID: &p.serviceConfig.PrimarySecurityGroupID},
 			},
 		},
+	}
+
+	if p.serviceConfig.KeyID != "" {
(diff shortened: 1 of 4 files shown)


stevenhorsman:
Thanks for this change @jtumber-ibm - it looks really handy - I wonder if it's worth updating the e2e logic for the ibmcloud provisioner to reduce the number of fields that are mandatory? I just need to decide that that might negatively effect the self-managed approach too much?

jtumber-ibm:
> Thanks for this change @jtumber-ibm - it looks really handy - I wonder if it's worth updating the e2e logic for the ibmcloud provisioner to reduce the number of fields that are mandatory? I just need to decide that that might negatively effect the self-managed approach too much?

We could add in the node labels to the selfmanaged cluster so the same "discovery" is possible?

stevenhorsman:
> We could add in the node labels to the selfmanaged cluster so the same "discovery" is possible?

Yeah, that sounds like a nice way to resolve it given that we control the self-managed cluster set-up

stevenhorsman:
> > > We could add in the node labels to the selfmanaged cluster so the same "discovery" is possible?
> > 
> > 
> > Yeah, that sounds like a nice way to resolve it given that we control the self-managed cluster set-up
> 
> I'll update the self-managed cluster with IKS-like labels and the provisioner-cli in a separate PR. Assuming this gets merged

Thanks. BTW the code looks good to me here, but I'm struggling with time at the moment in order to build and test this before giving my green tick. Sorry for the delay - too many threads means my time slicing is getting small!

stevenhorsman (approved):
LGTM. Thanks JT for the PR and Jordan for verifying and testing!

pkg/adaptor/cloud/ibmcloud/provider.go
```diff
@@ -60,6 +60,23 @@ func NewProvider(config *Config) (cloud.Provider, error) {
 		return nil, fmt.Errorf("either an IAM API Key or Profile ID needs to be set")
 	}
 
+	nodeName, ok := os.LookupEnv("NODE_NAME")
+	var nodeLabels map[string]string
+	if ok {
+		var err error
+		nodeLabels, err = k8sops.NodeLabels(context.TODO(), nodeName)
```
yoheiueda, line 67: The necessity of `context.TODO()` here probably indicates that the `cloud.Provider` interface is not properly  defined.

https://github.com/confidential-containers/cloud-api-adaptor/blob/a771f96274c8b98f9f450b9477ad0572902ae94e/pkg/adaptor/cloud/cloud.go#L36-L40

I will raise an PR to refactor the interface after this PR gets merged. Probably, we need the methods below, and initialization code will be moved to `Setup`.
```
 	Setup(ctx context.Context) error 
 	Teardown(ctx context.Context) error 
```

yoheiueda (approved):
LGTM. Thank you very much for this great enhancement!

I found an opportunity for future refactoring, but I think this PR can be merged as it is.

PR #1088, opened 2023-06-16
Add support for flexible instance types

bpradipt:
This is the initial version implemented only for AWS to kickstart discussion on the approach.

### DIFF
--- a/pkg/adaptor/cloud/aws/provider.go
+++ b/pkg/adaptor/cloud/aws/provider.go
@@ -36,9 +36,13 @@ type ec2Client interface {
 	CreateTags(ctx context.Context,
 		params *ec2.CreateTagsInput,
 		optFns ...func(*ec2.Options)) (*ec2.CreateTagsOutput, error)
+	// Add DescribeInstanceTypes method
+	DescribeInstanceTypes(ctx context.Context,
+		params *ec2.DescribeInstanceTypesInput,
+		optFns ...func(*ec2.Options)) (*ec2.DescribeInstanceTypesOutput, error)
 }
-
 type awsProvider struct {
+	// Make ec2Client a mockable interface
 	ec2Client     ec2Client
 	serviceConfig *Config
 }
@@ -59,6 +63,10 @@ func NewProvider(config *Config) (cloud.Provider, error) {
 	provider := &awsProvider{
 		ec2Client:     ec2Client,
 		serviceConfig: config,
+	}
+
+	if err = provider.updateInstanceTypeSpecList(); err != nil {
+		return nil, err
 	}
 
 	return provider, nil
@@ -86,7 +94,7 @@ func getIPs(instance types.Instance) ([]net.IP, error) {
 	return podNodeIPs, nil
 }
 
-func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator) (*cloud.Instance, error) {
+func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator, spec cloud.InstanceTypeSpec) (*cloud.Instance, error) {
 
 	instanceName := util.GenerateInstanceName(podName, sandboxID, maxInstanceNameLen)
 
@@ -97,6 +105,11 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 
 	//Convert userData to base64
 	userDataEnc := base64.StdEncoding.EncodeToString([]byte(userData))
+
+	instanceType, err := p.selectInstanceType(ctx, spec)
+	if err != nil {
+		return nil, err
+	}
 
 	var input *ec2.RunInstancesInput
 
@@ -114,7 +127,7 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 			MinCount:         aws.Int32(1),
 			MaxCount:         aws.Int32(1),
 			ImageId:          aws.String(p.serviceConfig.ImageId),
-			InstanceType:     types.InstanceType(p.serviceConfig.InstanceType),
+			InstanceType:     types.InstanceType(instanceType),
 			SecurityGroupIds: p.serviceConfig.SecurityGroupIds,
 			SubnetId:         aws.String(p.serviceConfig.SubnetId),
 			UserData:         &userDataEnc,
@@ -185,4 +198,64 @@ func (p *awsProvider) DeleteInstance(ctx context.Context, instanceID string) err
 
 func (p *awsProvider) Teardown() error {
 	return nil
+}
+
+// Add SelectInstanceType method to select an instance type based on the memory and vcpu requirements
+func (p *awsProvider) selectInstanceType(ctx context.Context, spec cloud.InstanceTypeSpec) (string, error) {
+
+	return cloud.SelectInstanceTypeToUse(spec, p.serviceConfig.InstanceTypeSpecList, p.serviceConfig.InstanceTypes, p.serviceConfig.InstanceType)
+}
+
+// Add a method to populate updateMachineTypeList for all the instanceTypes
+func (p *awsProvider) updateInstanceTypeSpecList() error {
+
+	// Get the instance types from the service config
+	instanceTypes := p.serviceConfig.InstanceTypes
+
+	// If instanceTypes is empty then populate it with the default instance type
+	if len(instanceTypes) == 0 {
+		instanceTypes = append(instanceTypes, p.serviceConfig.InstanceType)
+	}
+
+	// Create a list of instancetypespec
+	var instanceTypeSpecList []cloud.InstanceTypeSpec
+
+	// Iterate over the instance types and populate the instanceTypesTupleList
+	for _, instanceType := range instanceTypes {
+		vcpus, memory, err := p.getInstanceTypeInformation(instanceType)
+		if err != nil {
+			return err
+		}
+		instanceTypeSpecList = append(instanceTypeSpecList, cloud.InstanceTypeSpec{InstanceType: instanceType, VCPUs: vcpus, Memory: memory})
+	}
+
+	// Sort the instanceTypesTupleList by Memory and update the serviceConfig
+	p.serviceConfig.InstanceTypeSpecList = cloud.SortInstanceTypesOnMemory(instanceTypeSpecList)
+	logger.Printf("InstanceTypeSpecList (%v)", p.serviceConfig.InstanceTypeSpecList)
+	return nil
+}
+
+// Add a method to retrieve cpu, memory, and storage from the instance type
+func (p *awsProvider) getInstanceTypeInformation(instanceType string) (vcpu int64, memory int64, err error) {
+
+	// Get the instance type information from the instance type using AWS API
+	input := &ec2.DescribeInstanceTypesInput{
+		InstanceTypes: []types.InstanceType{
+			types.InstanceType(instanceType),
+		},
+	}
+	// Get the instance type information from the instance type using AWS API
+	result, err := p.ec2Client.DescribeInstanceTypes(context.Background(), input)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	// Get the vcpu and memory from the result
+	if len(result.InstanceTypes) > 0 {
+		vcpu = int64(*result.InstanceTypes[0].VCpuInfo.DefaultVCpus)
+		memory = int64(*result.InstanceTypes[0].MemoryInfo.SizeInMiB)
+		return vcpu, memory, nil
+	}
+	return 0, 0, fmt.Errorf("instance type %s not found", instanceType)
+
 }
--- a/pkg/adaptor/cloud/aws/provider_test.go
+++ b/pkg/adaptor/cloud/aws/provider_test.go
@@ -5,6 +5,7 @@ package aws
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"reflect"
 	"testing"
@@ -69,6 +70,36 @@ func (m mockEC2Client) TerminateInstances(ctx context.Context,
 	return &ec2.TerminateInstancesOutput{}, nil
 }
 
+// Create a mock EC2 DescribeInstanceTypes method
+func (m mockEC2Client) DescribeInstanceTypes(ctx context.Context,
+	params *ec2.DescribeInstanceTypesInput,
+	optFns ...func(*ec2.Options)) (*ec2.DescribeInstanceTypesOutput, error) {
+
+	// Take instance type from params
+	instanceType := params.InstanceTypes[0]
+	// Check if instance type is t2.medium, else return an error
+	if instanceType != "t2.medium" {
+		return nil, fmt.Errorf("Unsupported instance type")
+	}
+
+	// Return a mock DescribeInstanceTypesOutput
+	return &ec2.DescribeInstanceTypesOutput{
+		InstanceTypes: []types.InstanceTypeInfo{
+			{
+				InstanceType: instanceType,
+				// Add vCPU info for t2.medium
+				VCpuInfo: &types.VCpuInfo{
+					DefaultVCpus: aws.Int32(2),
+				},
+				// Add memory info for t2.medium
+				MemoryInfo: &types.MemoryInfo{
+					SizeInMiB: aws.Int64(4096),
+				},
+			},
+		},
+	}, nil
+}
+
 // Create a serviceConfig struct without public IP
 var serviceConfig = &Config{
 	Region: "us-east-1",
@@ -80,6 +111,38 @@ var serviceConfig = &Config{
 	SecurityGroupIds: []string{"sg-1234567890abcdef0"},
 	// Add image ID to serviceConfig
 	ImageId: "ami-1234567890abcdef0",
+	// Add InstanceTypes to serviceConfig
+	InstanceTypes: []string{"t2.small", "t2.medium"},
+}
+
+// Create a serviceConfig struct with invalid instance type
+var serviceConfigInvalidInstanceType = &Config{
+	Region: "us-east-1",
+	// Add instance type to serviceConfig
+	InstanceType: "t2.small",
+	// Add subnet ID to serviceConfig
+	SubnetId: "subnet-1234567890abcdef0",
+	// Add security group ID to serviceConfig
+	SecurityGroupIds: []string{"sg-1234567890abcdef0"},
+	// Add image ID to serviceConfig
+	ImageId: "ami-1234567890abcdef0",
+	// Add InstanceTypes to serviceConfig
+	InstanceTypes: []string{"t2.large", "t2.medium"},
+}
+
+// Create a serviceConfig with emtpy InstanceTypes
+var serviceConfigEmptyInstanceTypes = &Config{
+	Region: "us-east-1",
+	// Add instance type to serviceConfig
+	InstanceType: "t2.small",
+	// Add subnet ID to serviceConfig
+	SubnetId: "subnet-1234567890abcdef0",
+	// Add security group ID to serviceConfig
+	SecurityGroupIds: []string{"sg-1234567890abcdef0"},
+	// Add image ID to serviceConfig
+	ImageId: "ami-1234567890abcdef0",
+	// Add InstanceTypes to serviceConfig
+	InstanceTypes: []string{},
 }
 
 type mockCloudConfig struct{}
@@ -94,11 +157,11 @@ func TestCreateInstance(t *testing.T) {
 		serviceConfig *Config
 	}
 	type args struct {
-		ctx          context.Context
-		podName      string
-		sandboxID    string
-		cloudConfig  cloudinit.CloudConfigGenerator
-		instanceType string
+		ctx         context.Context
+		podName     string
+		sandboxID   string
+		cloudConfig cloudinit.CloudConfigGenerator
+		spec        cloud.InstanceTypeSpec
 	}
 	tests := []struct {
 		name    string
@@ -118,11 +181,11 @@ func TestCreateInstance(t *testing.T) {
 				serviceConfig: serviceConfig,
 			},
 			args: args{
-				ctx:          context.Background(),
-				podName:      "podtest",
-				sandboxID:    "123",
-				cloudConfig:  &mockCloudConfig{},
-				instanceType: "t2.small",
+				ctx:         context.Background(),
+				podName:     "podtest",
+				sandboxID:   "123",
+				cloudConfig: &mockCloudConfig{},
+				spec:        cloud.InstanceTypeSpec{InstanceType: "t2.small"},
 			},
 			want: &cloud.Instance{
 				ID:   "i-1234567890abcdef0",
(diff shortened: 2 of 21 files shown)


bpradipt:
cc @jtumber-ibm @yoheiueda @huoqifeng @mkulke @surajssd @snir911 
kindly take a look. This implements the simplest approach of specifying allowed instance types for podvm in CAA. The required instance type is provided in the pod spec via annotation. If no annotation present then CAA uses the default.

Subsequently, we can make it more generic by providing a tuple for accepted instance types based on @mkulke suggestion here - https://github.com/confidential-containers/cloud-api-adaptor/issues/321#issuecomment-1291843670

Since the webhook removes the resource spec and adds the extended resources, it looks like a logical location to find the best fit instance types based on cpu/mem requests to add the annotation.

pkg/adaptor/cloud/aws/provider.go
```diff
@@ -98,6 +98,25 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
...
+
+	// If instanceTypes is empty and instanceType is not default instance type, return error
+	if len(p.serviceConfig.InstanceTypes) == 0 && instanceType != p.serviceConfig.InstanceType {
+		return nil, fmt.Errorf("instance type %q is not supported.", instanceType)
+	}
+
+	// Check if instanceType is among the supported instance types only if instanceTypes is not empty.
+	if len(p.serviceConfig.InstanceTypes) > 0 {
+		if !util.Contains(p.serviceConfig.InstanceTypes, instanceType) {
+			return nil, fmt.Errorf("instance type %q is not supported", instanceType)
```
jtumber-ibm, line 116: Do we want to be more lenient here and use the default if an incorrect failure is specified? 
If not could/should that error be added to the pods event list so it is clearer to the end user why provisioning failed?

bpradipt: @jtumber-ibm good idea. I think being lenient will be more helpful along with a warning ?

snir911:
LGTM overall, do you think it will be possible to utilize this approach to handle also the issue where instance type isn't available  for the region (and fallback to the next type for example)?

pkg/adaptor/cloud/aws/provider_test.go
```diff
@@ -132,7 +164,111 @@ func TestCreateInstance(t *testing.T) {
...
+			want: &cloud.Instance{
+				ID:   "i-1234567890abcdef0",
+				Name: "podvm-podinvalidinstance-123",
+				IPs:  []net.IP{net.ParseIP("10.0.0.2")},
+			},
+			// Test should not return an error
+			wantErr: false,
+		},
+		// Test creating an instance with empty InstanceTypes
+		// The instance type is not set to default value
```
mkulke, line 193: Is this assertion actually tested? the tests just verifies that creation doesn't fail, if I understand it correctly

pkg/adaptor/cloud/aws/provider.go
```diff
@@ -98,6 +98,22 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
...
 	userDataEnc := base64.StdEncoding.EncodeToString([]byte(userData))
 
+	logger.Printf("instance type requested %q", instanceType)
+
+	// If instanceType is empty, set instanctType to default. Non-empty instanceType is set via annotation
+	// Or
+	// If instanceTypes is empty and instanceType is not default, set instanceType to default
+	// Or
+	// If instanceTypes is not empty and instanceType is not among the supported instance types, set instanceType to default
+	if instanceType == "" ||
```
mkulke, line 108: I would factor this out into a function and add a unit test to spec the desired behaviour

mkulke (commented):
I was also wondering about the (temporary) unavailability of an instance-type/region combination (e.g. confidential instances could be in high demand and not available in a given region temporarily). Maybe CAA should not be smart about finding a proper fallback, but propagate parseable errors (`requested: x, configured: [x, y]`, `requested: z, configured:  [x, y]`)  so that higher level components/users (i.e. those which manage the annotation) can take  decisions.

We can maybe also just simplify the picking logic in the first iteration:
- If specified, it needs to be part of the configured list, otherwise there is an error
- If unspecified, we pick the first entry in the list
- We do not deal with fallbacks at this point

pkg/adaptor/cloud/aws/provider_test.go
bpradipt: Right. It just verifies the creation and the related conditions. It's not actual creation of instance

pkg/adaptor/cloud/aws/provider.go
bpradipt: How about if I make this a generic utility method that can be called by any provider?

bpradipt:
Forgot to remove the webhook changes and move it into a separate PR.  I'll do that in the next iteration.

pkg/util/cloud_test.go
```diff
@@ -0,0 +1,74 @@
...
+		{
+			name: "instanceType=, instanceTypes=[], defaultInstanceType=t2.medium",
+			args: args{
+				instanceType:        "",
+				instanceTypes:       []string{},
+				defaultInstanceType: "t2.medium",
+			},
+			want: "t2.medium",
+		},
+		// Add test case with instanceType="t2.small", instanceTypes=[], defaultInstanceType="t2.medium"
```
mkulke, line 56: Thanks for adding the unit test, that helps to grok the logic. Could we conflate `instanceTypes` and `defaultInstanceType`? e.g. make the order of `instanceTypes` semantic and declare the first entry of `instanceTypes` to be the default?

If we want to have fallback logic in the future (we cannot start a VM in region X b/c of some reason) we could pick the next entry in the list. On the other hand this might be not intended (user requested spot instance, got very pricey GPU instance).

jtumber-ibm:
> > If we do error I think we'd also need to add in the mechanism to add the relevant information to the pod. They shouldn't have to scroll through cloud-api-adaptor logs to find out why their pod didn't provision if it was due to a user-configuration error.
> 
> Do you have any pointers on what needs to be done ? I can take a look and then we can decide the way forward

We could create events against the pod, however I believe we will need the pods `uid` which isn't currently passed in the annotations. But we could add into the remote hyperv, [this is the relevant annotation](https://github.com/containerd/containerd/blob/96de54db4385279a5f8474b5de6658f6336b6fd8/pkg/cri/annotations/annotations.go#LL70C18-L70C18)  I believe (haven't had a chance to test this yet though). 

Anyway this would definitely be more work, which is why I am in favor of defaulting rather than erroring.

mkulke:
> Anyway this would definitely be more work, which is why I am in favor of defaulting rather than erroring.

Maybe we could address those things in parallel, at this stage of the project maybe it's fine for users to grep the daemonset logs? and we can deal with proper propagation of error to k8s pod separately.

We have to keep in mind that those fallbacks can incur significant surprise costs to users. The user maybe entered a typo somewhere and didn't intend a long running batch job workload to be silently executed on costly machines, that are also part of the configured list. Alternatively, we make that behavior very transparent, maybe as a comment in the config file template and suggest users to create different caa runtime-classes (is that possible?) per "class".

bpradipt:
Looks like returning an error and getting it logged in the pod events is simple in our case.  We just need to return the error from CreateInstance and it'll get logged. Check for the sample here

```
Events:
  Type     Reason                  Age                From               Message
  ----     ------                  ----               ----               -------
  Normal   Scheduled               2m18s              default-scheduler  Successfully assigned kata-webserver/echo-kata-b764bbdbb-jk2ll to ocp-4.13-worker-0.testocp.com
  Normal   AddedInterface          93s                multus             Add eth0 [10.135.1.58/23] from ovn-kubernetes
  Normal   AddedInterface          92s                multus             Add eth0 [10.135.1.58/23] from ovn-kubernetes
  Warning  FailedCreatePodSandBox  91s (x2 over 92s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: remote hypervisor call failed: rpc error: code = Unknown desc = creating an instance : Requested instance type ("t3.small") is not available: unknown
```

pkg/util/cloud_test.go
bpradipt: How about handling this in a subsequent PR ? Making this change will require modifying the existing options (and any related changes).  Imho once we have an initial support for flexible instance types across the providers, we can conflate the two options into one.
But I'm open to make this change in this PR. Let me know what you prefer.

bpradipt:
For reference kata runtime PR - https://github.com/kata-containers/kata-containers/pull/7141

pkg/util/cloud_test.go
mkulke: Ah, right `InstanceType` already exists in the config, so we use that as `defaultInstanceType` if I understood it correctly. In this case we probably want to do it seperately, since it breaks existing config maps. I'll create a follow-up issue once this is merged.

mkulke (approved):

pkg/util/cloud.go
```diff
@@ -62,3 +63,43 @@ func GetPodNamespace(annotations map[string]string) string {
...
+
+// Method to check if a string exists in a slice
+func Contains(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
```
yoheiueda, line 80: Can we use the `slices` package? This package will be a part of the standard Go library in Go 1.21.
https://pkg.go.dev/golang.org/x/exp/slices#Contains

jtumber-ibm:
> Looks like returning an error and getting it logged in the pod events is simple in our case. We just need to return the error from CreateInstance and it'll get logged. Check for the sample here
> 
> ```
> Events:
>   Type     Reason                  Age                From               Message
>   ----     ------                  ----               ----               -------
>   Normal   Scheduled               2m18s              default-scheduler  Successfully assigned kata-webserver/echo-kata-b764bbdbb-jk2ll to ocp-4.13-worker-0.testocp.com
>   Normal   AddedInterface          93s                multus             Add eth0 [10.135.1.58/23] from ovn-kubernetes
>   Normal   AddedInterface          92s                multus             Add eth0 [10.135.1.58/23] from ovn-kubernetes
>   Warning  FailedCreatePodSandBox  91s (x2 over 92s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: remote hypervisor call failed: rpc error: code = Unknown desc = creating an instance : Requested instance type ("t3.small") is not available: unknown
> ```

Oh awesome, didn't know that would work

pkg/adaptor/cloud/cloud.go
```diff
@@ -34,7 +34,7 @@ const (
 var logger = log.New(log.Writer(), "[adaptor/cloud] ", log.LstdFlags|log.Lmsgprefix)
 
 type Provider interface {
-	CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator) (instance *Instance, err error)
+	CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator, instanceType string) (instance *Instance, err error)
```
yoheiueda, line 37: I think it's OK for the first implementation, but I think we need a mechanism to select an instance type based on required vcpus and memory size.

How about adding a new method like `SelectInstanceType` in the `Provider` interface, instead of adding a new argument to `CreateInstance`?

```
type Provider interface {
	CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator) (instance *Instance, err error)
	DeleteInstance(ctx context.Context, instanceID string) error
	Teardown() error
	SelectInstanceType(vcpus, memory int) string
}
```

bpradipt: Few thoughts on using vcpus and memory size to decide on the instance type:
1. If we want to use cpu/mem request/limit data from resource attribute in the pod spec then we need to update the webhook to not remove resource attribute from the spec.  
2. If we want to use cpu/mem annotation, then we will need to pass the required annotations to CAA and figure out if we need to reconcile with any cpu/mem resource request/limit in the spec. Most likely we can have the same logic as with Kata local hypervisor but without the hotplugging stuff.
3. We'll need to work out a best fit function which will return the closest instance type match from the configured list based on the cpu/mem request. This was something which @mkulke had suggested initially

@yoheiueda I'm ok to rewrite this PR and incorporate the additional interface if it avoids immediate rewrite later on. Let me know wdyt?

@mkulke @snir911 @jtumber-ibm since you have acked this PR, wanted to hear your thoughts as well.

bpradipt: I think both the approaches will be needed - instance type and cpu/mem will be helpful to the end user. Personally I find it hard to remember the cpu/mem of the instance types.

mkulke: I wonder whether we can retrieve the vCPU + memory characteristics from provider APIs, so when we process the ConfigMap initially, the provided instances types are enriched with this data. So we don't need the user to provide this, and we can sort out invalid instance types.

I wouldn't go beyond some verification check like "will pod fit into the given instance type" and minimal logic "if no instance type is given, pick the smallest possible from the configured available instance types" (that decision will probably be a bit arbitrary).

bpradipt: @mkulke it's possible to get the vcpu and memory characteristics.
Also I was discussing with @yoheiueda on updating the CreateInstance method signature. It'll be needed since the provider apis need the instance type for VM creation. 

I'll update this PR to also include vcpu/mem annotation handling for instance type selection which we can keep improving going forward

jtumber-ibm: Could we also add architecture into the specifications for choosing a instance type? Do we then also need to include confidential VM selection too? Also, gpus might be important use-case in the future?

bpradipt: You can use the instance type annotation to select gpu or any other instance without using cpu/mem

bpradipt:
I have incorporated most of the feedback (I think :-).  Here is a summary of the changes in this iteration
- Handles the following annotations in the pod spec `io.katacontainers.config.hypervisor.machine_type` `io.katacontainers.config.hypervisor.default_vcpus` and `io.katacontainers.config.hypervisor.default_memory`
- If both `io.katacontainers.config.hypervisor.default_vcpus` and `io.katacontainers.config.hypervisor.default_memory` annotations are present, it's given preference in selecting the instance type
- The best fit instance type is selected from a pre-configured list (which is passed as an option to CAA)
- The cpu/mem details of the instance type from the pre-configured list is automatically populated
- Updated unit tests as relevant

@jtumber-ibm @mkulke @snir911 @yoheiueda ptal

pkg/adaptor/cloud/aws/provider.go
```diff
@@ -186,3 +203,63 @@ func (p *awsProvider) DeleteInstance(ctx context.Context, instanceID string) err
...
+	if len(instanceTypes) == 0 {
+		instanceTypes = append(instanceTypes, p.serviceConfig.InstanceType)
+	}
+
+	// Create a list of instance type tuples
+	var instanceTypesTupleList []cloud.InstanceTypeTuple
+
+	// Iterate over the instance types and populate the instanceTypesTupleList
+	for _, instanceType := range instanceTypes {
+		vcpu, memory, err := p.getInstanceTypeInformation(instanceType)
```
jtumber-ibm, line 229: This could error because an specified instance-type isn't available in the region, in that situation would it be preferable to continue/skip that instance type rather return and not return a valid provider at all?

bpradipt: Or we assume that admin is correct and returning the error will force the admin to provide the correct instance types ?

bpradipt: @jtumber-ibm I am tending to favour keeping the code as-is, ie returning an error if there is a failure in populating the details of the instance type . The error gives an indication to the admin of what is wrong with the input instance types (like unavailability, typo etc)
And since the error will be during the start of CAA, it'll be a good indication of a wrong config as well.

pkg/adaptor/cloud/cloud.go
```diff
@@ -34,9 +35,11 @@ const (
...
 	DeleteInstance(ctx context.Context, instanceID string) error
 	Teardown() error
+	// Add SelectInstanceType method based on vCPU and memory to Provider interface
+	SelectInstanceType(ctx context.Context, vCPU int64, memory int64) (instanceType string, err error)
```
yoheiueda, line 42: I suggested to add the  `SelectInstanceType` method, but it turns out that adding the method seems the design complicated.

I think we can directly pass to required instance size information to the `CreateInstance` as follows, instead of adding the `instanceType` argument.

```
CreateInstance(ctx context.Context, podName, sandboxID string, cloudConfig cloudinit.CloudConfigGenerator, requirement InstanceSpec) (instance *Instance, err error)
```

```
type InstanceSpec struct {
	VCPUs       int
	Memory      int
	MachineType string
}
```

@bpradipt What do you think?

pkg/adaptor/cloud/cloud.go
```diff
@@ -75,6 +78,14 @@ type sandbox struct {
...
 	instanceID   string
 	netNSPath    string
+	instanceType string
+}
+
+type InstanceTypeTuple struct {
+	InstanceType string
+	Vcpu         int64
+	// In MiB
+	Memory int64
```
yoheiueda, line 88: We can possiblyuse the `InstanceSpec` type here like this.

```
type InstanceType struct {
	Name string
	Spec InstanceSpec
}
```

yoheiueda:
Do we need to update containerd configuration? I think all pod annotations are currently ignored for the remote hypervisor due to this PR.
https://github.com/kata-containers/kata-containers/pull/6970

pkg/adaptor/cloud/cloud.go
bpradipt: I thought about it and felt SelectInstanceType method is helpful. This will give flexibility to implement instance selection logic per provider as needed. 
Now what I can do is passing the entire InstanceTuple or InstanceSpec to the method instead of just cpu/mem which will provide flexibility w.r.to customising the method as needed.

pkg/adaptor/cloud/cloud.go
bpradipt: The naming is an issue though. Some cloud providers (eg. aws) use InstanceType as the name/flavour and hence I kept instancetype as the name and InstanceTypeTuple as the detailed spec

How about using MachineType and MachineTypeSpec to be generic ?
```
type MachineType struct {
     Name string
     Spec MachineTypeSpec
}

type MachineTypeSpec struct {
      Name string
      Vcpus int64
      Memory int64
}
```

pkg/adaptor/cloud/cloud.go
yoheiueda: I think implementing an instance selection logic per provider is possible with the approach I mentioned above by including `InstanceType` string in the struct.

```
type InstanceSpec struct {
	InstanceType string
	VCPUs        int
	Memory       int
	MachineType  string
}
```

A provider can prepare a private function `selectInstanceType` instead of a public one, and call the function at the beginning of `CreateInstance` like this.

```
instanceType := requirement.InstanceType
if instanceType == "" {
	instanceTypeName = selectInstanceType(requirement.VCPUs, requirement.Memory)
}
```



Another benefit of passing a requirement as a struct data is that we can add additional filed in the future without changing the method signature like this.

```
type InstanceSpec struct {
	InstanceType string
	VCPUs       int
	Memory      int
	MachineType string
	GPUs        int
}
```

pkg/adaptor/cloud/cloud.go
yoheiueda: Could you show some example data for each struct? I thought machine type is an architecture string like x86_64, s390x, ..., but it looks like something else.

bpradipt: I used MachineType to be in sync with the kata annotation of machine_type which we are using for instance type name. For qemu it's the qemu machine type like q35, microvm etc. For remote hypervisor it becomes the instance types (names like t2.small etc). We can use arch to indicate the architecture.

bpradipt:
> Do we need to update containerd configuration? I think all pod annotations are currently ignored for the remote hypervisor due to this PR. [kata-containers/kata-containers#6970](https://github.com/kata-containers/kata-containers/pull/6970)

It might be needed. We can do it as a separate PR

pkg/adaptor/cloud/cloud.go
bpradipt: As I already mentioned, I agree with passing of struct for flexibility and to avoid method signature change again.
Ok I'll remove the public method. Any other comments before I make the changes to limit rework ?

pkg/adaptor/cloud/cloud.go
bpradipt: Example data
```
MachineType {
   Name: "t2.small",
   Spec: MachineTypeSpec{
        Name:   "t2.small",
        Vcpus:  1,
        Memory: 2048,
    },
}
```

bpradipt: @yoheiueda does this make sense, or you prefer InstanceType?

bpradipt: On rethinking, I do feel machinetype can be confusing with arch. 
So may be we can use InstanceTypeSpec 
```
type InstanceTypeSpec struct {
	InstanceType string
	VCPUs     int
	Memory   int
	Arch         string
	GPUs       int
}
```

pkg/adaptor/cloud/cloud.go
yoheiueda: Thank you!  I have no other comment for now.

pkg/adaptor/cloud/cloud.go
yoheiueda: It looks good to me.

pkg/adaptor/cloud/cloud.go
bpradipt: Thanks @yoheiueda I updated the code based on your comments. Kindly take a look

pkg/adaptor/cloud/util.go
```diff
@@ -18,3 +24,89 @@ func DefaultToEnv(field *string, env, fallback string) {
...
+	if spec.VCPUs != 0 && spec.Memory != 0 {
+		instanceType, err = GetBestFitInstanceType(specList, spec.VCPUs, spec.Memory)
+		if err != nil {
+			return "", fmt.Errorf("failed to get instance type based on vCPU and memory annotations: %w", err)
+		}
+		logger.Printf("Instance type selected by the cloud provider based on vCPU and memory annotations: %s", instanceType)
+	} else if spec.InstanceType != "" {
+		instanceType = spec.InstanceType
+		logger.Printf("Instance type selected by the cloud provider based on instance type annotation: %s", instanceType)
+	}
```
yoheiueda, line 81: Do vcpus and memory have a higher precedence than instance type (machine type annotation)?

I thought the instance type annotation has a higher precedence, since it means that a user explicitly sets a specific instance type in a pod definition. I also think that we can make these annotations exclusive options, and causes an error when both annotations are specified.

@bpradipt wdyt?

bpradipt: @yoheiueda yes vcpus and mem have higher precedence than the annotation. I added that as a comment and will also include it in the docs. It's debatable thought. My reasoning has been that if the user is taking the initiative to specify vcpus and memory then it likely indicates a need for best fit and should take precedence. 
I don't think we need to make the options mutually exclusive. Precedence will provide flexibility and ease of use during deployment.

yoheiueda: OK, I see.

yoheiueda (approved):
LGTM. Thank you very much!

Issue #1108, opened 2023-06-22
Add a release checklist


wainersm:
Now that peer pods is fully (or almost?) integrated with CoCo's operator,  the two projects are hooked with respect to release. @stevenhorsman has improved the [CoCo's release checklist](https://github.com/confidential-containers/community/blob/main/.github/ISSUE_TEMPLATE/release-check-list.md) (see https://github.com/confidential-containers/community/pull/97) to include steps and information about the changes needed to release peer pods alongside.  But perhaps we should have our own checklist; which beyond leaving the process better documented, it will allow others to help on the release (our releases falling on @stevenhorsman and @bpradipt 's shoulders, let's share the burden ;)

stevenhorsman:
Yeah - there is an outstanding issue on me to document our release process in here and then work on improving it, so I'll take this issue. Thanks for chasing my Wainer and welcome back!

Issue #1109, opened 2023-06-22
Create version tags of images on release, so users can install stable versions


stevenhorsman:
At the moment, on release we build the podvm container images and tag them as `latest`, but don't create a version tag e.g. `v0.6.0` of these images, so users can't go back to previous release if they want to.

We also don't produce any tags for the webhook and cloud-api-adaptor images, so if changes are made post release for new function that introduces incompatibilities, then users won't be able to easily find the release version.

To fix these I think we should create tags for each release and we can update the `kustomization` overlays to point to the version based release. This would mean people could use the release tag to install the stable version of peer pods.

wainersm:
@stevenhorsman on the 0.7.0 I pinned the cloud-api-adaptor image (https://github.com/confidential-containers/cloud-api-adaptor/pull/1239) but I used the commit sha-1 of the built release candidate image. Here we have chicken-and-eggs problem (again) to automate this: if the `cloud-adaptor:<RELEASE>` image is built at the time we cut the release, how can we have the kustomization overlays updated before tagging? :)

I think we will have a similar problem for the webhook image.

Are the podvm images referenced somewhere in the code? If not, yeah, I need no problem on tagging them at the time we cut the release.

Issue #852, opened 2023-04-21
Add release tags for submodules 


yoheiueda:
A git tag is effective for the main Go module of the repository, and is not effective for sub modules like `peerpod-ctrl`.

Because of this reason, `go get` command for sub modules will fail like this.
```
$ go get github.com/confidential-containers/cloud-api-adaptor/peerpod-ctrl@v0.5.0
go: github.com/confidential-containers/cloud-api-adaptor/peerpod-ctrl@v0.5.0: invalid version: unknown revision peerpod-ctrl/v0.5.0
```

To add release tag for submodules, we need to create a git tag like `peerpod-ctrl/v0.5.0`

https://stackoverflow.com/a/60601402
https://groups.google.com/g/golang-nuts/c/W3e5MTwC7Js

PR #1110, opened 2023-06-22
doc: release doc

stevenhorsman:
Document the release process

Fixes: #1108

### DIFF
--- a/docs/Release-Process.md
+++ b/docs/Release-Process.md
@@ -26,14 +26,14 @@ the tag of the kata-containers release candidate.
 - The `kata-containers/src/runtime` go module that we include in the main `cloud-api-adaptor` [`go.mod`](../go.mod),
 the `peerpod-ctl` [`go.mod`](../peerpod-ctrl/go.mod) and the `csi-wrapper` [`go.mod`](../volumes/csi-wrapper/go.mod).
 This can be done by running
-  ```
-  go get github.com/kata-containers/kata-containers/src/runtime@<release candidate branch e.g. CCv0>
-  go mod tidy
-  ```
+    ```
+    go get github.com/kata-containers/kata-containers/src/runtime@<release candidate branch e.g. CCv0>
+    go mod tidy
+    ```
 in the top-level repo directory, and the `peerpod-ctl` and `volumes/csi-wrapper` directories.
 > **Note:** If there are API changes in the kata-runtime go modules and we need to cloud-api-adaptor to implement,
 then it may be necessary to temporarily get the peerpod-ctrl and csi-wrapper to self-reference the parent folder to
-avoid compilation errors. This can be done by running 
+avoid compilation errors. This can be done by running: 
 > ```
 > go mod edit -replace github.com/confidential-containers/cloud-api-adaptor=../
 > go mod tidy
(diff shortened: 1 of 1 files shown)


stevenhorsman:
@bpradipt - I've finally gotten around to write up the release process, thanks to Wainer's prompting, so if you could review it and let me know which bits I've got wrong! Thanks

docs/Release-Process.md
```diff
@@ -0,0 +1,63 @@
...
+### Post-release
+
+After the release has been cut the `peerpod-ctrl` and `volumes/csi-wrapper` go modules should be updated to remove any local replace references, and be updated to use the release version of the `cloud-api-adaptor` by running:
+```
+go get github.com/confidential-containers/cloud-api-adaptor
+go mod tidy
+```
+from in the `peerpod-ctrl` and `volumes/csi-wrapper` directories.
+
+## Improvements
```
wainersm, line 59: I am wondering if wouldn't be better to cut the release from a release branch, this way we don't need to revert all the changes to go back picking up the latest dependencies after the release. However I don't know how it would create additional problems, for example, having to change some github workflows for CI/release.

wainersm (approved):
Thanks @stevenhorsman for this write up. I had no idea on how complex the release was.

docs/Release-Process.md
stevenhorsman: Yeah, at the moment we effectively just create the release on the main branch and it's frozen forever there with the tag, but there is updates to latest needed. This is pretty similar to the situation with kata, the operator etc where we end up having the released versions baked in for a while post release (and in kata for example we sometimes don't bump image-rs until the next release cycle). Some could argue that this is bad and hides dev incompatibilities, but I guess others could argue that it means we are building on stable versions.

The self-reference with the cloud-api-adaptor and peerpod-ctrl and csi-wrapper is a bit different though, the issue is just that we can't point to a tag that we haven't created yet and the act of updating them, would change the sha. If they were in separate repos it would help the problem, but just defer it to us having more repos to tag and update. It might be interesting to see how this plays out with image-rs, attestation-agent and ocicrypt-rs from now on to see if they have any issues with this.

@yoheiueda - no I recall you saying you had an idea for this?

stevenhorsman:
> Thanks @stevenhorsman for this write up. I had no idea on how complex the release was.

I probably just made it sound difficult. I managed to do it, so it can't be too tough ;)

wainersm:
> > Thanks @stevenhorsman for this write up. I had no idea on how complex the release was.
> 
> I probably just made it sound difficult. I managed to do it, so it can't be too tough ;)

BTW, I'm volunteering myself to pilot the release of 0.7.0. If we two can make it then it is not definitively tough :)

docs/Release-Process.md
wainersm: > Yeah, at the moment we effectively just create the release on the main branch and it's frozen forever there with the tag, but there is updates to latest needed. This is pretty similar to the situation with kata, the operator etc where we end up having the released versions baked in for a while post release (and in kata for example we sometimes don't bump image-rs until the next release cycle). Some could argue that this is bad and hides dev incompatibilities, but I guess others could argue that it means we are building on stable versions.
>

Got it. I thought we were reverting to latest just after the release so I thought releasing from a branch would be better. From a CI point of view, I always advocate to build on stable versions :)

> The self-reference with the cloud-api-adaptor and peerpod-ctrl and csi-wrapper is a bit different though, the issue is just that we can't point to a tag that we haven't created yet and the act of updating them, would change the sha. If they were in separate repos it would help the problem, but just defer it to us having more repos to tag and update. It might be interesting to see how this plays out with image-rs, attestation-agent and ocicrypt-rs from now on to see if they have any issues with this.
> 

Definitively I prefer less repos. I don't know Golang enough to propose a better solution (if any exists). Just hoping @yoheiueda got an good idea :)

> @yoheiueda - no I recall you saying you had an idea for this?

bpradipt (approved):
/lgtm

docs/Release-Process.md
```diff
@@ -0,0 +1,63 @@
+TODO line length to 120
```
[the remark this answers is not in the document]
stevenhorsman: Sorry - I failed to push my last commit that removed this and did the adjustment!

yoheiueda:
@stevenhorsman could you also add some description of the specific steps to create a tag for a release?

As described in https://github.com/confidential-containers/cloud-api-adaptor/issues/852, we need to create a tag per module. For example, we need the tags listed below for the v0.7.0 release.

- v0.7.0
- peerpod-ctrl/v0.7.0
- peerpodconfig-ctrl/v0.7.0
- volumes/csi-wrapper/v0.7.0
- webhook/v0.7.0

Note that these tags do not necessarily point to the same commit. Each tag can point to a different commit.

So, I think we first create and merge a PR to update `go.mod` of the main `cloud-api-adaptor` module, and prepare a tag for it. Then, prepare and merge another PR to update the remaining `go.mod` using the tag for the reference to the main `cloud-api-adaptor` module, and create tags for the sub modules.

The dependency among the modules are as follows. peerpod-ctrl has a circular dependency, and we can't use the new for the third one. The third one is a reference to the PeerPod API for Kubernetes from the main `cloud-api-adaptor` module, so I guess using a old tag for the third one does not cause a problem unless there is API change.

- csi-wrapper  → cloud-api-adaptor
- peerpod-ctrl  → cloud-api-adaptor
- cloud-api-adaptor → peerpod-ctrl

I think this is out of scope of this PR, but one possible solution to break the cyclic dependency is to introduce a new sub module. The reference from peerpod-ctrl  to cloud-api-adaptor is the `pkg/adaptor/cloud` package. If we can introduce a new sub module `cloud`, then the dependency will looks like this.

- csi-wrapper  → cloud-api-adaptor
- peerpod-ctrl  → cloud
- cloud-api-adaptor → peerpod-ctrl
- cloud-api-adaptor → cloud

docs/Release-Process.md
stevenhorsman: > I thought we were reverting to latest just after the release so I thought releasing from a branch would be better.

I just to argue the counterpoint - if we start doing the update to overlays and webhook to pin their release versions I proposed in #1109 then that will be quite a few files that we will want to revert to latest soon after release, so maybe the branch is the best approach. My concern is us ended up with a lot of branches that would make things confusing, but I guess it's not disimilar to kata other than the frequency of releases is higher. Definitely an approach to discuss

stevenhorsman:
Thanks @yoheiueda - this is really useful information and I hadn't realised that we had issues with the tagging of the submodules. It's definitely more complicated. So am I correct in thinking that the short-term flow for cutting the release could be:

- Update the `kata-containers/src/runtime` go module that we include in the main `cloud-api-adaptor` [`go.mod`](../go.mod),
the `peerpod-ctl` [`go.mod`](../peerpod-ctrl/go.mod) and the `csi-wrapper` [`go.mod`](../volumes/csi-wrapper/go.mod).
  *If this introduces a breaking change them temporarily add a local references
- Merge this PR
- Tag v<version> as the cloud-api-adaptor/go.mod tag
- Tag peerpodconfig-ctrl/v<version> as the peerpodconfig-ctrl submodule tag on the same commit as it doesn't have any other dependencies
- Tag webhook/v<version> as the webhook submodule tag on the same commit as it doesn't have any other dependencies
- Update csi-wrapper and peerpod-ctrl to use the tagged version of cloud-api-adapter, by running
  ```
  go get github.com/confidential-containers/cloud-api-adaptor@v<version>
  go mod tidy
  ```
in their directories and removing the local reference if we needed to add it earlier
- Merge that PR
- Tag this commit with peerpod-ctrl/v0<version> and volumes/csi-wrapper/v<version>
- Create a v<version> release based on this commit

As you say the cloud-api-adaptor → peerpod-ctrl is a complication here that I guess we bump ad-hoc, but it would be good to break the cycle.

Does that work, or have I missed anything important again?

stevenhorsman:
An open question I have is should we create all these tags for the release candidate (ie something like `v0.7.0-alpha.1`), or just rely on commit hashes?

I think I probably favour tags - it is a little bit more overhead and will result in us creating double the number of tags, but I don't think that's a bad price to pay if it adds clarity

stevenhorsman:
@yoheiueda, @bpradipt, @wainersm - I've updated the documentation with the information about go submodules tagging. It has made the process more complicated though!

bpradipt (approved):
/lgtm

wainersm (approved):
Hi @stevenhorsman , thanks for promptly creating this release guide. Very needed and I will be able to verify the steps on the 0.7 release.

PR #1123, opened 2023-06-28
Add flexible VM sizes support for Azure provider

bpradipt:


bpradipt:
This is based on the initial PR for AWS #321. The difference is in the parameter names. Instead of instance-type Azure uses instance-sizes. The logic is same as for AWS.

yoheiueda (approved):
LGTM. Thank you very much!

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -358,3 +367,52 @@ func (p *azureProvider) deleteNetworkInterface(ctx context.Context, nicName stri
...
+
+	// Iterate over the page and populate the instanceSizeSpecList for all the instanceSizes
+	for pager.More() {
+		nextResult, err := pager.NextPage(context.Background())
+		if err != nil {
+			return fmt.Errorf("getting next page of VM sizes: %w", err)
+		}
+		for _, vmSize := range nextResult.VirtualMachineSizeListResult.Value {
+			if util.Contains(instanceSizes, *vmSize.Name) {
+				instanceSizeSpecList = append(instanceSizeSpecList, cloud.InstanceTypeSpec{InstanceType: *vmSize.Name, VCPUs: int64(*vmSize.NumberOfCores), Memory: int64(*vmSize.MemoryInMB)})
```
snir911, line 409: Sorry if it has been discussed before, what if multiple instance types matches same {vCPU, Mem} set?

snir911 (approved):
Added one question, overall LGTM though

pkg/adaptor/cloud/azure/provider.go
bpradipt: The first one in the list should get selected. We can try this scenario by updating the unit tests of the common cloud function to select bestfit instance type.

PR #1132, opened 2023-07-03
Azure: Use env file for secrets in kustomization

### LINKED ISSUE #1131: azure: secrets should not be inlined in kustomization
While we provide a service principal as a secret for the daemonset, we should not inline the variables in `kustomization.yaml`. There is a risk of leaking/pushing secrets by accident. Secrets should be put in a discrete file, which is covered by gitignore.

mkulke:
This PR uses the kustomization facility to reference an env file for a key/value map. We can then put the env file into gitignore.

fixes #1131

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -409,10 +409,8 @@ configMapGenerator:
 secretGenerator:
 - name: peer-pods-secret
   namespace: confidential-containers-system
-  literals:
-  - AZURE_CLIENT_ID="${AZURE_CLIENT_ID}"
-  - AZURE_CLIENT_SECRET="${AZURE_CAA_CLIENT_SECRET}"
-  - AZURE_TENANT_ID="${AZURE_TENANT_ID}"
+  envs:
+  - service-principal.env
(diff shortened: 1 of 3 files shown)


azure/README.md
```diff
@@ -409,10 +409,8 @@ configMapGenerator:
 secretGenerator:
 - name: peer-pods-secret
   namespace: confidential-containers-system
-  literals:
-  - AZURE_CLIENT_ID="${AZURE_CLIENT_ID}"
-  - AZURE_CLIENT_SECRET="${AZURE_CAA_CLIENT_SECRET}"
-  - AZURE_TENANT_ID="${AZURE_TENANT_ID}"
+  envs:
+  - service-principal.env
```
bpradipt, line 413: /lgtm
nice @mkulke .I'm yet to test though in my setup. Will do it shortly.
Can you please create tracker issues to add a similar approach for the other providers.

bpradipt (approved):

surajssd (approved):

Issue #1107, opened 2023-06-21
Libvirt test e2e failing due to lack of node-role.kubernetes.io/worker label


wainersm:
Today I ran the latest (commit 79d309dbdc) libvirt e2e test but it failed as:

```
<OMITTED>

Wait at least one worker be Ready
Wait system pods be running or completed

time="2023-06-21T15:37:22-03:00" level=info msg="Deploy the Cloud API Adaptor"
time="2023-06-21T15:37:22-03:00" level=info msg="Install the controller manager"
Wait for the cc-operator-controller-manager deployment be available
time="2023-06-21T15:38:29-03:00" level=info msg="Customize the overlay yaml file"
time="2023-06-21T15:38:32-03:00" level=info msg="Install the cloud-api-adaptor"
Wait for the cc-operator-daemon-install DaemonSet be available
F0621 15:39:32.565344   78821 env.go:369] Setup failure: timed out waiting for the condition
FAIL    github.com/confidential-containers/cloud-api-adaptor/test/e2e   795.612s
FAIL
make: *** [Makefile:91: test-e2e] Error 1
```

It turned out the cc-operator-daemon-install daemonSet didn't get available because the worker node hadn't the `node-role.kubernetes.io/worker` label:

```
$ kubectl get nodes
NAME                   STATUS   ROLES           AGE    VERSION
peer-pods-ctlplane-0   Ready    control-plane   129m   v1.27.3
peer-pods-worker-0     Ready    <none>          127m   v1.27.3
```

In order to solve that problem, it will be needed to apply the node label after the cluster is created (see https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/provision_libvirt.go#L86).

The Azure implementation also apply the label to the node (see https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/provision_azure.go#L308) however it uses `kubectl`. It would be nice if we could have a Go implementation instead, that new method could be hosted in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/e2e/common.go

How to reproduce the problem:

```
$ git clone https://github.com/confidential-containers/cloud-api-adaptor
$ cd cloud-api-adaptor
$ TEST_PROVISION=yes CLOUD_PROVIDER=libvirt make test-e2e
```

bpradipt:
Related issues
https://github.com/confidential-containers/cloud-api-adaptor/issues/780

[...]

PR #1152, opened 2023-07-07
tests: add worker label through go

bookinabox:
libvirt e2e tests did not properly label worker nodes with the worker label.

This creates a function in test/provisioner/common.go that adds the worker label to the libvirt e2e createCluster.

Fixes: #1107

### DIFF
--- a/test/provisioner/common.go
+++ b/test/provisioner/common.go
@@ -1,0 +1,54 @@
+// (C) Copyright Confidential Containers Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package provisioner
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+type patchLabel struct {
+	Op    string `json:"op"`
+	Path  string `json:"path"`
+	Value string `json:"value"`
+}
+
+// Adds the worker label to all workers nodes in a given cluster
+func AddNodeRoleWorkerLabel(ctx context.Context, clusterName string, cfg *envconf.Config) error {
+	fmt.Printf("Adding worker label to nodes belonging to: %s\n", clusterName)
+	client, err := cfg.NewClient()
+	if err != nil {
+		return err
+	}
+
+	nodelist := &corev1.NodeList{}
+	if err := client.Resources().List(ctx, nodelist); err != nil {
+		return err
+	}
+	// Use full path to avoid overwriting other labels (see RFC 6902)
+	payload := []patchLabel{{
+		Op: "add",
+		// "/" must be written as ~1 (see RFC 6901)
+		Path:  "/metadata/labels/node.kubernetes.io~1worker",
+		Value: "",
+	}}
+	payloadBytes, _ := json.Marshal(payload)
+	workerStr := clusterName + "-worker"
+	for _, node := range nodelist.Items {
+		if strings.Contains(node.Name, workerStr) {
+			if err := client.Resources().Patch(ctx, &node, k8s.Patch{PatchType: types.JSONPatchType, Data: payloadBytes}); err != nil {
+				return err
+			}
+		}
+
+	}
+	return nil
+}
--- a/test/provisioner/provision_libvirt.go
+++ b/test/provisioner/provision_libvirt.go
@@ -89,6 +89,7 @@ func NewLibvirtProvisioner(properties map[string]string) (CloudProvisioner, erro
 }
 
 func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Config) error {
+
 	cmd := exec.Command("/bin/bash", "-c", "./kcli_cluster.sh create")
 	cmd.Dir = l.wd
 	cmd.Stdout = os.Stdout
@@ -104,6 +105,11 @@ func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Con
 	home, _ := os.UserHomeDir()
 	kubeconfig := path.Join(home, ".kcli/clusters", clusterName, "auth/kubeconfig")
 	cfg.WithKubeconfigFile(kubeconfig)
+
+	if err := AddNodeRoleWorkerLabel(ctx, clusterName, cfg); err != nil {
+
+		return fmt.Errorf("labeling nodes: %w", err)
+	}
 
 	return nil
 }


wainersm:
Hi @bookinabox !

You will need to fix the error found by the golangci-lint:

```
./hack/golangci-lint.sh
Error: test/provisioner/common.go:26:14: ineffectual assignment to err (ineffassign)
		clientset, err = kubernetes.NewForConfig(config)
		           ^
make: *** [Makefile:113: golangci-lint] Error 1
```

wainersm:
I am also getting an error:

```
Adding worker label to nodes belonging to: peer-pods
W0710 15:07:13.356706   29753 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W0710 15:07:13.356717   29753 client_config.go:623] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
E0710 15:07:13.356787   29753 env.go:352] Recovering from panic and running finish actionsruntime error: invalid memory address or nil pointer dereference
```

I don't have `KUBECONFIG` exported on my environment. I let the e2e framework to create the cluster for me.

bookinabox:
> I am also getting an error:
> 
> ```
> Adding worker label to nodes belonging to: peer-pods
> W0710 15:07:13.356706   29753 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
> W0710 15:07:13.356717   29753 client_config.go:623] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
> E0710 15:07:13.356787   29753 env.go:352] Recovering from panic and running finish actionsruntime error: invalid memory address or nil pointer dereference
> ```
> 
> I don't have `KUBECONFIG` exported on my environment. I let the e2e framework to create the cluster for me.

Going to source the kcli_cluster.sh file then, so that the export has an effect outside the script (it doesn't look like its doing anything otherwise). I could also hard code the same path in, but I think to keep everything consistent, that might be best to only have the path declared in one place.

test/provisioner/common.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("initialized variables\n")
+	nodelist, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+
+	if err != nil {
+		return err
+	}
```
katexochen, line 41:
```suggestion
	nodelist, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
	if err != nil {
		return err
	}
```

test/provisioner/provision_libvirt.go
```diff
@@ -100,6 +101,16 @@ func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Con
...
 
+	if err != nil {
+		return err
+	}
+
+	err = AddNodeRoleWorkerLabel(clusterName, cfg)
+
+	if err != nil {
+		return fmt.Errorf("labeling nodes: %w", err)
+	}
```
katexochen, line 112:
```suggestion
	if err := AddNodeRoleWorkerLabel(clusterName, cfg); err != nil {
		return fmt.Errorf("labeling nodes: %w", err)
	}
```

test/provisioner/provision_libvirt.go
```diff
@@ -100,6 +101,16 @@ func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Con
 	kubeconfig := path.Join(home, ".kcli/clusters", clusterName, "auth/kubeconfig")
 	cfg.WithKubeconfigFile(kubeconfig)
 
+	if err != nil {
+		return err
+	}
+
```
katexochen, line 107: What error are we checking here?

test/provisioner/common.go
```diff
@@ -0,0 +1,64 @@
...
+
+	label := map[string]string{"node-role.kubernetes.io/worker": ""}
```
katexochen, line 43: Depending on merge order, this might needs to updated to fit the changes in https://github.com/confidential-containers/cloud-api-adaptor/pull/784

test/provisioner/provision_libvirt.go
bookinabox: oops, overlooked that. Thanks!

bookinabox:
updated to new worker label

katexochen (commented):
Sorry, missed this one on the first review. Thanks for updating to the new label.

test/provisioner/provision_libvirt.go
```diff
@@ -100,6 +101,11 @@ func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Con
 	kubeconfig := path.Join(home, ".kcli/clusters", clusterName, "auth/kubeconfig")
 	cfg.WithKubeconfigFile(kubeconfig)
 
+	if err := AddNodeRoleWorkerLabel(clusterName, cfg); err != nil {
```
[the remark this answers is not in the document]
bookinabox: Updated!

test/provisioner/common.go
```diff
@@ -0,0 +1,62 @@
...
+type patchLabel struct {
+	Op    string            `json:"op"`
+	Path  string            `json:"path"`
+	Value map[string]string `json:"value"`
+}
+
+func AddNodeRoleWorkerLabel(ctx context.Context, clusterName string, cfg *envconf.Config) error {
+	fmt.Printf("Adding worker label to nodes belonging to: %s\n", clusterName)
+
+	config, err := clientcmd.BuildConfigFromFlags("", cfg.KubeconfigFile())
```
wainersm, line 26: @bookinabox I tested this PR, it is working. I've only a few comments to the code.

First, what if you get the client object like (https://pkg.go.dev/sigs.k8s.io/e2e-framework@v0.1.0/pkg/envconf#Config.NewClient):
```
	client, err := cfg.NewClient()
	if err != nil {
		return err
	}
```

test/provisioner/common.go
```diff
@@ -0,0 +1,62 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+type patchLabel struct {
+	Op    string            `json:"op"`
+	Path  string            `json:"path"`
+	Value map[string]string `json:"value"`
+}
+
+func AddNodeRoleWorkerLabel(ctx context.Context, clusterName string, cfg *envconf.Config) error {
```
wainersm, line 23: did you try to extract the cluster name from the Config and/or Client objects? If it is possible, it would save us one argument to this function :)

wainersm: While talking about this function, as it will have public access...could you please add documentation?

test/provisioner/common.go
```diff
@@ -0,0 +1,62 @@
+package provisioner
```
wainersm, line 1: Could you add a license header to this file?

See https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/provision.go as for example:
```
// (C) Copyright Confidential Containers Contributors
// SPDX-License-Identifier: Apache-2.0
```

test/provisioner/common.go
```diff
@@ -0,0 +1,62 @@
...
+		Path:  "/metadata/labels",
+		Value: label,
+	}}
+
+	payloadBytes, _ := json.Marshal(payload)
+	workerStr := clusterName + "-worker"
+
+	for _, node := range nodelist.Items {
+		if strings.Contains(node.Name, workerStr) {
+			_, err = clientset.CoreV1().Nodes().Patch(ctx, node.Name, types.JSONPatchType, payloadBytes, metav1.PatchOptions{})
```
wainersm, line 54: If the label already exists, will it fail? If yes then maybe it should check its existence first.

test/provisioner/common.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Unable to add label %s to node %s: %s\n", label, node.Name, err)
```
wainersm, line 56: are you not returning in case of error on purpose?

test/provisioner/provision_libvirt.go
```diff
@@ -100,6 +101,11 @@ func (l *LibvirtProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Con
 	kubeconfig := path.Join(home, ".kcli/clusters", clusterName, "auth/kubeconfig")
 	cfg.WithKubeconfigFile(kubeconfig)
 
+	if err := AddNodeRoleWorkerLabel(ctx, clusterName, cfg); err != nil {
+
```
wainersm, line 105: don't need this extra newline.

test/provisioner/common.go
bookinabox: Well, the clientset and the client returned from that function are different types. I don't know why there are different ways to do the same thing/what the pros/cons are but I will try to do it this way. I imagine there will be some major changes required to do this though.

test/provisioner/common.go
bookinabox: I think I was at one point, but its probably better to error out.

test/provisioner/common.go
bookinabox: Is there a specific reason why this way would be better? I'm having a little trouble getting the patch operation to work with resources (I'm sure there's an easy fix, I'm confused which k8s.object I need to pass into .Patch) and I'm not sure what the benefit to doing the above way would be.

test/provisioner/common.go
bookinabox: afaik, it shouldn't fail, at least with this API. It has not errored out for that reason when I tested it.

bookinabox:
I changed it to use the different API and cleaned up the code a bit. One thing I'm still looking into is the cluster name stuff. It looks like kcli creates a cluster with the generic "kubernetes" cluster name. I haven't looked deep enough into the code to determine how it figures out how to delete them, but I will investigate.

This makes the "clustername" argument misleading, so I'm thinking of a better way/how to work around this.

test/provisioner/common.go
wainersm: I saw that you managed to use the cfg.NewClient() which returns a `kclient` (https://github.com/kubernetes-sigs/e2e-framework/tree/main/klient), however I still own you some answers.

The e2e-framework provides some facilities over the standard k8s api. For example, when you start the test suite you can set the k8s namespace resources will be created (and destroyed on the end). The `kcliet` wraps that information so that you don't need to pass the namespace to every call of, for example, `Resources()`. The `kclient` has also special `wait` routines for resources. And so on.

There are other reasons to use 'kclient' instead of 'clientset`. For example, avoid dependence on another package, avoid code fragmentation by leveraging a single API thoughtout the code base. And so on.

wainersm:
> I changed it to use the different API and cleaned up the code a bit. One thing I'm still looking into is the cluster name stuff. It looks like kcli creates a cluster with the generic "kubernetes" cluster name. I haven't looked deep enough into the code to determine how it figures out how to delete them, but I will investigate.
> 
> This makes the "clustername" argument misleading, so I'm thinking of a better way/how to work around this.

Do you prefer to merge as is and you send another PR later? IMO let's try to merge this because it will ease the testing of the other pending PRs....

wainersm:
I tested the latest version of this PR, working fine on my env.

wainersm (approved):

bookinabox:
okay, let's just merge it for now then.

wainersm:
hi @katexochen ! Do you have more comments for this PR?

katexochen:
Sorry, I didn't find the time to debug my local setup, but code lgtm. :+1:

snir911 (approved):
LGTM thanks!

Issue #1192, opened 2023-07-12
test/provisioner/provision_libvirt: libvirt_ssh_key_file in properties adds new entry every time


bookinabox:
When running test-e2e with libvirt and a TEST_PROVISION_FILE that includes a libvirt_ssh_key_file, the libvirt_ssh_key_file is appended to secretGenerator files in install/overlay/libvirt/kustomization.yaml regardless of whether that file is already listed. This is not a huge issue, but having multiple of the same file causes:

`Setup failure: configmap ssh-key-secret illegally repeats the key 'id_rsa'`

One fix could be checking if the file is already listed before checking it in the SetKustomizeSecretGeneratorFile function in test/provisioner/kustomize.go. (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/kustomize.go#L206)

PR #1193, opened 2023-07-12
test/provisioner: avoid adding duplicate file in secretGenerator

bookinabox:
Adds a check to avoid adding a file to secretGenerator in kustomization.yaml if it already is listed there.

Originally, in the libvirtd tests, if libvirtd_ssh_key_file was set and TEST_PROVISION_FILE environment variable was used, it would fail if one did not remember to reset the kustomization.yaml between tests.

Fixes: #1192

### DIFF
--- a/test/provisioner/kustomize.go
+++ b/test/provisioner/kustomize.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


wainersm (approved):
That behavior always bothered me! Thanks for the fix @bookinabox !

snir911 (approved):
LGTM

Issue #1188, opened 2023-07-12
Integrate IBMCloud e2e test of CSI Wrapper to workflow


wyuany:
We have e2e test of CSI wrapper and get it tested on self managed cluster of IBMCloud. The test result can be integrated to workflow of open source community of https://github.com/confidential-containers/cloud-api-adaptor so that open source users are confident with the quality of CSI Wrapper.

PR #1194, opened 2023-07-13
e2e: add workflow for csi wrapper e2e test

wyuany:
fix: https://github.com/confidential-containers/cloud-api-adaptor/issues/1188

This enable workflow to display e2e test result of csi wrapper. The e2e test of csi wrapper includes creating self managed kubeadm cluster in IBMCloud, then installing cloud api adapater and csi wrapper to the cluster, at last testing csi wrapper.

Signed-off-by: yuanyuanwang wyuany@cn.ibm.com

### DIFF
--- a/.github/workflows/daily-e2e-tests-ibmcloud.yaml
+++ b/.github/workflows/daily-e2e-tests-ibmcloud.yaml
@@ -23,6 +23,7 @@ jobs:
           - type: amd64
           - type: s390x-non-secure-execution
           - type: s390x-secure-execution
+          - type: csi_wrapper_x86_64
     steps:
       - name: Checkout the code
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-e2e-tests-ibmcloud.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
           - type: amd64
           - type: s390x-non-secure-execution
           - type: s390x-secure-execution
+          - type: amd64-csi-wrapper
```
liudalibj, line 26: would you like use a different keyword amd64 here, it will make the pipeline use a wrong log file
maybe use:
```
          - type: csi-wrapper-x86_64
```

.github/workflows/daily-e2e-tests-ibmcloud.yaml
```diff
@@ -89,6 +90,9 @@ jobs:
             s390x-secure-execution)
               arch_string="s390x-se"
               ;;
+            amd64-csi-wrapper)
+              arch_string="amd64"
```
liudalibj, line 94: for csi-wrapper, case, we do NOT need download and push podvm image again, as we use the same image from `amd64` type.

liudalibj (approved):
LGTM, @wyuany thanks for add the e2e-test for csi-wrapper in amd64 env.

liudalibj:
Thanks @stevenhorsman for the comments.

> The workflow looks good to me. Sorry I'm late to this as it was decided elsewhere, but just to confirm:
> 
> * is the reason that csi_wrapper test is not part of the other test suites because it runs ona. self-managed cluster rather than IKS?

As we didn't run the e2e test on self-managed x86 cluster, it's better run a full set e2e-test here, so let's replace `go test ...` with `make test-e2e` @wyuany 

> * Are there plans for an s390x CSI wrapper test suite?

Yeah, we have plan to run s390x csi wrapper test in future, it's depend on when the s390x ibm-vpc-block-csi-driver docker images will be supported.
```
root@peer-pod1 ~ $ docker pull gcr.io/k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver:master --platform=s390x
master: Pulling from k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver
Digest: sha256:3e5b13e80e39229a58353f03fc1847802833cac31a0b8f270c29128e76569125
Status: Image is up to date for gcr.io/k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver:master
image with reference gcr.io/k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver:master was found but does not match the specified platform: wanted linux/s390x, actual: linux/amd64
root@peer-pod1 ~ $
```

huoqifeng (approved):
/lgtm
Thanks! @wyuany and @liudalibj

PR #1199, opened 2023-07-14
Automatically fetch config map values

esposem:
This PR gets rid of some ConfigMap parameters that must be provided to CAA in order to make it run on Azure/AWS.

More specifically:
- In Azure, ConfigMap was
```
    cat <<EOF | oc apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: ${CM}
  namespace: ${OP_NAME}
data:
  CLOUD_PROVIDER: "azure"
  VXLAN_PORT: "9000"
  AZURE_INSTANCE_SIZE: "Standard_D8as_v5" #set
  PROXY_TIMEOUT: "15m"
  AZURE_IMAGE_ID: ${AZURE_IMAGE_ID}
  AZURE_SUBNET_ID: ${AZURE_SUBNET_ID}
  AZURE_NSG_ID: ${AZURE_NSG_ID}
  DISABLECVM: "true"
EOF
```
and becomes now
```
    cat <<EOF | oc apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: ${CM}
  namespace: ${OP_NAME}
data:
  CLOUD_PROVIDER: "azure"
  VXLAN_PORT: "9000"
  AZURE_INSTANCE_SIZE: "Standard_D8as_v5" #set
  PROXY_TIMEOUT: "15m"
  AZURE_IMAGE_ID: ${AZURE_IMAGE_ID}
  DISABLECVM: "true"
EOF
```

### Current limitations of this patch

azure:
- tested on Azure

Fixes: #1200

### DIFF
--- a/pkg/adaptor/cloud/azure/provider.go
+++ b/pkg/adaptor/cloud/azure/provider.go
@@ -19,6 +19,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	armcompute "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/avast/retry-go/v4"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/adaptor/cloud"
 	"github.com/confidential-containers/cloud-api-adaptor/pkg/util"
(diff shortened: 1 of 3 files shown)


esposem:
@snir911 @bpradipt  the solution to the issues mentioned in the PR description are:
- Assume the user provided a AZURE_USER_RESOURCE_GROUP field in the peerpodssecret, which corresponds to the user-created azure RG. This is valid only when using ARO/AKS.
- Query for subnets/vnets/nsg but if more than one result is returned, then return an error and force the user to insert it via the ConfigMap
- This implies (and should be documented somewhere) that for ARO the subnet ID *must* be provided by the user, as workers and masters are in two different user-created subnets, and we need the worker subnet.

pkg/adaptor/cloud/azure/manager.go
```diff
@@ -39,6 +39,7 @@ func (_ *Manager) LoadEnv() {
 	cloud.DefaultToEnv(&azurecfg.SubscriptionId, "AZURE_SUBSCRIPTION_ID", "")
 	cloud.DefaultToEnv(&azurecfg.Region, "AZURE_REGION", "")
 	cloud.DefaultToEnv(&azurecfg.ResourceGroupName, "AZURE_RESOURCE_GROUP", "")
+	cloud.DefaultToEnv(&azurecfg.UserResourceGroupName, "AZURE_USER_RESOURCE_GROUP", "")
```
snir911, line 42: can you explain why it's needed?

snir911 (commented):
Thanks! i added some comments, also please look at the build and lint actions

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -438,3 +480,142 @@ func (p *azureProvider) updateInstanceSizeSpecList() error {
...
+
+	virtualNetworksClientNewListPager := virtualNetworksClient.NewListPager(resourceGroupName, nil)
+	for virtualNetworksClientNewListPager.More() {
+		nextResult, err := virtualNetworksClientNewListPager.NextPage(context.Background())
+		if err != nil {
+			return fmt.Errorf("getting virtualnetworks NextPage: %w", err)
+		}
+		if nextResult.VirtualNetworkListResult.Value != nil {
+			for _, VirtualNetwork := range nextResult.VirtualNetworkListResult.Value {
+				found++
```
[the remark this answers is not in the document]
esposem: I'll then make it print all networks found, use the first one and print a warning if there are more than one found

pkg/adaptor/cloud/azure/manager.go
esposem: We need to pass an AZURE_USER_RESOURCE_GROUP because for some Azure solutions like ARO the user might need to create a separate resource group that seems not to be able to be queried via the Azure SDK. In addition, Azure also creates an internal resource group that can be queried by the SDK. This user RG is needed because some resources like Subnet ID are created under that RG

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -38,10 +39,43 @@ type azureProvider struct {
 	serviceConfig *Config
 }
 
-func NewProvider(config *Config) (cloud.Provider, error) {
+// TODO: with Azure managed services like ARO/AKS, this won't work.
```
bpradipt, line 42:
```suggestion
// TODO: Add support for managed K8s - AKS and ARO
// AKS and ARO creates a separate resource group for the cluster resources and current code needs to be adapted to handle retrieving the cluster resource group and related settings
```

bpradipt: And you can remove all the subsequent TODOs

bpradipt:
@esposem can you please rebase this PR ?

Issue #1015, opened 2023-05-26
podvm: attestation agent start fails due to wrong flags in systemd unit


katexochen:
The AA is currently always [built with ttRPC](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/Makefile.inc#L129), but the systemd unit we put into the podvm [isn't configured to use a unix domain socket](https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/podvm/files/etc/systemd/system/attestation-agent.service#L9):

```
ExecStart=/usr/local/bin/attestation-agent --getresource_sock 127.0.0.1:50001
```

PR #1202, opened 2023-07-14
Azure community gallery support

jepio:
Several changes that will allow us to publish demo images for CAA:
- support passing community gallery id in caa
- ~allow building a second attestation-agent (w/ grpc)~
- ~add tdx dependencies to ubuntu podvm image (azure-only but should go everywhere) so that cc_kbc can be used~

### DIFF
--- a/pkg/adaptor/cloud/azure/provider.go
+++ b/pkg/adaptor/cloud/azure/provider.go
@@ -12,6 +12,7 @@ import (
 	"net/netip"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
@@ -213,6 +214,15 @@ func (p *azureProvider) CreateInstance(ctx context.Context, podName, sandboxID s
 		securityProfile = nil
 	}
 
+	imgRef := &armcompute.ImageReference{
+		ID: to.Ptr(p.serviceConfig.ImageId),
+	}
+	if strings.HasPrefix(p.serviceConfig.ImageId, "/CommunityGalleries/") {
+		imgRef = &armcompute.ImageReference{
+			CommunityGalleryImageID: to.Ptr(p.serviceConfig.ImageId),
+		}
+	}
+
 	vmParameters := armcompute.VirtualMachine{
 		Location: to.Ptr(p.serviceConfig.Region),
 		Properties: &armcompute.VirtualMachineProperties{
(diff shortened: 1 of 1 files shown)


mkulke (commented):
I tested the gallery image support and it worked for me. 🎉

I would recommend splitting off the image-related changes into their own PR though, to make it more discoverable for reviews and harmonize the approach among the images

azure/image/misc-settings.sh
```diff
@@ -131,6 +135,13 @@ TLS_OPTIONS=-cert-file /etc/certificates/tls.crt -cert-key /etc/certificates/tls
 END
 fi
 
+mkdir -p /etc/systemd/system/attestation-agent.service.d
+cat <<END >/etc/systemd/system/attestation-agent.service.d/10-grpc.conf
+[Service]
+ExecStart=
+ExecStart=/usr/local/bin/attestation-agent.grpc --getresource_sock 127.0.0.1:50001
+END
+
```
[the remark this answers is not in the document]
jepio: I'll move this to a separate PR but this is kind of a policy question. If none of the CAA users have a use for ttRPC I'll switch it to gRPC and make sure we have two.

For sure `/usr/local/bin/attestation-agent` needs to stay ttrpc so that kata-agent works (can be fixed but that's how it is right now).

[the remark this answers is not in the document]
katexochen: > If none of the CAA users have a use for ttRPC

The current unit has never worked with ttRPC, as it uses an IP address as socket and will always fail (https://github.com/confidential-containers/cloud-api-adaptor/issues/1015). So I assume it isn't used by anyone.

bpradipt, line 144: So this will override https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/files/etc/systemd/system/attestation-agent.service.  Do we want this ? ( I have no preference, hence checking)

pkg/adaptor/cloud/azure/provider.go
```diff
@@ -213,16 +214,23 @@ func (p *azureProvider) CreateInstance(ctx context.Context, podName, sandboxID s
 		securityProfile = nil
 	}
 
+	imgRef := &armcompute.ImageReference{
+		ID: to.Ptr(p.serviceConfig.ImageId),
+	}
+	if strings.HasPrefix(p.serviceConfig.ImageId, "/CommunityGalleries/") {
```
bpradipt, line 220: For my understand, is this the doc for community gallery - https://learn.microsoft.com/en-us/azure/virtual-machines/share-gallery-community?tabs=cli ?

azure/image/misc-settings.sh
katexochen: There is an discussion of this regard under this comment https://github.com/confidential-containers/cloud-api-adaptor/pull/1202#discussion_r1264939958 :wink:

pkg/adaptor/cloud/azure/provider.go
bpradipt: Thanks @jepio. I'll try the community gallery. And also get back on the community RHEL image.

mkulke (approved):

bpradipt (approved):
/lgtm
Thanks @jepio 

As a separate PR, if we can create an AKS quick start guide for the upcoming v0.7.0 release it'll be very helpful to people trying out CoCo in Azure

bpradipt:
@jepio Added this as a reference - https://github.com/confidential-containers/cloud-api-adaptor/issues/1213

Issue #1228, opened 2023-07-21
podvm image build is broken


liudalibj:
Today's test-e2e pipline are failed, dug the root cause:
the podvm image build is broken as we change the `KATA_SRC_BRANCH` from ccv0 to `9e036108b4d5d9344eedcd2a5dac78d316b79cac`, https://github.com/confidential-containers/cloud-api-adaptor/pull/1225

it make the command failed at https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/Dockerfile.podvm_builder#L59
```
 ---> Running in adcfca185e7a
Cloning into 'kata-containers'...
fatal: Remote branch 9e036108b4d5d9344eedcd2a5dac78d316b79cac not found in upstream origin
The command '/bin/sh -c git clone ${KATA_SRC} -b ${KATA_SRC_BRANCH} kata-containers' returned a non-zero code: 128
root@peerpod-se-image-vsi:~/cloud-api-adaptor/podvm#
```

PR #1229, opened 2023-07-21
podvm: fix podvm image build

liudalibj:
- use another way to checkout kata-container codes

Fixes: #1228


stevenhorsman (approved):
LGTM. Thanks for the quick fix @liudalibj!

jepio (approved):

Issue #1238, opened 2023-07-24
libvirt: Get LIBVIRT_IP automatically


bookinabox:
Currently, the static route to the libvirt network must be exported before running with libvirt or running libvirt tests (see /libvirt/README.md). I believe this should be able to be obtained automatically, and that this may be needed to implement the network for #1165

LIBVIRT_IP is used to determine LIBVIRT_URI (though in tests, it looks like the libvirt properties file would be able to override the uri)

PR #1244, opened 2023-07-25
libvirt: Get LIBVIRT_IP from network XML config

bookinabox:
LIBVIRT_IP is required to run install_operator.sh for libvirt. This ip address should refer to the ip address defined in the network XML configuration and error out if it is undefined.

Rather than manually exporting this, this can be determined by looking at the network's xmldump.

By default, this is typically 192.168.122.1, but can be changed.

Note that kcli, which is used to create the cluster in kcli_cluster.sh must be past commit 0160f70 to create a cluster using a network that is does belong under the cidr submask 192.168.122.0/24. see:
github.com/karmab/kcli/commit/0160f7000f2bcc09b52b449e837e3ad386006a6a

Fixes: #1238

### DIFF
--- a/libvirt/install_operator.sh
+++ b/libvirt/install_operator.sh
@@ -41,7 +41,7 @@ usage() {
 
 	Use the following environment variables to change the installation
 	parameters:
-	LIBVIRT_IP    (required)
+	LIBVIRT_IP    (default determined from LIBVIRT_NET)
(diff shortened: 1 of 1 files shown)


wainersm:
@bpradipt @snir911 this one is very simple too :)

bpradipt (approved):

Issue #1234, opened 2023-07-21
cloud-api-adaptor image release build taking a long time


stevenhorsman:
During the release process Wainer pointed out that the cloud-api-adaptor [image workflow](https://github.com/confidential-containers/cloud-api-adaptor/actions/workflows/image.yaml) have grown to take >50mins now. Looking back it seems to have been triggered but the golang-fedora builder change: https://github.com/confidential-containers/cloud-api-adaptor/pull/1116 so we might want to revisit that and see if there are any steps we can do to improve release performance - particularly for multi-arch builds I think.

wainersm:
The last "fast" build took ~12min (https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/5424819216/job/14691935075) and the build times are:

 * ppc64le 
    ```
    #39 [linux/amd64->ppc64le builder 10/10] RUN make ARCH=ppc64le COMMIT=c10fa643486f98abd637e58003512eea08989df0 VERSION=v0.6.0-dev RELEASE_BUILD=true cloud-api-adaptor
   #39 DONE 553.9s
   ```

 * s390x
    ```
   #38 [linux/amd64->s390x builder 10/10] RUN make ARCH=s390x COMMIT=c10fa643486f98abd637e58003512eea08989df0 VERSION=v0.6.0-dev RELEASE_BUILD=true cloud-api-adaptor
   #38 DONE 573.6s
   ```

 * amd64 
    ```
    #40 [linux/amd64 builder 10/10] RUN make ARCH=amd64 COMMIT=c10fa643486f98abd637e58003512eea08989df0 VERSION=v0.6.0-dev RELEASE_BUILD=true cloud-api-adaptor
    #40 DONE 582.5s
   ```

[One of the last builds](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/5646326401/job/15293980614) took > 1h. The build times:


 * ppc64le 
    ```
    #35 [linux/ppc64le builder 8/8] RUN make ARCH=ppc64le COMMIT=6d7d2a3fe8243809b3c3a710792c8498292e2fc3 VERSION=v0.7.0-alpha.2-dev RELEASE_BUILD=true cloud-api-adaptor
   #35 DONE 2949.9s
    ```

 * s390x
    ```
[...]

PR #1116, opened 2023-06-23
image: use golang-fedora as builder

katexochen:
Previously, the docker build used the golang image, which is based
on debian, and copied the dynamically linked binary into a fedora
layer. This change switches the build to use a custom golang-fedora
image as builder.

Fixes #1093
Fixes #1113

Please check if this works for other platforms than amd64.

PR #1259, opened 2023-07-27
image: revert to use amd64 as build platform

katexochen:
The use of `--platform=$TARGETPLATFORM` was introduced by mistake in #1116, and caused long build times, as it emulates the target platform instead of using Go cross-compilation.

Fixes: #1234

Run finished in 19 min: https://github.com/katexochen/cloud-api-adaptor/actions/runs/5678850841

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,12 @@ ARG BUILD_TYPE=dev
 ARG BUILDER_BASE=quay.io/confidential-containers/golang-fedora:1.20.6-36
 ARG BASE=fedora:36
 
-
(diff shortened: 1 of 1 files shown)


wainersm (approved):
Thanks for the quick fix @katexochen !

bpradipt (approved):
/lgtm
Thanks @katexochen 
nit: Would adding a comment about TARGETPLATFORM and BUILDPLATFORM above the FROM statement help to prevent mistakenly changing the value?
Or adding a link to this excellent docker blog on emulation vs cross-compilation - https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/ ?

PR #1261, opened 2023-07-27
pkg/adaptor/libvirt: adds support for s390x based libvirt hosts

### LINKED ISSUE #1248: libvirt: support s390x architecture for domain XML
The domain XML created by the libvirt provider here: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/libvirt/libvirt.go#L162 has dependencies on the `x86_64` architecture:

- explicit value of `x86_64`
- use of `sata` and `ide` on the drivers

I suggest to make the following changes:
- detect the target architecture by consulting the libvirt connection
- if the target architecture is `s390x` create the libvirt XML with the required settings, in particular (probably not complete)
  - correct arch value on the DomainOSType
  - use `virtio` instead of `ide` and `sata`
  - add the `IOMMU: "on",` flag to the devices, this is important to support Secure Execution on s390x going forward
  - make sure no mem-balloon is used

CarstenLeue:
Fixes #1248

This PR creates a domainXML based on the architecture of the target host. First it finds the architecture via the GetNodeInfo method. Then if the architecture (aka Model) is s390x we'd create the domain XML with the s390x specifics, else it returns the x86_64 version as before.

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -506,7 +506,7 @@ az vm list \
   --output table
 ```
 
-Here you should see the vm associated with the pod `nginx`.
+Here you should see the vm associated with the pod `nginx`. If you run into problems then check the troubleshooting guide [here](../docs/troubleshooting/README.md).
 
 ## Cleanup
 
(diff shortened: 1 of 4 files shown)


CarstenLeue:
Closing in favor of https://github.com/confidential-containers/cloud-api-adaptor/pull/1262

PR #1265, opened 2023-07-28
pkg/adaptor/libvirt: cloud init ISO file using native go

### LINKED ISSUE #1250: libvirt: error handling in createCloudInitISO
The method https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/libvirt/libvirt.go#L23 handles errors by exiting the process.

I suggest to change error handling to return an error object and let the caller handle the situation.

### LINKED ISSUE #1251: libvirt: remove dependency on `/usr/bin/genisoimage` tool
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/libvirt/libvirt.go#L27 relies on the existence of the `/usr/bin/genisoimage` tool. The ISO creation step would fail if the tool were not present or if writing to intermediary files would fail.

I suggest to use a native go implementation to produce the ISO image, e.g. `github.com/kdomanski/iso9660`

Also we know that the size of the ISO image will be relatively small (there exists a limit of 64KB at least in the cloud for cidata ISO images), so we do not have to create anything on disk, this reduces the potential of failure. Instead we can create the ISO file in memory and directly upload it from there, e.g. like so:

```go
// CreateCloudInit produces a cloud init ISO file as a data blob with a userdata and a metadata section
func CreateCloudInit(userData, metaData []byte) ([]byte, error) {
	writer, err := iso9660.NewWriter()
	if err != nil {
		return nil, err
	}
	defer writer.Cleanup()

	err = writer.AddFile(bytes.NewReader(userData), userDataFilename)
	if err != nil {
		return nil, err
	}

	err = writer.AddFile(bytes.NewReader(metaData), metaDataFilename)
	if err != nil {
		return nil, err
	}

	err = writer.AddFile(bytes.NewReader([]byte{}), vendorDataFilename)
	if err != nil {
		return nil, err
	}

	var buf bytes.Buffer

	err = writer.WriteTo(&buf, ciDataVolumeName)
	if err != nil {
		return nil, err
	}

	// done
	return buf.Bytes(), nil
}

```

CarstenLeue:
Fixes #1251
Fixes #1250

This PR uses a native go implementation of an tool to create the cloud init ISO file, removing the dependency on an external tool. It also creates the ISO in-memory. This makes sense because the ISO file is very small and the content is backed in memory anyway. Creating the ISO file in-memory reduces the potential of errors, since nothing has to be written to the file system of the hosting container.
ISO file generation as well as memory based image uploader are backed by unit tests.

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -506,7 +506,7 @@ az vm list \
   --output table
 ```
 
-Here you should see the vm associated with the pod `nginx`.
+Here you should see the vm associated with the pod `nginx`. If you run into problems then check the troubleshooting guide [here](../docs/troubleshooting/README.md).
 
 ## Cleanup
 
--- a/docs/troubleshooting/README.md
+++ b/docs/troubleshooting/README.md
@@ -1,0 +1,35 @@
+# Troubleshooting
+
+This document describes generic troubleshooting steps after installation of Cloud API Adaptor.
+
+## Pod created but it stays in `ContainerCreating` state
+
+```console
+$ kubectl get pods -n confidential-containers-system -o wide
+NAME                                              READY   STATUS    RESTARTS   AGE     IP            NODE                                NOMINATED NODE   READINESS GATES
+cc-operator-controller-manager-585ffdbffd-2bmg6   2/2     Running   0          6m25s   10.244.0.11   aks-nodepool1-17035871-vmss000000   <none>           <none>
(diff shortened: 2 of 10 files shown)


CarstenLeue:
closing in favor of https://github.com/confidential-containers/cloud-api-adaptor/pull/1268

Issue #1266, opened 2023-07-28
ibmcloud: Podvm image-id list for region independent deployment


jtumber-ibm:
When a cluster admin is installing peerpods into a cluster they currently need to provide the correct image-id for each region as the images are region specific. Instead they should be able to specify a list of images and the correct one be selected at runtime.

Two possible approaches:
1. provide the image's region along side the image-ids, e.g. `uk:<image-id>,us:<image-id>` or `regions:uk,us,... images:<image-id>,<image-id>`
2. Only provide the list of images and check to see if the image exists in the current region at startup 

An additional benefit to the second approach is we could do further validation on the image at startup to trim the list for other criteria e.g. make sure the architecture matches the node architecture

PR #1267, opened 2023-07-28
ibmcloud: allow list of image-ids

jtumber-ibm:
Add the ability to specify a list of allowed image-ids Trim the list to remove image-ids not found in the same region as the node

Adds check that the ARCH matches the spec request, but that currently doesn't work as the instance spec ARCH value doesn't currently get populated.

Fixes: #1266

### DIFF
--- a/pkg/adaptor/cloud/ibmcloud/manager.go
+++ b/pkg/adaptor/cloud/ibmcloud/manager.go
@@ -24,7 +24,7 @@ func (*Manager) ParseCmd(flags *flag.FlagSet) {
 	flags.StringVar(&ibmcloudVPCConfig.ProfileName, "profile-name", "", "Default instance profile name to be used for the Pod VMs")
 	flags.Var(&ibmcloudVPCConfig.InstanceProfiles, "profile-list", "List of instance profile names to be used for the Pod VMs, comma separated")
 	flags.StringVar(&ibmcloudVPCConfig.ZoneName, "zone-name", "", "Zone name")
-	flags.StringVar(&ibmcloudVPCConfig.ImageID, "image-id", "", "Image ID")
+	flags.Var(&ibmcloudVPCConfig.Images, "image-id", "List of Image IDs, comma separated")
 	flags.StringVar(&ibmcloudVPCConfig.PrimarySubnetID, "primary-subnet-id", "", "Primary subnet ID")
 	flags.StringVar(&ibmcloudVPCConfig.PrimarySecurityGroupID, "primary-security-group-id", "", "Primary security group ID")
 	flags.StringVar(&ibmcloudVPCConfig.SecondarySubnetID, "secondary-subnet-id", "", "Secondary subnet ID")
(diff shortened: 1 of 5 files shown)


jtumber-ibm:
My own tests using `jamesnelson/cloud-api-adaptor:f882434d53109fbf0267260383bc3b0a61cc2e48`

Set `IBMCLOUD_PODVM_IMAGE_ID="r018-af5c0426-7bcb-435b-8605-017aae808057,r018-af5c0426-7bcb-435b-8605-017aae808080"` The first image-id is valid and the second is not.

From the logs 

```
2023/07/28 10:31:18 [adaptor/cloud/ibmcloud] skipping image (r018-af5c0426-7bcb-435b-8605-017aae808080), due to The requested resource does not exist, or cannot be accessed.
```
&
```
Images:ibmcloud.Images{ibmcloud.Image{ID:"r018-af5c0426-7bcb-435b-8605-017aae808057", Arch:"amd64", OS:"ubuntu-20-04-amd64"}}
```

pkg/adaptor/cloud/ibmcloud/manager.go
```diff
@@ -24,7 +24,7 @@ func (*Manager) ParseCmd(flags *flag.FlagSet) {
 	flags.StringVar(&ibmcloudVPCConfig.ProfileName, "profile-name", "", "Default instance profile name to be used for the Pod VMs")
 	flags.Var(&ibmcloudVPCConfig.InstanceProfiles, "profile-list", "List of instance profile names to be used for the Pod VMs, comma separated")
 	flags.StringVar(&ibmcloudVPCConfig.ZoneName, "zone-name", "", "Zone name")
-	flags.StringVar(&ibmcloudVPCConfig.ImageID, "image-id", "", "Image ID")
+	flags.Var(&ibmcloudVPCConfig.Images, "image-id", "List of Image ID, comma separated")
```
stevenhorsman, line 27: minor nit:
```suggestion
	flags.Var(&ibmcloudVPCConfig.Images, "image-id", "List of Image IDs, comma separated")
```

stevenhorsman (commented):
I did a super quick scan and will have a proper review later and test it out. Just a quick question on the expected behaviour - it allows a user to supply a list of VPC custom image IDs and it will filter them by region and then select the first one it finds? What is the behaviour if there are two valid images in a single region - failure, warning, or just pick the first one?

We might want to consider updating the e2e tests for this case, but short of enforcing an extra image uploaded and created in a different region I'm not sure I have a good suggestion there.

jtumber-ibm:
The expected use case is that there would normally only be one valid image per region, the implemented logic is just to pick the first valid one if there are multiple.

stevenhorsman:
> The expected use case is that there would normally only be one valid image per region, the implemented logic is just to pick the first valid one if there are multiple.

Do you think it is fair to log a warning if there are multiple valid ones?

jtumber-ibm:
> Do you think it is fair to log a warning if there are multiple valid ones?

Potentially, could be more explicit with which image gets selected? i.e. `selected <abc> out of # images`

stevenhorsman:
> > Do you think it is fair to log a warning if there are multiple valid ones?
> 
> Potentially, could be more explicit with which image gets selected? i.e. `selected <abc> out of # images`

Yeah - the main use case I want to avoid is a user appending a new image to their list mistakenly thinking that the others are all in other regions, but find they've missed updating one so ending running an old version. It might just be me that is likely to do that though!

Jordan9500 (approved):
LGTM! Thanks @jtumber-ibm.

stevenhorsman:
@jtumber-ibm - once you are back I think you might want to rebase this to pick up the latest version of go and resolve the govuln. Thanks!

PR #1262, opened 2023-07-27
pkg/adaptor/libvirt: adds support for s390x based libvirt hosts

CarstenLeue:
Fixes #1248

This PR creates a domainXML based on the architecture of the target host.
First it finds the architecture via the GetNodeInfo method. Then if the architecture (aka Model) is s390x we'd create the domain XML with the s390x specifics, else it returns the x86_64 version as before.

PR #1268, opened 2023-07-28
pkg/adaptor/libvirt: cloud init ISO file using native go

### LINKED ISSUE #1250: libvirt: error handling in createCloudInitISO

### LINKED ISSUE #1251: libvirt: remove dependency on `/usr/bin/genisoimage` tool

CarstenLeue:
Fixes #1251
Fixes #1250

This PR uses a native go implementation of an tool to create the cloud init ISO file, removing the dependency on an external tool. It also creates the ISO in-memory. This makes sense because the ISO file is very small and the content is backed in memory anyway. Creating the ISO file in-memory reduces the potential of errors, since nothing has to be written to the file system of the hosting container.
ISO file generation as well as memory based image uploader are backed by unit tests.

### DIFF
--- a/pkg/adaptor/cloud/libvirt/cloudinit.go
+++ b/pkg/adaptor/cloud/libvirt/cloudinit.go
@@ -1,0 +1,51 @@
+// (C) Copyright Confidential Containers Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package libvirt
+
+import (
+	"bytes"
+
+	"github.com/kdomanski/iso9660"
+)
+
+const (
+	userDataFilename   = "user-data"
+	metaDataFilename   = "meta-data"
+	vendorDataFilename = "vendor-data"
+	ciDataVolumeName   = "cidata"
+)
+
+// createCloudInit produces a cloud init ISO file as a data blob with a userdata and a metadata section
+func createCloudInit(userData, metaData []byte) ([]byte, error) {
+	writer, err := iso9660.NewWriter()
+	if err != nil {
+		return nil, err
+	}
+	defer writer.Cleanup() //nolint:errcheck // no need to check error in deferal
+
+	err = writer.AddFile(bytes.NewReader(userData), userDataFilename)
+	if err != nil {
+		return nil, err
+	}
+
+	err = writer.AddFile(bytes.NewReader(metaData), metaDataFilename)
+	if err != nil {
+		return nil, err
+	}
+
+	err = writer.AddFile(bytes.NewReader([]byte{}), vendorDataFilename)
+	if err != nil {
+		return nil, err
+	}
+
+	var buf bytes.Buffer
+
+	err = writer.WriteTo(&buf, ciDataVolumeName)
+	if err != nil {
+		return nil, err
+	}
+
+	// done
+	return buf.Bytes(), nil
+}
--- a/pkg/adaptor/cloud/libvirt/image.go
+++ b/pkg/adaptor/cloud/libvirt/image.go
@@ -8,13 +8,9 @@ package libvirt
 // Code copied from https://github.com/openshift/cluster-api-provider-libvirt
 
 import (
+	"bytes"
 	"fmt"
 	"io"
-	"net/http"
-	"net/url"
-	"os"
-	"strconv"
-	"strings"
 
 	libvirtxml "libvirt.org/go/libvirtxml"
 )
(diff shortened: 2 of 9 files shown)


pkg/adaptor/cloud/libvirt/image.go
```diff
@@ -25,109 +21,131 @@ type image interface {
 	string() string
 }
 
-type httpImage struct {
-	url *url.URL
+// type httpImage struct {
```
beraldoleal, line 24: If this code is no longer being used, you can remove it the same way you removed imports.

pkg/adaptor/cloud/libvirt/libvirt.go
```diff
@@ -9,56 +9,22 @@ import (
...
 
 	libvirt "libvirt.org/go/libvirt"
 	libvirtxml "libvirt.org/go/libvirtxml"
 )
 
-func createCloudInitISO(v *vmConfig, libvirtClient *libvirtClient) string {
-	logger.Printf("Create cloudInit iso\n")
-	cloudInitIso := libvirtClient.dataDir + "/" + v.name + "-cloudinit.iso"
+// createCloudInitISO creates an ISO file with a userdata and a metadata file. The ISO image will be created in-memory since it is small
+func createCloudInitISO(v *vmConfig) ([]byte, error) {
```
beraldoleal, line 21: All functions inside this file are libvirt related. If your proposal is changing the approach here, in terms of organization, maybe it makes sense to remove it from this file? Have you considered that?

pkg/adaptor/cloud/libvirt/volume.go
```diff
@@ -107,17 +105,17 @@ func newDefVolumeFromXML(s string) (libvirtxml.StorageVolume, error) {
...
+// 		ns, _ = strconv.Atoi(ts[1])
+// 	}
+// 	s, _ = strconv.Atoi(ts[0])
 
-	return time.Unix(int64(s), int64(ns))
-}
+// 	return time.Unix(int64(s), int64(ns))
+// }
 
 func uploadVolume(libvirtClient *libvirtClient, volumeDef libvirtxml.StorageVolume, img image) (volumeKey string, err error) {
```
beraldoleal, line 121: If we are going with this approach we probably should remove genisoimage from the following files, as well:
```

podvm/Dockerfile.podvm_builder.rhel
podvm/Dockerfile.podvm_builder.centos
Dockerfile
```

beraldoleal (commented):
Hello @caavery,

Thank you for this. I agree that moving this dependency from a package to a Go library could simplify the deployment process. Makes some sense to me.

I've left a few minor comments for your consideration, mostly related to the organization of the code.

pkg/adaptor/cloud/libvirt/libvirt.go
Carsten-Leue: I wanted to keep the changes to existing code as minimal as possible.

pkg/adaptor/cloud/libvirt/volume.go
wainersm: Dockerfile definitively as it will build the image where the cloud-api-adator service runs. I am not sure about `podvm/Dockerfile.podvm_builder.rhel` and `podvm/Dockerfile.podvm_builder.centos` as  packer might still need `genisoimage` to build the podvm qcow2.

Carsten-Leue:
> > @Carsten-Leue I will try to test this change still today.
> 
> My main concern at this point is with the changes on podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos . I tried to build the podvm with those changes but it failed due to an unrelated problem. I will put a 'hold' tag that should be lifted as soon as I test this change, ok?

@wainersm would it help if I simply undo the changes for  podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos? 

The only reason I put them in was this comment https://github.com/confidential-containers/cloud-api-adaptor/pull/1268#discussion_r1277909344

wainersm:
> > > @Carsten-Leue I will try to test this change still today.
> > 
> > 
> > My main concern at this point is with the changes on podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos . I tried to build the podvm with those changes but it failed due to an unrelated problem. I will put a 'hold' tag that should be lifted as soon as I test this change, ok?
> 
> @wainersm would it help if I simply undo the changes for podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos?
> 
> The only reason I put them in was this comment [#1268 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/pull/1268#discussion_r1277909344)

Yes @Carsten-Leue , please, remove those changes as I am not able to test them at the moment.

RedCarsten:
> > > > @Carsten-Leue I will try to test this change still today.
> > > 
> > > 
> > > My main concern at this point is with the changes on podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos . I tried to build the podvm with those changes but it failed due to an unrelated problem. I will put a 'hold' tag that should be lifted as soon as I test this change, ok?
> > 
> > 
> > @wainersm would it help if I simply undo the changes for podvm/Dockerfile.podvm_builder.rhel and podvm/Dockerfile.podvm_builder.centos?
> > The only reason I put them in was this comment [#1268 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/pull/1268#discussion_r1277909344)
> 
> Yes @Carsten-Leue , please, remove those changes as I am not able to test them at the moment.

Hi @wainersm 

alright, I have undone the changes for the mentioned files on this PR.

wainersm:
Hi @Carsten-Leue , I tested this pull request (ignored the changes of the podvm/Dockerfile.*) and I am getting the same error as reported in https://github.com/confidential-containers/cloud-api-adaptor/pull/1262#issuecomment-1668389479 . I'm investigating what is going wrong, my first hypothesis is that an update on kata runtime broke peer-pod's libvirt adaptor.

pkg/adaptor/cloud/libvirt/cloudinit.go
```diff
@@ -0,0 +1,51 @@
+// (C) Copyright Confidential Containers Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package libvirt
```
wainersm, line 4: Code wise, I have only one recommendation: we already have a common package for cloudinit (see `pkg/util/cloudinit`) , so what if move `createCloudInit()` and tests to that package?

RedCarsten: Hi @wainersm as far as I can tell the functionality in `pkg/util/cloudinit` deals with the creation of the `user-data` file while the function in the libvirt package is about the generation of the ISO file. 

I tend to not move that piece from libvirt, since that usecase to create the ISO in-memory and to assume an empty `vendor-data` is specific to the libvirt implementation.

wainersm:
hi @Carsten-Leue ! could you please re-send this PR with the conflicts resolved so that I can test it?

wainersm (approved):
hi @RedCarsten ! Thanks & sorry for taking longer to approve it.

wainersm:
@snir911 @bpradipt we need one more eyes here :)

CarstenLeue:
@wainersm @bpradipt I had to make another change to this PR, since I noticted that it was logging the content of the ISO file, which is not a good idea, since this could contain secrets.

```
logger.Printf("Uploading iso file: %s\n", isoVolName)
```

wainersm:
> @wainersm @bpradipt I had to make another change to this PR, since I noticted that it was logging the content of the ISO file, which is not a good idea, since this could contain secrets.
> 
> ```
> logger.Printf("Uploading iso file: %s\n", isoVolName)
> ```

ok, thanks. My approval stands and CI will finish soon to test that version does not break x86_64 (which won't for sure)

pkg/adaptor/cloud/libvirt/types.go
```diff
@@ -26,9 +26,9 @@ type vmConfig struct {
 	mem          uint
 	rootDiskSize uint64
 	userData     string
-	metaData     string
-	ips          []netip.Addr
-	instanceId   string //keeping it consistent with sandbox.vsi
+	// metaData     string
```
bpradipt, line 29: unintentional leftover ?

bpradipt: Rest looks good to me. Thanks @Carsten-Leue

bpradipt: @Carsten-Leue sorry It'll need a rebase again. And I think you'll need to remove the commented out variable as well and we should be good to merge this.

CarstenLeue: @bpradipt I removed the line and tried by best to rebase correctly. Could you please take a fresh look.

bpradipt (approved):
/lgtm
Thanks @CarstenLeue

Issue #1277, opened 2023-07-31
podvm: `error: package time v0.3.24 cannot be built because it requires rustc 1.67.0`


katexochen:
Error occurs when building with `AA_KBC=cc_kbc` :

```sh
#!/usr/bin/env bash
docker buildx build \
    -f Dockerfile.podvm_builder \
    -t "test-3107-build" \
    --no-cache \
    .
docker buildx build \
    -f Dockerfile.podvm_binaries \
    --build-arg "AA_KBC=cc_kbc" \
    --build-arg "BUILDER_IMG=test-3107-build" \
    .
```

Required by:
```
coco_keyprovider v0.1.0 (/guest-components/attestation-agent/coco_keyprovider)
└── reqwest v0.11.18
    ├── cookie v0.16.2
    │   ├── percent-encoding v2.3.0
    │   └── time v0.3.24
    └── cookie_store v0.16.2
        └── time v0.3.24 (*)

image-rs v0.1.0 (/guest-components/image-rs)
└── ocicrypt-rs v0.1.0 (/guest-components/ocicrypt-rs)
    └── josekit v0.8.3
        └── time v0.3.24 (*)
```

We are currently using rust 1.66.0. The AA is containerized with 1.67, but tests in guest-components run with rust stable, which is currently 1.71. Kata is using 1.69.

PR #1278, opened 2023-07-31
deps: upgrade rust to 1.69.0

katexochen:
Upgrading to the version currently used by kata. A dependency of attestation-agent required a upgrade to at least 1.67.

Fixes: #1277


bpradipt (approved):
/lgtm
Thanks @katexochen

Issue #1243, opened 2023-07-25
provisioner-ibmcloud: Add profile id support to the provisioner cli


Jordan9500:
As part of: https://github.com/confidential-containers/cloud-api-adaptor/issues/990. The ability to use trusted profile id's over ibmcloud api keys was added to the manual process. 

This should be added as an option to the ibmcloud provisioner cli and allow a user to use that instead.

PR #1279, opened 2023-07-31
provisioner-cli: Update provisioner files to include profile id support

tiashah1:
Uncomment profile id steps in kustomize.yaml
Add extra validation for steps which require API key to be set

Fixes: #1243

### DIFF
--- a/test/provisioner/provision_ibmcloud_initializer.go
+++ b/test/provisioner/provision_ibmcloud_initializer.go
@@ -22,6 +22,7 @@ import (
 type IBMCloudProperties struct {
 	IBMCloudProvider  string
 	ApiKey            string
+	IamProfileID      string
 	Bucket            string
 	CaaImageTag       string
 	ClusterName       string
@@ -63,6 +64,7 @@ func initProperties(properties map[string]string) error {
 	IBMCloudProps = &IBMCloudProperties{
 		IBMCloudProvider:  properties["IBMCLOUD_PROVIDER"],
 		ApiKey:            properties["APIKEY"],
+		IamProfileID:      properties["IAM_PROFILE_ID"],
 		Bucket:            properties["COS_BUCKET"],
 		CaaImageTag:       properties["CAA_IMAGE_TAG"],
 		ClusterName:       properties["CLUSTER_NAME"],
@@ -131,9 +133,10 @@ func initProperties(properties map[string]string) error {
 
 	log.Debugf("%+v", IBMCloudProps)
 
-	if len(IBMCloudProps.ApiKey) <= 0 {
-		return errors.New("APIKEY was not set.")
+	if len(IBMCloudProps.ApiKey) <= 0 && len(IBMCloudProps.IamProfileID) <= 0 {
+		return errors.New("APIKEY or IAM_PROFILE_ID must be set")
 	}
+
 	if len(IBMCloudProps.ResourceGroupID) <= 0 {
 		log.Info("[warning] RESOURCE_GROUP_ID was not set.")
 	}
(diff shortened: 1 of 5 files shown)


jtumber-ibm (approved):
Tested with specifying just IAM_PROFILE_ID for an existing cluster using

```
IBMCLOUD_PROVIDER="ibmcloud"
IAM_PROFILE_ID="---"
PODVM_IMAGE_ID="---"
INSTANCE_PROFILE_NAME="bz2-2x8"
CAA_IMAGE_TAG="7322a2ff36d57feec125ea1a4aadde3714c842d8"
```

Worked as expected. Also checked the same worked for APIKEY as before. 

I haven't tried the provisioner step though

test/provisioner/provision_ibmcloud_initializer.go
```diff
@@ -206,8 +208,14 @@ func initProperties(properties map[string]string) error {
 		}
 	}
 
-	if err := initVpcV1(); err != nil {
-		return err
+	if len(IBMCloudProps.ApiKey) > 0 {
+		if err := initClustersAPI(); err != nil {
+			return err
+		}
```
stevenhorsman, line 214: Why was this moved out of the `if strings.EqualFold(needProvisionStr` block? Doesn't it mean that `TEST_PROVISION` will be ignored and the cluster will be created whenever the api key is set?

stevenhorsman: So I think I've got this wrong after chatting to Jordan & Tia - this doesn't trigger the cluster creation, but just gets ready for it and it's needed for both provisioning and deprovisioning, so there is a change of behaviour, but it's because the deprovisioning would have failed previously which is a bit strange.

I think we basically want to get to the point that the provisioner-cli doesn't require any particular value of `TEST_PROVISION`, or `TEST_TEARDOWN` as we are explicitly calling those actions

test/provisioner/provision_ibmcloud_initializer.go
```diff
@@ -204,10 +203,20 @@ func initProperties(properties map[string]string) error {
 		if len(IBMCloudProps.CosServiceURL) <= 0 {
 			return errors.New("COS_SERVICE_URL was not set, example: s3.us.cloud-object-storage.appdomain.cloud")
 		}
+	} else {
```
mattarnoatibm, line 206: The `else` and `if` on the next line can be condensed into 1 line `else if`.

mattarnoatibm (changes requested):
Thanks for this contribution. I've made 1 review comment.

test/provisioner/provision_ibmcloud_initializer.go
stevenhorsman: In chatting through with Tia, she has tested that the provision and deprovision actions work irrespective of the `TEST_PROVISION`, or `TEST_TEARDOWN` values. I think there is probably some clarification which could be done more generally to improve the interaction between the environment variables and actions, but that is the outside the scope of this PR, so this conservation can be resolved

test/provisioner/provision_ibmcloud_initializer.go
tiashah1: Thanks, I have just made that change

mattarnoatibm (approved):
Looks good to me (LGTM)

stevenhorsman:
Can you `fixup`/`squash` the last three commits together please and update the commit message to reflect that it is fixing the `PODVM_IMAGE_ID` validation logic please? Thanks!

stevenhorsman (approved):
LGTM. Thanks @tiashah1!

Issue #1303, opened 2023-08-03
ibmcloud: Add wait condition for SSH key to be deleted completely in ibmcloud provisioner 


sudharshanibm3:
## Description
After completion of e2e test cases resources like clusters, SSH keys and VPC's must be deleted completely. Sometimes the VPC are failed to delete because ssh keys are still in use. 

```
time="2023-07-28T01:47:29Z" level=warning msg="Failed to delete vpc resources, err: SSH-key still in use by 1 instance(s). First instance ID: 0757_1ad6081d-xxxxxxx-xxxxx-xxxxx-xxxxxxxx."
```

PR #1258, opened 2023-07-27
ibmcloud: Wait condition for deleting subnets

sudharshanibm3:
Added wait condition to delete the subnet completely

Fixes: #1257

PR #1304, opened 2023-08-03
ibmcloud: wait condition for ssh deletion

sudharshanibm3:
Added wait condition for SSH key deletion

Fixes: #1303

### DIFF
--- a/test/provisioner/provision_ibmcloud.go
+++ b/test/provisioner/provision_ibmcloud.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"crypto/sha256"
+
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 
 	"github.com/confidential-containers/cloud-api-adaptor/test/utils"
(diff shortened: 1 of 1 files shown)


jtumber-ibm (approved):

stevenhorsman (commented):
Can we get a bit more information on this please in the commit message and/or the issue e.g. What is the problem we were seeing that means we want to wait 5 mins for the ssh delete to take place and after 5 minutes of waiting why do we just continue and not error if it's still not cleaned up? Thanks

sudharshanibm3:
Hi @stevenhorsman 
After completion of e2e test cases resources like clusters, SSH keys and VPC's must be deleted completely. Sometimes the VPC are failed to delete because ssh keys are still in use. 

```
time="2023-07-28T01:47:29Z" level=warning msg="Failed to delete vpc resources, err: SSH-key still in use by 1 instance(s). First instance ID: 0757_1ad6081d-xxxxxxx-xxxxx-xxxxx-xxxxxxxx."
```
In order to validate SSH key deletion, we added an extra loop for 5 minutes and breaks if its completely deleted
Implemented the same logic used for deletion of subnets - [PR](https://github.com/confidential-containers/cloud-api-adaptor/pull/1258)

stevenhorsman:
Thanks for the extra info, but in both the VPC and ssh key case, shouldn't the behaviour be that if, after 5mins the delete still hasn't worked, we should error rather than just carrying on? We'd hope that the delete won't take that long, but if it does then we want to avoid just hitting the error later on?

sudharshanibm3:
Hi @stevenhorsman ,
Modified the logic little bit now,
- If SSH key deleted completely, it returns nil
- else it will return error directly

test/provisioner/provision_ibmcloud.go
```diff
@@ -479,8 +480,23 @@ func deleteSshKey() error {
...
 		return err
 	}
-	log.Infof("SSH Key %s with ID %s is deleted.", IBMCloudProps.SshKeyName, IBMCloudProps.SshKeyID)
-	return nil
+	waitMinutes := 5
+	log.Infof("Waiting for SSH key  %s to be removed...", IBMCloudProps.SshKeyID)
+	for i := 0; i <= waitMinutes; i++ {
+		foundSsh, err := findSshKey(IBMCloudProps.SshKeyID)
+		if foundSsh == nil {
+			log.Infof("SSH %s is removed  ...", IBMCloudProps.SshKeyName)
```
stevenhorsman, line 488: Do we still this this line given we have the log below it that gives more info?

stevenhorsman (commented):
Thanks for the improvement. I've added a minor comment. If you can also add more description to the commit body that would be good as well. Something like:
```
- Add code to wait and check for the SSH key to be deleted to
avoid errors in later delete steps
- Wait for up to 5mins and return an error if the delete hasn't
completed
```
?

stevenhorsman (approved):
LGTM. Thanks for the updates!

stevenhorsman:
The Azure e2e tests are new and have never passed (https://github.com/confidential-containers/cloud-api-adaptor/actions/workflows/azure-e2e-test.yml), so I will merge this without them.

kartikjoshi21:
@stevenhorsman e2e-test were failing due to quota exceeded issue while building podvm image. It is passing now for other PRs. Thanks

Issue #1164, opened 2023-07-10
test/e2e/libvirt: allow to overwrite the Libvirt connection


wainersm:
Currently the URI used to connect to Libvirt is hard-coded to `qemu:///system`.  Let's allow to overwrite the URI (e.g. you might want to connect to a remote instance) via properties file.

This feature is very similar to https://github.com/confidential-containers/cloud-api-adaptor/issues/1163

wainersm:
The exact line where the connection is set: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/provision_libvirt.go#L68

bookinabox:
Thought I had made this PR a while ago, but I guess I did not.

PR #1307, opened 2023-08-04
test/libvirt: allow libvirt conn URI to be specified

bookinabox:
The libvirt tests currently assumes that it is always running on a local instance and hard codes "qemu:///system" as the URI to create the Connection object with.

This allows us to specify libvirt_conn_uri in the libvirt properties file to change it to a different instance.

Also updates test/provisioner/e2e/README.md to add this option to the documentation and clarify the difference with libvirt_uri.

Fixes: #1164

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -58,8 +58,9 @@ Use the properties on the table below for Libvirt:
 |---|---|---|
 |libvirt_network|Libvirt Network|"default"|
 |libvirt_storage|Libvirt storage pool|"default"|
-|libvirt_url|Libvirt connection URI|"qemu+ssh://root@192.168.122.1/system?no_verify=1"|
 |libvirt_vol_name|Volume name|"podvm-base.qcow2"|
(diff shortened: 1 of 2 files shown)


wainersm (approved):
LGTM. Thanks @bookinabox !

bookinabox:
I may need to modify install_operator as well.
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/libvirt/install_operator.sh#L102

bookinabox:
I have been having issues where the container is not properly created on the remote machine. Specifically, working through the libvirt README, the peer-pod is created and pod is created, but the pod gets stuck on containerCreating and will restart, spawning new vms each time. Unsure if this is an issue on my environment, but will investigate further.

wainersm:
@bpradipt @snir911 can I have your review here? I tested it, working fine and @bookinabox has used it to develop https://github.com/confidential-containers/cloud-api-adaptor/pull/1359

Issue #1311, opened 2023-08-08
ibmcloud: Error while unmarshalling CAA tag


sudharshanibm3:
## Description:
While running e2e-test cases, following error gets logged while unmarshalling CAA tag response
```
time="2023-08-08T00:29:51Z" level=error msg="json: cannot unmarshal number into Go struct field .tags.size of type string"
```

sudharshanibm3:
- On looking into the current code, the `QuayTagResponse` struct has property `size` which is set to `string` - [Ref](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/test/provisioner/provision_ibmcloud_kustomize.go#L29-L40)
- Also checked the API response of quay tags, the type of `size` is either `0` or `null`. Hence we are getting unmarshall error logs whenever the size is `0`
![Screenshot 2023-08-08 at 3.52.19 PM.png](https://zenhub.ibm.com/images/63d0c1b66bae02b9f5905ea2/b4f73878-5abe-46ae-ab72-7aea25e221b9)

PR #1312, opened 2023-08-08
ibmcloud: Auto update CAA tag

sudharshanibm3:
Removed extra properties in tag response other than Name and Manifest

Fixes: #1311

### DIFF
--- a/test/provisioner/provision_ibmcloud_kustomize.go
+++ b/test/provisioner/provision_ibmcloud_kustomize.go
@@ -28,13 +28,8 @@ type IBMCloudInstallOverlay struct {
 
 type QuayTagsResponse struct {
 	Tags []struct {
-		Name       string `json:"name"`
-		ModifiedAt string `json:"last_modified"`
-		Digest     string `json:"manifest_digest"`
-		Size       string `json:"size"`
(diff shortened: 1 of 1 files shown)


test/provisioner/provision_ibmcloud_kustomize.go
```diff
@@ -28,13 +28,9 @@ type IBMCloudInstallOverlay struct {
...
-		Name       string `json:"name"`
-		ModifiedAt string `json:"last_modified"`
-		Digest     string `json:"manifest_digest"`
-		Size       string `json:"size"`
-		Manifest   bool   `json:"is_manifest_list"`
-		Reversion  bool   `json:"reversion"`
-		StartTime  int    `json:"start_ts"`
+		Name     string                 `json:"name"`
+		Manifest bool                   `json:"is_manifest_list"`
+		Others   map[string]interface{} `json:"-"`
```
jtumber-ibm, line 33: Happy to be corrected, but I don't think `Others` gets populated with any values when unmarshalling so this will always be empty. I think it is clearer to just have Name and Manifest if that is all we need.

jtumber-ibm (approved):

Issue #1336, opened 2023-08-22
ibmcloud | doc: Update documentation for Test CSI Wrapper with ibm-vpc-block-csi-driver for peer pod demo on IBM Cloud VPC


sudharshanibm3:
- Test CSI Wrapper with ibm-vpc-block-csi-driver for peer pod demo on IBM Cloud VPC - [ref](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/volumes/csi-wrapper/examples/ibm/README.md#run-the-csi-wrapper-for-peerpod-storage-demo)
- The path for deployment yaml files has been changed from `/hack/ibm` to `examples/ibm`. Update the readme file with new file path.

PR #1337, opened 2023-08-22
ibmcloud: Updated path in Test CSI Wrapper documentation

sudharshanibm3:
Updated path from hack to examples in Test CSI Wrapper with ibm-vpc-block-csi-driver for peer pod demo on IBM Cloud VPC  documentation

Fixes: #1336


genjuro214 (approved):
LGTM

PR #1338, opened 2023-08-22
Azure: enable podvm images nightly build

kartikjoshi21:


.github/workflows/azure-podvm-image.yml
```diff
@@ -0,0 +1,122 @@
...
+      with:
+        # The paths to cache are documented here: https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci
+        path: |
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+        key: rust
```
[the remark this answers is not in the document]
surajssd: I think this cache is worth it, because it saves a few double digit seconds every time those files are downloaded. And yes, we do hit this cache because they key is static. Also those recommendations are from rust upstream: 
image

ci-infra/azure/main.tf
```diff
@@ -72,6 +72,16 @@ resource "azurerm_shared_image_gallery" "podvm_image_gallery" {
...
   resource_group_name = azurerm_resource_group.ci_rg.name
   location            = azurerm_resource_group.ci_rg.location
+
+  sharing {
+    permission = "Community"
+    community_gallery {
+	prefix = "cocopodvm"
+	eula   = "https://raw.githubusercontent.com/confidential-containers/confidential-containers/main/LICENSE"
+	publisher_uri = "https://github.com/confidential-containers/confidential-containers"
+	publisher_email = "kartikjoshi@microsoft.com"
```
surajssd, line 82: We need some generic email id isn't it?

ci-infra/azure/main.tf
```diff
@@ -87,4 +97,5 @@ resource "azurerm_shared_image" "podvm_image" {
   }
   hyper_v_generation        = "V2"
   confidential_vm_supported = true
+
```
surajssd, line 100: Seems unrelated? 
```suggestion
```

ci-infra/azure/outputs.tf
```diff
@@ -33,3 +33,7 @@ output "AZURE_PODVM_IMAGE_DEF_NAME" {
 output "AZURE_MANAGED_IDENTITY_NAME" {
   value = azurerm_user_assigned_identity.gh_action_user_identity.name
 }
+
+output "AZURE_COMMUNITY_GALLERY_NAME" {
+  value = azurerm_shared_image_gallery.podvm_image_gallery.sharing[0].community_gallery[0].name
+}
```
surajssd, line 39: We need to get these terraform changes in, before we merge the workflow. Because without the terraform changes, the workflow will publish an image but it won't be public. Which should not fail per se, but then we have to run the terraform commands before UTC midnight 😆

kartikjoshi21: Sooo terraform commit will get merged first if we merge this whole PR that should suffice right ?

ci-infra/azure/main.tf
kartikjoshi21: Yess it should be something generic, i saw other community gallery there we are using Magnus's id so for time being i used this one. What do you suggest which one we should use?

ci-infra/azure/outputs.tf
surajssd: I think it is ok, we can merge it together. We have a small window before the workflow runs.

ci-infra/azure/main.tf
surajssd: We can use the coco email id: cocoatmsft@outlook.com

surajssd: or if @bpradipt has some suggestions on this one?

bpradipt: [cocoatmsft@outlook.com](mailto:cocoatmsft@outlook.com) looks fine.

mkulke (commented):
Could we turn this into a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to share the code with the azure e2e test?

surajssd:
> Could we turn this into a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to share the code with the azure e2e test?

We should definitely do that. We don't have to maintain code in two places for the same thing.

mkulke:
> > Could we turn this into a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to share the code with the azure e2e test?
> 
> We should definitely do that. We don't have to maintain code in two places for the same thing.

I took the commits from this PR and consolidated it with the e2e test workflow to share the build job.

kartikjoshi21:
Merged along with [This](https://github.com/confidential-containers/cloud-api-adaptor/pull/1391#pullrequestreview-1609021629) !!

PR #1341, opened 2023-08-22
workflow: retry caa image build

wainersm:
The CAA image build on push is very important to not fail otherwise features and fixes don't properly get into devel/release cycle. But we recently have noticed errors like:

ERROR: failed to solve: DeadlineExceeded: failed to push quay.io/confidential-containers/cloud-api-adaptor:b277c1ae739ae318926ef4f39c88fac16050d13a: no active session for pj7teq3k1b3tqjew1y3tf5mgp: context deadline exceeded

Error that goes away when the workflow is manually re-trigger after a few minutes. So this added a retry mechanism that will try again the build command after 2 minutes and for the maximum of 3 tentatives.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -49,9 +49,16 @@ jobs:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
       - name: Build and push image
-        run: |
-          if [ ${{ matrix.type }} == "release" ]; then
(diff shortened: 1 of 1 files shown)


wainersm:
And obviously the commit's body length line exceeded the maximum allowed :D

stevenhorsman (approved):
LGTM and not a very invasive solution which is great. Thanks Wainer!

jtumber-ibm (approved):

PR #1359, opened 2023-08-25
WIP: Libvirt SEV-SNP support

bookinabox:
WIP Adds support for SEV-SNP.

For reference: https://libvirt.org/kbase/launch_security_sev.html

I currently also have a commit for remote libvirt connection cherry-picked on for testing purposes (using a remote machine with sev support, but running the cluster on my machine at the moment)

If LAUNCH_SECURITY is specified as "sev" and the machine suppports sev, it adds the necessary changes.

Some questions I have on the implementation:

- I have the sev changes be modifications of the x86_64 domainxml creation. I was debating whether it would be better to just create a whole separate domainxml creation function to isolate it more. I especially don't feel very comfortable with how I overwrote one of the disks and think that it might be better to make everything explicit. Thoughts? 
- The reason the disk had to be overwritten was because q35 machines do not support IDE connections. I tried connecting a disk with SATA, but it threw an error (bug?), so I switched to SCSI. Let me know if there are better ways to do this.
- In terms of guest policy (https://libvirt.org/formatdomain.html#launch-security), I left everything as disabled because I assumed this was currently mostly being used for testing and being permissive would be fine. Should this be set to anything specific or be specifiable?
- I assumed that secure boot should be on for the DomainLoader. Also, set it as stateless for boot attestation.
-  In https://libvirt.org/kbase/launch_security_sev.html#memory, it mentions two ways of addressing the extra memory needed. We chose to add 512 MiB on top of the 8 GiB and it seems boot successfully.

I need help determining what exactly I should be testing.

in terms of manual testing, I have been building an image and uploading it to a separate quay repository. I was having trouble setting up a cluster on a remote machine, so I am setting up a cluster locally and having the peer-pod vm be created on the remote machine. However, I have also been having trouble in containerCreating when creating a remote instance. Also, if you try this route, note that install_operator.sh assumes that you are not using a remote_uri and will overwrite libvirt_uri in the kustomization file with the local ip and user, so that will need to be manually changed back.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -117,6 +117,7 @@ ibmcloud_powervs() {
 libvirt() {
     test_vars LIBVIRT_URI
 
+    [[ "${DISABLECVM}" ]] && optionals+="-disable-cvm "
     set -x
     exec cloud-api-adaptor libvirt \
         -uri "${LIBVIRT_URI}" \
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/cri-api v0.23.1
 	libvirt.org/go/libvirt v1.8002.0
-	libvirt.org/go/libvirtxml v1.8002.0
+	libvirt.org/go/libvirtxml v1.9004.0
 )
 
 require (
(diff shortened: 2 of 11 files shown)


bpradipt:
I have few thoughts from a general design standpoint.
1. How about using `disable-cvm` CLI option (like in Azure provider) to indicate whether to launch a regular VM or a confidential VM ? This will bring in consistency.
2. Detecting whether the host/libvirt supports SEV, SEV-SNP or TDX capabilities in the code itself and based on that creating the VM using correct options

cc @wainersm @snir911

bookinabox:
> I have few thoughts from a general design standpoint.
> 
>     1. How about using `disable-cvm` CLI option (like in Azure provider) to indicate whether to launch a regular VM or a confidential VM ? This will bring in consistency.
> 
>     2. Detecting whether the host/libvirt supports SEV, SEV-SNP or TDX capabilities in the code itself and based on that creating the VM using correct options
> 
> 
> cc @wainersm @snir911

Thanks for the comments!

Afaik, libvirt has a single flag for whether sev is supported and I don't see anything about tdx.

bookinabox:
It does not actually look like libvirt currently has support for SNP, but I will just get SEV and the general confidential vm enabling for libvirt done.

bookinabox:
Moving this to #1397 since it looks like SNP is not supported yet.

PR #1366, opened 2023-08-28
podvm: install imds script also for the generic build

snir911:


aws/image/misc-settings.sh
```diff
@@ -38,85 +38,37 @@ then
...
+            apt-get update && apt-get install iptables -y
+            ;;
+        *)
+            echo "\"iptables\" is missing and cannot be installed, setup-nat-for-imds.sh is likely to fail 1>&2"
+            ;;
+        esac
+    fi
+
+    # Add a link to the service in multi-user.target.wants
+    ln -s /etc/systemd/system/setup-nat-for-imds.service /etc/systemd/system/multi-user.target.wants/setup-nat-for-imds.service
```
mkulke, line 71:
```suggestion
    systemctl enable setup-nat-for-imds
```

aws/image/misc-settings.sh
```diff
@@ -38,85 +38,37 @@ then
     esac
 fi
 
-if [[ "$CLOUD_PROVIDER" == "azure" && "$PODVM_DISTRO" == "ubuntu" ]]; then
+if [[ "$CLOUD_PROVIDER" == "azure" || "$CLOUD_PROVIDER" == "generic" ]] && [[ "$PODVM_DISTRO" == "ubuntu" ]]; then
```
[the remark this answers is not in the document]
mkulke: In this case I would also consider removing the iptables check and install it for every distro.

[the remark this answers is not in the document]
snir911: We don't have a target for it but we pass "generic" as the value for CLOUD_PROVIDER in the [workflow](https://github.com/confidential-containers/cloud-api-adaptor/blob/4a6d01090e808325bfa7b366cc1def43721f9d35/.github/workflows/podvm.yaml#L60)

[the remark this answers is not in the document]
bpradipt: Ok, I see. We use the `generic` in the workflow to add a prefix for the generated artefacts. But using `generic` might be confusing here. Or we'll have to add the description of `generic` cloud provider in the readme. Wdyt ?
What about removing the cloud provider check or just checking if it's not vsphere before adding the imds service ?

podvm/qcow2/misc-settings.sh
```diff
@@ -38,85 +38,37 @@ then
     esac
 fi
 
-if [[ "$CLOUD_PROVIDER" == "azure" && "$PODVM_DISTRO" == "ubuntu" ]]; then
+if [[ "$CLOUD_PROVIDER" == "azure" || "$CLOUD_PROVIDER" == "generic" ]] && [[ "$PODVM_DISTRO" == "ubuntu" ]]; then
     export DEBIAN_FRONTEND=noninteractive
+    curl -fsSL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo apt-key add -
+    echo "deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+
```
yoheiueda, line 45: I think this `podvm/qcow2/misc-settings.sh` is also used when s390x images are built with Packer.

@liudalibj @huoqifeng Is this correct?

liudalibj: yeah, the script be used at here https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/qcow2/ubuntu/qemu-ubuntu.pkr.hcl#L86
when s390x images are built with packer.

snir911: I won't be installed as CLOUD_PROVIDER=ibmcloud in this case, no?

aws/image/misc-settings.sh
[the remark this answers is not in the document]
snir911: If we are planning to have generic artifacts i think this would be the better option.
IIUC there's no problem with vSphere, the only reason it's separated is additional packages being installed (mabye we can install them for the generic use case? dunnu)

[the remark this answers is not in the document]
bpradipt: vSphere may need more additional packages. Who knows. Better to keep it separate imho

podvm/qcow2/misc-settings.sh
snir911: @yoheiueda @liudalibj should i remove it or this is ok?

liudalibj: @snir911 yeah, this is OK.

bpradipt (approved):
/lgtm
nit: I think updating the podvm image README to mention about the CLOUD_PROVIDER=generic option will be good.

podvm/qcow2/misc-settings.sh
liudalibj: it's ok here, I did a test to build one s390x image with this PR, the related logs
```
==> qemu.ubuntu: Uploading qcow2/misc-settings.sh => ~/misc-settings.sh
==> qemu.ubuntu: Provisioning with shell script: /tmp/packer-shell122513875
    qemu.ubuntu: passwd: password expiry information changed.
==> qemu.ubuntu: Removed /etc/systemd/system/timers.target.wants/apt-daily.timer.
==> qemu.ubuntu: Removed /etc/systemd/system/timers.target.wants/apt-daily-upgrade.timer.
==> qemu.ubuntu: Removed /etc/systemd/system/multi-user.target.wants/snapd.service.
==> qemu.ubuntu: Removed /etc/systemd/system/multi-user.target.wants/snapd.seeded.service.
==> qemu.ubuntu: Removed /etc/systemd/system/cloud-final.service.wants/snapd.seeded.service.
==> qemu.ubuntu: Removed /etc/systemd/system/multi-user.target.wants/snap.lxd.activate.service.
==> qemu.ubuntu: Uploading qcow2/build-s390x-se-image.sh => ~/build-s390x-se-image.sh
```

liudalibj: And when building one s390x podvm image the `CLOUD_PROVIDER` is empty inside the podvm_builder container.

liudalibj (approved):
LGTM

yoheiueda (approved):
LGTM

PR #1371, opened 2023-08-29
1092: Skip checking properties unnecessary for image upload are set if the provisioner only does the upload image action

### LINKED ISSUE #1092: Skip checking properties unnecessary for image upload are set if the provisioner only does the upload image action
If the provisoner CLI is only used to upload an image to the cloud provider (`caa-provisioner-cli -action=uploadimage`) then for the IBM Cloud provider the provisioner CLI requires properties to be set that shouldn't be necessary to upload the image to COS and create a VPC Custom Image from the image uploaded to COS.

If the environment variable `TEST_PROVISION` is set to `yes`, the properties file needs the following properties set, which shouldn't be needed:

`WORKER_OPERATION_SYSTEM`  - shouldn't be needed because when just usin upload image you're not creating a cluster
`KUBE_VERSION` - shouldn't be needed because when just usin upload image you're not creating a cluster
`ZONE` - shouldn't be needed because the custom images are not scoped to a single zone in a region, but they are scoped to a particular region

If the environment variable `TEST_PROVISION` is set to `no`, the properties file needs the following properties set, which shouldn't be needed

`PODVM_IMAGE_ID` - shouldn't be needed because we're uploading a new image whose ID won't be known until the new image has been created
`SSH_KEY_ID`
`VPC_SUBNET_ID`
`VPC_SECURITY_GROUP_ID`
`VPC_ID` - the last 4 parameters relate to Kubernetes clusters or VSIs and shouldn't be needed when creating a custom image

The Azure cloud provider may have similar issues to this if it supports the upload image action.

tiashah1:
Skip checking properties unnecessary for image upload are set if the provisioner only does the upload image action

Use the Action variable as a way to check if uploadimage is being performed 
Skip checks for unneeded variables if this is the action set
Set TEST_PROVISION to yes if the action is only uploadimage

Fixes #1092 

Signed-off-by: Tia Shah <tia.shah@ibm.com>

### DIFF
--- a/test/provisioner/provision_ibmcloud_initializer.go
+++ b/test/provisioner/provision_ibmcloud_initializer.go
@@ -135,10 +135,6 @@ func initProperties(properties map[string]string) error {
 
 	log.Debugf("%+v", IBMCloudProps)
 
-	if len(IBMCloudProps.ApiKey) <= 0 && len(IBMCloudProps.IamProfileID) <= 0 {
-		return errors.New("APIKEY or IAM_PROFILE_ID must be set")
-	}
-
 	if len(IBMCloudProps.ResourceGroupID) <= 0 {
 		log.Info("[warning] RESOURCE_GROUP_ID was not set.")
 	}
@@ -169,13 +165,16 @@ func initProperties(properties map[string]string) error {
 	log.Infof("IksServiceURL is: %s.", IBMCloudProps.IksServiceURL)
 
 	needProvisionStr := os.Getenv("TEST_PROVISION")
-	if strings.EqualFold(needProvisionStr, "yes") || strings.EqualFold(needProvisionStr, "true") {
+	if strings.EqualFold(needProvisionStr, "yes") || strings.EqualFold(needProvisionStr, "true") || Action == "uploadimage" {
 		if len(IBMCloudProps.ApiKey) <= 0 {
 			return errors.New("APIKEY is required for provisioning")
 		}
 		if len(IBMCloudProps.Region) <= 0 {
 			return errors.New("REGION was not set.")
 		}
+	}
+
+	if strings.EqualFold(needProvisionStr, "yes") || strings.EqualFold(needProvisionStr, "true") {
 		if len(IBMCloudProps.KubeVersion) <= 0 {
 			return errors.New("KUBE_VERSION was not set, get it via command: ibmcloud cs versions")
 		}
@@ -213,6 +212,10 @@ func initProperties(properties map[string]string) error {
 		}
 	} else if len(IBMCloudProps.PodvmImageID) <= 0 {
 		return errors.New("PODVM_IMAGE_ID was not set, set it with existing custom image id in VPC")
+	}
+
+	if len(IBMCloudProps.ApiKey) <= 0 && len(IBMCloudProps.IamProfileID) <= 0 {
+		return errors.New("APIKEY or IAM_PROFILE_ID must be set")
 	}
 
 	if len(IBMCloudProps.ApiKey) > 0 {
--- a/test/tools/provisioner-cli/main.go
+++ b/test/tools/provisioner-cli/main.go
@@ -43,15 +43,17 @@ func main() {
 		installDirectory = "../../install"
 	}
 
+	action := flag.String("action", "provision", "string")
+	flag.Parse()
+
+	pv.Action = *action
+
 	cfg := envconf.New()
 
 	provisioner, err := pv.GetCloudProvisioner(cloudProvider, provisionPropsFile)
 	if err != nil {
 		log.Fatal(err)
 	}
-
-	action := flag.String("action", "provision", "string")
-	flag.Parse()
 
 	if *action == "provision" {
 		log.Info("Creating VPC...")
@@ -97,6 +99,9 @@ func main() {
 
 	if *action == "uploadimage" {
 		log.Info("Uploading PodVM Image...")
+		if len(podvmImage) <= 0 {
+			log.Fatal("Environment variable TEST_PODVM_IMAGE must be set")
+		}
 		if _, err := os.Stat(podvmImage); os.IsNotExist(err) {
 			log.Fatal(err)
 		}
--- a/test/provisioner/common.go
+++ b/test/provisioner/common.go
@@ -21,6 +21,8 @@ type patchLabel struct {
 	Value string `json:"value"`
 }
 
+var Action string
+
 // Adds the worker label to all workers nodes in a given cluster
 func AddNodeRoleWorkerLabel(ctx context.Context, clusterName string, cfg *envconf.Config) error {
 	fmt.Printf("Adding worker label to nodes belonging to: %s\n", clusterName)


test/tools/provisioner-cli/main.go
```diff
@@ -97,6 +99,7 @@ func main() {
 
 	if *action == "uploadimage" {
 		log.Info("Uploading PodVM Image...")
+		os.Setenv("TEST_PROVISION", "yes")
```
mattarnoatibm, line 102: You'll want to set this, if the action is the `uploadimage` action, before getting the cloud provisioner as the get function initializes the cloud provisioner.

test/tools/Makefile
```diff
@@ -27,4 +27,4 @@ caa-provisioner-cli: force ## Build provisioner-cli for developer
 
 .PHONY: clean
 clean: ## Remove binaries.
-	rm -f $(BINARIES)
+	rm -f $(BINARIES)
```
mattarnoatibm, line 30: NP, add the final newline back.

test/tools/provisioner-cli/main.go
```diff
@@ -97,6 +99,7 @@ func main() {
 
 	if *action == "uploadimage" {
 		log.Info("Uploading PodVM Image...")
```
mattarnoatibm, line 101: Optional improvement - before stating the `podvmImage` file you could check `podvmImage` is not a zero length string, and error if it is zero length.

The error would be a hint that you need to set the `TEST_PODVM_IMAGE` environment variable when running the image upload action.

mattarnoatibm (changes requested):
Thanks for this contribution.

I left some review comments.

test/tools/Makefile
```diff
@@ -28,3 +28,4 @@ caa-provisioner-cli: force ## Build provisioner-cli for developer
 .PHONY: clean
 clean: ## Remove binaries.
 	rm -f $(BINARIES)
+
```
mattarnoatibm, line 31: Remove the tab character on the final newline.

mattarnoatibm (changes requested):
One NP left for my review comments.

mattarnoatibm:
Note - we don't have to put the issue number of the PR comments because its in the commit info already, as the `Fixes: #<issue_number>` line.

mattarnoatibm (approved):
LGTM

test/provisioner/provision_ibmcloud_initializer.go
```diff
@@ -176,11 +176,13 @@ func initProperties(properties map[string]string) error {
 		if len(IBMCloudProps.Region) <= 0 {
 			return errors.New("REGION was not set.")
 		}
-		if len(IBMCloudProps.KubeVersion) <= 0 {
-			return errors.New("KUBE_VERSION was not set, get it via command: ibmcloud cs versions")
-		}
-		if len(IBMCloudProps.WorkerOS) <= 0 {
-			return errors.New("WORKER_OPERATION_SYSTEM was not set, set it like: UBUNTU_20_64, UBUNTU_18_S390X")
+		if Action != "uploadimage" {
```
stevenhorsman, line 179: I would prefer for us to re-think this code and remove validation for the initializer and instead create some sort of runners interface with implementations for provision cluster, install, upload image, uninstall, deprovision cluster that would have (at least) a validate and 'run' function that could do the checks. However that is a big change, so I think this is fine for now.

test/tools/provisioner-cli/main.go
```diff
@@ -43,16 +43,22 @@ func main() {
...
 
+	action := flag.String("action", "provision", "string")
+	flag.Parse()
+
+	pv.Action = *action
+
 	cfg := envconf.New()
 
+	if *action == "uploadimage" {
+		os.Setenv("TEST_PROVISION", "yes")
```
stevenhorsman, line 54: What's the reason for setting TEST_PROVISION to "yes" for uploadImage. It feels like `TEST_PROVISION` should be linked to cluster creation rather than image actions? Is it because of some other tech debt restriction from elsewhere? If so, and it can't be updated, can you add this to the commit message body please, or as a comment in the code. Thanks!

stevenhorsman (changes requested):
Can you update the commit header to something like:
`test: Skip checking unnecessary properties` as it should be less than 50 chars and start with the "area of code being changed"

test/tools/provisioner-cli/main.go
mattarnoatibm: I suggested this addition. The reason I had in mind is if `TEST_PROVISION` is not set to `yes` or `true` then in `provision_ibmcloud_initializer.go` we check the existing VPC infrastructure parameters like `SSH_KEY_ID` and `VPC_SUBNET_ID` are set, but not `REGION` and `APIKEY` which you need to upload a new Custom Image.

If you want to be able to use the `uploadimage` action without setting `TEST_PROVISION` to `yes`, then for properties that are only checked when `TEST_PROVISION` is not set to `yes` you need, if you're doing `uploadimage`, to check `REGION` and `APIKEY` are set.

I don't suggest adding those whole paragraphs in as a code comment but we need to find an elegant way to work around the mix of the action parameter for the provisioner CLI and the environment variables.

Or maybe we solve the above by doing a bigger piece of refactoring out the use of environment variables in the provisoner CLI `main` package and the `provisioner` package where the properties are checked.

stevenhorsman: I've chatted to Tia about whether, instead of setting `TEST_PROVISION` in the main, we can update the `REGION` and `APIKEY` validation checks in `provision_ibmcloud_initializer.go` to run if `TEST_PROVISION=yes`, *or* `action == "uploadimage"`. I think this change to reduce the reliance on `TEST_PROVISION` is probably better and keeps the validation logic together.

test/provisioner/provision_ibmcloud_initializer.go
```diff
@@ -169,6 +169,15 @@ func initProperties(properties map[string]string) error {
 	log.Infof("IksServiceURL is: %s.", IBMCloudProps.IksServiceURL)
 
 	needProvisionStr := os.Getenv("TEST_PROVISION")
+	if strings.EqualFold(needProvisionStr, "yes") || strings.EqualFold(needProvisionStr, "true") || Action == "uploadimage" {
+		if len(IBMCloudProps.ApiKey) <= 0 {
```
stevenhorsman, line 173: As we are testing APIKEY and REGION here, we don't need them in the if block below anymore, so we should either remove them from below, or update this if check to just be `Action == "uploadimage"`. I think I slightly prefer the former, as there is less copy-paste, but I'm not too worried

tiashah1: Agree, removed it from the second check

stevenhorsman (approved):
The code looks good to me and I've tested it in the following ways:
- Happy path:
```
# ./caa-provisioner-cli -action=uploadimage
INFO[0000] IamServiceURL is: https://iam.cloud.ibm.com/identity/token.
INFO[0000] VpcServiceURL is: https://eu-gb.iaas.cloud.ibm.com/v1.
INFO[0000] IksServiceURL is: https://containers.cloud.ibm.com/global.
INFO[0001] Uploading PodVM Image...
INFO[0001] cos session initialized.
INFO[0004] qcow2 image file /root/go/src/github.com/confidential-containers/cloud-api-adaptor/podvm-s390x.qcow2 validated, sha256sum: 534a65f0cff85a78f2cad7cbc5be896503349a62254476db64300932a44e7e82.
INFO[0019] cosID cos://eu-gb/sh-pp-cos-bucket-eu-gb/podvm-s390x.qcow2, imageName podvm-s390x.
INFO[0022] Image podvm-s390x with PodvmImageID r018-262b71be-6008-4aa8-a7d2-b89db4c4b562 created from the bucket.
```
- Checking that the validation works for `TEST_PODVM_IMAGE`
```
# export TEST_PODVM_IMAGE=""
root@sh-pp-e2e1:~/go/src/github.com/confidential-containers/cloud-api-adaptor/test/tools# ./caa-provisioner-cli -action=uploadimage
INFO[0000] IamServiceURL is: https://iam.cloud.ibm.com/identity/token.
INFO[0000] VpcServiceURL is: https://eu-gb.iaas.cloud.ibm.com/v1.
INFO[0000] IksServiceURL is: https://containers.cloud.ibm.com/global.
INFO[0000] Uploading PodVM Image...
FATA[0000] Environment variable TEST_PODVM_IMAGE must be set
```
- Set APIKEY="" in properties:
```
# ./caa-provisioner-cli -action=uploadimage
FATA[0000] APIKEY or IAM_PROFILE_ID must be set
```
There is an argument that this error message is incorrect as the IAM_PROFILE_ID isn't enough, but if I do set it then I then get the correct error message:
```
./caa-provisioner-cli -action=uploadimage
INFO[0000] IamServiceURL is: https://iam.cloud.ibm.com/identity/token.
INFO[0000] VpcServiceURL is: https://eu-gb.iaas.cloud.ibm.com/v1.
INFO[0000] IksServiceURL is: https://containers.cloud.ibm.com/global.
FATA[0000] APIKEY is required for provisioning
```
If we'd like to change the validation in the future then it might be overkill to fix this up now, so I'll leave it to Tia to decide if it's worth pulling into scope.

stevenhorsman (approved):
LGTM. Thanks!

tiashah1:
I am unable to merge this as I do not have write access to the repository. Is there someone I can ask to merge / get the access required for this?
Thanks

PR #1388, opened 2023-08-31
Podvm version defaults

jtumber-ibm:
First part in using version.yaml as the source for all our defaults.

> N.B. this does still use the hardcoded versions for the CAA and KATA sources that are used in the podvm_builder step. That can be fixed later if we drive the docker builds from make.

Tested using the following commands

```
$ docker build -t podvm_builder -f podvm/Dockerfile.podvm_builder --build-arg CAA_SRC=https://github.com/jtumber-ibm/cloud-api-adaptor --build-arg CAA_SRC_REF=podvm-version-defaults . 
$ docker build --pull=false -t podvm_binaries -f podvm/Dockerfile.podvm_binaries --build-arg BUILDER_IMG=podvm_builder . 
$ docker build --pull=false -t podvm_amd64 -f podvm/Dockerfile.podvm --build-arg BUILDER_IMG=podvm_builder --build-arg BINARIES_IMG=podvm_binaries . 
```

After importing into ibmcloud with `./import podvm_amd64 ...` I had a working podvm vpc image.

```
$ ibmcloud is instances
Listing instances in all resource groups and region eu-gb under account James Tumber's Account as user james.tumber@ibm.com...
ID                                          Name                   Status    Profile   Image                     
07a7_4fb42e89-cbb4-4de8-a777-b5175d55124f   podvm-nginx-29a79ee4   running   bx2-2x8   podvm-581db2c-dirty-amd64 

$ kubectl get pods
NAME    READY   STATUS    RESTARTS   AGE
nginx   1/1     Running   0          7m32s
```

podvm-581db2c-dirty-amd64 with commit https://github.com/confidential-containers/cloud-api-adaptor/commit/581db2c4a9465f54d26579d69949d4570b090c9d


podvm/Dockerfile.podvm_builder
```diff
@@ -58,7 +60,7 @@ RUN echo $CAA_SRC
 
 RUN echo $CAA_SRC_REF
 
-RUN git clone ${CAA_SRC} -b ${CAA_SRC_REF} cloud-api-adaptor
+RUN git clone ${CAA_SRC} -b ${CAA_SRC_REF}  cloud-api-adaptor
```
katexochen, line 63:
```suggestion
RUN git clone ${CAA_SRC} -b ${CAA_SRC_REF} cloud-api-adaptor
```

podvm/Dockerfile.podvm_builder
```diff
@@ -29,6 +29,8 @@ RUN apt-get update -y && \
     echo "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee -a /etc/apt/sources.list && \
     apt-get update && apt-get install --no-install-recommends -y packer && \
     apt-get clean && \
+    wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 && \
+    chmod a+x /usr/local/bin/yq && \
```
katexochen, line 33: We should pin a version instead of pulling in latest.

Also, there is a very nice experimental syntax in docker, that can be used to add hash pinned dependencies:
https://github.com/confidential-containers/cloud-api-adaptor/blob/27af578f30e6d4f5151f6ee807e33e794e2edfde/hack/Dockerfile.golang#L1
 https://github.com/confidential-containers/cloud-api-adaptor/blob/27af578f30e6d4f5151f6ee807e33e794e2edfde/hack/Dockerfile.golang#L13

We can replace every wget in a dockerfile with this (must not necessarily happen in this pr).

podvm/Dockerfile.podvm_builder
katexochen, line 1:
```
ARG GO_VERSION="1.20.7"
ARG PROTOC_VERSION="3.11.4"
ARG RUST_VERSION="1.69.0"
```
There are still these hard-coded versions in the docker files. We should use the versions from the versions.yaml instead. (Same for the other builder files)

versions.yaml
stevenhorsman, line 1: I wonder if we should consider bumping the ubuntu release we use as we are 7 months behind the latest. I'm happy for that to be a separate PR though.

podvm/Dockerfile.podvm_builder
stevenhorsman, line 63: nit: remove extra added space

stevenhorsman (approved):
I've not tested it yet, though you have, but it looks good to me based on my initial read through of the commits

katexochen (changes requested):
Thanks, great work!

Makefile.defaults
```diff
@@ -0,0 +1,37 @@
...
+centos_RELEASE ?= stream-8
+
+ubuntu_amd64_IMAGE_URL := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).amd64.url)
+ubuntu_amd64_IMAGE_CHECKSUM := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).amd64.checksum)
+ubuntu_s390x_IMAGE_URL := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).s390x.url)
+ubuntu_s390x_IMAGE_CHECKSUM := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).s390x.checksum)
+centos_amd64_IMAGE_URL := $(call query,cloudimg.centos.$(centos_RELEASE).amd64.url)
+centos_amd64_IMAGE_CHECKSUM := $(call query,cloudimg.centos.$(centos_RELEASE).amd64.checksum)
+
+CAA_SRC := $(or $(CAA_SRC),$(call query,git.cloud-api-adaptor.url))
```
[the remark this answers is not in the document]
jtumber-ibm: So this value doesn't currently do anything (happy for now to remove it now to avoid confusion). 

My plan is that after this PR I will pick up #1148 which will include driving the docker build process via Make so that it can pass through the CAA_SRC value, which is also currently required for the `GO_VERSION`, `PROTOC_VERSION`, and `RUST_VERSION` values you mentioned in your other comment.

So the end goal being that from the dev machine or CI we could run.

```
make docker-build-podvm # example
```

Make would populate the CAA_SRC, GO_VERSION, PROTOC_VERSION & RUST_VERSION values from the local versions.yaml and then call docker e.g.

```
docker build -t podvm_builder -f Dockerfile.podvm_builder --build-arg CAA_SRC=https://github.com/... --build-arg CAA_SRC_REF=0000...  --build-arg  GO_VERSION=...  --build-arg  PROTOC_VERSION=...  --build-arg  RUST_VERSION=...  .
```

Inside the docker image we would clone the remote CAA repo and install the tooling at the right level specified in versions.yaml but then the remote versions.yaml (which should be the same but might not be) is used for the rest of the process. Or we can choose to use the local files which #1148 will introduce in which case the versions.yaml file used will be the same from start to finish.

podvm/Dockerfile.podvm_builder
jtumber-ibm: See: https://github.com/confidential-containers/cloud-api-adaptor/pull/1388#discussion_r1314644120

As we need this values before we get to the `make` step to install the tooling at the moment I think they should be left in until the next step in the process. But I am planning to remove these too after this PR has gone in

Makefile.defaults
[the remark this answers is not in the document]
katexochen: I would say that pulling the source from remote and then continuing the build is a bad idea in general, as the consequences are quite hard to understand as you end up combining outer Makefile and Docker image with inner Makefile that might not fit together.
Imo we should just switch to build from local source completely and remove `CAA_SRC`.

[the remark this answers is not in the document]
katexochen: But fine for me to discuss this in the next PR. :)

versions.yaml
jtumber-ibm: I'll do this in a separate PR

katexochen (approved):

PR #1400, opened 2023-09-04
Relative make includes

jtumber-ibm:
Ensure that the included Makefiles can always be found if run from other working directory.

Previously if you were to try to run the podvm make target from other directories it would fail to find Makefile.defaults

### DIFF
--- a/podvm/Makefile.inc
+++ b/podvm/Makefile.inc
@@ -4,12 +4,17 @@
 
 # This file contains the common build steps across the providers
 # Including building the required binaries from source
-# It is to be included via `include ../../podvm/Makefile.inc` in
-# each of the cloud provider podvm image Makefiles
+# It can be included via
+# ```
+# ROOT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))../../
+# include $(ROOT_DIR)podvm/Makefile.inc
+# ```
+# in each of the cloud provider podvm image Makefiles
 
+ROOT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))../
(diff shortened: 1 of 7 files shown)


mkulke (approved):

podvm/Makefile.inc
```diff
@@ -7,9 +7,10 @@
 # It is to be included via `include ../../podvm/Makefile.inc` in
```
stevenhorsman, line 7: Is it worth updating this comment to instruct about:
```
ROOT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))../../
include $(ROOT_DIR)podvm/Makefile.inc
```
?

podvm/Makefile.inc
```diff
@@ -7,9 +7,10 @@
...
 # each of the cloud provider podvm image Makefiles
 
+PARENT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))../
```
stevenhorsman, line 10: I'm not sure on this, but would `PODVM_DIR` be a clearer name than `PARENT_DIR`?

stevenhorsman (commented):
Just a couple of minor comments

podvm/Makefile.inc
jtumber-ibm: It is pointing to the parent of the podvm directory i.e. the cloud-api-adaptor root. I can use the ROOT_DIR here too if you think that is clearer?

stevenhorsman: Thanks - that's clearer to me and the matches the others.

stevenhorsman (approved):
LGTM. Thanks

PR #1397, opened 2023-09-01
libvirt: Add support for SEV

bookinabox:
Allows SEV to be enabled for libvirt peer-pods when hardware is supported.

When DISABLECVM is not enabled, the libvirt adaptor will attempt to determine the appropiate launchSecurity (sev vs s390pv).

This can be overwritten by specifying LIBVIRT_LAUNCH_SECURITY.

Otherwise, if DISABLECVM is enabled, then the VM will be started with no launchSecurity.

Fixes: #1395

PR #1403, opened 2023-09-05
Libvirt sev support tests

bookinabox:
Adds tests for #1397.


bookinabox:
Not sure if enabling confidential vms (DISABLECVM=false), should be automatically enabled for the test. This might require restarting the adaptor-daemonset pod within the test if DISABLECVM=true to start with.

ATM, I just check whether DISABLECVM=true and skip the e2e test if it is.

Any thoughts would be appreciated.

PR #1404, opened 2023-09-06
workflow: set context to build peerpod-ctrl image

genjuro214:
working-directory is not applied to steps.uses


snir911 (approved):
Thanks, yes, I think it should work as this what was in the original PR i sent which IIRC i tested and it was removed during the PR reviewing process

huoqifeng (approved):
/lgtm
thanks @genjuro214

huoqifeng:
I'll merge this PR based on the comments from @snir911 

> Thanks, yes, I think it should work as this what was in the original PR i sent which IIRC i tested and it was removed during the PR reviewing process

I'll merge this PR based on the comments from @snir911

PR #1408, opened 2023-09-06
podvm build: include AA binariy in BINARIES list

### LINKED ISSUE #1406: podvm-build: why do we exclude the attestation-agent from the binary cache?
https://github.com/confidential-containers/cloud-api-adaptor/blob/6ad7f19e79ec2f6d4eb78cf97e6b1b2643f34fc0/podvm/Makefile.inc#L61-L63

In builds w/ prebuilt binaries we do use `BINARIES= make image`, however we still have to rebuild AA from scratch, which is quite costly and sometimes not what you want.

What is the reason behind this exception? I don't see a reasoning by using git blame, other than the `AA_KBC` toggle that has since disappeared:

https://github.com/confidential-containers/cloud-api-adaptor/commit/722fa85ef74bf9c3035c4bbeec4b593f7332e02f

Can we just add it to the `BINARIES` list?

mkulke:
If you specify `make image BINARIES=` to skip the building of binaries, attestation-agent will be excluded, because it is readded to the list in any case.

fixes #1406


stevenhorsman (approved):
LGTM. I guess the old code was just a bit of a hangover from when the attestation-agent was conditionally added. Thanks!

PR #1409, opened 2023-09-06
Misc AWS fixes

bpradipt:
* Switch to Ubuntu 22.04 
* Allow specifying root volume in packer builds
* Allow specifying root volume size in cloud-api-adaptor


snir911 (approved):
LGTM, I'd add a section in AWS docs for both volume size changing methods

huoqifeng (approved):
/lgtm

bpradipt:
The go vulnerability check error is informational. Hence not fixing it in this PR

PR #1413, opened 2023-09-06
Libvirt e2e_on_pull workflow fixes

wainersm:
Two fixes, being the first really important:

* fix checkout on libvirt workflow
* skip libvirt tests only on CI

Cc @ldoktor the last being your suggestion :)

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -15,6 +15,11 @@ on:
         default: ''
         required: false
         type: string
+      git_ref:
+        default: 'main'
+        description: Git ref to checkout the cloud-api-adaptor repository. Defaults to main.
+        required: false
+        type: string
 
 env:
   CLOUD_PROVIDER: libvirt
@@ -27,6 +32,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.git_ref }}
 
       - name: Setup Golang version ${{ env.GO_VERSION }}
         uses: actions/setup-go@v4
--- a/.github/workflows/e2e_on_pull.yaml
+++ b/.github/workflows/e2e_on_pull.yaml
@@ -6,6 +6,14 @@
 name: e2e tests
 
 on:
+  # Note on repository checkout: pull_request_target sets `GITHUB_SHA` to the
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e_libvirt.yaml
```diff
@@ -15,6 +15,11 @@ on:
         default: ''
         required: false
         type: string
+      git_ref:
+        default: 'main'
+        description: Github ref to checkout the cloud-api-adaptor repository
```
mkulke, line 20:
```suggestion
        description: Git ref to checkout the cloud-api-adaptor repository. Defaults to main.
```

mkulke (approved):

.github/workflows/e2e_on_pull.yaml
```diff
@@ -59,6 +59,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
stevenhorsman, line 64: I might have misunderstood something, but I think we had a similar thing in kata, where we pulled the `pull_request.head.sha` and had problems where if the PR wasn't rebased then you get issues, so contributors need to rebase to get accurate CI. What we ended up doing (shout-out to Jeremi for the suggestion), was switch this to `github.event.pull_request.merge_commit_sha` (some detail in the REST API [doc](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request) and the Kata [PR](https://github.com/kata-containers/kata-containers/pull/7814#pullrequestreview-1606407631). So maybe we should consider that here as well?

stevenhorsman: Fabiano has just posted a potential issue with this approach on slack: https://katacontainers.slack.com/archives/C879ACQ00/p1694089545591799 so it's worth understanding that before making the change

.github/workflows/e2e_libvirt.yaml
wainersm: Done

.github/workflows/e2e_on_pull.yaml
stevenhorsman: Cool that sounds fine to me

stevenhorsman (approved):
LGTM. Thanks!

Issue #1329, opened 2023-08-16
Update Libvirt README Delete Section


bookinabox:
The deletion/uninstallation section of the Libvirt README is outdated.

For example it references install/yamls/deploy.yaml, which is no longer used. It should be updated to reflect current means of uninstalling the operator.

PR #1416, opened 2023-09-07
libvirt: update README uninstall

bookinabox:
Updates the uninstallation instructions for libvirt.

Fixes: #1329


wainersm (approved):

wainersm:
@snir911 very simple changes yet useful.

libvirt/README.md
```diff
@@ -232,22 +237,54 @@ cc-operator-daemon-install-fkkzz                 1/1     Running   0          20
...
+```
+
+It can take some minutes to get those pods deleted. Afterwards you will notice that only the *controller-manager* pod is
+still up. The ccruntime resource will also be deleted. This can be verified with:
+```
+kubectl get ccruntime
+```
+which should return nothing.
+
+To delete the *controller-manager*:
```
snir911, line 281: Why this is needed? user can just use `make delete` for this purpose no?

snir911 (approved):
LGTM, thanks, added minor comment

Issue #1417, opened 2023-09-07
Add additional check in cloud-api-adaptor startup probe


liudalibj:
startup probe was added in cloud-api-adaptor via PR https://github.com/confidential-containers/cloud-api-adaptor/pull/1323, it will check the UDS of the service and the legacy PeerPods instances and make sure UDS is ready and all PeerPods instances become running status before report a heath status to the checker. 

Actually, we need move the check more earlier at the time when cloud-api-adaptor starts up, so that we can sure the cloud-api-adaptor is in real heath status. So, in this story we're going to add more checks in startup probe, like:
- The podvm Image ID exist
- The authentication to IaaS api is good
...

PR #1418, opened 2023-09-07
ConfigVerifier: Veify cloud service config before Start cloud services

liudalibj:
- Check podvm image id is not empty
- Only verify cloud config when `CLOUD_CONFIG_VERIFY` env is `yes`

fixes: #1417

### DIFF
--- a/pkg/adaptor/cloud/ibmcloud-powervs/provider.go
+++ b/pkg/adaptor/cloud/ibmcloud-powervs/provider.go
@@ -142,6 +142,14 @@ func (p *ibmcloudPowerVSProvider) Teardown() error {
 	return nil
 }
 
+func (p *ibmcloudPowerVSProvider) ConfigVerifier() error {
+	ImageId := p.serviceConfig.ImageID
+	if len(ImageId) == 0 {
+		return fmt.Errorf("ImageId is empty")
+	}
+	return nil
+}
+
 func (p *ibmcloudPowerVSProvider) getVMIPs(ctx context.Context, instance *models.PVMInstance) ([]netip.Addr, error) {
 	var ips []netip.Addr
 	ins, err := p.powervsService.instanceClient(ctx).Get(*instance.PvmInstanceID)
--- a/pkg/adaptor/cloud/ibmcloud/provider.go
+++ b/pkg/adaptor/cloud/ibmcloud/provider.go
@@ -414,3 +414,11 @@ func (p *ibmcloudVPCProvider) DeleteInstance(ctx context.Context, instanceID str
 func (p *ibmcloudVPCProvider) Teardown() error {
 	return nil
 }
+
+func (p *ibmcloudVPCProvider) ConfigVerifier() error {
+	images := p.serviceConfig.Images.String()
+	if len(images) == 0 {
+		return fmt.Errorf("image-id is empty")
+	}
+	return nil
+}
--- a/cmd/cloud-api-adaptor/main.go
+++ b/cmd/cloud-api-adaptor/main.go
@@ -112,6 +112,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		flags.IntVar(&cfg.networkConfig.VXLANPort, "vxlan-port", vxlan.DefaultVXLANPort, "VXLAN UDP port number (VXLAN tunnel mode only")
 		flags.IntVar(&cfg.networkConfig.VXLANMinID, "vxlan-min-id", vxlan.DefaultVXLANMinID, "Minimum VXLAN ID (VXLAN tunnel mode only")
 		flags.StringVar(&cfg.serverConfig.AAKBCParams, "aa-kbc-params", "", "attestation-agent KBC parameters")
+		flags.BoolVar(&cfg.serverConfig.EnableCloudConfigVerify, "cloud-config-verify", false, "Enable cloud config verify - should use it for production")
 
 		cloud.ParseCmd(flags)
 	})
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -17,6 +17,7 @@ optionals+=""
 [[ "${TLS_SKIP_VERIFY}" ]] && optionals+="-tls-skip-verify "
 [[ "${PROXY_TIMEOUT}" ]] && optionals+="-proxy-timeout ${PROXY_TIMEOUT} "
 [[ "${AA_KBC_PARAMS}" ]] && optionals+="-aa-kbc-params ${AA_KBC_PARAMS} "
+[[ "${CLOUD_CONFIG_VERIFY}" == "true" ]] && optionals+="-cloud-config-verify "
 
 test_vars() {
     for i in "$@"; do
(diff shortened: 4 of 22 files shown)


pkg/adaptor/cloud/ibmcloud-powervs/provider.go
```diff
@@ -245,3 +245,11 @@ func (p *ibmcloudPowerVSProvider) getFromDHCPServer(ctx context.Context, instanc
 
 	return ip, nil
 }
+
+func (p *ibmcloudPowerVSProvider) Check() error {
```
huoqifeng, line 249: Shall we use StartupCheck specifically to avoid readiness or liveness probe by mistake in future?

pkg/adaptor/cloud/ibmcloud/provider.go
```diff
@@ -414,3 +414,11 @@ func (p *ibmcloudVPCProvider) DeleteInstance(ctx context.Context, instanceID str
 func (p *ibmcloudVPCProvider) Teardown() error {
 	return nil
 }
+
+func (p *ibmcloudVPCProvider) Check() error {
+	images := p.serviceConfig.Images.String()
```
huoqifeng, line 419: Possible to check the API-KEY and image existing via image ID?

pkg/adaptor/cloud/ibmcloud-powervs/provider.go
liudalibj: updated.

pkg/adaptor/cloud/ibmcloud/provider.go
liudalibj: it already be checked at [updateImageList](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/ibmcloud/provider.go#L125C20-L125C35) inside `NewProvider` function.

genjuro214 (approved):
LGTM

pkg/adaptor/cloud/ibmcloud/provider.go
huoqifeng: oh, I see, sounds we don't need extra checking any more because `NewProvider` guarantee the parameters works well? like:
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/cmd/cloud-api-adaptor/main.go#L30
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/ibmcloud/manager.go#L43
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/ibmcloud/provider.go#L125
If there is any wrong in `NewProvider`, the provider can not be initialized and the service can not start, therefore the UDS will not be opened. The `uds` has been checked in startup probe.

liudalibj: keep the simple checker here to keep consistent with other providers

stevenhorsman (approved):
LGTM if you fixup the `Check` rename into the previous commit before merging. Thanks

pkg/adaptor/cloud/ibmcloud/provider.go
huoqifeng: Maybe we don't need a new Check interface but add the check in existing `NewProvider`? It's already there and did some checking.

yoheiueda (approved):
LGTM

pkg/adaptor/cloud/ibmcloud/provider.go
huoqifeng: It's reasonable for me to keep a separate check function also if we considering `NewProvider` just set configuratins and `Check` to verify the configurations like do some async http calls.

huoqifeng (approved):
/lgtm

bpradipt:
@liudalibj so we now have two checkers 
1. startup probes
2. startup checker before starting the CAA daemon

As I understand we will need both these as it serves different purposes - startup probes checking for peer-pods to be in ready state to control the rolling updates and startup checker to verify if the required configs are available before starting the CAA daemon.

Startup checker is more like config verifier before starting the CAA daemon. Consider a development workflow where I just deploy the CAA with dummy config and then update the config with right values and restart the CAA pod to pickup the changes. This workflow is important from development velocity standpoint and will no longer be available with this change.

Otoh I do see value in startup checker to verify the configurations. But in it's current form I find it confusing and breaking an existing development flow with no significant benefit. So here is my suggestion:
1. Introduce an option for config verifier with default as false. When it's set, then it'll run the verifiers. We can also introduce verifiers set to run - like all, specific ones etc. This can be done subsequently. But the framework can be created now.
2. Can we rename the methods/feature to config verifier? Although startup checker is also fine, but I feel config verifier more closely aligns with the intended purpose.
3. Create a separate module which contains all the verifiers/checkers. Use them for both startup probes and config verifier. I don't know for sure it this will require a separate repo since the verifiers need to make use of the cloud provider apis and doing this in the same repo might introduce a cyclic dependency like we see with peerpod-ctrl. But if creating a separate repo is the right way then we go with that.

We can discuss on the community slack if it helps.

bpradipt:
I'm putting the `hold` label to discuss the design aspects mentioned in my previous comment.

liudalibj:
> @liudalibj so we now have two checkers
> 
> 1. startup probes
> 2. startup checker before starting the CAA daemon
> 
> As I understand we will need both these as it serves different purposes - startup probes checking for peer-pods to be in ready state to control the rolling updates and startup checker to verify if the required configs are available before starting the CAA daemon.
> 
> Startup checker is more like config verifier before starting the CAA daemon. Consider a development workflow where I just deploy the CAA with dummy config and then update the config with right values and restart the CAA pod to pickup the changes. This workflow is important from development velocity standpoint and will no longer be available with this change.
> 
> Otoh I do see value in startup checker to verify the configurations. But in it's current form I find it confusing and breaking an existing development flow with no significant benefit. So here is my suggestion:
> 
> 1. Introduce an option for config verifier with default as false. When it's set, then it'll run the verifiers. We can also introduce verifiers set to run - like all, specific ones etc. This can be done subsequently. But the framework can be created now.
> 2. Can we rename the methods/feature to config verifier? Although startup checker is also fine, but I feel config verifier more closely aligns with the intended purpose.
> 3. Create a separate module which contains all the verifiers/checkers. Use them for both startup probes and config verifier. I don't know for sure it this will require a separate repo since the verifiers need to make use of the cloud provider apis and doing this in the same repo might introduce a cyclic dependency like we see with peerpod-ctrl. But if creating a separate repo is the right way then we go with that.
> 
> We can discuss on the community slack if it helps.

Thanks @bpradipt you are right `Startup checker` is more like config verifier before starting the CAA daemon.
I will try to address your suggestion 1 and 2 first in this pr when I get time in next week, and try to use a separate module/repo in future prs.

liudalibj:
> I'm putting the `hold` label to discuss the design aspects mentioned in my previous comment.

Hi @bpradipt I updated this PR, addressed your comments item 1&2, would you like help to check the codes again? Thanks.
- updated function name to `ConfigVerifier`
- Only verify cloud config when CLOUD_CONFIG_VERIFY env is yes

bpradipt (approved):
/lgtm
Thanks @liudalibj 
Some CIs are failing..

PR #1419, opened 2023-09-07
podvm: Fix Makefile mistakes

jtumber-ibm:
IMAGE_URL/CHECKSUM weren't being set correctly with the docker build.

Apologies for not catching this in the last PR.


stevenhorsman (approved):
LGTM. Thanks!

podvm/Makefile
```diff
@@ -10,14 +10,14 @@ include $(SELF_DIR)Makefile.inc
 
 IMAGE_SUFFIX := .qcow2
 PODVM_DISTRO ?= ubuntu
-IMAGE_URL ?= $($(PODVM_DISTRO)_$(DEB_ARCH)_IMAGE_URL)
-IMAGE_CHECKSUM ?= $($(PODVM_DISTRO)_$(DEB_ARCH)_IMAGE_CHECKSUM)
+IMAGE_URL := $(or $(IMAGE_URL),$($(PODVM_DISTRO)_$(DEB_ARCH)_IMAGE_URL))
```
bpradipt, line 13: Will this change allow to override IMAGE_URL and IMAGE_CHECKSUM from CLI ?

jtumber-ibm: Yes that is what the `or` function checks for, but it is better than `?=` because it will work if the value is empty but set. Which happens in the Docker builds if you don't pass an argument in.

bpradipt: Great.. Thanks for the explanation @jtumber-ibm

bpradipt (approved):
/lgtm

Issue #1421, opened 2023-09-08
Fix vulnerabilities identified by govulncheck


stevenhorsman:
We've started seeing error in golvumcheck: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/6120859772/job/16613517203?pr=1377 so I think we need to update to the newest go release 1.20.8 that has fixes for these.

PR #1422, opened 2023-09-08
deps: Update Go to v1.20.8

stevenhorsman:
Resolve two vulnerabilites:
- GO-2023-2043 - Improper handling of special tags within script contexts in html/template
- GO-2023-2041 - Improper handling of HTML-like comments in script contexts in html/template

Fixes: #1421


.github/workflows/podvm_builder.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
         file: |
           podvm/${{ matrix.dockerfile }}
         build-args: |
-          "GO_VERSION=1.20.7"
+          "GO_VERSION=1.20.8"
```
wainersm, line 60: hi @stevenhorsman , for coincidence yesterday I changed that workflow to read the properties from versions.yaml but I kept it closed to send alongside other stuffs. Looking at this change now I realized on how many places some properties are hard-coded! So I think I will expand my changes to include other workflows, what do you think? 

Don't want to block this PR, btw.

wainersm (approved):
Thanks @stevenhorsman !

jtumber-ibm (approved):

.github/workflows/podvm_builder.yaml
stevenhorsman: Yeah, I was surprised how many places go version was still hard-coded, so anything we can do to reduce that is great

PR #1423, opened 2023-09-08
workflows: read properties from versions.yaml

wainersm:
A bunch of our workflows have hard-coded go version and bumping the value requires "search & replace" of many files. This is error prone. On the other hand, we got the `versions.yaml` that ideally should be the source of truth of external dependencies. Thus, in this PR I changed most workflows that had go version declared on them to read the value from `versions.yaml`

Related to https://github.com/confidential-containers/cloud-api-adaptor/pull/1422#discussion_r1319834023

Note:
 * Split in several commits to ease review. As the changes are repetitive I will squash them all before the merge
 * I didn't change Azure workflows as I don't know if they should be following `versions.yaml` (FYI @sudharshanibm3)
 * Looked at `ibmcloud-powervs/image/prereq.sh` but decided to not touch it because I don't know how it is used. Apparently it could be just installing `yq` to read `versions.yaml` (as long as that file is present on the machine, otherwise should be downloaded...)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,16 +21,18 @@ jobs:
           - release
         runner:
           - ubuntu-latest
-        go_version:
-          - "1.20.8"
-
     steps:
       - name: Checkout the pull request code
         uses: actions/checkout@v3
-      - name: Setup Golang version ${{ matrix.go_version }}
+      - name: Read properties from versions.yaml
+        run: |
+          go_version="$(yq '.tools.golang' versions.yaml)"
+          [ -n "$go_version" ]
+          echo "GO_VERSION=${go_version}" >> "$GITHUB_ENV"
+      - name: Setup Golang version ${{ env.GO_VERSION }}
         uses: actions/setup-go@v4
         with:
-          go-version: ${{ matrix.go_version }}
+          go-version: ${{ env.GO_VERSION }}
       - name: Install build dependencies
         if: matrix.type == 'dev'
         run: |
@@ -57,6 +59,6 @@ jobs:
         uses: actions/upload-artifact@v3
         if: matrix.type == 'dev'
         with:
-          name: tests_report_junit-${{ matrix.runner }}_${{ matrix.go_version }}
+          name: tests_report_junit-${{ matrix.runner }}_${{ env.GO_VERSION }}
           path: ${{ github.workspace }}/tests_report_junit.xml
           retention-days: 1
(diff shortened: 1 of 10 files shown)


stevenhorsman (commented):
It might be that `env.variable` works fine, but based on the `GITHUB_ENV` [example doc](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-writing-an-environment-variable-to-github_env) it looks like you can just reference the variable directly, so that might be cleaner?

stevenhorsman:
I guess you need to rebase now that the 1.20.8 bump that gone in. Sorry about that!

wainersm:
> I guess you need to rebase now that the 1.20.8 bump that gone in. Sorry about that!

Done.

wainersm:
> It might be that `env.variable` works fine, but based on the `GITHUB_ENV` [example doc](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-writing-an-environment-variable-to-github_env) it looks like you can just reference the variable directly, so that might be cleaner?

`GITHUB_ENV` will be available as shell variable in `run` steps, but outside in a `${{ }}` expression I should use a context (`env`, `github`, `inputs`...etc) to access a property. So `env.VARIABLE` to access the `VARIABLE`.

wainersm:
ok, `yq r versions.yaml 'tools.golang'` has failed with `invalid input text "r"`.... let me check what is the `yq` version installed in the runner and add some safe-guards to avoid undefined behaviors.

wainersm:
> ok, `yq r versions.yaml 'tools.golang'` has failed with `invalid input text "r"`.... let me check what is the `yq` version installed in the runner and add some safe-guards to avoid undefined behaviors.

The gh runner (ubuntu 22.04) installs yq `3.145.1` (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) which is incompatible with the version I tested. So I re-done the `yq` commands and added a safe-guard to ensure the property is not empty. You will see this pattern all over the places:

```
          go_version="$(yq '.tools.golang' versions.yaml)"
          [ -n "$go_version" ]
          echo "GO_VERSION=${go_version}" >> "$GITHUB_ENV"
```

wainersm:
fixed the controllers jobs that don't run on root directory so had to adjust the path to versions.yaml

jtumber-ibm (approved):
lgtm

stevenhorsman (approved):
LGTM. Thanks for the env explanation.

surajssd (approved):

wainersm:
Thanks @jtumber-ibm @stevenhorsman @surajssd !

PR #1434, opened 2023-09-13
e2e test: add test case for non-disruptive PeerPod upgrade

genjuro214:
initialize a new test TestCaaDaemonsetUpgrade

* Setup: 
  * prepare an IKS cluster with 2 workers, and install PeerPod operator  
  * create a nginx deployment with 2 replicas, so that there are 2 nginx PeerPods, and each pod for a worker
  * create a nginx service based on the deployment, and get the service cluster IP
  * create a verify pod to curl the cluster IP every 1 second; the verify pod will exit with error if curl return non-zero result
* Access:
  * edit **cloud-api-adaptor-daemonset** to increase `StartupProbe.FailureThreshold`, so that the daemonset will be updated, and the CAA pods on each worker will be recreated
  * wait for CAA daemonset updated, and the nginx deployment be available again
  * check the status of verify pod, if it's still running. If yes, that means nginx deployment is always available when updating CAA daemonset; otherwise, it's not expected, the upgrade is not non-disruptive

### DIFF
--- a/test/e2e/rolling_update_test.go
+++ b/test/e2e/rolling_update_test.go
@@ -1,0 +1,297 @@
+// (C) Copyright Confidential Containers Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/e2e-framework/klient"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	envconf "sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const WAIT_DEPLOYMENT_AVAILABLE_TIMEOUT = time.Second * 180
+
+func doTestCaaDaemonsetRollingUpdate(t *testing.T) {
+	runtimeClassName := "kata-remote"
+	namespace := envconf.RandomName("default", 7)
+	deploymentName := "nginx-deployment"
+	containerName := "nginx"
+	imageName := "nginx"
+	serviceName := "nginx-service"
+	portName := "port80"
+	rc := int32(2)
+	labelsMap := map[string]string{
+		"app": "nginx",
+	}
+	verifyPodName := "verify-pod"
+	verifyContainerName := "verify-container"
+	verifyImageName := "radial/busyboxplus:curl"
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      deploymentName,
+			Namespace: namespace,
+			Labels:    labelsMap,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &rc,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labelsMap,
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labelsMap,
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:            containerName,
+							Image:           imageName,
+							ImagePullPolicy: v1.PullAlways,
+						},
+					},
+					RuntimeClassName: &runtimeClassName,
+					Affinity: &v1.Affinity{
+						PodAntiAffinity: &v1.PodAntiAffinity{
+							RequiredDuringSchedulingIgnoredDuringExecution: []v1.PodAffinityTerm{
+								{
+									LabelSelector: &metav1.LabelSelector{
+										MatchExpressions: []metav1.LabelSelectorRequirement{
+											{
+												Key:      "app",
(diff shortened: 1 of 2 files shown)


test/e2e/common_suite_test.go
```diff
@@ -870,3 +874,253 @@ func doTestCreatePeerPodWithAuthenticatedImageWithoutCredentials(t *testing.T, a
 	pod := newPod(namespace, podName, podName, imageName, withRestartPolicy(v1.RestartPolicyNever))
 	newTestCase(t, "InvalidAuthImagePeerPod", assert, "Peer pod with Authenticated Image with Invalid Credentials has been created").withPod(pod).withAuthenticatedImage().withAuthImageStatus(expectedAuthStatus).withCustomPodState(v1.PodPending).run()
 }
+
+func doTestCaaDaemonsetUpgrade(t *testing.T) {
```
huoqifeng, line 878: We talked about to split the test cases as the common_suit_test.go grows greatly, maybe this is right PR to add a separate file.

huoqifeng (commented):
Thanks @genjuro214 brings the PR.

test/e2e/common_suite_test.go
```diff
@@ -870,3 +874,253 @@ func doTestCreatePeerPodWithAuthenticatedImageWithoutCredentials(t *testing.T, a
...
+	imageName := "nginx"
+	serviceName := "nginx-service"
+	portName := "port80"
+	rc := int32(2)
+	labelsMap := map[string]string{
+		"app": "nginx",
+	}
+	verifyPodName := "verify-pod"
+	verifyContainerName := "verify-container"
+	verifyImageName := "radial/busyboxplus:curl"
```
huoqifeng, line 892: Shall we tag the testing tools and the nginx image and push them to https://quay.io/repository/confidential-containers/? So that we can avoid noisy docker hub limitations?

test/e2e/common_suite_test.go
```diff
@@ -870,3 +874,253 @@ func doTestCreatePeerPodWithAuthenticatedImageWithoutCredentials(t *testing.T, a
 	pod := newPod(namespace, podName, podName, imageName, withRestartPolicy(v1.RestartPolicyNever))
 	newTestCase(t, "InvalidAuthImagePeerPod", assert, "Peer pod with Authenticated Image with Invalid Credentials has been created").withPod(pod).withAuthenticatedImage().withAuthImageStatus(expectedAuthStatus).withCustomPodState(v1.PodPending).run()
 }
+
+func doTestCaaDaemonsetUpgrade(t *testing.T) {
+	runtimeClassName := "kata-remote"
+	namespace := envconf.RandomName("default", 7)
+	deploymentName := "nginx-deployment"
+	containerName := "nginx"
+	imageName := "nginx"
```
huoqifeng, line 883: Shall we tag the testing tools and the nginx image and push them to https://quay.io/repository/confidential-containers/? So that we can avoid noisy docker hub limitations?

test/e2e/common_suite_test.go
```diff
@@ -870,3 +874,253 @@ func doTestCreatePeerPodWithAuthenticatedImageWithoutCredentials(t *testing.T, a
...
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      deploymentName,
+			Namespace: namespace,
+			Labels:    labelsMap,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &rc,
+			Selector: &metav1.LabelSelector{
```
huoqifeng, line 902: How did we make sure each nginx pods assigned to different workers?

test/e2e/common_suite_test.go
genjuro214: That would be good. But I don't have the access to  https://quay.io/repository/confidential-containers/. Maybe we can do it via workflow, but who can help to set the image as public?

huoqifeng: Lets take it as an action further.

test/e2e/common_suite_test.go
genjuro214: I did nothing. But I didn't noticed the 2 pods would be deployed on 1 worker if there are 2 schedulable workers

huoqifeng: @genjuro214 it's not always the case. It's possible the 2 pods assigned on same worker, I met it before.

genjuro214:
Move the test code to a separate file from `common_suite_test.go`

genjuro214:
@huoqifeng OK, I add `PodAntiAffinity` to make sure the pod with label "app: nginx" will not be assigned to the same worker.

test/e2e/rolling_update_test.go
```diff
@@ -0,0 +1,297 @@
...
+	imageName := "nginx"
+	serviceName := "nginx-service"
+	portName := "port80"
+	rc := int32(2)
+	labelsMap := map[string]string{
+		"app": "nginx",
+	}
+	verifyPodName := "verify-pod"
+	verifyContainerName := "verify-container"
+	verifyImageName := "radial/busyboxplus:curl"
```
liudalibj, line 45: can we use mulit-arch verify image here? so that we can run this case with s390x clusters.
it seems that `radial/busyboxplus:curl` only have amd64 docker image

liudalibj (approved):
LGTM
discussed with @genjuro214 we will try to build/push the used images to quay.io and make sure they are multi-arch ones in future.

test/e2e/rolling_update_test.go
```diff
@@ -0,0 +1,297 @@
...
+		Spec: v1.PodSpec{
+			RestartPolicy: v1.RestartPolicyNever,
+			Containers: []v1.Container{
+				{
+					Name:  verifyContainerName,
+					Image: verifyImageName,
+					Command: []string{
+						"/bin/sh",
+						"-c",
+						// Not complete command; will append later
```
huoqifeng, line 127: Is the command ready, I think we need it done before merge?

huoqifeng: oh, I see the where it's updated.

huoqifeng (approved):
/lgtm, thank you @genjuro214

Issue #1432, opened 2023-09-13
Add ibmcloud provider in peerpod-ctl


huoqifeng:
Add ibmcloud provider in the built in binaries and image for peerpod-ctl

PR #1439, opened 2023-09-14
peerpod-ctl: build multi-arch images for peerpod-ctl

huoqifeng:
Fixes: #1432

### DIFF
--- a/.github/workflows/peerpod-ctrl_image.yaml
+++ b/.github/workflows/peerpod-ctrl_image.yaml
@@ -43,6 +43,7 @@ jobs:
             quay.io/confidential-containers/peerpod-ctrl:${{ github.sha }}
           push: true
           context: peerpod-ctrl
+          platforms: linux/amd64, linux/s390x, linux/ppc64le
           build-args: |
-            "$(make list-build-args)"
+            GOFLAGS=-tags=aws,azure,ibmcloud,vsphere,libvirt
 
--- a/peerpod-ctrl/Dockerfile
+++ b/peerpod-ctrl/Dockerfile
@@ -1,12 +1,12 @@
 # Build the manager binary
-FROM quay.io/confidential-containers/golang-fedora:1.20.7-36 as builder
+FROM --platform=$TARGETPLATFORM quay.io/confidential-containers/golang-fedora:1.20.8-36 as builder
 ARG TARGETOS
 ARG TARGETARCH
 ARG CGO_ENABLED=1
 ARG GOFLAGS
 
 WORKDIR /workspace
-RUN if [ "$CGO_ENABLED" = 1 ] ; then dnf install -y libvirt-devel; fi
+RUN if [ "$CGO_ENABLED" = 1 ] ; then dnf install -y libvirt-devel && dnf clean all; fi
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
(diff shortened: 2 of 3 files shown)


huoqifeng:
I noticed the `License Compliance` failure in PR request, does anyone know the reason? Is this something new we introduced recently?

stevenhorsman:
> I noticed the `License Compliance` failure in PR request, does anyone know the reason? Is this something new we introduced recently?

We think it is a false positive, but the reason for it showing up here is that you push this source as a branch in upstream, rather than in your fork

.github/workflows/peerpod-ctrl_image.yaml
```diff
@@ -43,6 +43,7 @@ jobs:
             quay.io/confidential-containers/peerpod-ctrl:${{ github.sha }}
           push: true
           context: peerpod-ctrl
+          platforms: linux/amd64, linux/s390x
```
stevenhorsman, line 46: I wonder if we should have add `,linux/ppc64le` as well for Power?

peerpod-ctrl/Dockerfile
```diff
@@ -19,16 +19,17 @@ COPY main.go main.go
 COPY api/ api/
 COPY controllers/ controllers/
 
+RUN if [ "$TARGETARCH" = s390x ] ; then ln /usr/bin/gcc /usr/bin/s390x-linux-gnu-gcc; fi
```
jtumber-ibm, line 22: I haven't seen this as a requirement before for cross-compiling go programs

peerpod-ctrl/Dockerfile
```diff
@@ -1,12 +1,12 @@
 # Build the manager binary
-FROM quay.io/confidential-containers/golang-fedora:1.20.7-36 as builder
+FROM --platform=$TARGETPLATFORM quay.io/confidential-containers/golang-fedora:1.20.8-36 as builder
```
jtumber-ibm, line 2: Should this be `BUILDPLATFORM` instead?

peerpod-ctrl/Dockerfile
huoqifeng: We got error like this without the change https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/6184725451/job/16788917938:
```
#18 180.2 cgo: C compiler "s390x-linux-gnu-gcc" not found: exec: "s390x-linux-gnu-gcc": executable file not found in $PATH
```
I guess the reason is GHA is doing cross-platform build while the golang image `quay.io/confidential-containers/golang-fedora:1.20.8-36` is s390x, thus, gcc is real `gcc` rather than emulator `s390x-linux-gnu-gcc` in this image.

.github/workflows/peerpod-ctrl_image.yaml
huoqifeng: I'd like to add `linux/ppc64le`, did not add it because I'm not sure why there is no `power` provider added in the Makefile https://github.com/confidential-containers/cloud-api-adaptor/blob/main/Makefile#L25

peerpod-ctrl/Dockerfile
huoqifeng: Success record is https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/6185637431 with the change.

peerpod-ctrl/Dockerfile
huoqifeng: I think we use the same arch for builder when build the binaries, so $TARGETPLATFORM is right here? Which is linux/s390x for s390x.

huoqifeng:
> > I noticed the `License Compliance` failure in PR request, does anyone know the reason? Is this something new we introduced recently?
> 
> We think it is a false positive, and it's been around pretty much forever (Paul is talking to Dan Middleton about it), but the reason for it showing up here is that you pushed this source as a branch in upstream, rather than in your fork and that's were it gets triggered

Thanks @stevenhorsman , I'm using a branch to test the GHA. But later I noticed the builder image can be pulled without credential, so a branch is not required really.

peerpod-ctrl/Dockerfile
huoqifeng: Using `BUILDPLATFORM` for builder, error like this:  https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/6192335422/job/16812236353#step:6:732

.github/workflows/peerpod-ctrl_image.yaml
huoqifeng: Added `linux/ppc64le`.

peerpod-ctrl/Dockerfile
huoqifeng: I guess amd64 golang-fedora image does not have CGO compiler to compile s390x and ppc64le, so we need use corresponding arch image as builder to use CGO, added `CC` arg to avoid the symbol link.

huoqifeng: I think the GHA against above commits https://github.com/confidential-containers/cloud-api-adaptor/actions/workflows/peerpod-ctrl_image.yaml shows why we need use `TARGETPLATFORM`. Maybe there is alternatives, e.g. built all archs cgo compilers in the builder image -- `golang-fedora`?

jtumber-ibm: To do the cross complication you would also need to turn off cgo. But if this way works too then that's fine.

huoqifeng: @jtumber-ibm So I guess you're OK with the change? Would you  please approve the PR if there is no more concerns?

liudalibj (approved):
thanks @huoqifeng LGTM

jtumber-ibm (approved):

PR #1445, opened 2023-09-18
volumes: fix the error to create nginx pod for ibm

genjuro214:
Specify Command in yaml

Not sure why the entrypoint in image is not be used. Maybe we need dig into kata agent code for image later.


liudalibj (approved):
LGTM

huoqifeng (approved):
LGTM

Issue #1093, opened 2023-06-19
install: ERROR: `version LIBVIRT_QEMU_8.2.0 not found (required by cloud-api-adaptor)`


katexochen:
When deploying cloud-api-adaptor on AKS, I get the following error using latest staging:

```shell
cloud-api-adaptor: /lib64/libvirt-qemu.so.0: version `LIBVIRT_QEMU_8.2.0' not found (required by cloud-api-adaptor)  
```
Did any requirements changed due to the switch in how we deploy CAA?

katexochen:
I've found what causes this issue:

Different then when building CAA for releases, dev builds include libvirt as builtin cloud provider:
https://github.com/confidential-containers/cloud-api-adaptor/blob/79d309dbdc3d29a3032501d7cfc31ec5935873df/Makefile#L22-L27

When libvirt is in the set of builtin cloud provides, CGO is enabled and we are linking against libvirt:
https://github.com/confidential-containers/cloud-api-adaptor/blob/79d309dbdc3d29a3032501d7cfc31ec5935873df/Makefile#L63-L67

When building in Docker, we use `golang:1.19` as builder image and `fedora:36` for release:
https://github.com/confidential-containers/cloud-api-adaptor/blob/79d309dbdc3d29a3032501d7cfc31ec5935873df/Dockerfile#L2-L3

So we are linking against libvirt on debian, and copying the dynamically binary to fedora, which uses a different naming scheme for libraries:

```shell
# Inside container
$ file /lib64/libvirt.so.0
libvirt-qemu.so.0: symbolic link to libvirt-qemu.so.0.8001.0
$ objdump -p libvirt-qemu.so.0.8001.0 | grep LIBVIRT_QEMU
2 0x00 0x098c3fd3 LIBVIRT_QEMU_0.8.3
3 0x00 0x098c40d4 LIBVIRT_QEMU_0.9.4
        LIBVIRT_QEMU_0.8.3 
4 0x00 0x08c30f80 LIBVIRT_QEMU_0.10.0
        LIBVIRT_QEMU_0.9.4 
5 0x00 0x098b31d3 LIBVIRT_QEMU_1.2.3
        LIBVIRT_QEMU_0.10.0 
```

I couldn't figure out what triggered this issue yet. My previous CAA build was about a month old and worked, non of the mentioned lines were touched since then.

katexochen:
I guess the simplest way to solve this is to use fedora as builder, too.

PR #1448, opened 2023-09-19
podvm: Use local files for build and reference versions file

jtumber-ibm:
This introduces new Make targets in the root Makefile to create the podvm image.

If doing all from scratch `make docker-builder docker-binaries docker-image`. Else if binaries and builder already exist for current versions.yaml file then `make docker-image` will be enough.

_**I haven't update the rhel or centos versions yet**_


stevenhorsman:
Sorry for the basic question - is this supposed to replace the current direct docker approach? If so I think we need to update the podvm README.md doc to reflect this?

jtumber-ibm:
> Sorry for the basic question - is this supposed to replace the current direct docker approach? If so I think we need to update the podvm README.md doc to reflect this?

It is a helper rather than a complete replacement, but you're right the docs will need updating too.

wainersm:
hi @jtumber-ibm ,  I hope to have some time today or tomorrow to review this. In meanwhile I was trying a different solution to allow caching kata-agent/AA to speed up the CI jobs, I am glad you sent this because I was having a really hard time to solve some cases :)

Makefile
```diff
@@ -162,3 +164,45 @@ ifneq ($(CLOUD_PROVIDER),)
...
+DOCKER_OPTS := --load $(if $(filter $(PUSH),true),--push) $(EXTRA_DOCKER_OPTS)
+
+DOCKERFILE_SUFFIX := $(if $(filter $(PODVM_DISTRO),ubuntu),,.$(PODVM_DISTRO))
+BUILDER_DOCKERFILE := Dockerfile.podvm_builder$(DOCKERFILE_SUFFIX)
+BINARIES_DOCKERFILE := Dockerfile.podvm_binaries$(DOCKERFILE_SUFFIX)
+PODVM_DOCKERFILE := Dockerfile.podvm$(DOCKERFILE_SUFFIX)
+
+docker-builder:
+	docker build -t $(PODVM_BUILDER_IMAGE) -f podvm/$(BUILDER_DOCKERFILE) \
+	--build-arg GO_VERSION=$(GO_VERSION) \
```
katexochen, line 188: We should remove the default version of Go in the docker image. Could this be also the right PR to also inject the other versions like rust etc?

katexochen (changes requested):

Makefile
```diff
@@ -162,3 +164,45 @@ ifneq ($(CLOUD_PROVIDER),)
...
+	$(DOCKER_OPTS) .
+
+docker-image: 
+	docker build -t $(PODVM_IMAGE) -f podvm/$(PODVM_DOCKERFILE) \
+	--build-arg BUILDER_IMG=$(PODVM_BUILDER_IMAGE) \
+	--build-arg BINARIES_IMG=$(PODVM_BINARIES_IMAGE) \
+	--build-arg PODVM_DISTRO=$(PODVM_DISTRO) \
+	--build-arg ARCH=$(ARCH) \
+	$(DOCKER_OPTS) .
+
```
katexochen, line 208:
```suggestion
docker-binaries:
	docker build -t $(PODVM_BINARIES_IMAGE) -f podvm/$(BINARIES_DOCKERFILE) \
	--build-arg BUILDER_IMG=$(PODVM_BUILDER_IMAGE) \
	--build-arg PODVM_DISTRO=$(PODVM_DISTRO) \
	--build-arg ARCH=$(ARCH) \
	--build-arg AA_KBC=$(AA_KBC) \
	$(DOCKER_OPTS) .

docker-image:
	docker build -t $(PODVM_IMAGE) -f podvm/$(PODVM_DOCKERFILE) \
	--build-arg BUILDER_IMG=$(PODVM_BUILDER_IMAGE) \
	--build-arg BINARIES_IMG=$(PODVM_BINARIES_IMAGE) \
	--build-arg PODVM_DISTRO=$(PODVM_DISTRO) \
	--build-arg ARCH=$(ARCH) \
	$(DOCKER_OPTS) .
```
There are some whitespace issues in your file. What editor are you using? :)

jtumber-ibm: VS Code, although I recently switch workstations so quite a few settings were missed (I didn't have the settings sync enabled) :)

.dockerignore
```diff
@@ -12,10 +12,9 @@ ibmcloud/
 libvirt/
 peerpod-ctrl/
 peerpodconfig-ctrl/
-podvm/
 volumes/
 webwook/
 
-.git*
+
```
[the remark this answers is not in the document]
jtumber-ibm: It is annoying but in its current state it requires copying across the git directory into docker so we can get the commit hash for the `.git-commit` versioning in the top-level Makefile. 
We could refactor it to be able to parse in the commit calculated outside of docker.

Makefile
jtumber-ibm: Yes to removing the default Go versions from the docker image so we don't have to keep that up to date.

Not sure what you mean by the other versions, they are also being included in the next lines, or something I've missed?

Makefile
katexochen: I use the following in `settings.json` for normalizing whitespace:

```json
    "files.trimTrailingWhitespace": true,
    "files.trimFinalNewlines": true,
    "files.insertFinalNewline": true,
```

.dockerignore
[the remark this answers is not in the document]
katexochen: My comment was just to remove the second newline here. ;)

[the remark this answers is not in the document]
jtumber-ibm: Oh okay

wainersm:
@jtumber-ibm how does this PR is related with https://github.com/confidential-containers/cloud-api-adaptor/pull/1018 ? does this supersedes PR 1018?

wainersm:
@jtumber-ibm still trying to understand your changes, I just glanced at the code. One question that I have is about `f binaries and builder already exist for current versions.yaml file then make docker-image will be enough`...

Apart from `versions.yaml`,it seems that build flags and build scripts themselves that would invalidate a cache. For example, suppose we decide to change `ttrc` to false in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/Makefile.inc#L149 while keeping the same version of guest components.

Makefile
```diff
@@ -162,3 +164,46 @@ ifneq ($(CLOUD_PROVIDER),)
...
 endif
+
+### PODVM IMAGE BUILDING ###
+
+REGISTRY ?= quay.io/confidential-containers
+
+PODVM_DISTRO ?= ubuntu
+
+PODVM_BUILDER_IMAGE ?= $(REGISTRY)/podvm-builder-$(PODVM_DISTRO):$(VERSIONS_HASH)
+PODVM_BINARIES_IMAGE ?= $(REGISTRY)/podvm-binaries-$(ARCH):$(VERSIONS_HASH)
```
wainersm, line 175: Missing `$(PODVM_DISTRO)` in binaries image name.

jtumber-ibm:
> @jtumber-ibm how does this PR is related with #1018 ? does this supersedes PR 1018?

It does supersede #1018, I'll ask @yoheiueda if it is okay to close #1018

Makefile
jtumber-ibm: Are the binaries DISTRO specific?

jtumber-ibm:
> @jtumber-ibm still trying to understand your changes, I just glanced at the code. One question that I have is about `f binaries and builder already exist for current versions.yaml file then make docker-image will be enough`...
> 
> Apart from `versions.yaml`,it seems that build flags and build scripts themselves that would invalidate a cache. For example, suppose we decide to change `ttrc` to false in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/Makefile.inc#L149 while keeping the same version of guest components.

That is true, and a good point. I don't have a great answer for how to solve it though. Do you have a recommendation for how to track those changes / invalidate the cache when the build changes? My thought was to add a `build-version` entry into the versions.yaml but that isn't easy to track against the actual changes

Makefile
katexochen: If they are dynamically linked they are, yes.

katexochen: See https://github.com/confidential-containers/cloud-api-adaptor/issues/1093#issuecomment-1600991335 for an example.

katexochen:
> > @jtumber-ibm still trying to understand your changes, I just glanced at the code. One question that I have is about `f binaries and builder already exist for current versions.yaml file then make docker-image will be enough`...
> > Apart from `versions.yaml`,it seems that build flags and build scripts themselves that would invalidate a cache. For example, suppose we decide to change `ttrc` to false in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/Makefile.inc#L149 while keeping the same version of guest components.
> 
> That is true, and a good point. I don't have a great answer for how to solve it though. Do you have a recommendation for how to track those changes / invalidate the cache when the build changes? My thought was to add a `build-version` entry into the versions.yaml but that isn't easy to track against the actual changes

I don't think we should build an entire build system from scratch. If we want such features, we will have to use a real build system like Nix or Bazel.

Makefile
jtumber-ibm: Ah interesting, thanks. I'll make the name distro specific then

wainersm:
> > > jtumber-ibm still trying to understand your changes, I just glanced at the code. One question that I have is about `f binaries and builder already exist for current versions.yaml file then make docker-image will be enough`...
> > > Apart from `versions.yaml`,it seems that build flags and build scripts themselves that would invalidate a cache. For example, suppose we decide to change `ttrc` to false in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/Makefile.inc#L149 while keeping the same version of guest components.
> > 
> > 
> > That is true, and a good point. I don't have a great answer for how to solve it though. Do you have a recommendation for how to track those changes / invalidate the cache when the build changes? My thought was to add a `build-version` entry into the versions.yaml but that isn't easy to track against the actual changes
> 
> I don't think we should build an entire build system from scratch. If we want such features, we will have to use a real build system like Nix or Bazel.

Yes, sure, don't want to re-invent the wheels. I am not sure about Nix, but even with bazel we would still have problems because the main concern is the build inside containers.

okay, back to the problem... ignore it? i.e. assume in a rare situation we might need to change build parameters but not `versions`, then defer the solution?

If not ignore then I can think of either:
 * Add a new property to `versions` to version the build configuration (e.g. `build_version=N`) that we can bump in case of we want to force the rebuild for whatever reason.
 * Apart from `versions.yaml`, compute the build files (Makefiles, etc) in the `VERSION_HASH`. This is the approach used by kata containers, btw.

Both approach are error prone: one might forget to update the `build_version`. One might forget to include new build files on the computation of the `VERSION_HASH`.

BTW, @fidencio has recently added (https://github.com/kata-containers/kata-containers/pull/8100)  the infra to allow building kata-agent separated from the entire rootfs and, IIUC, the idea is to cache the kata-agent binary just like the other components. In other words, soon we might be able to consume kata-agent binaries... In Fabiano we trust, In fabiano we rest our hopes!

wainersm:
@jtumber-ibm perhaps one last thing not relate with code properly saying. When we (@bpradipt and I) wrote the first implementation, the idea of pulling the sources from git tree was to enable reproducibility. Switching to copying the local sources to the image we lose reproducibility if I am not mistaken. As a developer I love the simplicity that this solution brings to my life (I can even feel the inner peace on my heart growing...) but I'm not sure how it impacts on peer pods' distributors. Is this (loosing reproducibility) something already discussed on the community? I missed many discussions recently so it might be case....

katexochen:
> When we (bpradipt and I) wrote the first implementation, the idea of pulling the sources from git tree was to enable reproducibility. 

IMO pulling code from GitHub has nothing to do with reproducibility. You can totally have reproducible builds from a local checkout. If you pull things from remote, you have to pin everything you pull in (by hash!). We never did that, and the way we pulled things, and still pull things, is not reproducible at all. Some examples from current main:

We don't even pin the base image by hash. The meaning of the tag changes multiple times a week:

https://github.com/confidential-containers/cloud-api-adaptor/blob/cf07e8adb388523a81c39483974fd2b8d0e03a13/podvm/Dockerfile.podvm_builder#L9


We pull CAA from main. If I build two times and you merge a PR in between, I get different results:

https://github.com/confidential-containers/cloud-api-adaptor/blob/cf07e8adb388523a81c39483974fd2b8d0e03a13/podvm/Dockerfile.podvm_builder#L57

Kata isn't pinned as well, as is any other of the many tools we are building:

https://github.com/confidential-containers/cloud-api-adaptor/blob/cf07e8adb388523a81c39483974fd2b8d0e03a13/podvm/Dockerfile.podvm_builder#L60

Currently, the only reproducible bit in our builds yq (and the stuff we already pinned in versions.yaml):

https://github.com/confidential-containers/cloud-api-adaptor/blob/cf07e8adb388523a81c39483974fd2b8d0e03a13/podvm/Dockerfile.podvm_builder#L14-L15

katexochen:
> I am not sure about Nix, but even with bazel we would still have problems because the main concern is the build inside containers.

I don't get that point. You wouldn't use those tools within a container, they come with their own sandbox.

Makefile
jtumber-ibm: I have moved the `yq` values out as well

Makefile.defaults
```diff
@@ -5,11 +5,19 @@ ifeq (, $(shell command -v $(YQ_COMMAND) 2> /dev/null))
 $(error "$(YQ_COMMAND) not found, consider doing snap install yq")
 endif
 
-YQ_MAJOR_VERSION := $(shell $(YQ_COMMAND) --version | sed -nE 's/.* v?([0-9]+).*/\1/p')
+# As we need `yq` to be installed to fetch the values from versions.yaml
+# these values are hardcoded here, all other versions/references should be
+# stored in versions.yaml
+YQ_VERSION := v4.35.1
+YQ_CHECKSUM := "sha256:bd695a6513f1196aeda17b174a15e9c351843fb1cef5f9be0af170f2dd744f08"
```
katexochen, line 12: Can't we move this into `versions.yaml` ?

jtumber-ibm: Do you think we should add a new section for remote files/binaries in the versions.yaml? i.e. At the moment the `tools` section doesn't have a field for checksums. Obviously YAML is flexible that we could just do that the `yq` ...

```yaml
tools:
  golang: 1.20.8
  rust: 1.72.0
  protoc: 3.11.4
  yq:
    version: ...
    checksum: ...
```

Although we don't have a strict schema to follow be nice to keep it consistent.

katexochen: Your right. Let's skip it then. I think the versions.yaml thing won't be for every anyway, as it just can't replace a full build system.

katexochen (approved):
LGTM. Local build was successful, full build took 38 min.

stevenhorsman (commented):
I've not had a chance to look through all the code, but I've run it locally with `ARCH=s390x make podvm-builder podvm-binaries podvm-image` and it created an image at least. I hit one problem on the way and I think we might need to update the documentation of how to build the podvm based on this as an easier way to run it locally. I hope to review it more fully next week. Thanks and sorry for the delays in the review

Makefile.defaults
```diff
@@ -5,11 +5,19 @@ ifeq (, $(shell command -v $(YQ_COMMAND) 2> /dev/null))
...
 
-ifneq (4, $(YQ_MAJOR_VERSION))
-$(error "yq major version should be 4, is: $(YQ_MAJOR_VERSION)")
+MINIMUM_YQ_MAJOR_VERSION := 4
```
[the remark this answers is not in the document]
katexochen: One of those problems we could solve with a reproducible dev environment provided by Nix...

surajssd:
What is blocking this PR?

stevenhorsman:
> What is blocking this PR?

I think JT needs to rebase once they are back from vacation and it needs another review, which I intend to do once it's rebased to test the images that are created, and run it on s390x and x86, rather than just built a single one to test out the process. I think some doc updates might be important as well as I mention in my comment last week.

Makefile.defaults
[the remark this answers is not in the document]
jtumber-ibm: https://github.com/mikefarah/yq/issues/1158#issuecomment-1081245199

We could specify `eval (e)` in our commands instead to ensure we are compatible with all the 4.x minor versions

jtumber-ibm:
@surajssd @stevenhorsman rebased and doc update added

stevenhorsman (approved):
I recreated an s390x podvm based on the documentation and tried it out locally and it seems to pass the tests, so LGTM. Thanks for the updates!

PR #1460, opened 2023-09-22
forwarder: customize the listener port

Amulyam24:
By default the port is 15150, this adds the support to customize it to a specific port. Currently changes added for AWS, Azure and IBM Cloud PowerVS providers.

Fixes: #1457

### DIFF
--- a/ibmcloud-powervs/image/Makefile
+++ b/ibmcloud-powervs/image/Makefile
@@ -5,6 +5,8 @@
 SKOPEO_SRC =
 SKOPEO_BIN = /usr/bin/skopeo
 
+export FORWARDER_PORT ?=
+
 ROOT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))../../
 include $(ROOT_DIR)podvm/Makefile.inc
 
--- a/aws/README.md
+++ b/aws/README.md
@@ -27,6 +27,11 @@ For example if you want to set the volume size to 40 GiB, then do the following:
 export VOLUME_SIZE=40
 ```
 
+If you want to use a specific port or address for `agent-protocol-forwarder`, set `FORWARDER_PORT` environment variable.
+```
+export FORWARDER_PORT=<port-number>
+```
+
 - Create a custom AWS VM image based on Ubuntu 22.04 having kata-agent and other dependencies
 
 > **NOTE**: For setting up authenticated registry support read this [documentation](../docs/registries-authentication.md).
--- a/aws/image/rhel/aws-rhel.pkr.hcl
+++ b/aws/image/rhel/aws-rhel.pkr.hcl
@@ -91,7 +91,8 @@ build {
     environment_vars = [
       "CLOUD_PROVIDER=${var.cloud_provider}",
       "PODVM_DISTRO=${var.podvm_distro}",
-      "DISABLE_CLOUD_CONFIG=${var.disable_cloud_config}"
+      "DISABLE_CLOUD_CONFIG=${var.disable_cloud_config}",
+      "FORWARDER_PORT=${var.forwarder_port}"
     ]
     inline = [
       "sudo -E bash ~/misc-settings.sh"
--- a/aws/image/rhel/variables.pkr.hcl
+++ b/aws/image/rhel/variables.pkr.hcl
@@ -62,3 +62,8 @@ variable "enable_nvidia_gpu" {
   type    = string
   default = env("ENABLE_NVIDIA_GPU")
 }
+
+variable "forwarder_port" {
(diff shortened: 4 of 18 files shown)


yoheiueda (commented):
I think `FORWARDER_ADDRESS` can also be configuration.

How about providing `FORWARDER_LISTEN` or `FORWARDER_LISTEN_ADDR` instead of `FORWARDER_PORT` only?

The default value of `FORWARDER_LISTEN` will be `0.0.0.0:15150`.

Amulyam24:
> The default value of `FORWARDER_LISTEN` will be `0.0.0.0:15150`.

Hi @yoheiueda, are you suggesting we pass `FORWARDER_LISTEN`(addr+port) by default to the agent-protocol-forwarder?

> I think `FORWARDER_ADDRESS` can also be configuration.

Okay

yoheiueda:
@Amulyam24 

My point was that it is preferable to make the `0.0.0.0` part configurable in addition to the  `15150` part. 

>  are you suggesting we pass FORWARDER_LISTEN(addr+port) by default to the agent-protocol-forwarder?

So, when a user want to use the default `0.0.0.0:15150`, I don't think we need to explicitly specify the value in the command line argument of `agent-protocol-forwarder`.

yoheiueda (approved):
LGTM.  Thank you very much!

bpradipt:
Is using a customisable listen address (not port) applicable ? Since CAA uses the node IP (public or private) and the port to connect to the APF. 
@yoheiueda @Amulyam24  is there a scenario that I'm missing?

Amulyam24:
> applicable ? Since CAA uses the node IP (public or private) and the port to connect to the APF.

That's right @bpradipt. I missed this point. I will let @yoheiueda answer it if he has any specific scenario case for it

yoheiueda:
@bpradipt @Amulyam24 Oh, that's right. I don't come up with a specific scenario. I think we can go ahead without `FORWARDER_ADDRESS`.

When a pod VM image supports multiple network interfaces in the future, we will possibly provide an option to select a specific network address. But, the current deployment mechanism only supports a single network, so there are no appropriate value to be set to`FORWARDER_ADDRESS`.

bpradipt:
@Amulyam24 can you please fix the commit check

podvm/Makefile.inc
```diff
@@ -32,6 +32,10 @@ RUST_TARGET := $(RUST_ARCH)-unknown-linux-$(LIBC)
 CDH_RESOURCE_PROVIDER ?= kbs
 SEALED_SECRET ?= yes
 
+export FORWARDER_PORT ?=
+export FORWARDER_ADDRESS := 0.0.0.0
+export FORWARD_LISTEN = $(FORWARDER_ADDRESS):$(FORWARDER_PORT)
+
```
bpradipt, line 38: Do we really need these variables in the Makefile? Can't we just use this as env variable for packer to pickup ?

Amulyam24: Yes, they are not needed here. Moved to PowerVS specific Makefile and modifies packer variables accordingly, PTAL!

Amulyam24: Looking into the packer failures

Amulyam24:
Hi @bpradipt, did you get a chance to verify this option for AWS/Azure?

bpradipt:
@Amulyam24 sorry this PR got missed. Can you please rebase and resolve the conflicts ?

ibmcloud-powervs/image/Makefile
```diff
@@ -5,6 +5,10 @@
 SKOPEO_SRC =
 SKOPEO_BIN = /usr/bin/skopeo
 
+FORWARDER_ADDRESS := 0.0.0.0
+export FORWARDER_PORT ?=
+export FORWARDER_LISTEN = $(FORWARDER_ADDRESS):$(FORWARDER_PORT)
+
```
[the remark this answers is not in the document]
bpradipt: FORWARDER_ADRESS is not needed - https://github.com/confidential-containers/cloud-api-adaptor/pull/1460#issuecomment-1742326305
Only the port is configurable.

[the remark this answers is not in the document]
Amulyam24: Right, the `FORWARDER_ADDRESS`, missed removing it here. Addressed, PTAL.

bpradipt (approved):
/lgtm
Thanks @Amulyam24

stevenhorsman (approved):
Looks okay to me.

Amulyam24:
Rebased the PR to resolve conflicts.

PR #1470, opened 2023-09-25
azure: build podvm with cc_kbc by default

mkulke:
We probably want to have `cc_kbc` (being able to talk to kbs) in the nightly images instead of `offlinefs_kbc`, which is the default. it's not trivial to have a build that supports that feature, so it would be good to have it baked into the image

### DIFF
--- a/.github/workflows/azure-podvm-image-build.yml
+++ b/.github/workflows/azure-podvm-image-build.yml
@@ -32,10 +32,11 @@ env:
   PODVM_IMAGE_NAME: "peerpod-image-${{ github.run_id }}-${{ github.run_attempt }}"
   SSH_USERNAME: "peerpod"
   VM_SIZE: "Standard_D2as_v5"
+  AA_KBC: "cc_kbc_az_snp_vtpm"
 
 jobs:
   build-podvm-image:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     defaults:
       run:
         working-directory: cloud-api-adaptor/azure/image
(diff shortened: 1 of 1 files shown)


surajssd:
With this feature we will be able to start doing attestation? Talking to KBS?

bpradipt:
@mkulke do you also want to add another commit to build the image with `DISABLE_CLOUD_CONFIG=true` ? (Can be in a separate PR also)

mkulke:
> With this feature we will be able to start doing attestation? Talking to KBS?

yes `cc_kbc` is the kbc module that will interact with a CoCo KBS. strictly, attestation would be also triggered with other kbs modules like `eaa_kbc`, but `cc_kbc` is probably the only one interesting in this context atm.

mkulke:
> @mkulke do you also want to add another commit to build the image with `DISABLE_CLOUD_CONFIG=true` ? (Can be in a separate PR also)

That would require changes to the CAA configuration, so we probably want to do that in another PR.

bpradipt (approved):
/lgtm
Thanks @mkulke

mkulke:
Note: We cannot merge this atm, because the resulting podvm images do not contain tdx attestation libraries. We don't need tdx support atm, but you cannot build a cc_kbc-enabled binary without linking to those libraries. attestation-agent will fail to start

mkulke:
ok, adding `CLOUD_PROVIDER=azure` to the environment when starting packer will install the required libs. It's probably a bit redundant that we have to do that, since `misc-settings.sh` is an azure-specific file, but I wouldn't bother looking into this, it might introduce breakage if we remove this condition.

However: in my tests image decryption **did not work** using cc_kbc

cc_kbc initializes an attestation, because it requests a dummy `auth.json` file from kbs (this is the desired behaviour if you set `aa_kbc_params` atm). This is the kbs log for an unencrypted image:

image

But image-rs does not request a decryption key for encrypted layers. I suspect something in the parsing code changed.

I would opt for merging it anyway since it's not a regression. Unencrypted images continue to work. Further investigation needs to be performed to find out why image decryption does not work like intended, but cc_kbc is able to do a remote attestation.

surajssd (approved):

mkulke:
> misc-settings.sh is a common file that is copied across all providers. Ideally we should just figure out a way to refer to it from podvm directory itself.

👍 

> ok to merge then ?

I'm testing against the latest CDH + versions.yaml updates that were just merged. There was a change in the build-system of attestation-agent, just specifying `AA_KBC=cc_kbc` will not include any real attesters, the fix is probably trivial (`AA_KBC=cc_kbc_az_snp_vtpm`), but I have to test

mkulke:
Decrypting images work again with the latest AA changes, so that's good.

image

Unencrypted images will only request the dummy auth.json:

image

Pulling unecrypted images also works with an empty `AA_KBC_PARAMS` setting.

kartikjoshi21 (approved):

Issue #688, opened 2023-03-08
Encrypting pod network traffic between worker node and pod VM


bpradipt:
The default networking solution extends pod network by creating a VXLAN tunnel between worker node and pod VM. 
Does this need to be encrypted? If yes what are the available mechanisms ?

bpradipt:
Few additional questions:
1. When pod-to-pod traffic is secured via mTLS using service mesh, does it makes sense to encrypt the tunnel traffic ?
2. If the workload is TLS enabled then does it makes sense to encrypt the tunnel traffic ?

As for the options to encrypt the tunnel traffic, I could find ipsec and [macsec](https://developers.redhat.com/blog/2016/10/14/macsec-a-different-solution-to-encrypt-network-traffic)

The fundamental question is do we really need to encrypt the tunnel traffic ?

yoheiueda:
I think it depends on how Kubernetes pod network is configured.

In a typical case, inter-node communication of Kubernetes pod network is not encrypted. In this case, we don't need to encrypt the tunnel traffic.

If inter-node communication of Kubernetes pod network is encrypted, the tunnel traffic should be encrypted. For example, Calico supports encryption of in-cluster pod traffic. https://docs.tigera.io/calico/3.25/network-policy/encrypt-cluster-pod-traffic

yoheiueda:
WireGuard is gaining popularity for secure tunnels.

It looks like we can implement WireGuard tunnels for CAA in a relatively straightforward way.

https://www.wireguard.com/quickstart/
https://www.wireguard.com/netns/

bpradipt:
> I think it depends on how Kubernetes pod network is configured.
> 
> In a typical case, inter-node communication of Kubernetes pod network is not encrypted. In this case, we don't need to encrypt the tunnel traffic.
> 
> If inter-node communication of Kubernetes pod network is encrypted, the tunnel traffic should be encrypted. For example, Calico supports encryption of in-cluster pod traffic. https://docs.tigera.io/calico/3.25/network-policy/encrypt-cluster-pod-traffic

@yoheiueda this makes sense.  We can provide it as an optional capability for cases where pod-to-pod traffic is not encrypted by default. For cases where pod-to-pod traffic is encrypted we can use wireguard.

I'm assuming mTLS using service mesh should work as-is. Although I have personally not tested service-mesh on a peer-pods environment.

huoqifeng:
[...]

PR #1474, opened 2023-09-27
fips support in podvm builds

snir911:
for now tested only with RHEL, some adaptions required for ubuntu support

### DIFF
--- a/aws/image/misc-settings.sh
+++ b/aws/image/misc-settings.sh
@@ -99,6 +99,24 @@ ExecStartPre=
 END
 fi
 
+# enable FIPS mode
+if [ "$BOOT_FIPS" == "true" ]; then
+# Add vsphere specific commands to execute on remote
+    case $PODVM_DISTRO in
+    centos)
+        #fallthrough
+        ;&
+    rhel)
+        fips-mode-setup --enable
+        ;;
+    ubuntu)
+        # TODO: configure the bootloader to add kernel options "fips=1"
+        ;;
+    *)
+        ;;
+    esac
+fi
+
 # Disable unnecessary systemd services
 
 case $PODVM_DISTRO in
--- a/podvm/Dockerfile.podvm_binaries.rhel
+++ b/podvm/Dockerfile.podvm_binaries.rhel
@@ -41,6 +41,6 @@ RUN cd cloud-api-adaptor/podvm && \
 RUN tar czvf /podvm-binaries.tar.gz -C /src/cloud-api-adaptor/podvm/files usr/ etc/
 RUN tar czvf /pause-bundle.tar.gz -C /src/cloud-api-adaptor/podvm/files pause_bundle/
 
-FROM registry.access.redhat.com/ubi8/ubi:8.7
+FROM registry.access.redhat.com/ubi9/ubi:9.2
 COPY --from=podvm_builder /podvm-binaries.tar.gz /
 COPY --from=podvm_builder /pause-bundle.tar.gz /
--- a/versions.yaml
+++ b/versions.yaml
@@ -14,6 +14,12 @@ cloudimg:
       amd64:
         url: https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-GenericCloud-8-20220913.0.x86_64.qcow2
         checksum: "sha256:8717251f8e4d2fe3e5032799caae89358c1ba68d65a16b5128a59ec6003aac1c"
+  rhel:
+    9.2: # dummy link, get trial image from: https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/server/trial
+      amd64:
+        url: https://access.redhat.com/downloads/content/479/ver=/rhel---9/9.2/x86_64/product-software
+        checksum: "sha256:34ff41b5274692c984e3860b21136af8b6ae502744c6c7578dda82002fba0287"
+
 tools:
   golang: 1.20.8
   rust: 1.72.0
--- a/Makefile.defaults
+++ b/Makefile.defaults
@@ -18,6 +18,7 @@ endef
 
 ubuntu_RELEASE ?= focal
 centos_RELEASE ?= stream-8
+rhel_RELEASE ?= 9.2
 
 ubuntu_amd64_IMAGE_URL := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).amd64.url)
 ubuntu_amd64_IMAGE_CHECKSUM := $(call query,cloudimg.ubuntu.$(ubuntu_RELEASE).amd64.checksum)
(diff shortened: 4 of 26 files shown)


podvm/Dockerfile.podvm_binaries.rhel
```diff
@@ -41,6 +41,6 @@ RUN cd cloud-api-adaptor/podvm && \
 RUN tar czvf /podvm-binaries.tar.gz -C /src/cloud-api-adaptor/podvm/files usr/ etc/
 RUN tar czvf /pause-bundle.tar.gz -C /src/cloud-api-adaptor/podvm/files pause_bundle/
 
-FROM registry.access.redhat.com/ubi8/ubi:8.7
```
bpradipt, line 44: Can you also add a new commit for CentOS to use centos9 stream and codeready-builder ?

bpradipt (approved):
/lgtm
Few suggestions to include either in this PR or next
1. Updating CentOS to CentOS 9 stream
2. Adding ENABLE_FIPS in the README

bpradipt:
@snir911 it might be helpful to paste test results with FIPS enabled CVM to be sure nothing broke from a functionality standpoint.

mkulke:
I don't have practical experience with deploying FIPS-enabled applications, but I suspect this has implications beyond the kernel and needs to be enforced in the userland, so this applies to all kinds of staticly linked binaries like CAA, kata-agent and guest-components. AFAIK there are no FIPS certified rust crypto-modules, not sure about go. We can link against fips-mode openssl probably, but we'd have to do this consistently.

Should we limit the scope of the flag (`FIPS_KERNEL`) and maybe abort the image build if `ENABLE_FIPS` is set, but we don't support it for a given distro yet?

IMO, we could be more succinct in naming our flags: e.g. `ENABLE_FIPS=true` and `DISABLE_CLOUD_CONFIG=true` could bw shortened to `FIPS=true` `CLOUD_CONFIG=false`.

katexochen:
>  not sure about go

Go can link against boringSSL for FIPS compliant crypto.

I know some customers/use cases require FIPS, but letting a 'I don't like fips' here anyway. :wink: 

Other things of interest might be the plan to use wireguard for podvm tunnel encryption (https://github.com/confidential-containers/cloud-api-adaptor/issues/688), as wg isn't FIPS compliant. Would likely need replacement with IPSec for FIPS.

aws/image/misc-settings.sh
```diff
@@ -99,6 +99,24 @@ ExecStartPre=
...
+# Add vsphere specific commands to execute on remote
+    case $PODVM_DISTRO in
+    centos)
+        #fallthrough
+        ;&
+    rhel)
+        fips-mode-setup --enable
+        ;;
+    ubuntu)
+        # TODO: configure the bootloader to add kernel options "fips=1"
```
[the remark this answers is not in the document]
snir911: It seems there's a helper tool only for Pro subscription, i'm not sure how handle ATM

snir911:
Yes, kernel wouldn't be enough for a whole FIPS compatibility, maybe it would be better to name the variable FIPS_BOOT to make it more clear, WDYT?
I was aware to the binaries incompatibility (mostly in rust) but i didn't think of the tunneling

bpradipt:
> Yes, kernel wouldn't be enough for a whole FIPS compatibility, maybe it would be better to name the variable FIPS_BOOT to make it more clear, WDYT? I was aware to the binaries incompatibility (mostly in rust) but i didn't think of the tunneling

We don't yet use wg based tunnelling. But it's an important consideration.

podvm/Dockerfile.podvm_binaries.rhel
snir911: lets do CentOS in separate i hit weird issue with pkgs namings

snir911:
I add a note in docs and changes the var name to BOOT_FIPS.
I intentionally don't add the variable to dockerfile until we'll have better coverage also for Ubuntu (and binaries?)

versions.yaml
```diff
@@ -14,6 +14,12 @@ cloudimg:
       amd64:
         url: https://cloud.centos.org/centos/8-stream/x86_64/images/CentOS-Stream-GenericCloud-8-20220913.0.x86_64.qcow2
         checksum: "sha256:8717251f8e4d2fe3e5032799caae89358c1ba68d65a16b5128a59ec6003aac1c"
+  rhel:
+    9.2: # dummy link, get trial image from: https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/server/trial
+      amd64:
+        url: https://access.redhat.com/downloads/content/479/ver=/rhel---9/9.2/x86_64/product-software
+        checksum: "sha256:34ff41b5274692c984e3860b21136af8b6ae502744c6c7578dda82002fba0287"
```
bpradipt, line 21: nit: may be you want to edit the commit on dummy link and also add any other prereq to get the image like rhel id etc

snir911: I'm not sure i follow, if you use trial no prerequisites are needed IIUC

jensfr (approved):
Hi @snir911, changes look good to me!

snir911:
@mkulke @katexochen are you ok with this? i modified the var name to BOOT_FIPS to make it clear it's fips boot compatibility

katexochen:
> mkulke katexochen are you ok with this? i modified the var name to BOOT_FIPS to make it clear it's fips boot compatibility

Okay for me, thanks for asking. I still don't fully get what the benefit of this is. If your userland isn't FIPS compliant, what do you get from a FIPS compliant kernel? The podvm still isn't compliant (and won't be in the near future). Is there any benefit of partial compliance for customers? Just trying to understand the use of this..

mkulke:
> > mkulke katexochen are you ok with this? i modified the var name to BOOT_FIPS to make it clear it's fips boot compatibility
> 
> Okay for me, thanks for asking. I still don't fully get what the benefit of this is. If your userland isn't FIPS compliant, what do you get from a FIPS compliant kernel? The podvm still isn't compliant (and won't be in the near future). Is there any benefit of partial compliance for customers? Just trying to understand the use of this..

yup, same for me. My concern was mostly the broad claim of `FIPS=true`, I understand sometimes there are box-checking exercises which would mandate a FIPS kernel, even if the full scope of the application is not covered

snir911:
Clearly it requires much more work to have CoCo, as whole, FIPS complaint, however, i think it can still useful for PeerPods usecases
* This will allow mimicking FIPS environment for FIPS complaint workloads for testing, CI etc..
* You can allow partial FIPS compatibility which can be helpful in certain cases
* once we'll be able to provide FIPS compatible binaries we'll be already in a better place, for rust IIUC  openssl, boringssl and aws-lc has some sort of FIPS compatability, it requires some investigation

katexochen:
> This will allow mimicking FIPS environment for FIPS complaint workloads for testing, CI etc..

My concern here is that each of such options adds an additional dimension to our build and test matrix. We have a ton of options how things can be configured, but we don't have test to test them. And as many options aren't tested, things are silently broken and the whole thing is difficult to use. If we would actually want to test all the combinations, our CI hours would explode.

That's also why I really aim to keep the matrix low with the new mkosi images: one distro, ideally only one configuration. However, I think mkosi will also make it much easier to add additional configuration to images out of tree.

>  I understand sometimes there are box-checking exercises which would mandate a FIPS kernel, even if the full scope of the application is not covered

If this is required anyway for the box-checkers, I totally understand.

mkulke:
> That's also why I really aim to keep the matrix low with the new mkosi images: one distro, ideally only one configuration. However, I think mkosi will also make it much easier to add additional configuration to images out of tree.

I want to underline this point, every toggle that we add permutations to the testing burden, which is already pretty large atm (cloud vendor, operating system, ...)

snir911:
ok, thanks for your comments, I'll be closing it for now (at least until we'll know how to adapt the binaries for fips as well)

Issue #1467, opened 2023-09-25
RFC: Remove cloud-config / cloud-init from PodVM Image?


mkulke:
# Problem Statement

Currently, cloud-config is running at VM startup to configure the system.

## How we use it

- If `DISABLE_CLOUD_CONFIG` is unset, we populate the fs with configuration files using the `write_files` directive.
- We use it to start CSP-specific agents, like Azure's `waagent`.
- Potentially network configuration

## Why we might not want to use it

Cloud Config is a venue to [execute](https://cloudinit.readthedocs.io/en/latest/reference/examples.html#run-commands-on-first-boot) arbitrary untrusted code at startup.

# Suggested Solution

Replace the required functionality (config file creation, cloud callback, network configuration) with systemd units. We started this work by providing an option to write config files by parsing user-data directly (`DISABLE_CLOUD_CONFIG=true`).

# Alternative Solutions

- Measure cloud-config data at early boot stages, prior to processing.

# Open Problems

I'm not sure whether alternatives to cloud-config are available. I think in some cases a virtual iso is attached and we'd need to support that one as a source in addition to `process-user-data`.

katexochen:
I agree we should remove cloud init. We have some hacks in Constellation to run without `waagent`.

mkulke:
Yes, that's possible and documented as a valid agent-less deployment now. The support situation for those VMs might be tricky for some deployments. So I'm not sure what e.g. the plans for RHEL CVMs & Cloud-Config are, since I suspect Cloud-Config in a TEE as problematic for all flavors of CVMs, not just PeerPod PodVMs.

FWIW, I've been running tests using a discrete [report-ready tool](https://github.com/kinvolk/azure-cvm-tooling/pull/34/files). PodVMs start reliably without cloud-config.

There might be CAA use cases for Cloud Config which we need to cover (or reconsider). Admin users and SSH pubkeys, configured in CAA, are currently provisioned via Cloud-Config to the PodVM. We could certainly support that in some way, but we don't want to rewrite cloud-config probably.

katexochen:
>  The support situation for those VMs might be tricky for some deployments. So I'm not sure what e.g. the plans for RHEL CVMs & Cloud-Config are, since I suspect Cloud-Config in a TEE as problematic for all flavors of CVMs, not just PeerPod PodVMs.

@bpradipt what is your opinion in this regard?

[...]

Issue #1035, opened 2023-06-02
Build system organization


katexochen:
There is currently quite some movement in the build system:

- #994 moves the build system to Docker first
- #1018 improves building local code (with some discussion about isolation in https://github.com/confidential-containers/cloud-api-adaptor/pull/994#issuecomment-1563755080)
- #598 aims to improve versioning (with the goal of supply-chain security and reproducible builds)

to only name a few. 

I propose we discuss the goals we want to reach with the build system (isolation, reproducibility, usability, etc.) and the tools to reach these goals as well as the greater organization of our build in this thread.

katexochen:
We likely should agree on what those goals mean to us. 

Like, when we are talking about reproducible builds, are we talking about bit-by-bit reproducibility? Do we want to archive this for all components? Should be easy for binaries. More effort will be needed for the container images, but doable. Reproducible podvm image? I don't think this is possible with packer.

katexochen:
I will add this to the agenda of the next peerpod meeting.

katexochen:
Related issue regarding the organization of podvm files/templates: https://github.com/confidential-containers/cloud-api-adaptor/issues/899

jtumber-ibm:
I've created https://github.com/confidential-containers/cloud-api-adaptor/pull/1057 which is one way to pin the versions used in the podvm build process. But this is more of a PoC at this point

mkulke:
Thanks for picking this up. To make this actionable, I would suggest to untangle as much as possible.

1) We probably have to deal with the BUSL license issue (replacing packer, terraform), to be CNCF compliant, but we should address this separately.
1) My understanding is that there is a particular problem with building images using **QEMU** + **x360** arch + **Ubuntu 22.04+**. libguestfs looks intriguing, if we can limit ourselves to just provisioning a bunch of static binaries and configuration files on top of an image. At the moment there is some conditional logic in scripts which run on the vm during image build (`misc-settings.sh`) This would need to be extracted and I'm not sure whether we can get around install some dependencies.
2) I agree that a caa_src is rather confusing and not having the option would reduce complexity in the build process.
3.1) Unless that's a requirement, I wouldn't say so. Maybe it would speed up uncached builds, since the individual bins can be built in parallel, but I don't think that justifies the additional complexity. Having the static images in a container that is tagged by a versions.yaml hash would be fine, imo.
[...]

PR #1489, opened 2023-10-02
podvm-mkosi: build podvm images from scratch with mkosi

katexochen:
This PR adds support for building podvm images from scratch with mkosi. The targeted distribution is fedora. With some additional work, these images will be fully measurable on systems with vTPM.

The toolchain is currently pinned using Nix. Feel free to ping me here or on slack in case you have any questions regarding setting up Nix or building the image.

#### From Source

- Rebuild CAA image with secure boot disabled (from this PR)
- Follow `podvm-mkosi/README.md` to build the podvm image.
When building the builder container, use `CAA_SRC` and `CAA_SRC_REF` to point to this branch.

- Upload the image _with the right settings_

#### Prebuilt

Prebuilt image for Azure: 
```
/CommunityGalleries/mkosi-87e1d07e-86a1-45fc-8151-bc3a6242f8e8/Images/mkosi_testing/Versions/0.4.2
```
Prebuilt CAA image: 
```
ghcr.io/katexochen/cloud-api-adaptor/cloud-api-adaptor:dev-249439d8c04a71058d965e294915a285808e425f-dirty
```

#### Accessing the pod VM

The images currently have serial console enabled. You can access the pod VM through the cloud providers portal or with `virsh console` for libvirt.


pkg/adaptor/cloud/azure/provider.go
```diff
@@ -419,7 +419,8 @@ func (p *azureProvider) getVMParameters(instanceSize, diskName, b64EncData strin
 		securityProfile = &armcompute.SecurityProfile{
 			SecurityType: to.Ptr(armcompute.SecurityTypesConfidentialVM),
 			UefiSettings: &armcompute.UefiSettings{
-				SecureBootEnabled: to.Ptr(true),
+				// Turn off, we want to do measured boot. Secure boot brings no additional security.
+				SecureBootEnabled: to.Ptr(false),
```
bpradipt, line 423: Can we use a config flag ( like `disable-cvm` ) for this setting with default set to false? I'm thinking about some scenarios where this setting could be a checklist item.

katexochen: I've added a flag `enable-secure-boot` that is false by default.

podvm-mkosi/README.md
```diff
@@ -0,0 +1,46 @@
...
+    -o type=local,dest="./resources/binaries-tree" \
+    - < ../podvm/Dockerfile.podvm_binaries.fedora
+```
+
+### Activating the Nix development shell
+
+This puts the dependencies from the Nix flake into your path.
+
+```sh
+nix develop ./..#podvm-mkosi
```
mkulke, line 33: On my nix installation I had to use:

```bash
nix develop ./..#podvm-mkosi --extra-experimental-features nix-command --extra-experimental-features flakes
```

also I had to install `zstd` & `uidmap` on my buíld system. Are we able to put this into the nix recipe?

podvm-mkosi/README.md
```diff
@@ -0,0 +1,46 @@
...
+You can exit the environment later using <kbd>Ctrl</kbd> + <kbd>D</kbd>
+
+### Build the image
+
+```
+mkosi
+```
+
+
+docker buildx build -f Dockerfile.podvm_builder.fedora --build-arg CAA_SRC=https://github.com/katexochen/cloud-api-adaptor --build-arg CAA_SRC_REF=feat/mkosi-builds --no-cache -t podvm-builder-fed .
```
mkulke, line 46: leftover?

podvm-mkosi/README.md
katexochen: Mkosi requires `newuidmap` for rootless user namespaces. This cannot be part of the Nix shell, as you cannot give setuid privileges to something in `/nix/store`.

podvm-mkosi/README.md
```diff
@@ -0,0 +1,46 @@
...
+Currently, mksoi and other related tools are provided through a [Nix](https://nixos.org/) flake. Nix ensures all tools used in the build of the image are itself reproducible and pinned. mkosi requires a very recent systemd version, so using tools installed on the host is usually not possible.
+
+The recommended way to install Nix on Linux systems is the [Determinate Systems nix-installer](https://github.com/DeterminateSystems/nix-installer).
+
+
+## Building an image
+
+### Building PodVM binaries and place them in the binaries-tree
+
+```sh
```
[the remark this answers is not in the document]
katexochen: As far as I understand, this funkyness is manly in combination with `--platform` / cross-arch builds, right? Or isn't the given command working for you? I don't want to stick with this dockerfile for long and rather pull in the individual artifacts directly, but I thought it would be a good first step to give something familiar.

podvm-mkosi/README.md
katexochen: yes, removed.

podvm-mkosi/README.md
katexochen: added `zstd` to the shell.

podvm-mkosi/README.md
```diff
@@ -0,0 +1,46 @@
...
+docker buildx build \
+    -t podvm-builder-fedora \
+    --load \
+    - < ../podvm/Dockerfile.podvm_builder.fedora
```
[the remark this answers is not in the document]
katexochen: I guess we can also add `which` to the builder instead.

podvm-mkosi/README.md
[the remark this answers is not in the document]
mkulke: I had to switch to the docker driver for buildx as default, otherwise itwould not find the local image built in the first step (the builder img)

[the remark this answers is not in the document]
katexochen: What exactly did you change? Somehow this works out of the box for me.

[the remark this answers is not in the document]
mkulke:
```
$ docker buildx use default
```

pkg/adaptor/cloud/azure/provider.go
katexochen: @bpradipt could you check if its working out for you this way?

podvm-mkosi/README.md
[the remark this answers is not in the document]
katexochen: So would you suggest to push? Or should we set `--builder default` for the build?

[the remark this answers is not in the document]
katexochen: We could also just merge into a single Dockerfile.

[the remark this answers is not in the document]
mkulke: I think it's fine to add this line as a remark in the readme

mkulke:
Until we have the tool available, this is the `./podvm-mkosi/Makefile` i'm using to publish mkosi images to a registry, in case that helps with testing:

```Makefile
VHD_FILE := system-$(GALLERY_IMAGE_VERSION).vhd
VHD_ARTIFACT := build/$(VHD_FILE)
VHD_URL := https://$(AZURE_STORAGE_ACCOUNT).blob.core.windows.net/$(AZURE_STORAGE_CONTAINER)/$(VHD_FILE)
IMAGE_SIZE := 2G
BYTES := $(shell numfmt --from=iec $(IMAGE_SIZE))
MB := $$((1024 * 1024))
ROUNDED_SIZE := $$(($(BYTES) / $(MB) * $(MB)))
NOW_PLUS_1H := $(shell date -u -d '+1 hour' '+%Y-%m-%dT%H:%MZ')
SAS_TOKEN := $(shell az storage blob generate-sas \
	--blob-url $(VHD_URL) \
	--permissions rw \
	--expiry $(NOW_PLUS_1H) \
	--output tsv)

build/system.raw:
	mkosi

.PHONY: $(VHD_ARTIFACT)
$(VHD_ARTIFACT): build/system.raw
	qemu-img resize --shrink -f raw build/system.raw $(ROUNDED_SIZE) && \
	qemu-img convert -f raw -o subformat=fixed,force_size -O vpc build/system.raw $(VHD_ARTIFACT)

.PHONY: upload
upload: $(VHD_ARTIFACT)
	azcopy cp $(VHD_ARTIFACT) "$(VHD_URL)?$(SAS_TOKEN)"

.PHONY: publish
publish: upload
	az sig image-version create \
		--resource-group $(AZURE_RESOURCE_GROUP) \
		--gallery-name $(GALLERY_NAME) \
		--gallery-image-definition $(GALLERY_IMAGE_DEF_NAME) \
		--gallery-image-version $(GALLERY_IMAGE_VERSION) \
		--target-regions $(AZURE_REGION) \
		--os-vhd-uri $(VHD_URL) \
	    --os-vhd-storage-account $(AZURE_STORAGE_ACCOUNT)
```

mkulke:
I've been able to build my images and use them successfully, including remote attestation via cc_kbc. it's very clean & simple, I like it a lot. great work!

I'll post further comments with my findings

mkulke:
I noticed that `DISABLE_CLOUD_CONFIG=true` is implied on those images. is that correct at the moment? If so, we have to mention this in the Readme, since the peerpod configmap needs to be set accordingly.

mkulke:
The podvm takes some 75s to come up, which is longer than the usual time w/ packer-based podvm images (somewhere around 45s I think):

```bash
2023/10/06 14:10:19 [adaptor/cloud/azure] CreateInstance: name: "podvm-busybox-caa-7fcb8d777f-h9bq4-34de2b65"
2023/10/06 14:11:36 [adaptor/cloud/azure] created VM successfully: /subscriptions/redacted/resourceGroups/mgns/providers/Microsoft.Compute/virtualMachines/podvm-busybox-caa-7fcb8d777f-h9bq4-34de2b65
```

katexochen:
> I noticed that DISABLE_CLOUD_CONFIG=true is implied on those images. is that correct at the moment?

It should work, cloud init is installed in the image. I didn't understand yet why we have cloud init, and why it is possible to disable it. I would prefer to remove it completely, but I'm not sure about the implications.

I'll check again why cloud init might not be working correctly.

katexochen:
> Peeking into the podvms (I had to add openssh-server and sudo to the fedora packages to do this, later we probably want to add a debug build toggle)

Currently, serial console is enabled with auto login, so you can attach there to debug without additional requirements. Should have mentioned that. I think on the long run, I'd like to support a version with serial console enabled for debugging.

> I noticed it's in degraded state due to a failed TPM measurement unit in initrd

Is that with the latest commit? I noticed this issue and added `tpm-tools` to the initrd.

mkulke:
> > I noticed that DISABLE_CLOUD_CONFIG=true is implied on those images. is that correct at the moment?
> 
> It should work, cloud init is installed in the image. I didn't understand yet why we have cloud init, and why it is possible to disable it. I would prefer to remove it completely, but I'm not sure about the implications.
> 
> I'll check again why cloud init might not be working correctly.

cloud api adaptor by default uses cloud-config to populate the fs with certain config files. that is it will define user-data as a specific [yaml](https://cloudinit.readthedocs.io/en/latest/reference/examples.html) that starts with `#cloud-config`, so cloud-init will start processing it and distribute the files. `DISABLE_CLOUD_CONFIG` is a build (podvm) _and_ runtime (configmap) flag that changes that behaviour. We parse user-data explicitly and put the files into their places ourselves and thus we don't have to wait for cloud-init (cloud-config) to finish and the startup is quicker.

it's confusing, since we support both options and things will fail in non-obvious ways if you mix a certain image with the wrong caa config. In hindsight this should have been an experimental flag or replace cloud-config.

katexochen:
>  In hindsight this should have been an experimental flag or replace cloud-config. 

We should reduce that to a single way of doing it, ideally without cloud init. Could you bring https://github.com/confidential-containers/cloud-api-adaptor/issues/1467 in the next call? I won't be able to attend.

mkulke:
> > Peeking into the podvms (I had to add openssh-server and sudo to the fedora packages to do this, later we probably want to add a debug build toggle)
> 
> Currently, serial console is enabled with auto login, so you can attach there to debug without additional requirements. Should have mentioned that. I think on the long run, I'd like to support a version with serial console enabled for debugging. I've added that to the PR description.
> 
> > I noticed it's in degraded state due to a failed TPM measurement unit in initrd
> 
> Is that with the latest commit? I noticed this issue and added `tpm-tools` to the initrd.

yup. that's fixed now. thx.

katexochen:
Regarding the cloud init in the image:

cloud init seems to work in some way, not sure it does everything that is expected:

<details>
<summary>journalctl -u cloud-init</summary>

```
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: Starting cloud-init.service - Initial cloud-init job (metadata service crawler)...
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: Cloud-init v. 23.1.2 running 'init' at Fri, 06 Oct 2023 15:26:54 +0000. Up 28.59 seconds.
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: ++++++++++++++++++++++++++++++++++++++Net device info++++++++++++++++++++++++++++++++++++++
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +--------+------+------------------------------+-------------+--------+-------------------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: | Device |  Up  |           Address            |     Mask    | Scope  |     Hw-Address    |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +--------+------+------------------------------+-------------+--------+-------------------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |  eth0  | True |           10.9.0.5           | 255.255.0.0 | global | 60:45:bd:a2:43:b0 |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |  eth0  | True | fe80::6245:bdff:fea2:43b0/64 |      .      |  link  | 60:45:bd:a2:43:b0 |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   lo   | True |          127.0.0.1           |  255.0.0.0  |  host  |         .         |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   lo   | True |           ::1/128            |      .      |  host  |         .         |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +--------+------+------------------------------+-------------+--------+-------------------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: ++++++++++++++++++++++++++++++Route IPv4 info+++++++++++++++++++++++++++++++
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-----------------+----------+-----------------+-----------+-------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: | Route |   Destination   | Gateway  |     Genmask     | Interface | Flags |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-----------------+----------+-----------------+-----------+-------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   0   |     0.0.0.0     | 10.9.0.1 |     0.0.0.0     |    eth0   |   UG  |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   1   |     10.9.0.0    | 0.0.0.0  |   255.255.0.0   |    eth0   |   U   |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   2   |     10.9.0.1    | 0.0.0.0  | 255.255.255.255 |    eth0   |   UH  |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   3   |  168.63.129.16  | 10.9.0.1 | 255.255.255.255 |    eth0   |  UGH  |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   4   | 169.254.169.254 | 10.9.0.1 | 255.255.255.255 |    eth0   |  UGH  |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-----------------+----------+-----------------+-----------+-------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +++++++++++++++++++Route IPv6 info+++++++++++++++++++
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: | Route | Destination | Gateway | Interface | Flags |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   0   |  fe80::/64  |    ::   |    eth0   |   U   |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   2   |    local    |    ::   |    eth0   |   U   |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: |   3   |  multicast  |    ::   |    eth0   |   U   |
Oct 06 15:26:54 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: new group: name=peerpod, GID=1000
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: new user: name=peerpod, UID=1000, GID=1000, home=/home/peerpod, shell=/bin/bash, from=none
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to group 'adm'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to group 'wheel'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to group 'systemd-journal'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to shadow group 'adm'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to shadow group 'wheel'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 useradd[546]: add 'peerpod' to shadow group 'systemd-journal'
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: 2023-10-06 15:26:59,736 - util.py[WARNING]: Failed generating key type rsa to file /etc/ssh/ssh_host_rsa_key
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: 2023-10-06 15:26:59,740 - util.py[WARNING]: Failed generating key type dsa to file /etc/ssh/ssh_host_dsa_key
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: 2023-10-06 15:26:59,743 - util.py[WARNING]: Failed generating key type ecdsa to file /etc/ssh/ssh_host_ecdsa_key
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 cloud-init[485]: 2023-10-06 15:26:59,747 - util.py[WARNING]: Failed generating key type ed25519 to file /etc/ssh/ssh_host_ed25519_key
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: Finished cloud-init.service - Initial cloud-init job (metadata service crawler).

```
</details>

But process user data seems to be failing:

<details>
<summary>journalctl -u process-user-data</summary>

```
Oct 06 15:26:53 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: Starting process-user-data.service - Process user data...
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: provider: azure, userDataUrl: http://169.254.169.254/metadata/instance/compute/userData?api-version=2021-01-01&format=text
Oct 06 15:26:59 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 0: invalid user data
Oct 06 15:27:04 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 1: invalid user data
Oct 06 15:27:09 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 2: invalid user data
Oct 06 15:27:14 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 3: invalid user data
Oct 06 15:27:19 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 4: invalid user data
Oct 06 15:27:24 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 5: invalid user data
Oct 06 15:27:29 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 6: invalid user data
Oct 06 15:27:34 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 7: invalid user data
Oct 06 15:27:39 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 8: invalid user data
Oct 06 15:27:44 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Retry attempt 9: invalid user data
Oct 06 15:27:44 podvm-test-5c64579748-xfb47-36b4c618 process-user-data[477]: Error: failed to get valid user data
Oct 06 15:27:44 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: process-user-data.service: Control process exited, code=exited, status=1/FAILURE
Oct 06 15:27:44 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: process-user-data.service: Failed with result 'exit-code'.
Oct 06 15:27:44 podvm-test-5c64579748-xfb47-36b4c618 systemd[1]: Failed to start process-user-data.service - Process user data.

```
</details>

mkulke:
there is something odd going on, somehow the api-server-rest symlink is not copied to the image, hence it's not enabled on the podvm

```bash
$ find resources/binaries-tree | grep api-server-rest
resources/binaries-tree/etc/systemd/system/api-server-rest.service
resources/binaries-tree/etc/systemd/system/multi-user.target.wants/api-server-rest.service
resources/binaries-tree/usr/local/bin/api-server-rest
$ rm -r build/*
$ mkosi
...
$ sudo losetup -P -f build/system.raw
$ sudo mount /dev/loop0p2 /tmp/mkosi-image
$ sudo find /tmp/mkosi-image | grep api-server-rest
/tmp/mkosi-image/etc/systemd/system/api-server-rest.service
/tmp/mkosi-image/usr/local/bin/api-server-rest
```

curiously there is a message in the build logs that mentions all our enabled CAA units but api-server-rest:

```
Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/process-user-data.service is added as a dependency to a non-existent unit multi-user.target.
Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/run-image.mount is added as a dependency to a non-existent unit multi-user.target.
Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/run-kata\x2dcontainers.mount is added as a dependency to a non-existent unit multi-user.target.
Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/setup-nat-for-imds.service is added as a dependency to a non-existent unit multi-user.target.
Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/kata-agent.service is added as a dependency to a non-existent unit multi-user.target.
...
```

katexochen:
> there is something odd going on, somehow the api-server-rest symlink is not copied to the image, hence it's not enabled on the podvm
> 
> 
> curiously there is a message in the build logs that mentions all our enabled CAA units but api-server-rest:
> 
> ```
> Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/process-user-data.service is added as a dependency to a non-existent unit multi-user.target.
> Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/run-image.mount is added as a dependency to a non-existent unit multi-user.target.
> Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/run-kata\x2dcontainers.mount is added as a dependency to a non-existent unit multi-user.target.
> Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/setup-nat-for-imds.service is added as a dependency to a non-existent unit multi-user.target.
> Unit /home/magnuskulke/dev/cloud-api-adaptor/podvm-mkosi/mkosi.workspace/.mkosi-tmp313wnsf3/root/etc/systemd/system/kata-agent.service is added as a dependency to a non-existent unit multi-user.target.
> ...
> ```

There is `podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system-preset/30-coco.preset`, which enables units. I need to add the new stuff there.

kartikjoshi21:
Tried this PR as well. Nice work. Thanks @katexochen. Also @mkulke is it possible to include Makefile you mentioned to upload vhd to azure gallery as part of some README until we don't have any tool available ?

kartikjoshi21:
> @kartikjoshi21 Tool was just published: https://github.com/edgelesssys/uplosi I'll add it to the readme.

Ohh Great, i will try it too. Thanks.

katexochen:
@mkulke The `process-user-data` service is still failing, and I'm not sure why. Also, the `api-server-rest` is failing as it cannot find `/run/confidential-containers/cdh.sock`, which might be related to the process-user-data fail. Any ideas where to look? I've also rebased on latest main.

.github/workflows/podvm_mkosi.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Build nix shell to cache dependencies
+        run: nix build .#devShells.x86_64-linux.podvm-mkosi
+
+      - name: Install crane
+        run: nix profile install nixpkgs#crane
+
+        # This removes the checkout and creates a btrfs volume with maximized
+        # build space.
+      - name: Maximize build space
```
mkulke, line 62: I suspect this is due to limited space available on the builder for images. Or why do we need that? locally my build folder is rather small (2GB)

mkulke (commented):
We have to be cautious about overspending the novelty budget in this PR. I'm generally ok with the choice of nix, but IMO this is better introduced as a conscious choice rather than en-passant, and at the expense of something that we remove instead (e.g. building binaries w/ docker).

Would it be possible to execute mkosi via docker (see df sample below) in the scope of this PR and address introduce nix separately? 

Also, are we fine with adding mkosi-builds merely as yet another additional option or should it replace the packer-based builds for x86_64 instead?

That said, I'm also fine w/ merging it as-is, but I'd like others to chime in.

```Dockerfile
FROM debian:unstable-20230919-slim
RUN apt-get update && apt-get install -y \
	cpio \
	mkosi \
	mtools \
	zstd
RUN mkdir /mkosi
COPY mkosi.presets /mkosi/mkosi.presets
COPY mkosi.skeleton /mkosi/mkosi.skeleton
COPY mkosi.conf /mkosi/mkosi.conf
WORKDIR /mkosi
```

katexochen:
> Also, are we fine with adding mkosi-builds merely as yet another additional option or should it replace the packer-based builds for x86_64 instead?

So for me, this is definitely meant as replacement. However, I don't think we should already remove other code in this PR. Some things required for the full replacement:

- Testing by a wider range of contributors. I don't want to break the full project in case things don't work out as expected.
- Migrate CI. This PR only has a simple build action, but we need to migrate image publishing, e2e tests etc.
- Security. Compared to the non-mkosi podvm images, there is currently a security regression. I would like to
  -  Create a production and debug variant of the mkosi image. The current image allows access through serial console
  - Add measured boot to replace the disabled secure boot.
  
If people engage, I think we can reach these this in the next release period and remove the old x86 podvm builds before the release.

.github/workflows/podvm_mkosi.yaml
mkulke: > Are you building on ext4 locally?
yup, maybe we can take a look at the mkosi CI jobs as a reference?

katexochen: The images they build in their CI are quite minimal. There aren't any actions regarding disk size, so I don't think it is a problem for them. We have the same space problem in Constellation. I would like to investigate at some point, but I couldn't find the cycles yet to dig into it.

katexochen:
> Would it be possible to execute mkosi via docker (see df sample below) in the scope of this PR and address introduce nix separately?

So **regarding Nix**. I thought about the docker thing, and there are several reasons why I don't think it is a good idea (or even practical).

---

>  FROM debian:unstable-20230919-slim

First, I've tried that container, it doesn't work for me (and I think it won't work at all with docker):

```
root@3e69dc3f74b0:/mkosi# mkosi
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

There might be a way to get this running with podman, but that would introduce another tool, too.

Also this images uses debian unstable. While mkosi 18 is already available there, afaik updates in that channel are enrolled automatically without any review, including major version bumps. There have been 3 major version releases in the last 2 months. When getting mkosi from unstable, builds would immediately break and there isn't anything we can do to prevent this or to roll back to a previous version in unstable.

Next, I highly doubt that builds on debian are reproducible. I've spent the last months getting this reproducible with Nix. The mkosi version in nixpkgs uses a patched version of systemd (with commits that are upstream, but will only be part of the next release). We also fixed many of the tools mkosi uses to get things reproducible.
I maintain the mkosi package in nixpkgs with a colleague. We have nightly tests running on nixpkgs master + mkosi main to ensure things stay reproducible and working.

---

> and at the expense of something that we remove instead (e.g. building binaries w/ docker).

Nix does many things. In this PR, we use nix to pin our tooling and provide a convenient dev environment. Using nix for building the binaries is something I really like to do, but a story for own. If we want to use nix, we need to introduce it incrementally anyway. We can also build container images with nix, and the podvm image itself could run nixos. But we cannot do this in a single PR.

---

> We have to be cautious about overspending the novelty budget in this PR. 

> Would it be possible to execute mkosi via docker (see df sample below) in the scope of this PR and address introduce nix separately?

I presented the plan to integrate mkosi using nix in the meeting two weeks ago, and this is the first feedback I got in this regard. As mentioned in the previous comment, there are many opens to get this to a point where it actually usable. Reworking this to use some other tooling feels like a step back. I don't think I got the cycles to implement that and reinvent the wheel.

We can discuss if we would like to use nix before merging this. I've tried to kick that discussion of months ago in https://github.com/confidential-containers/cloud-api-adaptor/issues/1035, but there hasn't been much interaction on the general point about build system, only on fixes for the pressing issues.

mkulke:
My concern is not about the merits of nix and its applicability for the project. To the contrary I think it's a good fit and we should adopt it. My concern is simple repository and PR hygiene and containing the sprawl of different options and tools. From my observation, in its current state, it's pretty hard for even experienced engineers to contribute to or even just deploy CAA. The cognitive overhead to make a meaningful contribution is rather big.

If a build tool is introduced alongside a functional equivalent, with the best intents of applying the new tool everywhere in the future, technical dept is introduced implicitly. Rewriting Makefile and CI yamls is a thankless task, and there's a risk this won't receive the attention it needs.

My suggestion would be to attempt to run the mkosi build in Docker somehow (afaiu it's a stated design-goal of mkosi to run in a container, maybe that's not docker) and introduce nix in a separate PR that will specifically address the question of reproducible builds. This suggestion only makes sense if mkosi can be decoupled with minimum effort from the proposed nix build-environment and run in a docker container, using podman instead wouldn't make sense.

I'll verify again whether images can be built successfully w/ docker (I've seen the pid0 error, which makes sense. but the mkosi run was still successful). Debian unstable might not be a good target indeed, I didn't spend much thought into what base image would make more sense.

That said, even if that turns out to work and there is a good base image, I'm still open to merge the PR as-is, however there should be a few more eyes on this

cc @surajssd @stevenhorsman @wainersm

katexochen:
> My concern is simple repository and PR hygiene

I can cherry-pick the first commit into its own PR to discuss nix and merge that one first.

> From my observation, in its current state, it's pretty hard for even experienced engineers to contribute to or even just deploy CAA. The cognitive overhead to make a meaningful contribution is rather big.

Exactly, and that's why I'd rather not want to introduce another make-docker-something layer of confusion.

katexochen:
@mkulke @bpradipt @surajssd 

Following yesterdays protocol it seems like you require to separate this from Nix, so I've updated the PR to use docker to unblock.

As the CI action to build was relying on Nix, I removed it.

I've also opened https://github.com/confidential-containers/cloud-api-adaptor/issues/1516 to formally propose that we introduce Nix as build system.

mkulke (approved):
LGTM, thanks for accommodating and adding an alternative docker based build step.

bpradipt:
> @mkulke @bpradipt @surajssd
> 
> Following yesterdays protocol it seems like you require to separate this from Nix, so I've updated the PR to use docker to unblock.
> 
> As the CI action to build was relying on Nix, I removed it.
> 
> I've also opened #1516 to formally propose that we introduce Nix as build system.

Thanks @katexochen .. 

I have also started reviewing/testing this PR again as I didn't look at the latest changes. Hope to complete it soon.

bpradipt:
I successfully tested an image build on AWS. For AWS, the AMI created from the raw image needs to have `BootMode` set to `uefi` .

Snippet from a successful run 

```
023/10/13 06:47:35 [adaptor/proxy]         org.systemd.property.DefaultDependencies: true
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.SandboxName: k8s_bp-test_default_993edf75-61ff-499a-b319-36dde107ae31_0
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.Image: docker.io/library/ubuntu@sha256:a9038002b9b29c77c93b1c562ca2bee51313c6f5208c8d90b91929db62e96930
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.ImageName: docker.io/library/ubuntu:latest
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.container.hash: c8db32e2
2023/10/13 06:47:35 [adaptor/proxy]         io.katacontainers.pkg.oci.container_type: pod_container
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.container.ports: [{"containerPort":8888,"protocol":"TCP"}]
2023/10/13 06:47:35 [adaptor/proxy]         kubernetes.io/config.seen: 2023-10-13T06:46:22.983946387Z
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.ContainerID: 5c827feddf29438de41fa0cd4e2204363fbfcfc8cc9845eaefd45794c480595d
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.pod.terminationGracePeriod: 30
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.pod.namespace: default
2023/10/13 06:47:35 [adaptor/proxy]         io.katacontainers.pkg.oci.bundle_path: /run/containers/storage/overlay-containers/5c827feddf29438de41fa0cd4e2204363fbfcfc8cc9845eaefd45794c480595d/userdata
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.TTY: false
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.pod.uid: 993edf75-61ff-499a-b319-36dde107ae31
2023/10/13 06:47:35 [adaptor/proxy]         kata.peerpods.io/instance_type: t2.small
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.SeccompProfilePath:
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.container.terminationMessagePath: /dev/termination-log
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.container.restartCount: 0
2023/10/13 06:47:35 [adaptor/proxy]         io.kubernetes.cri-o.Name: k8s_ubuntu_bp-test_default_993edf75-61ff-499a-b319-36dde107ae31_0
2023/10/13 06:47:35 [adaptor/proxy]         org.systemd.property.After: ['crio.service']
2023/10/13 06:47:35 [adaptor/proxy]         kubectl.kubernetes.io/last-applied-configuration: {"apiVersion":"v1","kind":"Pod","metadata":{"annotations":{},"labels":{"app":"test"},"name":"bp-test","namespace":"default"},"spec":{"containers":[{"args":["infinity"],"command":["sleep"],"env":[{"name":"NVIDIA_VISIBLE_DEVICES","value":"all"}],"image":"ubuntu","name":"ubuntu","ports":[{"containerPort":8888}]}],"runtimeClassName":"kata-remote-cc"}}
2023/10/13 06:47:35 [adaptor/proxy] getImageName: got image from annotations: docker.io/library/ubuntu:latest
2023/10/13 06:47:35 [adaptor/proxy] CreateContainer: calling PullImage for "docker.io/library/ubuntu:latest" before CreateContainer (cid: "5c827feddf29438de41fa0cd4e2204363fbfcfc8cc9845eaefd45794c480595d")
2023/10/13 06:47:37 [adaptor/proxy] CreateContainer: successfully pulled image "docker.io/library/ubuntu:latest"
2023/10/13 06:47:37 [adaptor/proxy] StartContainer: containerID:5c827feddf29438de41fa0cd4e2204363fbfcfc8cc9845eaefd45794c480595d
```

bpradipt (approved):
/lgtm
Nice work @katexochen

mkulke:
Should we extend the secureboot param to AWS and other providers and maybe cherry-pick this into its own PR?

katexochen:
@mkulke for AWS, secure boot configuration is part of the image upload, and therefore not part of our code (but part of uplosi). Regarding cherry-picking: I can do that, but I don't think the commit makes much sense out of the context of this PR.

mkulke:
> @mkulke for AWS, secure boot configuration is part of the image upload, and therefore not part of our code (but part of uplosi). Regarding cherry-picking: I can do that, but I don't think the commit makes much sense out of the context of this PR.

I see, thx for clarifying. We might be interested in the secureboot toggle for other cases, but if we merge this soon, that should be ok.

PR #1500, opened 2023-10-05
podvm: bump guest-components version

mkulke:
Since confidential-containers/guest-components/pull/371 has been merged, we need to update guest components in the podvm. Otherwise remote attestation via cc_kbc will not work for peer pods.


katexochen (approved):

bpradipt (approved):
/lgtm

Issue #1450, opened 2023-09-20
Failed to run nginx container in peer pod


genjuro214:
It's not always reproduced. 

When creating a peer pod with nginx container, sometimes the nginx may exit with error.

The nginx logs looks like below:
```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
dpkg-query: no packages found matching nginx
10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Ignoring /docker-entrypoint.d/.wh..wh..opq
/docker-entrypoint.sh: Configuration complete; ready for start up
2023/09/19 13:56:25 [emerg] 1#1: getpwnam("nginx") failed in /etc/nginx/nginx.conf:2
nginx: [emerg] getpwnam("nginx") failed in /etc/nginx/nginx.conf:2
```

stevenhorsman:
From https://github.com/NginxProxyManager/nginx-proxy-manager/issues/398 is looks like there might be an issue with the nginx user not being added, but I'm not sure why that would only sometimes be hit.

Internally there has been a suggestion to switch out test pods to be busybox, rather than nginx, but I'm concerned that this is hiding the problem rather than solving it.

sudharshanibm3:
Hi @stevenhorsman ,
Based on this [nginx-issue](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/398),  Nginx container starts before the container that creates the "nginx" user finishes. In some cases, the container may not have completed user creation by the time Nginx starts, leading to this `[emerg] getpwnam("nginx") failed` error.

So can we make a init container to enable "nginx" user exists before starting the Nginx service in order to avoid the instablity we are facing now ?

https://gist.github.com/sudharshanibm3/caba60a426b94a2d522e465382703903

[...]

PR #1502, opened 2023-10-06
ibmcloud: Included Readiness Probe along with nginx containers

sudharshanibm3:
Included Readiness Probe along with nginx containers

Fixes:  #1450

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -8,6 +8,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 type podOption func(*corev1.Pod)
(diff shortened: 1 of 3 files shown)


mkulke:
Why would the dockerhub's nginx image - arguably the most testes oci image - require an init container?

test/e2e/ibmcloud_test.go
```diff
@@ -308,6 +308,7 @@ func newPodWithPVCFromIBMVPCBlockDriver(namespace, podName, containerName, image
 					Name:            containerName,
 					Image:           imageName,
 					ImagePullPolicy: corev1.PullAlways,
+					Ports:           []corev1.ContainerPort{{ContainerPort: 80}},
```
mkulke, line 311: Should we add that to all test suites?

sudharshanibm3: No I think, I added ContainerPort to test cases which only executes commands inside container

mkulke: I mean not just for IBM, but make this the default everywhere.

sudharshanibm3: Hi @mkulke ,
Sorry I updated the code now
- Here I added `containerport` and `ReadinessProbe` manually for ibm specific test cases - [Ref](https://github.com/confidential-containers/cloud-api-adaptor/pull/1502/files#diff-96d31bf66f849bfcdfdd93f021bd2de1bf27e2d9cf6ceb7e79f2afceef4fc3a7R312-R322)
- But for other test cases created a common function (withContainerPort) to add `containerport` and `ReadinessProbe` in [common.go](https://github.com/confidential-containers/cloud-api-adaptor/pull/1502/files#diff-2c7fc9c9c12632c5e550abeb31614356abd17710d948a8664f9d2f2163e38b0dR22-R38)

mkulke: Ok, I guess we can add it to other clouds & libvirt in separate PRs. I was just concerned to introduce different test behavior in the individual backends.

cc @kartikjoshi21 @wainersm

mkulke (approved):

test/e2e/common.go
```diff
@@ -40,6 +58,19 @@ func withImagePullSecrets(secretName string) podOption {
...
+// 		p.Spec.InitContainers = []corev1.Container{
+// 			{
+// 				Name:    "useradd-init-container",
+// 				Image:   "busybox",
+// 				Command: []string{"adduser", "-D", "nginx"},
+// 			},
+// 		}
+// 	}
+// }
+
```
stevenhorsman, line 73: Can we remove the commented out code from this PR please as we aren't using it.

stevenhorsman (approved):
LGTM. I agree with Magnus that adding the probe to the other cloud providers makes sense, but we can start with it here and monitor the results first. Thanks!

PR #1517, opened 2023-10-12
peerpod-ctrl: check and clean old PeerPod objects 

### LINKED ISSUE #1510: Need clean old Peer Pod VMs after updating CAA DeamonSet
When testing the case to update CAA DeamonSet, we found the old Peer Pod VMs are not deleted after updating.

To update CAA DeamonSet, the pod with CAA container will be recreated. The new CAA container will create new Peer Pod VMs for existing Peer Pods on the node. However, the old Peer Pod VMs for those Peer Pods still exist.

For example, if there is a nginx peer pod on a worker, after updating CAA pod on the worker, there would be 2 `PeerPod` objects (we have deployed **peerpod-ctrl**) owned by the nginx peer pod, and 2 Pod VMs, one is created before updating while another is created after updating.

If we delete the nginx peer pod, **peerpod-ctrl** will help to delete the old Pod VM. It works as expected.

However, for CAA DeamonSet updating case, we expect the old Pod VMs could be deleted after the update completed, rather than waiting until the peer pod is deleted.

genjuro214:
For Create and Update events without DeletionTimestamp,
check if there is old PeerPod object owned by the same Pod,
and delete it to clean the dangling Pod VM

Fixes #1510

### DIFF
--- a/peerpod-ctrl/controllers/peerpod_controller.go
+++ b/peerpod-ctrl/controllers/peerpod_controller.go
@@ -82,6 +82,25 @@ func (r *PeerPodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	if pp.ObjectMeta.GetDeletionTimestamp() == nil { // TODO: consider filter events
+		// Create or Update events without DeletionTimestamp
+		// Check if existing old PeerPod owned by current Pod,
+		// and delete it to clean the dangling VM.
+		// It may caused by CAA update, unplanned down or crashes
+		ppList := confidentialcontainersorgv1alpha1.PeerPodList{}
+		if err := r.List(ctx, &ppList); err != nil {
+			logger.Info("Failed to get PeerPod list", "error", err)
+			return ctrl.Result{}, err
+		}
+
+		for _, item := range ppList.Items {
+			if isOldPeerPod(item, pp) {
+				logger.Info("Found old PeerPod object owned by the same Pod", "old PeerPod", item)
+				if err := r.Delete(ctx, &item); err != nil {
+					logger.Info("Failed to delete old PeerPod", "error", err)
+				}
+			}
+		}
+
 		return ctrl.Result{}, nil
 	}
 
@@ -153,4 +172,10 @@ func SetProvider() (cloud.Provider, error) {
 	}
 
 	return nil, fmt.Errorf("cloudmgr: %s cloud provider not supported", cloudName)
+}
+
+func isOldPeerPod(pp, cur confidentialcontainersorgv1alpha1.PeerPod) bool {
+	return pp.OwnerReferences[0].UID == cur.OwnerReferences[0].UID && // Same owner
+		pp.UID != cur.UID && // Not cur itself
+		pp.CreationTimestamp.Before(&cur.CreationTimestamp) // Created before cur
 }
--- a/test/e2e/rolling_update_test.go
+++ b/test/e2e/rolling_update_test.go
@@ -27,7 +27,7 @@ import (
 )
 
 const WAIT_DEPLOYMENT_AVAILABLE_TIMEOUT = time.Second * 180
-const OLD_VM_DELETION_TIMEOUT = time.Second * 15
+const OLD_VM_DELETION_TIMEOUT = time.Second * 60
 
 func doTestCaaDaemonsetRollingUpdate(t *testing.T, assert RollingUpdateAssert) {
 	runtimeClassName := "kata-remote"
@@ -229,6 +229,10 @@ func doTestCaaDaemonsetRollingUpdate(t *testing.T, assert RollingUpdateAssert) {
 				t.Fatal(fmt.Errorf("verify pod is not running"))
 			}
 
+			time.Sleep(OLD_VM_DELETION_TIMEOUT)
+			log.Info("Verify old VM instances have been deleted:")
+			assert.VerifyOldVmDeleted(t)
+
 			return ctx
 		}).
 		Teardown(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
@@ -251,10 +255,6 @@ func doTestCaaDaemonsetRollingUpdate(t *testing.T, assert RollingUpdateAssert) {
 			if err = client.Resources().Delete(ctx, deployment); err != nil {
 				t.Fatal(err)
 			}
-
-			time.Sleep(OLD_VM_DELETION_TIMEOUT)
-			log.Info("Verify old VM instances have been deleted:")
-			assert.VerifyOldVmDeleted(t)
 
 			return ctx
 		}).Feature()


liudalibj (approved):
LGTM

snir911:
Thanks @genjuro214 , few questions
1. at the time you have 2 PeerPod objects you also have two Pod (nginx) objects or only one?
2. if there's only one Pod, when you delete it manually, PeerPod objects remains? gets deleted? one/both?
3. Can you explain how to reproduce? I'm not familar with the updating procedure

Notice that in such case deletion should happen when PeerPod controller gets a delete event for the owned PeerPod object by the GC so it may take some time

huoqifeng:
@snir911 to reproduce the case, you can try restart the cloud-api-adaptor ds after creating a PeerPod. 
I guess here is the problem of the logic is:
1. PeerPod created.
2. cloud-api-adaptor ds shutdown
3. PeerPod instance disconnected from the VM instance
4. PeerPod reports error to kubelet and api-server
5. api-server/controller issues containers restart in PeerPod -- which will be blocked as caa is shut down
6. cloud-api-adaptor ds stand up
7. api-server/controller issues containers restart in PeerPod and succeed, which cause another CRD creating. I guess which is because the pause restart.

huoqifeng:
@snir911 so above logic (I guess), causes the pod has multiple CRD and multiple VM instances. Like:
           |--  PeerPod CRD1 - VM1
Pod-- |-- PeerPod CRD2 - VM2

huoqifeng:
@yoheiueda may you help have a look at this also?

genjuro214:
@snir911 
For #1, after updating, 1 Peer Pod will own 2 PeerPod objects if we don't clean the old one
For #2, if deleting Peer Pod, all the PeerPod objects owned by this Pod will be deleted for [Background cascading deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#background-deletion)
For #3, see @huoqifeng 's comment above

yoheiueda:
@genjuro214 @huoqifeng 

As @snir911 pointed out, I think a PeerPod object will be automatically deleted after the corresponding Pod object is deleted, since owner reference to the Pod object is set in the PeerPod object, as far as I understand correctly.

So, I am wondering the status of the Pod object for nginx after CAA is restarted. A new Pod object is created by the API server? If so, what is the status of the old Pod object?

huoqifeng:
@yoheiueda after caa restart, the pod kept as same, but a new PeerPod-CRD created and new VM instance created. Which means we have same Pod but 2 PeerPod-CRD and 2 VM. I guess which is because of the `step 7` I listed in https://github.com/confidential-containers/cloud-api-adaptor/pull/1517#issuecomment-1759649010

yoheiueda:
Thank you, @huoqifeng. I understand the situation now.

So, this proposed solution can only handle graceful restating of CAA. Right?

Is it possible to handle the cases of unplanned down or crashes of CAA?

One possible way is to check whether a Pod object already has a corresponding PeerPod object when a CAA creates a new PeerPod object. If there is an existing PeerPod object with dangling VM, CAA needs to delete the old PeerPod object before creating the old one.

What do you think?

huoqifeng:
@yoheiueda @genjuro214 I agree `CAA needs to delete the old PeerPod object before creating the old one.`, which can guarantee only ONE PeerPod-CRD per one Pod in any case.

huoqifeng:
But I guess there is some technical problem to solve for this approach? @genjuro214

genjuro214:
I think @yoheiueda 's proposal may be doable. We can check if there is old PeerPod object owned by the Pod before creating a new PeerPod object in PeerPodService

snir911:
@genjuro214 I think it may be less intrusive to do it from the peerpod-ctrl, IIRC it's easier to get owner (Pod) from the owned (PeerPod) than the opposite, also we should be able to recognize easily new PeerPods creation in peerpod-ctrl then, you can get its owner Pod and go over the other existing PeerPods and execute Delete for the older instances.
I'm not sure it's better approach and it requires more thinking but I'd consider it as i think it would be easier to implement this way, WDYT?

genjuro214:
Update code as @yoheiueda suggested, try to delete old PeerPod objects owned by the pod before creating new PeerPod object.

@snir911 , I think it will be better if we can check and delete old PeerPod objects in **peerpod-ctrl**, but I'm not sure how to implement it. In current `Reconcile` func, I don't know how to confirm it's called for "create" events or not?

snir911:
@genjuro214 actually it seems **Reconciliation shouldn't be driven off changes in individual Events** however, we are using [GetDeletionTimestamp](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/peerpod-ctrl/controllers/peerpod_controller.go#L84C19-L84C39) to identify deletion events, additionally IIUC it should be possible to filter events in [SetupWithManager](https://github.com/confidential-containers/cloud-api-adaptor/blob/a95750555d8907ced96b4c29c6095f106ad6a020/peerpod-ctrl/controllers/peerpod_controller.go#L108) so that it will be registered only to create and delete and then this can be implemented

genjuro214:
@snir911 @yoheiueda , I tried to use [WithEventFilter](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.2/pkg/builder#Builder.WithEventFilter) to filter events in `SetupWithManager`. However, I just found there seems no "Delete" events. If I filter other events and just keep "Delete", `Reconcile` will not be invoked during CAA update.

And we need to handle "Update" event which will set "DeletionTimestamp", I think.

So I changed the login in `Reconcile` as below:

If "DeletionTimestamp" is not set, go through all PeerPod objects, and check:
* if there is other PeerPod object with the same owner
* if this PeerPod object is created before current PeerPod object

If both YES, try to delete the old PeerPod object to clean the dangling VM.

I also revert the changes in adaptor. The changes in **peerpod-ctrl** can fix the CAA update issue in https://github.com/confidential-containers/cloud-api-adaptor/issues/1510

yoheiueda:
@genjuro214 Thank you very much for the update!

@snir911 I have a basic question about PeerPod objects.

https://github.com/confidential-containers/cloud-api-adaptor/blob/631de585e8c7fa0040e933538a99f4510bea2af4/pkg/adaptor/k8sops/peerpod.go#L68

A PeerPod object name is set to the same name as its corresponding Pod object with a random suffix. This allows us to create multiple PeerPod objects for a single Pod object. If we use the exactly same name without randomness for the PeerPod object name,  we can guarantee 1-to-1 mapping between PeerPod and Pod objects.

Is there any specific reason to allows multiples PeerPod objects for a single Pod object?

yoheiueda:
I also come up with a comment from a broader perspective not specific to this PR. 

I think there is an inconsistency among management of VM creation/deletion and PeerPod object creation/deletion.

|              |  VM instance    | PeerPod object|
|--------| -------- | ------- |
| Create | CAA        | CAA   |
| Delete | peerpod-ctrl     |  (Deleted by owner reference)  |


If we can make peerpod-ctrl  responsible for both creation and deletion of VM instances like below, I think the architecture becomes simpler.

|              |  VM instance    | PeerPod object|
|--------| -------- | ------- |
| Create | peerpod-ctrl         | CAA   |
| Update\* | peerpod-ctrl        | CAA   | 
| Delete | peerpod-ctrl     |  (Deleted by owner reference)  | 

\* When CAA updates a PeerPod object, peerpod-ctrl recreates a VM instance

Actually I adopt this approach in our internal prototype as commented below, and it worked fine.
https://github.com/confidential-containers/cloud-api-adaptor/issues/328#issuecomment-1306865473
https://github.com/confidential-containers/cloud-api-adaptor/issues/375#issuecomment-1334894017

I implemented CAA that creates VM instances using cloud API, but it was just because we needed a quick implementation without controller code at that time. I don't think we have a strong reason to create VM instances from CAA now, since we already have a controller.

I don't think tghis kind of refactoring needs to be covered in this PR, but I think it is a good idea to rethink our design sometime in the future.

huoqifeng:
> I also come up with a comment from a broader perspective not specific to this PR.
> 
> I think there is an inconsistency among management of VM creation/deletion and PeerPod object creation/deletion.
> 	VM instance 	PeerPod object
> Create 	CAA 	CAA
> Delete 	peerpod-ctrl 	(Deleted by owner reference)
> 
> If we can make peerpod-ctrl responsible for both creation and deletion of VM instances like below, I think the architecture becomes simpler.
> 	VM instance 	PeerPod object
> Create 	peerpod-ctrl 	CAA
> Update* 	peerpod-ctrl 	CAA
> Delete 	peerpod-ctrl 	(Deleted by owner reference)
> 
> * When CAA updates a PeerPod object, peerpod-ctrl recreates a VM instance
> 
> Actually I adopt this approach in our internal prototype as commented below, and it worked fine. [#328 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/issues/328#issuecomment-1306865473) [#375 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/issues/375#issuecomment-1334894017)
> 
> I implemented CAA that creates VM instances using cloud API, but it was just because we needed a quick implementation without controller code at that time. I don't think we have a strong reason to create VM instances from CAA now, since we already have a controller.
> 
> I don't think tghis kind of refactoring needs to be covered in this PR, but I think it is a good idea to rethink our design sometime in the future.

@yoheiueda Thanks for the comments, would you mind open an issue to track the refine and describe the idea?

snir911:
@genjuro214 indeed i'm afraid update may be an issue if filtering is not possible, have you tried to generate update event?

>A PeerPod object name is set to the same name as its corresponding Pod object with a random suffix. This allows us to create multiple PeerPod objects for a single Pod object. If we use the exactly same name without randomness for the PeerPod object name, we can guarantee 1-to-1 mapping between PeerPod and Pod objects.
>
>Is there any specific reason to allows multiples PeerPod objects for a single Pod object?

I don't think so, actually i think it's good & simple idea, the only thing is the PeerPod naming must contain also the namespace i suppose (or using the pod ID instead of name?) @genjuro214 WDYT?

about the redesign, this indeed may be a blessed change. BTW another option i thought of before is just to delete the instances from CAA based on the value of PeerPod (no controller) but i'm not sure it's good practice in k8s.

genjuro214:
@snir911 , I do tested to update the old PeerPod object. Since we also checked the CreationgTimestamp, so the updating of old PeerPod will not cause the new PeerPod to be deleted.

I think it's good enough for current PeerPod name

snir911 (approved):
I see now @genjuro214 , Thanks!
In regards to the naming i was referring to @yoheiueda suggestion to simply use Owner (Pod) based name for PeerPod and then IIUC updating the already existing object's resources will generate update event which would allow to Delete the previous PeerPod resources .
Having said that, your current suggestion LGTM as well, i'd be fine with both approaches

huoqifeng (approved):
LGTM

huoqifeng:
As discussed in the thread, a new issue/pr will used to track the CRD object naming, map things.

yoheiueda:
@huoqifeng @liudalibj @snir911 I raise a new issue about the new design. Any comments are appreciated!

https://github.com/confidential-containers/cloud-api-adaptor/issues/1534

PR #1520, opened 2023-10-12
WIP: copy-files: coco-default.rego as default policy

danmihai1:
The Kata Containers default agent policy allows access to all the agent endpoints - including SetPolicy and ExecProcess. Therefore, change the default policy to coco-default.rego, that denies access to the same endpoints that agent-config.toml denies - in preparation for the future removal of the endpoint restriction support from agent-config.toml.

### DIFF
--- a/aws/image/copy-files.sh
+++ b/aws/image/copy-files.sh
@@ -26,3 +26,9 @@ sudo mkdir -p /usr/local/bin
 sudo cp -a /tmp/files/usr/* /usr/
 
 sudo cp -a /tmp/files/pause_bundle /
+
+# Use the policy file providing the same agent enpoint restrictions as agent-config.toml
+default_policy="/etc/kata-opa/default-policy.rego"
+if [ -f "${default_policy}" ]; then
+	ln -sf "coco-default.rego" "${default_policy}"
+fi
--- a/azure/image/copy-files.sh
+++ b/azure/image/copy-files.sh
@@ -26,3 +26,9 @@ sudo mkdir -p /usr/local/bin
 sudo cp -a /tmp/files/usr/* /usr/
 
 sudo cp -a /tmp/files/pause_bundle /
+
+# Use the policy file providing the same agent enpoint restrictions as agent-config.toml
+default_policy="/etc/kata-opa/default-policy.rego"
+if [ -f "${default_policy}" ]; then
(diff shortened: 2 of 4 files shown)


stevenhorsman (commented):
Hey Dan, are you planning to add the default policy in a subsequent PR? I know that we aren't based on main at the moment, so are not processing it with rego, but we could add it here now?

We will want to start using policies in future for peer pods, so I don't have a problem with introducing a default one now, but It's also worth noting that the peer pod agent-config.toml (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/files/etc/agent-config.toml) isn't actually blocking any endpoints (well it would block setPolicy as it doesn't exist in `CCv0` so far, so to keep the current behaviour we don't necessarily need a rego policy, we could just wait until agent-config isn't process and then delete it.

danmihai1:
> Hey Dan, are you planning to add the default policy in a subsequent PR? I know that we aren't based on main at the moment, so are not processing it with rego, but we could add it here now?
> 
> We will want to start using policies in future for peer pods, so I don't have a problem with introducing a default one now, but It's also worth noting that the peer pod agent-config.toml (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/podvm/files/etc/agent-config.toml) isn't actually blocking any endpoints (well it would block setPolicy as it doesn't exist in `CCv0` so far, so to keep the current behaviour we don't necessarily need a rego policy, we could just wait until agent-config isn't process and then delete it.

Thanks for pointing out that this copy of agent-config.toml is a little different than agent-config.toml from rootfs.sh!

I improved https://github.com/kata-containers/kata-containers/pull/8220 based on your idea. If that other PR looks good to reviewers, I think we should just abandon this PR, then wait for the merge with main. Please let me know in case you have better ideas here.

stevenhorsman:
> I improved [kata-containers/kata-containers#8220](https://github.com/kata-containers/kata-containers/pull/8220) based on your idea. If that other PR looks good to reviewers, I think we should just abandon this PR, then wait for the merge with main. Please let me know in case you have better ideas here.

Yeah, I think this is fine - once we are consuming the kata-agent from main and we can drop the endpoint config entirely, we can turn POLICY="YES" on in the agent build in the peer-pod and build and pick up the default policy rego files :)

PR #1522, opened 2023-10-16
Misc controller fixes

bpradipt:
This series includes misc fixes for webhook and peerpodconfig-ctrl
1. Updates the Makefile and kustomization.yaml to use the default images. This makes it easy to deploy these controllers via simple `make -C <dir> <target>` invocation
2. Create a gh-action job to build and push peerpodconfig-ctrl image to quay.io
3. Adds missing daemonset related rbac rules for peerpodconfig-ctrl
4. Fix the webhook gh-action jobs to point to the correct location for versions.yaml

### DIFF
--- a/peerpodconfig-ctrl/config/rbac/role.yaml
+++ b/peerpodconfig-ctrl/config/rbac/role.yaml
@@ -56,6 +56,26 @@ rules:
   - update
   - watch
 - apiGroups:
+  - apps
+  resourceNames:
+  - peerpodconfig-ctrl-caa-daemon
+  resources:
+  - daemonsets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets/finalizers
+  verbs:
+  - update
+- apiGroups:
   - confidentialcontainers.org
   resources:
   - peerpodconfigs
--- a/.github/workflows/peerpodconfig-ctrl_image.yaml
+++ b/.github/workflows/peerpodconfig-ctrl_image.yaml
@@ -1,0 +1,49 @@
+# Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Push peerpodconfig-ctrl image
+---
+name: peerpodconfig-ctrl image push
+
+on:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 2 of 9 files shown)


peerpodconfig-ctrl/config/rbac/role.yaml
```diff
@@ -55,6 +55,27 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - apps
+  resources:
```
huoqifeng, line 60: Shall we use `resourceNames` to limit the resources instances here?

peerpodconfig-ctrl/config/rbac/role.yaml
```diff
@@ -55,6 +55,27 @@ rules:
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
```
huoqifeng, line 75: Same as above, shall we use `resourceNames` to limit the resources instances here?

bpradipt: Like setting it to the following - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/peerpodconfig-ctrl/controllers/peerpodconfig_controller.go#L139 ?

huoqifeng: @bpradipt Like https://github.com/confidential-containers/cloud-api-adaptor/blob/main/peerpod-ctrl/config/rbac/role.yaml#L11
As explained as below in https://kubernetes.io/docs/reference/access-authn-authz/rbac/
```
You can also refer to resources by name for certain requests through the resourceNames list. When specified, requests can be restricted to individual instances of a resource.
```

huoqifeng: I guess peerpodconfig-ctrl should only be grant necessary permission to change specific instances which should be limited by `resourceNames` list?

bpradipt: I updated the PR to explicitly mention the daemonset created

peerpodconfig-ctrl/config/rbac/role.yaml
bpradipt: @huoqifeng I updated to use the resourceName for the daemonset

snir911 (approved):

Issue #1516, opened 2023-10-12
Use Nix as build system


katexochen:
Our current build with make and docker are not only chaotic, but also aren't reproducible.

[Nix](https://nixos.org/) is a build system that enables reproducible builds, which is a hard requirement if we want to do meaningful remote attestation. 

Nix can build 
- Software packages
- Container images 
- OS images

in a reproducible way.

I propose we use Nix as a build system. 
A transition to Nix would need to be done incrementally.

#### Considered alternatives

Major alternative to consider instead of Nix is Bazel. We have been using Bazel for quite some time at Edgeless, but recently added Nix in addition, as Bazel could not fulfill all the requirements. As CAA has many similarities to Constellation regarding builds of binaries, containers and OS images, I think we would hit the same limitations.
There are other systems like [buck2](https://github.com/facebook/buck2), but those are rather niche ecosystems.

katexochen:
In general, Nix builds aren't as fine granular, compared to Bazel builds for example. Nix doesn't share a cache between different cargo builds (but there is upstream work on this). However, it is input-addressed, so you only rebuild packages that changed and their dependents. For our use case, fast/cached builds within a single binary aren't that important. We rather need to ensure we only rebuild a binary when necessary, and that is done by nix through only rebuilding a package when its inputs change. And while Bazel has the fancier caching, it comes at a high cost, as you basically have to declare every single input and output file in Starlark, whereas you can just invoke cargo/make/cmake/whatever in Nix.

So build time would for sure be reduced (by large, I assume), as we update the source of the binaries rarely.

In other scenarios where we might not want to build _within_ nix right now, but with the help of nix, nix could enable us to use a local cache while also using pinned toolchains. For example, if we use `nix develop` with mkosi.

mkulke:
Yup, I understand that we can expect some speedup over building with docker-containers, but in any case don't expect any regression 👍

mkulke:
two questions that came to my mind:

In my limited experience with nix, when building binaries that are dynamically linked it's not possible to deploy them to non-nix system, as nix uses its own dynamic loader. I used patchelf to replace it with the system's linux-ld to work around it, but that's probably not the proper way. How would we handle this in our scenario?

In some cases we might have exotic build dependencies e.g. for tdx attestation. How would be handle that, if the CPU vendor only provides ubuntu packages, would it be on us to maintain a nix package for those?

katexochen:
[...]

PR #1523, opened 2023-10-16
podvm-mkosi: use toolchain from nixpkgs

katexochen:
If we want to have reproducible OS image build, we need to use the toolchain from nixpkgs. The mkosi version in nixpkgs uses a patched version of systemd (with commits that are upstream, but will only be part of the next release). We also fixed many of the tools mkosi uses to get things reproducible.
I maintain the mkosi package in nixpkgs with a colleague. We have nightly tests running on nixpkgs master + mkosi main to ensure things stay reproducible and working.

The use of Nix in general is discussed in #1516. Nevertheless, as I already have this ready, I wanted to open this at least as a draft. I don't think using Nix must be an all-or-nothing decision. I don't think there is currently a way to get around using it for what it is used in this PR if we want reproducible images.


mkulke (approved):
LGTM. built an image and tested it in a peerpod deployment.

I didn't test the GH action code. I'm not entirely sure about the build-space issue on a worker, locally my `podvm-mkosi` folder consumes 2.5G, and this includes the vhd copy. I'd suspected docker build leftovers, but we should be on a fresh node 🤔

bpradipt:
I tried the gh workflow using nektos/act  and it failed during the build image step.
Binaries got built successfully.

```
warning: Git tree '/home/ubuntu/cloud-api-adaptor' is dirty
error: getting status of '/nix/store/7pqci4rwn43ng88610r6a79xxx1j2irm-source/flake.nix': No such file or directory
[Create a Pod VM image with mkosi/Build image   ]   ❌  Failure - Main Build nix shell to cache dependencies
[Create a Pod VM image with mkosi/Build image   ] exitcode '1': failure
[Create a Pod VM image with mkosi/Build image   ] ⭐ Run Post Install Nix
[Create a Pod VM image with mkosi/Build image   ]   🐳  docker cp src=/home/ubuntu/.cache/act/cachix-install-nix-action@v22/ dst=/var/run/act/actions/cachix-install-nix-action@v22/
[Create a Pod VM image with mkosi/Build image   ]   ✅  Success - Post Install Nix
[Create a Pod VM image with mkosi/Build image   ] 🏁  Job failed
Error: Job 'Build image' failed
```

bpradipt:
Another question on the gh-action, is it possible to reuse a pre-built binaries for the podvm "build image" job ? In my local env, re-running the "build image" job results in rebuilding the binaries. And it took roughly 30 min to build the binaries on a VM with 2vcpus and 16GB mem.

katexochen:
> I tried the gh workflow using nektos/act and it failed during the build image step. Binaries got built successfully.

I haven't tried this with act, not sure what caused the error you observed. If you want to test this workflow, you can just trigger it on your fork. Container images are automatically pushed to the registry of the repo the workflow is executed. Here is a successful run on my fork: https://github.com/katexochen/cloud-api-adaptor/actions/runs/6928972339

katexochen:
> Another question on the gh-action, is it possible to reuse a pre-built binaries for the podvm "build image" job ? In my local env, re-running the "build image" job results in rebuilding the binaries. And it took roughly 30 min to build the binaries on a VM with 2vcpus and 16GB mem.

Yes, this definitely can be optimized. Should we just add a input for the image name? Then we can skip the first job if the image is passed as input.

As we've already discussed a few times, we should get rid of the binaries container image build step, build binaries in separate and cache them in the mid-term.

bpradipt:
> > Another question on the gh-action, is it possible to reuse a pre-built binaries for the podvm "build image" job ? In my local env, re-running the "build image" job results in rebuilding the binaries. And it took roughly 30 min to build the binaries on a VM with 2vcpus and 16GB mem.
> 
> Yes, this definitely can be optimized. Should we just add a input for the image name? Then we can skip the first job if the image is passed as input.

Yeah, this might be helpful.
> 
> As we've already discussed a few times, we should get rid of the binaries container image build step, build binaries in separate and cache them in the mid-term.

Would it make sense to split the workflow into two - one for building the binaries and another for the podvm image ?

katexochen:
> Would it make sense to split the workflow into two - one for building the binaries and another for the podvm image ?

Not sure about this, that would place the burden on managing cache invalidation on the developer - which hasn't been working out well with the old image pipeline, where we would only update these images once or twice a release cycle.

bpradipt:
> > Would it make sense to split the workflow into two - one for building the binaries and another for the podvm image ?
> 
> Not sure about this, that would place the burden on managing cache invalidation on the developer - which hasn't been working out well with the old image pipeline, where we would only update these images once or twice a release cycle.

I see..

.github/workflows/podvm_mkosi.yaml
```diff
@@ -0,0 +1,116 @@
...
+  workflow_dispatch:
+    inputs:
+      binaries-image:
+        description: "Prebuild fedora binaries image, as produced by this workflow under ghcr.io/confidential-containers/cloud-api-adaptor/podvm/binaries-fedora"
+        required: false
+
+jobs:
+  build-binaries:
+    name: Build binaries
+    if : ${{ github.event.inputs.binaries-image == '' }}
```
bpradipt, line 13: nice.. thanks for adding this.

bpradipt (approved):
/lgtm
Thanks @katexochen

PR #1524, opened 2023-10-16
podvm-mkosi: production and debug variant

katexochen:
This separates the podvm images build with mkosi into two variants: one for production use with a minimal set an required packages, and one for debugging, adding some useful tools and allowing access to the serial console. The debug variant ins fully additive to the production one.

`make build` builds the production image.
`make build-debug` builds the debug image.


podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
```diff
@@ -0,0 +1,16 @@
...
+
+[Content]
+Autologin=true
+KernelCommandLine=rd.shell
+KernelCommandLine=systemd.setenv=SYSTEMD_SULOGIN_FORCE=1
+Packages=
+    nano
+    vim
+    strace
+    openssh-server
```
mkulke, line 15: if we have openssh we probably want sudo, or can the key be provisioned for root?

mkulke: I think on the current image we'd also need a passwordless sudoers directive:

```
cat etc/sudoers.d/wheel-no-passwd
%wheel  ALL=(ALL)       NOPASSWD: ALL
```

katexochen: SSH is now working as expected for root and the key embedded in the debug image. I've decided to ignore the cloud init key, will work on throwing that out next.

podvm-mkosi/README.md
```diff
@@ -23,6 +23,16 @@ If you want to use the image with libvirt, run the following to convert to qcow2
 qemu-img convert -f raw -O qcow2 build/system.raw build/system.qcow2
 ```
 
+## Debug image
```
bpradipt, line 26: Adding the makefile target for the debug image will be helpful here

bpradipt:
@katexochen is it possible to name the raw output image differently for regular and debug image?
I see the name defaults to `system.raw`

katexochen:
> @katexochen is it possible to name the raw output image differently for regular and debug image?
> I see the name defaults to `system.raw`

What exactly would be the benefit of naming the image?

The thing is, we are building two images here, the initrd and the system image. The initrd is a dependency of the system image. If we use ImageId to name the output, both outputs will be renamed and the dependency would need to be declared separate for debug and production image. That's doable, but I wanted to keep the complexity of the config as low as possible.

Also, having an output with a constant name makes ist very handy in CI and scripts, e.g. in the uplosi upload.

bpradipt:
> > @katexochen is it possible to name the raw output image differently for regular and debug image?
> > I see the name defaults to `system.raw`
> 
> What exactly would be the benefit of naming the image?

It'll avoid rewriting the existing image. This is exactly what happened with me ;) 

> 
> The thing is, we are building two images here, the initrd and the system image. The initrd is a dependency of the system image. If we use ImageId to name the output, both outputs will be renamed and the dependency would need to be declared separate for debug and production image. That's doable, but I wanted to keep the complexity of the config as low as possible.
> 

I see. Then if you can just add a comment in the README for people like me, it'll avoid the surprise post building the image  :-) 

> Also, having an output with a constant name makes ist very handy in CI and scripts, e.g. in the uplosi upload.

Fair point.

katexochen:
> It'll avoid rewriting the existing image. This is exactly what happened with me ;) 

Would it be helpful to build a production and debug image side by side? Currently,  we just rm -rf the build directory before triggering another build.

bpradipt:
> > It'll avoid rewriting the existing image. This is exactly what happened with me ;)
> 
> Would it be helpful to build a production and debug image side by side? Currently, we just rm -rf the build directory before triggering another build.

I think, for now, it's not needed. And as you said, let's keep it simple for now. We can revisit this later if there is a need

katexochen:
@bpradipt I've added the make target and a note about the fact that images are overwritten to the readme.

bpradipt (approved):
/lgtm
Thanks @katexochen

bpradipt:
@katexochen I couldn't get SSH login to work with the debug image. I can login via the the serial console (EC2) and see that the public key is there but SSH login fails
```
root@ip-20-0-0-244 .ssh]# cat authorized_keys 
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDXIqsl+VbNg9MZwGYyCXIWhF/9Q6kP6o6ZJ6fPnJ6HwKyMjQv3QRPdXUBzjX0afEDVx12x/WUjUQEO/P4b0cYJBbg5nE0DyvfmdNFBpO1I6f1P0P4QKfhEH4XV9COvtUUJnoxctxp8eG2CuATI0+yJ7A0VVdrffaoYQfrXFFrZbRYUfH6RDc888UBI4HSjCI9Sfv0spN9Ej/j8JRCgHYhGM2T63JeoJH6/yVLq3zP5bUqK30N5fDlv448xlLa9GE8FB++bX1+qzPZtiKbNmyRVFsIh7b4SlH0SUACr/TJBsUDVkN7x5sNOfk5mZ1ESWIQ2x0HOevQVNV/AA5Bi4r+7 bpradipt@Pradiptas-MBP-2.Dlink
[root@ip-20-0-0-244 .ssh]# pwd
/root/.ssh
```

I'm yet to debug what else is needed for the ssh config to work.

katexochen:
> @katexochen I couldn't get SSH login to work with the debug image. I can login via the the serial console (EC2) and see that the public key is there but SSH login fails

Did you try to connect via public IP? I'm not sure what your setup looks like on AWS, but it could be related to some security group settings.

I usually use one of the other pods to access the podVM from the pod subnet, as the podVMs on Azure aren't provisioned with public IPs.

bpradipt:
> > @katexochen I couldn't get SSH login to work with the debug image. I can login via the the serial console (EC2) and see that the public key is there but SSH login fails
> 
> Did you try to connect via public IP? I'm not sure what your setup looks like on AWS, but it could be related to some security group settings.
> 
> I usually use one of the other pods to access the podVM from the pod subnet, as the podVMs on Azure aren't provisioned with public IPs.

I used public IP. Basically created a VM using the image and connected to it.  SSH connection gets initialised but then it gets closed stating no supported method found. I'll rebuild the image and try again.

katexochen:
> I used public IP. Basically created a VM using the image and connected to it. SSH connection gets initialised but then it gets closed stating no supported method found. I'll rebuild the image and try again.

Not sure the image is supposed to work stand-alone. This image is still using cloud init, so networking might not be setup correctly when cloud metadata is missing. Try `networkctl` over the serial console to see if network was set up correctly, and list failed systemd units with `systemctl list-units --state=failed`.

bpradipt:
The problem was due to ownership permission of the authorized_keys file, the source of which was my resources/authorized_keys

From journalctl

```
Authentication refused: bad ownership or modes for file /root/.ssh/authorized_keys
```
Setting the mode correctly got SSH working
```
[root@ip-20-0-0-70]# chmod 0400 /root/.ssh/authorized_keys 
```

podvm-mkosi/README.md
```diff
@@ -23,6 +23,26 @@ If you want to use the image with libvirt, run the following to convert to qcow2
...
+# for building the binaries
+make fedora-binaries-builder
+make binaries
+# for building a debug image
+make image-debug
+```
+
+Notice that building a debug image will overwrite any previous existing debug or production image.
+
+For using SSH, create a file `resources/authorized_keys` with your SSH public key. The user within the
```
bpradipt, line 43:
```suggestion
For using SSH, create a file `resources/authorized_keys` with your SSH public key. Ensure the permissions are set to `0400` for the `authorized_keys` file. The user within the
```

mkulke (commented):
The debug image will boot, but there is something curious about ssh, root login will not work, because this statement is prefixed to the authorized_keys file. 

when inspecting the `system.raw` image, that prefix doesn't exist at `/root/.ssh/authorized_keys`, it's just the verbatim file as added to the `./resources` folder. so I suspect something is prefixing the file at first boot. would this make sense?

```bash
cat .ssh/authorized_keys 
no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ...
```

katexochen:
> The debug image will boot, but there is something curious about ssh, root login will not work, because this statement is prefixed to the authorized_keys file.
> 
> when inspecting the `system.raw` image, that prefix doesn't exist at `/root/.ssh/authorized_keys`, it's just the verbatim file as added to the `./resources` folder. so I suspect something is prefixing the file at first boot. would this make sense?
> 
> ```shell
> cat .ssh/authorized_keys 
> no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ...
> ```

It is expected that cloud-init will append that to the  `authorized_keys` file. You key should be in the first line of that file and won't get overridden. That annoyance will stop in https://github.com/confidential-containers/cloud-api-adaptor/pull/1532.

I've re-checked, and SSH is working as expected for me. Did you build a debug image?

mkulke:
> > The debug image will boot, but there is something curious about ssh, root login will not work, because this statement is prefixed to the authorized_keys file.
> > when inspecting the `system.raw` image, that prefix doesn't exist at `/root/.ssh/authorized_keys`, it's just the verbatim file as added to the `./resources` folder. so I suspect something is prefixing the file at first boot. would this make sense?
> > ```shell
> > cat .ssh/authorized_keys 
> > no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ...
> > ```
> 
> It is expected that cloud-init will append that to the `authorized_keys` file. You key should be in the first line of that file and won't get overridden. That annoyance will stop in #1532.
> 
> I've re-checked, and SSH is working as expected for me. Did you build a debug image?

Yes, I did build a debug image,the key is in the image and sshd is running there. but the authorized_keys file seems to be prepended, not appended with that statement. does this make sense? my authorized_keys file has just a single line:

```bash
wc -l ./resources/authorized_keys
1 ./resources/authorized_keys
```

katexochen:
@mkulke could you send me the content of `/root/.ssh/authorized_keys` after boot?

mkulke:
> @mkulke could you send me the content of `/root/.ssh/authorized_keys` after boot?

```bash
cat ~/.ssh/authorized_keys
no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAA.... magnuskulke@MININT-QQ6OGSF
```

podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
```diff
@@ -0,0 +1,17 @@
...
+Autologin=true
+KernelCommandLine=rd.shell
+KernelCommandLine=systemd.setenv=SYSTEMD_SULOGIN_FORCE=1
+SkeletonTrees=../../mkosi.skeleton-debug
+Packages=
+    nano
+    vim
+    strace
+    openssh-server
+    file
```
mkulke, line 17: Do we need to include dnf/yum or something to be able to install additional commands? I found myself often requiring to install additional debugging utilities on podvms.

katexochen: Doesn't make much sense, as the disk will be read-only in a few PRs.

katexochen:
> > @mkulke could you send me the content of `/root/.ssh/authorized_keys` after boot?
> 
> ```shell
> cat ~/.ssh/authorized_keys
> no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAA.... magnuskulke@MININT-QQ6OGSF
> ```

The key from `resources/authorized_keys` wasn't included in the image. The appended key in this line is the key that is set via cloud-init. The cloud-init ssh key is not supposed to work.

podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
mkulke: I see, I feel we need a facility for that, though - maybe an overlayroot backed by tmpfs? workarounds like privileged sidecar wouldn't help if we have agent problems.

mkulke:
> > > @mkulke could you send me the content of `/root/.ssh/authorized_keys` after boot?
> > 
> > 
> > ```shell
> > cat ~/.ssh/authorized_keys

> The key from `resources/authorized_keys` wasn't included in the image. The appended key in this line is the key that is set via cloud-init. The cloud-init ssh key is not supposed to work.

hmm... i mounted the produced image.raw and the file was there:

```bash
sudo losetup -P /dev/loop0 podvm-mkosi/build/system.raw
mkdir /tmp/podvm
sudo mount /dev/loop0p2 /tmp/podvm/
sudo cat /tmp/podvm/root/.ssh/authorized_keys
ssh-rsa AAA.... magnuskulke@MININT-QQ6OGSF
ssh-rsa AAA.... magnuskulke@MININT-QQ6OGSF
```

I suspect it gets overwritten. To make sure the key doesn't come from somewhere else I set the peerpod' ssh-key-secret to empty and put same the key on 2 lines in my authorized_keys file. on a booted instance it looks looks like this:

```bash
cat /root/.ssh/authorized_keys
no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA.... magnuskulke@MININT-QQ6OGSF
no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA... magnuskulke@MININT-QQ6OGSF
```

podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
mkulke: If we fail to start a containers via kata + remote-hypervisor, we'd need something to start debug containers locally, after ssh'ing into the instance. It doesn't need to be super straightforward, or even builtin. but we should document a way to do this. not having the option to access additional tooling would be prohibitive for debugging peerpod issues from my experience.

katexochen:
> I suspect it gets overwritten. To make sure the key doesn't come from somewhere else I set the peerpod' ssh-key-secret to empty and put same the key on 2 lines in my authorized_keys file. on a booted instance it looks looks like this:
> 
> ```shell
> cat /root/.ssh/authorized_keys
> no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA.... magnuskulke@MININT-QQ6OGSF
> no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA... magnuskulke@MININT-QQ6OGSF
> ```

Weird.  Can you `cat podvm-mkosi/build/system.manifest | grep -A 3 cloud-init` to check the cloud-init version? The installed packages aren't pinned yet.

I'd suggest we merge this anyway and revisit if this problem still exists after cloud-init is removed. I would rather not spend more time on fixing cloud-init related stuff in this PR.

mkulke:
> > I suspect it gets overwritten. To make sure the key doesn't come from somewhere else I set the peerpod' ssh-key-secret to empty and put same the key on 2 lines in my authorized_keys file. on a booted instance it looks looks like this:
> > ```shell
> > cat /root/.ssh/authorized_keys
> > no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA.... magnuskulke@MININT-QQ6OGSF
> > no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"peerpod\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa AAAA... magnuskulke@MININT-QQ6OGSF
> > ```
> 
> Weird. Can you `cat podvm-mkosi/build/system.manifest | grep -A 3 cloud-init` to check the cloud-init version? The installed packages aren't pinned yet.
> 
> I'd suggest we merge this anyway and revisit if this problem still exists after cloud-init is removed. I would rather not spend more time on fixing cloud-init related stuff in this PR.

```bash
cat podvm-mkosi/build/system.manifest | grep -A 3 cloud-init
      "name": "cloud-init",
      "version": "23.1.2-1.fc38",
      "architecture": "noarch"
    },
    ```

katexochen:
> ```shell
> cat podvm-mkosi/build/system.manifest | grep -A 3 cloud-init
>       "name": "cloud-init",
>       "version": "23.1.2-1.fc38",
>       "architecture": "noarch"
>     },
> ```

Same here, I can't explain why it does something different on your side. ¯\_(ツ)_/¯ Good decision to remove it.

mkulke (approved):

katexochen:
> albeit we still need a way to get a proper debug shell on those pod-vm's. but this can be addressed in a discrete pr.

Proper in the sense of enough packages installed? Or packages installable at runtime?

mkulke:
> > albeit we still need a way to get a proper debug shell on those pod-vm's. but this can be addressed in a discrete pr.
> 
> Proper in the sense of enough packages installed? Or packages installable at runtime?

I'd say packages "installable" at runtime, I don't think it makes to anticipate what we might need. maybe we can install podman?

podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
bpradipt: fwiw, for one of my debugging scenarios, I did add dnf during the image build to make it easier to install packages during debugging.

PR #1532, opened 2023-10-20
podvm-mkosi: remove cloud-init

### LINKED ISSUE #1467: RFC: Remove cloud-config / cloud-init from PodVM Image?

katexochen:
This is based on #1524, which should be reviewed and merged first.

Fixes #1467 (for x86, we should check with the IBM folks if they want to remove cloud init for s390x as well)

This requires `DISABLE_CLOUD_CONFIG=true` in your `install/overlay/${CSP}/kustomization.yaml`

### DIFF
--- a/podvm-mkosi/README.md
+++ b/podvm-mkosi/README.md
@@ -65,5 +65,6 @@ Packages=
 The following limitations apply to these images. Notice that the limitations are intentional to
 reduce complexity of configuration and CI and shall not be seen as open to-dos.
 
-- `DISABLE_CLOUD_CONFIG=false` is implied. The mkosi images are currently using
-    cloud init/cloud config, and there is no possibility to disable it.
+- `DISABLE_CLOUD_CONFIG=true` is implied. The mkosi images are using
+    our own mechanism to fetch metadata from the IMDS API. Cloud init is not supported.
+    Ensure that `DISABLE_CLOUD_CONFIG=true` in your `install/overlay/${CSP}/kustomization.yaml`.
--- a/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
+++ b/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
@@ -15,3 +15,6 @@ Packages=
     strace
     openssh-server
     file
+    iputils
+    curl
+    wget
--- a/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf
+++ b/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf
@@ -22,4 +22,4 @@ Packages=
     tpm2-tools
     iproute
     iptables
-    cloud-init
+    afterburn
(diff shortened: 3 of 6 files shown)


pkg/adaptor/cloud/azure/provider.go
```diff
@@ -232,7 +232,7 @@ func (p *azureProvider) CreateInstance(ctx context.Context, podName, sandboxID s
 
 	result, err := p.create(ctx, vmParameters)
 	if err != nil {
-		if err := p.deleteDisk(ctx, diskName); err != nil {
+		if err := p.deleteDisk(context.Background(), diskName); err != nil {
```
mkulke, line 235: unrelated?

podvm-mkosi/README.md
```diff
@@ -65,5 +65,5 @@ Packages=
 The following limitations apply to these images. Notice that the limitations are intentional to
 reduce complexity of configuration and CI and shall not be seen as open to-dos.
 
-- `DISABLE_CLOUD_CONFIG=false` is implied. The mkosi images are currently using
-    cloud init/cloud config, and there is no possibility to disable it.
+- `DISABLE_CLOUD_CONFIG=true` is implied. The mkosi images are using
+    our own mechanism to fetch metadata from the IMDS API. Cloud init is not supported.
```
mkulke, line 69: I'd add the note from the PR description on how to disable it.

pkg/adaptor/cloud/azure/provider.go
katexochen: Yes, thought this is to minor for a separate PR, but can cherry-pick if you like.

podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora-debug.conf
```diff
@@ -15,3 +15,6 @@ Packages=
     strace
     openssh-server
     file
+    iputils
+    curl
+    wget
```
mkulke, line 20: unrelated?

mkulke (approved):
thanks, tested successfully

bpradipt (approved):
/lgtm
Thanks @katexochen

PR #1539, opened 2023-10-30
Misc CVM fixes for AWS

bpradipt:
1. Add uefi support when registering AMI created from raw image. This is tested with mkosi fedora image
2. Adjust disable-cvm option to align with other providers

### DIFF
--- a/pkg/adaptor/cloud/aws/manager.go
+++ b/pkg/adaptor/cloud/aws/manager.go
@@ -34,9 +34,7 @@ func (_ *Manager) ParseCmd(flags *flag.FlagSet) {
 	// Add a parameter to indicate the root volume size for the Pod VMs
 	// Default is 30GiBs for free tier. Hence use it as default
 	flags.IntVar(&awscfg.RootVolumeSize, "root-volume-size", 30, "Root volume size (in GiB) for the Pod VMs")
-	// Setting disable-cvm to true as there are still some rough edges with CVMs.
-	// Once the issues are fixed, we can set it to false by default
-	flags.BoolVar(&awscfg.DisableCVM, "disable-cvm", true, "Use non-CVMs for peer pods")
+	flags.BoolVar(&awscfg.DisableCVM, "disable-cvm", false, "Use non-CVMs for peer pods")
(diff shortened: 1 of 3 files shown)


pkg/adaptor/cloud/aws/manager.go
```diff
@@ -36,7 +36,7 @@ func (_ *Manager) ParseCmd(flags *flag.FlagSet) {
 	flags.IntVar(&awscfg.RootVolumeSize, "root-volume-size", 30, "Root volume size (in GiB) for the Pod VMs")
 	// Setting disable-cvm to true as there are still some rough edges with CVMs.
```
katexochen, line 37: I guess we should also remove the comment?

bpradipt: Ah yes. Thanks for spotting it

snir911 (approved):

katexochen (approved):
Code lgtm, but didn't test as I currently don't have a cluster available.

bpradipt:
I tested it with a fedora image generated using mkosi and instance type of m6a.large

The following image shows that AMD SEV-SNP is enabled
image

The following image shows the SNP tags automatically added by AWS
image

What is still pending, is to get vtpm based attestation working. 
@katexochen do you have any idea from your experience with constellation ?

katexochen:
> What is still pending, is to get vtpm based attestation working. @katexochen do you have any idea from your experience with constellation ?

Did you use uplosi for the image upload? The image should be tagged in the right way to enable vTPM: https://github.com/edgelesssys/uplosi/blob/main/aws/uploader.go#L376

Also maybe we need to use boot mode "uefi" instead of "uefi-preferred".

You can check if an AMI has vTPM enabled by [following this AWS doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-nitrotpm-support-on-ami.html).

bpradipt:
> https://github.com/edgelesssys/uplosi/blob/main/aws/uploader.go#L376

I didn't use uplosi for AWS. I'll try and let you know.

bpradipt:
Changed from `uefi-preferred` to `uefi` in the image upload script.

PR #1540, opened 2023-10-30
caa: install yq from source for crossarch builds

jtumber-ibm:
As we are using emulation for the build step (sorry I didn't notice this before), because of requiring CGO, we need to build yq for each arch instead of using upstream releases.


katexochen (approved):

huoqifeng (approved):
LGTM

Issue #1543, opened 2023-10-31
Use terraform instead of own provisioner code


katexochen:
Terraform/OpenTofu is a common and well established provisioning mechanism. The provisioning code in terraform could be used in the e2e tests, but would be test agnostic, so we could also use it in development to setup a cluster in a predefined way, and even expose it to users as entry point to set up the needed infrastructure.

I know some of us already have some terraform code or deployment scripts out of tree, including me. Providing a common and simple solution to provision the infra would also lower the hurdle for new developers.

katexochen:
Terraform should cover all supported cloud providers, including libvirt.

stevenhorsman:
I think the theory behind switching to the e2e framework code for provisioning was to try and keep it all consumed in the same place (and also we only had terraform for IBMCloud). Are you proposing making the e2e tests two step again, or calling the terraform from the e2e framework (and setting up the terraform.tfvars (or whatever the OpenTofu equivalent is) from the e2e properties input if required?

katexochen:
> Are you proposing making the e2e tests two step again

This. Infra provisioning and e2e testing should happen in two separate steps. First do the provider specific provisioning, then execute the generic tests based on the kubeconfig.

katexochen:
The azure e2e seems to already be divided into these steps.

katexochen:
> are you just factoring the cluster creation into the provisioning stage, or does it include the installation of the operator and kustomizisation of the cloud provider?

I'm only talking about infrastructure here. So no kubectl steps are done within terraform.

stevenhorsman:
> > are you just factoring the cluster creation into the provisioning stage, or does it include the installation of the operator and kustomizisation of the cloud provider?
> 
> I'm only talking about infrastructure here. So no kubectl steps are done within terraform.

Ok, the operator install is in the provisioner code at the moment, so I wanted to check.

It would be a bit frustrating to revert the changes that we did at the beginning of the year to switch the ibm cloud provisioning back from code to terraform, but that's not a reason to not do it.

katexochen:
> It would be a bit frustrating to revert the changes that we did at the beginning of the year to switch the ibm cloud provisioning back from code to terraform, but that's not a reason to not do it.

What was the reason for the switch?

stevenhorsman:
> What was the reason for the switch?

[...]

PR #1561, opened 2023-11-07
test: Add createcluster action to provisioner cli

mkulke:
The test itself will perform an installation of the operator/CAA on the cluster as a [setup step](https://github.com/confidential-containers/cloud-api-adaptor/blob/b18675f52f54ad7d8eae72ce7a6668be81329c11/test/e2e/main_test.go#L120), so installing it twice is redundant.

The installation should be idempotent, but having 2 places performing the installation is probably suboptimal and might lead to undesired effects if different provisioner config files are used. If we want to test whether a re-installation works, we should have a dedicated test for that.

I would opt for removing the CAA installation from the `provision` action, but a `createcluster` action might be a good middle ground.

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -153,7 +153,7 @@ jobs:
       working-directory: test/tools
       run: |
         make caa-provisioner-cli
-        ./caa-provisioner-cli -action=provision
+        ./caa-provisioner-cli -action=createcluster
 
   run-e2e-test:
     runs-on: ubuntu-latest
--- a/test/tools/provisioner-cli/main.go
+++ b/test/tools/provisioner-cli/main.go
@@ -55,6 +55,13 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if *action == "createcluster" {
+		log.Info("Creating Cluster...")
+		if err := provisioner.CreateCluster(context.TODO(), cfg); err != nil {
+			log.Fatal(err)
+		}
+	}
+
 	if *action == "provision" {
 		log.Info("Creating VPC...")
 		if err := provisioner.CreateVPC(context.TODO(), cfg); err != nil {


stevenhorsman:
@katexochen might be interested in this in relation to https://github.com/confidential-containers/cloud-api-adaptor/issues/1543

mkulke:
> @katexochen might be interested in this in relation to #1543

If we replace cluster provisioning w/ tf code, this PR wouldn't be needed. Not sure how much of an effort the tf conversion will be, maybe that could be a temporary fix?

stevenhorsman (approved):
LGTM. Thanks!

mkulke:
> * Create cluster (which could be replaced with terraform)
> * Install CoCo operator including peer-pods
> * Run e2e tests
> * Uninstall CoCo operator including peer-pods
> * Delete cluster (again could terraform)

Maybe there's value in keeping steps 2-3 in the test suite, so the tests would always be coupled to a known configuration of a peer-pod setup. step 4 we probably only need to provide in the provisioner cli in tests. removing clusters should be sufficient.

we might also need a step 6 which is garbage collecting resource leftovers, bugs in CAA will leak resources that aren't cleaned up properly. we cannot avoid having those bugs, so we need to account for that in the tests.

stevenhorsman:
> Maybe there's value in keeping steps 2-3 in the test suite, so the tests would always be coupled to a known configuration of a peer-pod setup. step 4 we probably only need to provide in the provisioner cli in tests. removing clusters should be sufficient.
> 
> we might also need a step 6 which is garbage collecting resource leftovers, bugs in CAA will leak resources that aren't cleaned up properly. we cannot avoid having those bugs, so we need to account for that in the tests.

Yeah, those are great points. My thoughts around this were to have the separate actions/units available and independently runnable, but that the test suite would be set to run 2 and 3, but things like the 1 and 5 with are controlled by environment variables would be decoupled. Anyway thoughts for the future!

surajssd (approved):
I have not tested this myself, but looks good at first impression. Thanks @mkulke

PR #1563, opened 2023-11-08
Add separate nydus-snapshotter test

stevenhorsman:
- Given that nydus-snapshotter doesn't seem stable, we want to
    add a separate test for the nydus snapshotter
    so that we can consider skipping it for cloud-providers that aren't
    working
 - Given that on AKS the runtime-handler annotation seems to cause issues
    and the nydus-snapshotter isn't stable, we don't want to add it
    to every pod.job anymore

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -76,6 +76,12 @@ func TestAwsCreateSimplePod(t *testing.T) {
 	doTestCreateSimplePod(t, assert)
 }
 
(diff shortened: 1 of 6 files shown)


wainersm (approved):
yes @stevenhorsman , that's what I had in mind. The SimplePod test should pass on all providers as the "non snapshotter" way is supposed to work on all.

huoqifeng (approved):
lgtm

Issue #1558, opened 2023-11-03
centos ssh timeout on podvm image build


jtumber-ibm:
During release build testing I've noticed that the centos image build is not able to complete.

Extract from the packer logs which might be relevant 

```
rcu: INFO: rcu_preempt detected stalls on CPUs/tasks:
rcu: 	0-...!: (0 ticks this GP) idle=4268/0/0x0 softirq=173/173 fqs=0 (false positive?)
	(detected by 1, t=60005 jiffies, g=-1119, q=6021 ncpus=2)
Sending NMI from CPU 1 to CPUs 0:
NMI backtrace for cpu 0 skipped: idling at default_idle+0xb/0x10
rcu: rcu_preempt kthread timer wakeup didn't happen for 60002 jiffies! g-1119 f0x0 RCU_GP_WAIT_FQS(5) ->state=0x402
rcu: 	Possible timer handling issue on cpu=0 timer-softirq=15
rcu: rcu_preempt kthread starved for 60005 jiffies! g-1119 f0x0 RCU_GP_WAIT_FQS(5) ->state=0x402 ->cpu=0
rcu: 	Unless rcu_preempt kthread gets sufficient CPU time, OOM is now expected behavior.
rcu: RCU grace-period kthread stack dump:
task:rcu_preempt     state:I stack:0     pid:17    ppid:2      flags:0x00004000
Call Trace:
 <TASK>
 __schedule+0x20a/0x550
 ? __pfx_rcu_gp_kthread+0x10/0x10
 schedule+0x2d/0x70
 schedule_timeout+0x88/0x160
 ? __pfx_process_timeout+0x10/0x10
 rcu_gp_fqs_loop+0x114/0x460
 rcu_gp_kthread+0xfb/0x190
 kthread+0xe0/0x100
 ? __pfx_kthread+0x10/0x10
 ret_from_fork+0x2c/0x50
 </TASK>
rcu: Stack dump where RCU GP kthread last ran:
Sending NMI from CPU 1 to CPUs 0:
NMI backtrace for cpu 0 skipped: idling at default_idle+0xb/0x10
```

snir911:
Hi, running `PODVM_DISTRO=centos make podvm-image` on a Fedora machine at main (098d5be692d0e3de0ac5995f61c1d692b3a88be5) worked for me.
Something unusual in your environment?
```
packer 1.9.4
qemu 7.0.0
```

jtumber-ibm:
[...]

PR #1564, opened 2023-11-09
enable root ssh login for centos

lysliu:
CentOS disallow root ssh access by default. See /etc/ssh/sshd_config.

sed -i "s/#PermitRootLogin yes/PermitRootLogin yes/g" /etc/ssh/sshd_config

fix the issue of https://github.com/confidential-containers/cloud-api-adaptor/issues/1558


snir911:
@lysliu, thanks, did this change fix the issue for you? it seems packer's default is not "root":
https://github.com/confidential-containers/cloud-api-adaptor/blob/af7ac2173c1187f11a0f48023eecc4fd9181c0e0/podvm/qcow2/centos/variables.pkr.hcl#L42-L44
Also it worked for me locally without allowing root so i suspect this is different issue..

bpradipt:
@lysliu are you using centos? With support for creating fedora image via mkosi I was hoping we can remove centos and avoid the maintenance overhead. Wdyt ?

bpradipt:
> @lysliu are you using centos? With support for creating fedora image via mkosi I was hoping we can remove centos and avoid the maintenance overhead. Wdyt ?

@lysliu @stevenhorsman @wainersm your thoughts please..
Is it ok if centos is removed? It's already removed from aws and azure packer builds.

lysliu:
Thanks,  I'm ok to remove centos

stevenhorsman:
> @lysliu @stevenhorsman @wainersm your thoughts please..
Is it ok if centos is removed?

Yes - Wainer has an update to comment it out in https://github.com/confidential-containers/cloud-api-adaptor/pull/1595 which I hope will be merged soon. Maybe rather than comment it, we should just remove it forever in this workflow though?

wainersm:
> > @lysliu @stevenhorsman @wainersm your thoughts please..
> > Is it ok if centos is removed?
> 
> Yes - Wainer has an update to comment it out in #1595 which I hope will be merged soon. Maybe rather than comment it, we should just remove it forever in this workflow though?

@stevenhorsman @bpradipt @lysliu I'm preparing a commit to wipe out CentOS from this project completely \o/

PR #1570, opened 2023-11-13
Bump 0.8 release versions

stevenhorsman:


stevenhorsman:
> hi @stevenhorsman !
> 
> Are you going to pin the operator (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/Makefile#L167) later?

I can pin it. I'm not sure if it will have much effect due to https://github.com/confidential-containers/operator/issues/262, but that doesn't mean we should avoid it, so I'll add it to this PR now

wainersm:
> > hi @stevenhorsman !
> > Are you going to pin the operator (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/Makefile#L167) later?
> 
> I can pin it. I'm not sure if it will have much effect due to [confidential-containers/operator#262](https://github.com/confidential-containers/operator/issues/262), but that doesn't mean we should avoid it, so I'll add it to this PR now. Thanks for the reminder!

hmmm  I wasn't aware of  [confidential-containers/operator#262](https://github.com/confidential-containers/operator/issues/262). Even though, we will be pinning the other components like pre-reqs and kata-runtime IIUC.

stevenhorsman:
> hmmm I wasn't aware of [confidential-containers/operator#262](https://github.com/confidential-containers/operator/issues/262). Even though, we will be pinning the other components like pre-reqs and kata-runtime IIUC.

Yes - I was being a bit unfair on the limited effect of it. Definitely worth doing and documented in the release process, so I should have remember to do it this morning!

mkulke (approved):

PR #1572, opened 2023-11-13
adaptor/azure: use context.Background for cleanup

katexochen:
Otherwise the cleanup won't be executed if the context was canceled.


mkulke (approved):

bpradipt (approved):
/lgtm

PR #1581, opened 2023-11-15
set defaults from environment to instanceTypes

snir911:
and setting default CVM type for Azure

setting using env fixes peerpod-ctrl error:
```
The following supplied instance types do not exist: []
```


pkg/adaptor/cloud/aws/provider.go
```diff
@@ -168,6 +164,10 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 		b64EncData = base64.StdEncoding.EncodeToString([]byte(userData))
 	}
 
+	if err = p.updateInstanceTypeSpecList(); err != nil {
+		return nil, err
+	}
+
```
bpradipt, line 170: This method is time consuming and calling it for every CreateInstance call will be suboptimal. 
Populating the instance spec should be during initialisation imho.

snir911: do you have an alternative idea?

bpradipt: Not yet :-(

PR #1582, opened 2023-11-15
azure: aws: set defaults from environment to instanceTypes 

snir911:
and setting default CVM type for Azure

setting using env fixes peerpod-ctrl error:

`The following supplied instance types do not exist: []`


bpradipt (approved):
/lgtm
I think it's better to update the default here as well for consistency - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/install/overlays/azure/kustomization.yaml#L23

surajssd (approved):

mkulke (approved):

bpradipt:
Thanks @mkulke @surajssd

PR #1585, opened 2023-11-18
Remove duplicate config scripts from packer configs and misc cleanups

bpradipt:
This PR includes the following:
1. Remove duplicate config scripts for aws and azure
2. Remove centos packer config from aws and azure as we are no longer using it
3. Minor adjustment in initialising packer for aws


aws/image/Makefile
```diff
@@ -19,6 +19,7 @@ $(IMAGE_FILE): setopts $(BINARIES) $(FILES)
 		-var region=${AWS_REGION} \
 		-var instance_type=${INSTANCE_TYPE} \
 		-var volume_size=${VOLUME_SIZE} \
+		-var config_script_src=$(ROOT_DIR)podvm/qcow2 \
```
snir911, line 23: nit: I'd add "/" although not really needed

azure/image/Makefile
```diff
@@ -25,7 +25,9 @@ $(IMAGE_FILE): $(BINARIES) $(FILES)
 	# export PKR_VAR_az_gallery_image_version="${GALLERY_IMAGE_VERSION}"
 
 	packer init ./${PODVM_DISTRO}/
-	packer build ./${PODVM_DISTRO}/
+	packer build \
+		-var config_script_src=$(ROOT_DIR)podvm/qcow2 \
```
snir911, line 29: same comment about adding "/"

azure/image/rhel/variables.pkr.hcl
```diff
@@ -118,3 +118,8 @@ variable "az_gallery_image_version" {
   type    = string
   default = ""
 }
+
+variable "config_script_src" {
+  type    = string
+  default = ""
+}
```
snir911, line 125: Is this needed? can't we just drop this from variables file? same question for all  files

snir911:
also, about centos? what about support for qcow based centos? it seems it's used:
https://github.com/confidential-containers/cloud-api-adaptor/pull/1564

aws/image/Makefile
bpradipt: Hmm, since the ROOT_DIR definition already has "/" I didn't add it. Although no harm in adding.

azure/image/rhel/variables.pkr.hcl
bpradipt: The alternative is to have relative path (like https://github.com/confidential-containers/cloud-api-adaptor/blob/main/aws/image/ubuntu/aws-ubuntu.pkr.hcl#L43), which I wanted to avoid. Is there any other alternative ?

bpradipt:
> also, about centos? what about support for qcow based centos? it seems it's used: #1564

I'll create a separate issue to remove centos from qcow2.

azure/image/rhel/variables.pkr.hcl
snir911: I mean, if you already set config_script_src/others through the "-var" packer cmd option, is it needed also here?

bpradipt: Ah I see. Afaik, the variables must be there in .pkr files for it to passed via command line

snir911 (approved):
LGTM thanks!

Issue #1588, opened 2023-11-20
Show e2e-test result for libvirt Cloud Provider in opensource e2e-pipelines


liudalibj:
We created one internal job to run e2e-test for libvirt Cloud Provider base on libvirt self-managed amd64 cluster inside IBM, we pushed to test result to COS already, in this issue we want to show the e2e-test result for libvirt Cloud Provider in opensource e2e-pipelines.
- https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/daily-e2e-tests-ibmcloud.yaml

PR #1589, opened 2023-11-20
e2e: daily libvirt e2e-test with podvm-generic-ubuntu-amd64 podvm image

liudalibj:
- show e2e-test result for libvirt cloud_provider base on self-managed amd64 cluster base libvirt
- only push images when it not in registry

fixes: #1588

### DIFF
--- a/.github/workflows/daily-e2e-tests-ibmcloud.yaml
+++ b/.github/workflows/daily-e2e-tests-ibmcloud.yaml
@@ -20,7 +20,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - type: amd64
+          - type: e2e_amd64
+          - type: libvirt_amd64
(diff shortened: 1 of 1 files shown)


liudalibj:
tested at here: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/6926459077/job/18838647518

huoqifeng (approved):
lgtm

huoqifeng:
I saw `License Compliance` error in PR check but I think it's a known issue. I also noticed errors in the the test result but not a e2e pipeline issue either, so, I think it's good, thanks @liudalibj

stevenhorsman:
> I saw License Compliance error in PR check but I think it's a known issue.

Yes, this comes up as this branch was pushed to the upstream, rather than a fork, so it's best to follow that approach in future

PR #1590, opened 2023-11-20
podvm: Enable download-image on multi-arch

stevenhorsman:
- We only publish the podvm containing container images as amd64 arch images, but as @tumberino has pointed out it doesn't contain any executable code, so when downloading it, we can just force the platform to be amd64.


tumberino (approved):

bpradipt (approved):
/lgtm

PR #1592, opened 2023-11-21
terraform: add new variables for existing VPC and subnet

genjuro214:
Current IBM Cloud terraform to create self-managed cluster will always create new VPC and subnet for VSIs.

As a result, it's easy to see vpc quota exceeded errors because by default only 10 VPCs could be created in a region under an account.

In this PR, we introduced 2 new variables to specify existing VPC and subnet names. If they are set, will create VSIs in existing VPC; otherwise, will create new VPC and subnet as usual


liudalibj (approved):
/LGTM

Thanks @genjuro214

huoqifeng:
@tumberino would you please have a look at that? thanks!

PR #1600, opened 2023-11-23
Introduce the release and on_push workflows

### LINKED ISSUE #1109: Create version tags of images on release, so users can install stable versions

wainersm:
In this PR you will find:

- Added a workflow to be triggered when a release is created. It can be used for automation tasks on release, for example, build and publish the project images. Current implementation build and push tagged images of caa, podvm and peerpod-ctrl
- Added an unified workflow to be triggered when a pull request is merged. Current implementation build and push images of caa and peerpod-ctrl

Fixes #1109


wainersm:
This is still WiP because a) I wanted to check with @stevenhorsman if that's what he had in mind b) some jobs failed on my tests (see more below), and c) more images should be published by the release workflow (e.g. webhooks)

I did some fake releases on my fork of this repository to test. As you can see in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6969554029/job/18966656793 two jobs failed:

 * podvm build for CentOS (known issue, it should be disabled in https://github.com/confidential-containers/cloud-api-adaptor/pull/1595)
 * "image build manifests" of the caa job. It fails to pull some image for s390x that is supposed to had been built. Any idea what's wrong?

stevenhorsman:
> a) I wanted to check with @stevenhorsman if that's what he had in mind

Yes - at initial glance that looks great and better than I envisaged!

wainersm:
ok, updated this PR to include all the images we release (I think).

Tested the release workflow in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6984183876 :
 
 *  The [caa / build and push (dev-amd64, linux/amd64)](https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6984183876/job/19006574031#logs) started to fail today because it isn't able to pull Fedora 36 packages, I hope it is a temporary issue.
 * The  `image build manifests` job didn't run because it depends on ^^^. However, on previous tentatives it was failing, for example in  https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6969554029/job/18966656793

And here I tested the workflow for push: https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6984176502
 * I pushed changed to all components (i.e. csi-wrapper, peerpod-ctrl..etc) to check the triggering based on changed files was working
 * I also pushed commits that touches only one component to still check the filter by changed files. For example, in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/6983988123 I triggered only the peerpod-ctrl

wainersm:
@stevenhorsman @surajssd @mkulke If you have time to review this....also I appreciate if you don't care about merging with some nits because it took me a lot of time to test those workflows and I'm afraid I won't have much more time to work on them.

.github/workflows/publish_images_on_push.yaml
```diff
@@ -0,0 +1,81 @@
...
+      git_ref: ${{ github.sha }}
+    secrets: inherit
+
+  peerpod-ctrl:
+    needs: [get_changed_files]
+    uses: ./.github/workflows/peerpod-ctrl_image.yaml
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'push' &&
+      contains(needs.get_changed_files.outputs.changed_files, 'peerpod-ctrl/'))
```
mkulke, line 53: that looks pretty elegant, but will this be sufficient to cover changes that should trigger a rebuild? if changes have occured e.g. in `pkg/*` we'd also want an updated build, no?

mkulke (approved):
lgtm. tiny question with regards to the path selection logic

stevenhorsman (approved):
Looks good to me and I'm okay for us to iterate on this once merged and we've been able to see it in action a couple of times.

wainersm:
@stevenhorsman @mkulke update the on_push workflow to always build the images regardless the changed files.

Testing in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/7005944528 and https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/7006016281 (notice on this last link that I came from the future to announce the v0.9.0 release was a success! )

wainersm:
thanks @stevenhorsman @mkulke !

PR #1602, opened 2023-11-24
packer: Enable support for adding nvidia gpu to podvm image

bpradipt:
This commit enables NVIDIA gpu support for aws and azure podvm image. For example enabling gpus etc.

 NVIDIA GPU support is enabled by default when building packer based
    image for aws and azure
    A default build of podvm in aws took 10 min

  ```
    ==> Wait completed after 10 minutes 16 seconds

    ==> Builds finished. The artifacts of successful builds are:
    --> peer-pod-ubuntu.amazon-ebs.ubuntu: AMIs were created:
    us-east-2: ami-0463ae5aa8d5b3606

    rm -fr toupload

    real    10m34.352s
    user    0m18.919s
    sys     0m10.044s

```

If you want to disable, then run with ENABLE_NVIDIA_GPU=no
For example:
cd azure/image
PODVM_DISTRO=ubuntu ENABLE_NVIDIA_GPU=no make image

Few additional work is required going forward which I plan to handle in subsequent PRs.
1. Add gpu verification test case
2. Add gpu support in mkosi build

### DIFF
--- a/aws/image/Makefile
+++ b/aws/image/Makefile
@@ -7,9 +7,12 @@ include $(ROOT_DIR)podvm/Makefile.inc
 
 .PHONY: image clean default-vpc
 
-INSTANCE_TYPE ?= t3.small
+INSTANCE_TYPE ?= c4.xlarge
 
 VOLUME_SIZE ?= 30
+
+ENABLE_NVIDIA_GPU ?= yes
+export ENABLE_NVIDIA_GPU
 
 image: $(IMAGE_FILE)
 
@@ -21,6 +24,7 @@ $(IMAGE_FILE): setopts $(BINARIES) $(FILES)
 		-var instance_type=${INSTANCE_TYPE} \
 		-var volume_size=${VOLUME_SIZE} \
 		-var config_script_src=$(ROOT_DIR)/podvm/qcow2 \
+		-var addons_script_src=$(ROOT_DIR)/podvm/addons \
 		-var ami_name=${IMAGE_NAME} ./${PODVM_DISTRO}/
 	rm -fr toupload
 
--- a/podvm/addons/README.md
+++ b/podvm/addons/README.md
@@ -1,0 +1,7 @@
+## Introduction
+
+The addons directory is used to enable different addons for the podvm image.
+Each addon and its associated files (binaries, configuration etc) should be under
+specific sub-dir under `addons`. 
+
+Each addon sub-dir needs to have `setup.sh` for setting up the addon.
--- a/aws/image/rhel/aws-rhel.pkr.hcl
+++ b/aws/image/rhel/aws-rhel.pkr.hcl
@@ -97,4 +97,42 @@ build {
       "sudo -E bash ~/misc-settings.sh"
     ]
   }
+
+  # Addons
+  # To avoid multiple conditionals, copying the entire addons directory
+  # Individual addons are installed based on environment_vars by setup_addons.sh
+  provisioner "shell-local" {
+    command = "tar cf toupload/addons.tar -C ../../podvm addons"
+  }
+
+  provisioner "file" {
+    source      = "toupload"
+    destination = "/tmp/"
+  }
+
+  provisioner "shell" {
+    inline = [
+      "cd /tmp && tar xf toupload/addons.tar",
+      "rm toupload/addons.tar"
+    ]
+  }
+
+  provisioner "file" {
+    source      = "${var.addons_script_src}/setup_addons.sh"
+    destination = "~/setup_addons.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    environment_vars = [
+      "CLOUD_PROVIDER=${var.cloud_provider}",
+      "PODVM_DISTRO=${var.podvm_distro}",
+      "DISABLE_CLOUD_CONFIG=${var.disable_cloud_config}",
+      "ENABLE_NVIDIA_GPU=${var.enable_nvidia_gpu}"
+    ]
+    inline = [
+      "sudo -E bash ~/setup_addons.sh"
+    ]
+  }
+
 }
--- a/aws/image/rhel/variables.pkr.hcl
+++ b/aws/image/rhel/variables.pkr.hcl
@@ -52,3 +52,13 @@ variable "config_script_src" {
   type    = string
   default = ""
 }
+
+variable "addons_script_src" {
+  type    = string
+  default = ""
+}
+
+variable "enable_nvidia_gpu" {
+  type    = string
+  default = env("ENABLE_NVIDIA_GPU")
+}
--- a/aws/image/ubuntu/aws-ubuntu.pkr.hcl
+++ b/aws/image/ubuntu/aws-ubuntu.pkr.hcl
@@ -83,4 +83,41 @@ build {
       "sudo -E bash ~/misc-settings.sh"
     ]
   }
+
+  # Addons
+  # To avoid multiple conditionals, copying the entire addons directory
+  # Individual addons are installed based on environment_vars by setup_addons.sh
+  provisioner "shell-local" {
+    command = "tar cf toupload/addons.tar -C ../../podvm addons"
+  }
+
+  provisioner "file" {
+    source      = "toupload"
+    destination = "/tmp/"
+  }
+
+  provisioner "shell" {
+    inline = [
+      "cd /tmp && tar xf toupload/addons.tar",
+      "rm toupload/addons.tar"
+    ]
+  }
+
+  provisioner "file" {
+    source      = "${var.addons_script_src}/setup_addons.sh"
+    destination = "~/setup_addons.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    environment_vars = [
+      "CLOUD_PROVIDER=${var.cloud_provider}",
+      "PODVM_DISTRO=${var.podvm_distro}",
+      "DISABLE_CLOUD_CONFIG=${var.disable_cloud_config}",
+      "ENABLE_NVIDIA_GPU=${var.enable_nvidia_gpu}"
+    ]
+    inline = [
+      "sudo -E bash ~/setup_addons.sh"
+    ]
+  }
 }
--- a/aws/image/ubuntu/variables.pkr.hcl
+++ b/aws/image/ubuntu/variables.pkr.hcl
@@ -53,3 +53,12 @@ variable "config_script_src" {
   default = ""
 }
 
+variable "addons_script_src" {
+  type    = string
+  default = ""
+}
+
+variable "enable_nvidia_gpu" {
+  type    = string
+  default = env("ENABLE_NVIDIA_GPU")
+}
--- a/azure/image/Makefile
+++ b/azure/image/Makefile
@@ -7,6 +7,8 @@ include $(ROOT_DIR)podvm/Makefile.inc
 
 .PHONY: image clean
 
+ENABLE_NVIDIA_GPU ?= yes
+
 image: $(IMAGE_FILE)
 
 $(IMAGE_FILE): $(BINARIES) $(FILES)
@@ -27,6 +29,7 @@ $(IMAGE_FILE): $(BINARIES) $(FILES)
 	packer init ./${PODVM_DISTRO}/
 	packer build \
 		-var config_script_src=$(ROOT_DIR)/podvm/qcow2 \
+		-var addons_script_src=$(ROOT_DIR)/podvm/addons \
 		./${PODVM_DISTRO}/
 	rm -fr toupload
 
--- a/azure/image/rhel/azure-rhel.pkr.hcl
+++ b/azure/image/rhel/azure-rhel.pkr.hcl
@@ -110,6 +110,43 @@ build {
     ]
   }
 
+  # Addons
+  # To avoid multiple conditionals, copying the entire addons directory
+  # Individual addons are installed based on environment_vars by setup_addons.sh
+  provisioner "shell-local" {
+    command = "tar cf toupload/addons.tar -C ../../podvm addons"
+  }
+
+  provisioner "file" {
+    source      = "toupload"
+    destination = "/tmp/"
+  }
+
+  provisioner "shell" {
+    inline = [
+      "cd /tmp && tar xf toupload/addons.tar",
+      "rm toupload/addons.tar"
+    ]
+  }
+
+  provisioner "file" {
+    source      = "${var.addons_script_src}/setup_addons.sh"
+    destination = "~/setup_addons.sh"
+  }
+
+  provisioner "shell" {
+    remote_folder = "~"
+    environment_vars = [
+      "CLOUD_PROVIDER=${var.cloud_provider}",
+      "PODVM_DISTRO=${var.podvm_distro}",
+      "DISABLE_CLOUD_CONFIG=${var.disable_cloud_config}",
+      "ENABLE_NVIDIA_GPU=${var.enable_nvidia_gpu}"
+    ]
+    inline = [
+      "sudo -E bash ~/setup_addons.sh"
+    ]
+  }
+
   provisioner "shell" {
     execute_command = "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'"
     inline = [
--- a/azure/image/rhel/variables.pkr.hcl
+++ b/azure/image/rhel/variables.pkr.hcl
@@ -123,3 +123,13 @@ variable "config_script_src" {
   type    = string
   default = ""
 }
+
+variable "addons_script_src" {
+  type    = string
+  default = ""
+}
+
+variable "enable_nvidia_gpu" {
+  type    = string
+  default = env("ENABLE_NVIDIA_GPU")
+}
--- a/azure/image/ubuntu/azure-ubuntu.pkr.hcl
+++ b/azure/image/ubuntu/azure-ubuntu.pkr.hcl
@@ -90,6 +90,44 @@ build {
     ]
   }
 
+  # Addons
+  # To avoid multiple conditionals, copying the entire addons directory
+  # Individual addons are installed based on environment_vars by setup_addons.sh
+  provisioner "shell-local" {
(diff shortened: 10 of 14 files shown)

### CODE REVIEW

**Comment by katexochen** [comment]
As far as I understood, we plan to remove these images at the end of this release cycle, so I'm not sure why we are adding new features.

What are the security considerations of mounting blob fuse? How is this secure in the CC attacker model?

**Comment by mkulke** [comment]
> > To my understanding, we plan to remove these images at the end of this release cycle, so I'm not sure why we are adding new features.
> 
> Right and hence I want to hear everyone's thoughts. I can start exploring how to have similar functionality with mkosi. But unfortunately there is no way to work on other use cases till that time. Also I have made an attempt to keep it isolated and only include in packer builds. By default nothing gets included in the packer generated image.

IMO it's ok to conduct experiments like until we don't formally declare the packer-related code deprecated and freeze it. I'm not entirely sure we're committed to dropping packer, as it might be required for certain archs. We should reach and document a consensus about that,

What I think we have reached an agreement on so far: we're committed to provide mkosi-based image builds, possibly as a reference architecture. what we do with the existing packer-based infra needs to be decided (like deprecate or move it to a downstream project after 0.9). We should keep discussing this topic in #1326 

This PR probably covers 2 aspects:
1) a facility for having addons in the podvm
2) sample addons using that facility

I think we can look at those individually.

For 1) I think we should probably consider a similar mechanism for configuring mkosi builds. maybe we can have something like the `debug` profile? maybe we don't want that level of configurability of mkosi builds, because it leads to an explosion in image permutations or in a large tcb?

In any case, 2) looks intriguing. I'll have to check those out. I guess they will also set the bar for what we have to support with fedora-based mkosi builds to make a switch feasible.

**Comment by katexochen** [comment]
> I guess they will also set the bar for what we have to support with fedora-based mkosi builds to make a switch feasible.

Continuously raising the bar for mkosi images by continuing adding features to the legacy approach does not seem reasonable to me.

**Comment by mkulke** [comment]
> > I guess they will also set the bar for what we have to support with fedora-based mkosi builds to make a switch feasible.
> 
> Continuously raising the bar for mkosi images by continuing adding features to the legacy approach does not seem reasonable to me.

admittedly that's annoying, but I think that's tolerable if temporary. We probably want to decide on concrete timeline/criteria on when to switch over in a coming community call, so there's less ambiguity. OTOH, at this point of the project, I also wouldn't want to curb/hamper exciting tinkering, like the GPU integration, by enforcing a sanctioned-only way to do images. It's a balance, I guess.

File: podvm/addons/gpu/README.md [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+Example pod definition:
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-test
+  labels:
+    app: test
+  annotations:
+    io.katacontainers.config.hypervisor.machine_type: Standard_NC4as_T4_v3
```
mkulke, line 27:
```suggestion
    io.katacontainers.config.hypervisor.machine_type: Standard_NC4as_T4_v3
    io.containerd.cri.runtime-handler: kata-remote
```

don't we need this?

mkulke (commented):
- Tested a build without an addon => built successfully, was able to start a pod with it on AKS
- Tested a build with gpu addon => built successfully, couldn't start pod due to missing gpu quota

So, in general things seem to at least not break. A couple of questions/comments:

- afaict we have no test-coverage for those, so the addon-feature or individual addons might be broken inadvertently by future code changes. manual testing is very cumbersome and in the GPU case, depends on available quota.
- do we intend to support non-confidential GPU instances in the project?
- does it make sense to introduce a "*-contrib" structure (maybe not a repo, can be a folder) in which can stage experimental features that are not supported/systematically tested yet?

bpradipt:
> @bpradipt can we add addon-support to mkosi build process as well?
> 
> What are the blockers to support the GPU artifacts for mkosi based image builds?

This is next in my plan to include both GPU and opa for mkosi based image builds. The only catch is testing the Fedora packages as most of my GPU testing have been with either RHEL or Ubuntu

podvm/addons/opa/README.md
```diff
@@ -0,0 +1,6 @@
+## Introduction
```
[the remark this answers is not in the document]
bpradipt: Yeah..

podvm/addons/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Each addon needs to have `setup.sh` for setting up the addon.
+To enable the addon create a file named `.enable` in the top level podvm/addons
+directory and in the specific addon sub-dir.
+
+The addons are setup by the `qcow2/setup_addons.sh` script.
+
+Two addons are provided as examples:
+- `gpu`: This addon enables nvidia GPU support in the podvm image
+- `blobfuse`: This addon enables Azure blobfuse to mount blobs inside pod VM for use by the pods
```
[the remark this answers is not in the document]
bpradipt: Yeah, I'll update. I removed the blobfuse and mountpoint for now.

bpradipt:
> * Tested a build without an addon => built successfully, was able to start a pod with it on AKS
> * Tested a build with gpu addon => built successfully, couldn't start pod due to missing gpu quota
> 
> So, in general things seem to at least not break. A couple of questions/comments:
> 
> * afaict we have no test-coverage for those, so the addon-feature or individual addons might be broken inadvertently by future code changes. manual testing is very cumbersome and in the GPU case, depends on available quota.
> * do we intend to support non-confidential GPU instances in the project?
> * does it make sense to introduce a "*-contrib" structure (maybe not a repo, can be a folder) in which can stage experimental features that are not supported/systematically tested yet?

Yeah. I don't have a good idea on how that should be. This applies for packer (which will be removed going forward) and mkosi as well. 
Is it ok if the restructuring can be taken up later on after more discussions on the best way forward - like separate directory or a project ?

podvm/addons/gpu/README.md
bpradipt: In my testing it was not needed. The annotation will be used by CAA to spin up the specific instance and the runtimeclass is part of of the pod spec itself.

podvm/addons/README.md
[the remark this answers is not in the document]
bpradipt: I have removed the opa from this series and will add it as part of the policy enablement PR I plan to send after this.

bpradipt:
I have removed the `opa` addon from this series as I feel it's best suited to be used with the policy enablement pr - https://github.com/confidential-containers/cloud-api-adaptor/pull/1607.

Both gpu and opa addons need to be enabled for mkosi builds. This is next in my todo list.

podvm/addons/gpu/README.md
mkulke: I think it will break once we switch to kata#main. AFAIU the nydus-snapshotter requires this.

podvm/addons/README.md
```diff
@@ -0,0 +1,21 @@
...
+The addons are setup by the `qcow2/setup_addons.sh` script.
+
+An example skeleton addon is provided as an example
+
+By default all addons are disabled. If you want to enable an addon, then you will need to do the following:
+
+```
+export ADDON_NAME=REPLACE_ME
+touch podvm/addons/.enable
+touch podvm/addons/${ADDON_NAME}.enable
```
mkulke, line 20: Do we need to introduce this `.enable` mechanism or can we cover this with makefile flags?

podvm/addons/.gitignore
```diff
@@ -0,0 +1 @@
+.enable
```
mkulke, line 1: If this is in .gitignore you cannot build images with addons on a CI.

mkulke (commented):
> Yeah. I don't have a good idea on how that should be. This applies for packer (which will be removed going forward) and mkosi as well.
Is it ok if the restructuring can be taken up later on after more discussions on the best way forward - like separate directory or a project ?

I'm not sure I grok the intent of this proposal yet. Is it purely provisional or is intended to be a future-proof image extension mechanism? 

In general, we probably want to reduce the configurable surface of podvm images to make testing more manageable and have some confidence about working configurations. This PR would go in an opposite direction. Maybe it's warranted for GPU-enabled podvm images, as a special case. But then I would cover this specific case with some `make INCLUDE_GPU_STACK=1 image` parameterization, rather than introducing a generic addon mechanism.

podvm/addons/.gitignore
bpradipt: I see..

podvm/addons/README.md
bpradipt: Possible to cover with makefile. Like for policy, based on AGENT_POLICY=yes, these files will be created using instructions in the Makefile.

bpradipt:
> > Yeah. I don't have a good idea on how that should be. This applies for packer (which will be removed going forward) and mkosi as well.
> > Is it ok if the restructuring can be taken up later on after more discussions on the best way forward - like separate directory or a project ?
> 
> I'm not sure I grok the intent of this proposal yet. Is it purely provisional or is intended to be a future-proof image extension mechanism?
> 

I would not say future proof. But at least a workable method to add extensions. I open for other ideas for extensibility.

> In general, we probably want to reduce the configurable surface of podvm images to make testing more manageable and have some confidence about working configurations. This PR would go in an opposite direction. Maybe it's warranted for GPU-enabled podvm images, as a special case. But then I would cover this specific case with some `make INCLUDE_GPU_STACK=1 image` parameterization, rather than introducing a generic addon mechanism.

Agree. I would start with the premise that anything which is  turned on by default should be testable in CI. 
Also parameterization will soon become unmanageable for different cases. Like the existing misc script with conditionals.

podvm/addons/README.md
bpradipt: https://github.com/confidential-containers/cloud-api-adaptor/pull/1607/files#diff-6bc6c17c0f0e3f8d3827c3664d191076319cc9059c50dee659ea9392344ed602R38-R41

podvm/addons/.gitignore
mkulke: yes, but then you have to alter the workflow. often workflows and OIDC identities are tied to a specific branch (main), so you'll run the workflow on main, but specify a "ref" that will be used in the `checkout` step to build a specific revision.

podvm/addons/README.md
bpradipt: Similarly, for addons which we decide to enable always, we can add an option in makefile to create the `.enable` file and include it.

I'm thinking the same approach can be used when switching to mkosi although I'm not yet sure how.

mkulke:
> I would not say future proof. But at least a workable method to add extensions. I open for other ideas for extensibility.

Do we need such an extension mechanism, or is it sufficient to document how an image can be customized?

In Systemd there's Sysext and Confext: https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html to extend os images in a systematic fashion (like updating a bundled container runtime or system library) but it's not package manager, things easily break if conflicting system extensions are applied.

bpradipt:
> https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html



> > I would not say future proof. But at least a workable method to add extensions. I open for other ideas for extensibility.
> 
> Do we need such an extension mechanism, or is it sufficient to document how an image can be customized?

Hmm, challenge is the packer config. Will it be possible via only documentation to provide the changes to packer hcl ? 

> 
> In Systemd there's Sysext and Confext: https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html to extend os images in a systematic fashion (like updating a bundled container runtime or system library) but it's not package manager, things easily break if conflicting system extensions are applied.

Are you suggesting enabling this option in the podvm image to allow extensions ?
Also reading about it, I don't think this will work for rpms/debs. Might work for binaries and configs, but even then I'm not sure what's the mechanism to provide these extensions. Do you have some ideas ?

mkulke:
> Hmm, challenge is the packer config. Will it be possible via only documentation to provide the changes to packer hcl ?

I don't think it's gets a lot easier to customize a build than what is supported right now in packer + hcl if you copy a script and execute it:

```hcl
  provisioner "file" {
    source      = "/install-some-deps.sh"
    destination = "~/install-some-deps.sh"
  }

  provisioner "shell" {
    remote_folder = "~"
    inline = [
      "sudo bash ~/install-some-deps.sh"
    ]
  }
```

> Are you suggesting enabling this option in the podvm image to allow extensions ? Also reading about it, I don't think this will work for rpms/debs. Might work for binaries and configs, but even then I'm not sure what's the mechanism to provide these extensions. Do you have some ideas ?

No indeed. sys/confext are only for binaries and configs, nothing is being executed here. rpms/debs you would install in the mkosi or packer build-process. sysext would merely allow extending an image by overlaying some files, without interfering in the build process itself.

Again, I probably don't understand the use case well, to understand the requirements atm. like who would want to extend a podvm image with what? 

Is an extension something like a debug image, like a fixed, known-by-us build-time option? could it be `make image-gpu` like `make image-debug` or do we require something more composable `make image DEBUG=1 GPU=1` then? I think either would work without developing our own extension mechanism.

Or is an extension something unforeseeable, like a bespoke monitoring agent that a customer has to install and configure in their environment? In this case both packer and mkosi provide convenient ways to do this kind of customization, we can hardly improve on those in terms of ux.

bpradipt:
> > Hmm, challenge is the packer config. Will it be possible via only documentation to provide the changes to packer hcl ?
> 
> I don't think it's gets a lot easier to customize a build than what is supported right now in packer + hcl if you copy a script and execute it:
> 
> ```terraform
>   provisioner "file" {
>     source      = "/install-some-deps.sh"
>     destination = "~/install-some-deps.sh"
>   }
> 
>   provisioner "shell" {
>     remote_folder = "~"
>     inline = [
>       "sudo bash ~/install-some-deps.sh"
>     ]
>   }
> ```
> 
> > Are you suggesting enabling this option in the podvm image to allow extensions ? Also reading about it, I don't think this will work for rpms/debs. Might work for binaries and configs, but even then I'm not sure what's the mechanism to provide these extensions. Do you have some ideas ?
> 
> No indeed. sys/confext are only for binaries and configs, nothing is being executed here. rpms/debs you would install in the mkosi or packer build-process. sysext would merely allow extending an image by overlaying some files, without interfering in the build process itself.
> 
> Again, I probably don't understand the use case well, to understand the requirements atm. like who would want to extend a podvm image with what?
> 
> Is an extension something like a debug image, like a fixed, known-by-us build-time option? could it be `make image-gpu` like `make image-debug` or do we require something more composable `make image DEBUG=1 GPU=1` then? I think either would work without developing our own extension mechanism.
> 
> Or is an extension something unforeseeable, like a bespoke monitoring agent that a customer has to install and configure in their environment? In this case both packer and mkosi provide convenient ways to do this kind of customization, we can hardly improve on those in terms of ux.

Let me ask a different question - what if we add gpu and opa in the default podvm image ? And instead of `.enable` I can include the addons based on AGENT_POLICY and ENABLE_GPU options. By default both these will be enabled. 
Using flexible instance types we can always test gpu in CI if needed (although cost might be a factor).
Other than image build time and size, do you see any drawback ?

And for extensions we can rethink in the context of mkosi and not spend too much time in doing it with packer. Wdyt ?

mkulke:
> Let me ask a different question - what if we add gpu and opa in the default podvm image ? And instead of `.enable` I can include the addons based on AGENT_POLICY and ENABLE_GPU options. By default both these will be enabled. Using flexible instance types we can always test gpu in CI if needed (although cost might be a factor). Other than image build time and size, do you see any drawback ?
> 
> And for extensions we can rethink in the context of mkosi and not spend too much time in doing it with packer. Wdyt ?

yes, as built-time options those do make sense to me. About what feature is supposed to be opt-in and opt-out there can be a separate discussion once we port these to mkosi.

bpradipt:
@mkulke I changed this PR to add support for NVIDIA GPU in packer based podvm images.  Updated the main description as well.

NVIDIA gpu support is enabled by default, meaning the packer built images will have the NVIDIA gpu drivers and should be usable depending on the instance type selected. 
Let me know what you think and if we can move ahead with this one for now.

Few additional work is required going forward which I plan to handle in subsequent PRs.

- Add gpu verification test case
- Add gpu support in mkosi build

podvm/files/etc/agent-config.toml
```diff
@@ -7,7 +7,7 @@ server_addr="unix:///run/kata-containers/agent.sock"
 
 # This field sets up the KBC that attestation agent uses
 # This is replaced in the makefile steps so do not set it manually
-aa_kbc_params = ""
+aa_kbc_params = "cc_kbc::null"
```
mkulke, line 10: unrelated, I assume? might have side-effects

mkulke (approved):
that makes sense to me, thanks for accommodating.

nit: I doubt we need the `addon` folder. the `enable_nvidia_gpu` flag would be enough. since we copy the necessary scripts on the build image anyway, we can just check for the ENV in `misc-settings.sh` and invoke `nvidia_gpu/setup.sh` if necessary.

podvm/files/etc/agent-config.toml
bpradipt: yeah, unrelated

bpradipt:
> that makes sense to me, thanks for accommodating.
> 

No problems. I'm glad we could arrive at a consensus.

> nit: I doubt we need the `addon` folder. the `enable_nvidia_gpu` flag would be enough. since we copy the necessary scripts on the build image anyway, we can just check for the ENV in `misc-settings.sh` and invoke `nvidia_gpu/setup.sh` if necessary.

I thought about whether to have a separate folder or not, and felt it's cleaner to have a separate folder and keep the config files under it to accommodate support for other GPUs (eg Intel ones) or other accelerators requiring podvm image customisation.  The naming is a challenge though. Thought about misc, contrib etc but then went with addons since there is already misc-settings :-).

aws/image/Makefile
```diff
@@ -7,10 +7,13 @@ include $(ROOT_DIR)podvm/Makefile.inc
 
 .PHONY: image clean default-vpc
 
-INSTANCE_TYPE ?= t3.small
+INSTANCE_TYPE ?= c4.xlarge
```
snir911, line 10: Do we want this to be the default ? or only ENABLE_NVIDIA_GPU=yes

snir911 (approved):
LGTM overall, added minor comment, thanks!

PR #1603, opened 2023-11-24
Unify build workflows

wainersm:
* Merged the controllers workflow into build
* Build and run tests for webhook and volumes/csi-wrapper as part of the build workflow


.github/workflows/build.yaml
```diff
@@ -62,3 +62,48 @@ jobs:
           name: tests_report_junit-${{ matrix.runner }}_${{ env.GO_VERSION }}
           path: ${{ github.workspace }}/tests_report_junit.xml
           retention-days: 1
+
+  controllers:
+    name: controllers
+    runs-on: ubuntu-latest
```
snir911, line 68: I just noticed it's running on ubuntu, Dockerfile uses quay.io/confidential-containers/golang-fedora:1.20.8-38, maybe worth to make these match as part of this series, WDYT?

beraldoleal (approved):
LGTM, thanks.

.github/workflows/build.yaml
wainersm: @snir911 it seems a good idea. There aren't github-hosted Fedora runners, we would need to use VMs on Azure. And I just took look at the azure's marketplace but couldn't find a suitable Fedora 38 image. Thus, we would need to create and push a Fedora image first. Mind if we merge this PR without such as change then I can open an issue to track (and hopefully) and change the runner later?

wainersm:
Updated this PR to fix a conflict on versions.yaml.

.github/workflows/build.yaml
snir911: oh, sorry i thought i acked, sure go ahead

snir911 (approved):

stevenhorsman (approved):
LGTM. Thanks!

PR #1625, opened 2023-12-13
docs: Add docs for e2e tests on CAA deployment

mkulke:
Adding instruction on how tests can be triggered against an existing deployment of CAA. This is useful for debugging tests and testing local changes.


test/e2e/README.md
```diff
@@ -129,4 +129,34 @@ func TestCloudProviderCreateSimplePod(t *testing.T) {
...
+
+Run the test suite with the respective flags:
+
+```bash
+CLOUD_PROVIDER=azure \
+TEST_TEARDOWN=no \
+TEST_PROVISION=no \
+TEST_INSTALL_CAA=no \
+TEST_PROVISION_FILE=./skip-provisioning.properties \
+go test -v -tags=$CLOUD_PROVIDER -timeout 10m -count=1 ./
```
stevenhorsman, line 161: Not a big deal, but I wonder if we should make this just run `make test-e2e` for consistency and decoupling incase we want to update the test command in future. I guess your instructions assumes that we will be in `test/e2e` (which is a safe assumption as that's where this doc file lives), so if you do switch this we might need to make it clear that is runs from the project root directory (which we might need for other places in this doc too)?

stevenhorsman (approved):
LGTM. Thanks!

bpradipt (approved):
/lgtm

kartikjoshi21 (approved):

PR #1627, opened 2023-12-19
podvm: revert agent-config path to /etc

mkulke:
The ./podvm kata-agent unit has been set to use a config file in /etc. Having the kata-agent config file in /run will break the CAA libvirt tests, since we have dependencies that rely on the config being in /etc. ./podvm-mkosi will override this path to a configuration in /run.

Tested w/ PodVM images built w/ packer + mkosi

### DIFF
--- a/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf
+++ b/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf
@@ -1,0 +1,10 @@
+# On a read-only fs the kata-agent config is created in /run/peerpod, since it contains
+# a parameter that can be set at pod creation time.
+[Unit]
+ConditionKernelCommandLine=
+
+[Service]
+ExecStart=
+ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
+ExecStop=
+ExecStopPost=/usr/local/bin/kata-agent-clean --config /run/peerpod/agent-config.toml
--- a/podvm/files/etc/systemd/system/kata-agent.service
+++ b/podvm/files/etc/systemd/system/kata-agent.service
@@ -6,9 +6,9 @@ After=netns@podns.service process-user-data.service
 
 [Service]
 ExecStartPre=mkdir -p /run/kata-containers
-ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
+ExecStart=/usr/local/bin/kata-agent --config /etc/agent-config.toml
 ExecStartPre=-umount /sys/fs/cgroup/misc
-ExecStopPost=/usr/local/bin/kata-agent-clean --config /run/peerpod/agent-config.toml
+ExecStopPost=/usr/local/bin/kata-agent-clean --config /etc/agent-config.toml
 # Now specified in the agent-config.toml Environment="KATA_AGENT_SERVER_ADDR=unix:///run/kata-containers/agent.sock"
 SyslogIdentifier=kata-agent
 
--- a/cmd/process-user-data/types.go
+++ b/cmd/process-user-data/types.go
@@ -5,7 +5,7 @@ const (
 	providerAzure = "azure"
 	providerAws   = "aws"
 
-	defaultAgentConfigPath  = "/run/peerpod/agent-config.toml"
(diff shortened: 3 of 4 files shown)


stevenhorsman (approved):
Code LGTM, I'm trying it out on s390x with packer just to test, but if libvirt-e2e ci test pass then that's good enough for me.

stevenhorsman:
I guess there is still the question of whether the guest-components that read directly from `/etc/agent-config.toml` will be broken in the mkosi build if that is still using the new `/run/peerpods/` path?

mkulke:
> I guess there is still the question of whether the guest-components that read directly from `/etc/agent-config.toml` will be broken in the mkosi build if that is still using the new `/run/peerpods/` path?

yes, independently from this PR, cc_kbc doesn't work w/ mkosi at the moment, as the relevant paths are hardcoded in guest-components, so we need to add a fix for that still.

podvm/files/etc/systemd/system/kata-agent.service
```diff
@@ -6,9 +6,9 @@ After=netns@podns.service process-user-data.service
 
 [Service]
 ExecStartPre=mkdir -p /run/kata-containers
-ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
+ExecStart=/usr/local/bin/kata-agent --config /etc/agent-config.toml
```
bpradipt, line 9: I think the process-user-data will still update the default file - /run/peerpod/agent-config.toml (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/cmd/process-user-data/types.go#L8).

Does the systemd unit also need to create a link 
`ln -s  /etc/agent-config.toml /run/peerpod/agent-config.toml`  ?

mkulke: oh, good catch. yes, we also need to undo that, otherwise cc_kbc will also be broken on packer-based builds

podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf
```diff
@@ -0,0 +1,10 @@
+# On a read-only fs the kata-agent config is created in /run/peerpod, since it contains
+# a parameter that can be set at pod creation time.
+[Unit]
+ConditionKernelCommandLine=
+
+[Service]
+ExecStart=
+ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
+ExecStop=
+ExecStopPost=/usr/local/bin/kata-agent-clean --config /run/peerpod/agent-config.toml
```
bpradipt, line 10: The systemd file for process-user-data also needs to be overridden to specify the config file as `/run/peerpod/agent-config.toml` 

```
[Service]
ExecStart=
ExecStart=/usr/local/bin/process-user-data update-agent-config --agent-config-file /run/peerpod/agent-config.toml

```

podvm/files/etc/systemd/system/kata-agent.service
mkulke: Ok, with the path reset in `types.go` cc_kbc works again on a packer-built image. It doesn't work w/ an mkosi-image, but that's expected.

podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf
mkulke: tested w/ mkosi. The unit doesn't fail anymore. (CDH still fails, but that's expected)

bpradipt, line 10: Override for process-user-data service is missing.

liudalibj (approved):
LGTM, with this fix, it will unblock the daily e2e-test for ibmcloud and libvirt.
Thanks @mkulke

podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf
mkulke: indeed. thx

surajssd (approved):

PR #1636, opened 2023-12-20
Update AWS setup instructions

bpradipt:
This series provides the following
1. Update instructions for AWS
2. Use env file for secrets (similar to Azure)

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -1,33 +1,42 @@
-# Setup instructions
+# Cloud API Adaptor (CAA) on AWS
+
+This documentation will walk you through setting up CAA (a.k.a. Peer Pods) on Amazon Elastic Kubernetes Service (EKS). It explains how to deploy:
+
+- One worker EKS
+- CAA on that Kubernetes cluster
+- An Nginx pod backed by CAA pod VM
+
+> **Note**: Run the following commands from the root of this repository.
+
 ## Prerequisites
 
-- Set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` for AWS cli access
-
+- Install `aws` CLI [tool](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
+- Install `eksctl` CLI [tool](https://eksctl.io/installation/)
+- Set `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_REGION` for AWS cli access
 - Install packer by following the instructions in the following [link](https://learn.hashicorp.com/tutorials/packer/get-started-install-cli)
-
 - Install packer's Amazon plugin `packer plugins install github.com/hashicorp/amazon`
 
-## Build Pod VM Image
-
-### Option-1: Modifying existing marketplace image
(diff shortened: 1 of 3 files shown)


mkulke (approved):
LGTM, i'm not able to test it though

aws/README.md
```diff
@@ -62,64 +71,208 @@ docker build -t aws \
...
+AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+EOF
 ```
-mkdir -p qcow2-img && cd qcow2-img
 
-curl -LO https://raw.githubusercontent.com/confidential-containers/cloud-api-adaptor/staging/podvm/hack/download-image.sh
+### Update the `kustomization.yaml` file
+
+Run the following command to update the [`kustomization.yaml`](../install/overlays/aws/kustomization.yaml) file:
```
stevenhorsman, line 140: Hey Pradipta - is this intended to be a point in time reference, or something longer lasting? If it's the later then I don't know whether it would be better to instruction them to edit the existing `install/overlays/aws/kustomization.yaml` file rather than overwrite it, so things like the CAA new are not fixed?

You might be able to do something with the caa provisioner to help with the config, but I understand if you don't want to get into that.

bpradipt: Thanks for the suggestion Steve. It does make sense to instruct direct editing of the kustomization.yaml. However, currently for AWS we only need one variable to be provided and the rewriting helps to get started quickly. 
Going forward, I will explore `sed` commands to edit the values from CLI instead of re-writing. Hope this is ok..

stevenhorsman: Sure, if it's a short-term hard-coded doc then that's fine - I just don't want AWS users to get stuck on the 0.8 release of CAA once we've updated to 0.9 based on main

stevenhorsman (approved):
I don't have an AWS account to test, but the doc looks reasonable to me apart from the version hard-coding which Pradipta is planning to review in future. Thanks!

aws/README.md
snir911: BTW, another option would be to modify CM after creation, can be done in 2 commands, not sure it's better though...

aws/README.md
```diff
@@ -62,64 +71,208 @@ docker build -t aws \
...
+```
+
+## Deploy Kubernetes using EKS
+
+- Create cluster config file
+This config will create a single node cluster. Feel free to change it as per your requirement.
 
 ```
-aws ec2 describe-images --query "Images[*].[ImageId]" --filters "Name=name,Values=peer-pod-ami" --region ${AWS_REGION} --output text
+cat > ekscluster-config.yaml <<EOF
```
snir911, line 100: maybe it worth considering having some templates folder and then using something like `envsubst < tmps/eks-config.yaml| oc apply -f-`

snir911 (approved):
LGTM, thanks!

aws/README.md
bpradipt: Yeah, either this or some sed fu to make changes

PR #1662, opened 2024-01-12
docs: update docs to point to CoCo instructions

beraldoleal:
This ensures users and developers always access the most up-to-date instructions, which is already documented at the CoCo level, eliminating the risk of outdated or duplicated content in our documentation.


stevenhorsman (approved):
Hey @beraldoleal - your second commit is failing the checks as it has no body, but otherwise these changes look good to me. Thanks for the clarifications

PR #1686, opened 2024-02-03
Misc fixes

bpradipt:
1. The provisioning command used a wrong ordering of daemonconfig and authJson files, resulting in startup errors for process-user-data
2. Change AWS default instance type to CVM to align with the default  disable-cvm option
3. Fix handling of describeimages


huoqifeng (approved):
LGTM

snir911 (approved):

PR #1698, opened 2024-02-16
packer: add replication region to shared image gallery

cmaf:
Add eastus2 to replication regions in order to access the image from there.

Note: My understanding is that this will make the nightly build image available in eastus2 as well as eastus, is that the case? Per this comment from another thread https://github.com/confidential-containers/cloud-api-adaptor/issues/1696#issuecomment-1934942535


azure/image/ubuntu/azure-ubuntu.pkr.hcl
```diff
@@ -30,6 +30,7 @@ source "azure-arm" "ubuntu" {
     image_name           = "${var.az_gallery_image_name}"
     image_version        = "${var.az_gallery_image_version}"
     storage_account_type = "Standard_LRS"
+    replication_regions  = "eastus2"
```
mkulke, line 33: note: replication_regions is an array, I think it makes sense to add eastus2 for TDX availability, we can also add `westeurope`, since it supports both tdx and snp vms
```suggestion
    replication_regions  = ["eastus2", "westeurope"]
```

cmaf: @mkulke added, thanks for the feedback!

mkulke (approved):
lgtm

bpradipt (approved):
/lgtm
Thanks @cmaf

cmaf:
@bpradipt @mkulke how quickly does this take effect? Is it the nightly build?

mkulke:
yup should be replicated in the build. please note, if you plan to use it for tdx, the image might not be configured properly yet: https://github.com/confidential-containers/cloud-api-adaptor/issues/1703

PR #1715, opened 2024-02-26
ci: reuse az nightly image in az e2e test

mkulke:
draft until #1713 is merged

For nightly e2e-tests we can re-use the nightly-built podvm image
without rebuilding it from scratch.

This change will convert the e2e test into a reusable workflow and
invoke it nightly from a meta workflow "azure nightly build". this
workflow will call both the podvm-image-build and the e2e test
subsequently.


mkulke:
good idea, doesn't work. the output from one reusable wf doesn't seem to be usable as input for another, it will be in call cases. I don't think it's worth debugging.

PR #1718, opened 2024-02-26
policy: fix broken default policy symlink on mkosi

mkulke:
The podvm makefile would be invoked with an empty default policy file setting from the mkosi docker build, in which the docker-build arg is set to an empty string. The podvm makefile would hence not pick the default, but attempt to use the empty string in the creation of the symbolic link, resulting in a symbolic link that points to itself.

The change will make sure that the env is only set in the Dockerfile invocation if the variable is set in the mkosi makefile. Also we quote the the source file for the symbolic link creation, so we bail out properly, instead of creating a self-pointing symbolic link.


stevenhorsman (approved):
LGTM. Thanks for tracking it down!

bpradipt (approved):
/lgtm

PR #1731, opened 2024-03-11
kustomize: upgrade kustomize to v5.3.0

liudalibj:
- upgrade kustomize to v5.3.0 which has s390x arch support

fixes https://github.com/confidential-containers/cloud-api-adaptor/issues/1732

### DIFF
--- a/peerpod-ctrl/Makefile
+++ b/peerpod-ctrl/Makefile
@@ -194,7 +194,7 @@ CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
-KUSTOMIZE_VERSION ?= v3.8.7
+KUSTOMIZE_VERSION ?= v5.3.0
 CONTROLLER_TOOLS_VERSION ?= v0.10.0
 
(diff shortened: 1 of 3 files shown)


snir911 (approved):
Thanks! I wonder why peerpodsconfig doesn't have it, old operator-sdk? maybe it should be bumped ?

liudalibj:
> Thanks! I wonder why peerpodsconfig doesn't have it, old operator-sdk? maybe it should be bumped ?

peerpodsconfig using another way to install kustomize, and it is using kustomize@v4 latest version, that is `v4.5.7`
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/peerpodconfig-ctrl/Makefile#L166
and it's has s390x support: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.5.7

huoqifeng (approved):
LGTM

liudalibj:
> Thanks! I wonder why peerpodsconfig doesn't have it, old operator-sdk? maybe it should be bumped ?

update the get dep tools method in peerpodsconfig-ctrl, make it same as the one in peerpods-ctrl

stevenhorsman (approved):
LGTM. Thanks for unifying the approach that all three components take!

PR #1739, opened 2024-03-13
Build: Update build.sh

davidhadas:
Support podman  - no `--push` flag

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -59,6 +59,7 @@ function build_caa_payload_image() {
 		--build-arg COMMIT="${commit}" \
 		--build-arg YQ_VERSION="${YQ_VERSION}" \
 		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
+		-- load \
 		-f Dockerfile \
 		${tag_string} \
 		.
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -61,8 +61,8 @@ function build_caa_payload_image() {
 		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
 		-f Dockerfile \
 		${tag_string} \
-		--push \
 		.
+	docker push  ${registry}/${name}  
 	popd
 }
 
@@ -106,8 +106,8 @@ function build_caa_payload_arch_specific() {
 		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
 		-f Dockerfile \
 		${tag_string} \
-		--push \
 		.
+  	docker push  ${registry}/${name}
 	popd
 }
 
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -61,8 +61,8 @@ function build_caa_payload_image() {
 		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
 		-f Dockerfile \
 		${tag_string} \
-		--push \
```
liudalibj, line 64: here maybe we need `--load` to replace `--push`

hack/build.sh
```diff
@@ -106,8 +106,8 @@ function build_caa_payload_arch_specific() {
...
```
liudalibj, line 109: here maybe we need `--load` to replace `--push`

davidhadas:
How can I force the tests to be redone?

hack/build.sh
davidhadas: adding `--load` works for podman
I cannot identify a difference between running with or without it.

liudalibj: docker buildx build docker image inside one container instance, eg.
```
docker buildx ls
NAME/NODE        DRIVER/ENDPOINT             STATUS  BUILDKIT             PLATFORMS
multi-builder    docker-container
  multi-builder0 unix:///var/run/docker.sock running v0.12.5              linux/amd64*, linux/s390x*, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
default *        docker
  default        default                     running v0.11.6+0a15675913b7 linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
```
Without the `--load` option, the built out image only inside the container can't be list on the host machine, so the new `docker push  ${registry}/${name}` will be failed, as it try to push image from the host.

hack/build.sh
```diff
@@ -61,8 +61,8 @@ function build_caa_payload_image() {
 		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
```
davidhadas, line 61:
```suggestion
		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
		-- load \
```

hack/build.sh
```diff
@@ -106,8 +106,8 @@ function build_caa_payload_arch_specific() {
...
```
davidhadas, line 106:
```suggestion
		--build-arg YQ_CHECKSUM="${YQ_CHECKSUM}" \
		--load \
```

hack/build.sh
davidhadas: Added --load

hack/build.sh
davidhadas: Added --load

davidhadas:
Replaced by #1774

Issue #1596, opened 2023-11-22
Attempt to update kata-runtime to point to the `main` version


stevenhorsman:
As part of the merge to main effort, we have https://github.com/kata-containers/kata-containers/pull/7046 which is adding the remote hypervisor feature to the kata runtime. Once this is merge we should test out whether the CAA can re-vendor on it and see what issues there are. We also know that as part of these changes we want to remove the gogoprotobuf workaround https://github.com/confidential-containers/cloud-api-adaptor/blob/eb1b368f84825bb83b9033f07228e04cdef3ceb1/go.mod#L162-L164 and align with kata where we had to do the same thing.

stevenhorsman:
After a lot of compilation issues, I've got it all compiling not and built a CAA OCI image from it, but when testing it doesn't work:
```
Events:
  Type     Reason                  Age                 From               Message
  ----     ------                  ----                ----               -------
  Normal   Scheduled               35m                 default-scheduler  Successfully assigned default/alpine to sh-libvirt-s390x-e2e-22-04-test-4
  Warning  FailedCreatePodSandBox  7s (x160 over 35m)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: remote hypervisor call failed: ttrpc: closed: unknown
```
so I've clearly broken something during the change

stevenhorsman:
Beraldo has recommended re-generating the hypervisor protos with ttrpc rather than grpc, so I've created a branch that has that change in kata-runtime: https://github.com/kata-containers/kata-containers/compare/main...stevenhorsman:hypervisor-ttrpc?expand=1
I'm not reworking the changes to undo some of these related to the ttrpc -> grpc change.

stevenhorsman:
I've updated my branch to use my fork of kata runtime with the ttrpc changes and I think I get further now. as when I try and start a peer pods the error is:
```
Events:
  Type     Reason                  Age                    From               Message
  ----     ------                  ----                   ----               -------
  Normal   Scheduled               7m20s                  default-scheduler  Successfully assigned default/nginx-secret-pod to peer-pods-worker-0
  Warning  FailedCreatePodSandBox  114s (x26 over 7m20s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd container: create prepare snapshot dir: failed to create temp dir: stat /var/lib/containerd-nydus/snapshots: no such file or directory: unknown
```

[...]

PR #1754, opened 2024-03-20
Switch to use kata-containers main branch

stevenhorsman:


stevenhorsman:
If anyone is interested. I've created some instructions for how to test this PR out in https://github.com/confidential-containers/cloud-api-adaptor/issues/1596#issuecomment-2066919368

There are still some things to resolve though, so it's not fully ready yet:
- Resolve the double CAA ds issue
- Improve the caching on kata-containers to allow us to pick pinned releases
- Use caches for the guest components to speed up the podvm build
- Work out how to resolve the containerd & nydus caching limitation in our CI/for development

src/cloud-api-adaptor/install/yamls/caa-pod.yaml
```diff
@@ -1,91 +0,0 @@
-apiVersion: apps/v1
```
beraldoleal, line 1: @stevenhorsman,

By removing this file won't we break the TLS and CRI runtime endpoint overlays?

If you enable any of them in the `kustomization.yaml`. it will try to find a DaemonSet with "peerpodconfig-ctrl-caa-daemon" name, but is missing within the current execution, even tho it will be managed by

```
$ kubectl apply -k src/cloud-api-adaptor/src/cloud-api-adaptor/install/overlays/libvirt/
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
error: no matches for Id DaemonSet.v1.apps/peerpodconfig-ctrl-caa-daemon.confidential-containers-system; failed to find unique target for patch DaemonSet.v1.apps/peerpodconfig-ctrl-caa-daemon.confidential-containers-system
 ```
 
 Did I missed something?

**Review by beraldoleal — COMMENTED** [review]
Hi @stevenhorsman , first of all, huge and amazing work here.

Haven't finished the review yet, but I left a comment for now, that, iiuc, might be broken.

File: src/cloud-api-adaptor/install/yamls/caa-pod.yaml (hunk 1 above)
Comment (stevenhorsman) [reply src/cloud-api-adaptor/install/yamls/caa-pod.yaml:1]: I'll take a look today. Recently I've been using the scripts I updated rather than manually, so I probably missed this error, though the fact everything still seems to work is interesting.

Comment (stevenhorsman) [reply src/cloud-api-adaptor/install/yamls/caa-pod.yaml:1]: I can't reproduce this and doing all the steps manually still don't see the error:
```
# kubectl apply -k "github.com/confidential-containers/operator/config/overlays/peerpods/default"
# kubectl set env -n confidential-containers-system deployment/cc-operator-controller-manager RELATED_IMAGE_CAA=${CAA_IMAGE}
# kubectl rollout status deploy/cc-operator-controller-manager -n confidential-containers-system
# kubectl apply -k "github.com/confidential-containers/operator/config/samples/ccruntime/peer-pods"
# kubectl apply -k install/overlays/libvirt/
serviceaccount/cloud-api-adaptor created
clusterrole.rbac.authorization.k8s.io/node-viewer created
clusterrole.rbac.authorization.k8s.io/peerpod-editor created
clusterrole.rbac.authorization.k8s.io/pod-viewer created
clusterrolebinding.rbac.authorization.k8s.io/node-viewer created
clusterrolebinding.rbac.authorization.k8s.io/peerpod-editor created
clusterrolebinding.rbac.authorization.k8s.io/pod-viewer created
configmap/peer-pods-cm created
secret/auth-json-secret created
secret/peer-pods-secret created
secret/ssh-key-secret created
```

Let's chat when you are online and see where the differences are

Comment (beraldoleal) [reply src/cloud-api-adaptor/install/yamls/caa-pod.yaml:1]: @stevenhorsman you need to enable the patch here:

https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/overlays/libvirt/kustomization.yaml#L60

Comment (stevenhorsman) [reply src/cloud-api-adaptor/install/yamls/caa-pod.yaml:1]: Beraldo has let me know that we need to uncomment the cri-runtime patch in https://github.com/confidential-containers/cloud-api-adaptor/blob/0db87d03d6887884d2c27a0b86dccc1a51bfa8f7/src/cloud-api-adaptor/install/overlays/libvirt/kustomization.yaml#L60 and with that I can now reproduce the error:
```
# kubectl apply -k install/overlays/libvirt/
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
error: no matches for Id DaemonSet.v1.apps/peerpodconfig-ctrl-caa-daemon.confidential-containers-system; failed to find unique target for patch DaemonSet.v1.apps/peerpodconfig-ctrl-caa-daemon.confidential-containers-system
```
I'm not 100% what it is needed for, but I guess we should try and resolve it if possible

File: src/cloud-api-adaptor/pkg/adaptor/proxy/proxy.go [hunk 2]
```diff
@@ -22,7 +22,7 @@ import (
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
-	criapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	criapi "k8s.io/cri-api/pkg/apis/runtime/v1"
```
beraldoleal: I guess this deserves its own commit, or if needed by this one document, in the commit message, why we are removing the alpha2.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -140,50 +83,8 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
-			if err != nil {
-				logger.Printf("PullImage fails: %v", err)
-				return nil, err
-			}
-			// kata-agent uses this annotation to fix the image bundle path
-			// https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L694-L696
-			req.OCI.Annotations[cri.ImageName] = imageName
-		}
+	if !pullImageInGuest {
+		logger.Printf("Pulling image separately not support on main")
```
beraldoleal, line 85: Could we improve a little bit this message? Maybe:

"Pulling image separately is not supported. Please consider using...."

src/csi-wrapper/pkg/wrapper/nodeservice.go
```diff
@@ -11,7 +11,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/proto/podvminfo"
+	podvminfo "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/proto/podvminfo"
```
beraldoleal, line 14: Is this necessary by this commit or deserves its own?

src/cloud-api-adaptor/Makefile
```diff
@@ -121,8 +121,8 @@ image-with-arch: .git-commit ## Build the per arch image
 .PHONY: deploy
 deploy: ## Deploy cloud-api-adaptor using the operator, according to install/overlays/$(CLOUD_PROVIDER)/kustomization.yaml file.
 ifneq ($(CLOUD_PROVIDER),)
-	kubectl apply -k "github.com/confidential-containers/operator/config/release?ref=v0.8.0"
-	kubectl apply -k "github.com/confidential-containers/operator/config/samples/ccruntime/peer-pods?ref=v0.8.0"
+	kubectl apply -k "github.com/confidential-containers/operator/config/overlays/peerpods/default"
+	kubectl apply -k "github.com/confidential-containers/operator/config/samples/ccruntime/peer-pods"
```
beraldoleal, line 130: I found some other files with `samples/ccruntime/peer-pods` reference, should we update them as well?

 * docs/Release-Process.md
 * src/cloud-api-adaptor/test/provisioner/provision.go

docs/Release-Process.md
```diff
@@ -124,9 +124,9 @@ instability and all these versions where tested in the release candidate testing
 For the cloud-api-adaptor we need to wait until the Kata Containers release tag has been created and the
```
beraldoleal, line 124: Ohh here it is one... I would suggest squashing this commit with the previous one.

src/cloud-api-adaptor/test/provisioner/provision.go
```diff
@@ -424,7 +424,7 @@ func (p *CloudAPIAdaptor) Delete(ctx context.Context, cfg *envconf.Config) error
 	}
```
beraldoleal, line 427: And here is the other... I would suggest squashing too.

src/cloud-api-adaptor/install/README.md
```diff
@@ -48,7 +48,7 @@ You can deploy the CoCo operator and cloud-api-adaptor with the `Makefile` by ru
 * `make deploy` deploys operator, runtime and cloud-api-adaptor pod in the configured cluster
     * validate kubectl is available in your `$PATH` and `$KUBECONFIG` is set
 
-> **Note:** `make delete` deletes the cloud-api-adaptor daemonset from the configured cluster (and peerpod-ctrl if RESOURCE_CTRL=true is set)
+> **Note:** `make delete` deletes the cloud-api-adaptor daemonset from the configured cluster
```
beraldoleal, line 54: nit again: those changes could be squashed with the previous commit.

src/cloud-api-adaptor/docker/README.md
```diff
@@ -80,7 +80,7 @@ For changing the CAA image to your custom built image (eg. `quay.io/myuser/cloud
 you can use the following:
 
 ```bash
-kubectl set image ds/cloud-api-adaptor-daemonset -n confidential-containers-system cloud-api-adaptor-con=quay.io/myuser/cloud-api-adaptor
+kubectl set image ds/peerpodconfig-ctrl-caa-daemon -n confidential-containers-system caa-pod=quay.io/myuser/cloud-api-adaptor
```
beraldoleal, line 83: nit: we could do all the renaming in a single commit.

src/cloud-api-adaptor/install/overlays/aws/cri_runtime_endpoint.yaml
```diff
@@ -1,15 +1,15 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: cloud-api-adaptor-daemonset
```
beraldoleal, line 4: You are deleting some files with 8338cb70f29058b32871ff6547699a5a61102142, so I guess part of this commit could be dropped.

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -163,7 +163,7 @@ func IsPulledWithNydusSnapshotter(ctx context.Context, t *testing.T, client klie
 		t.Fatal(err)
 	}
 	for _, pod := range podlist.Items {
-		if pod.Labels["app"] == "cloud-api-adaptor" && pod.Spec.NodeName == nodeName {
+		if pod.Labels["name"] == "peerpodconfig-ctrl-caa-daemon" && pod.Spec.NodeName == nodeName {
```
beraldoleal, line 167: Same here, merge with previous ones.

src/cloud-api-adaptor/go.mod
```diff
@@ -174,10 +205,6 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-// The following line is a workaround for the issue descrined in https://github.com/containerd/ttrpc/issues/62
-// We can remove this workaround when Kata stop using github.com/gogo/protobuf
-replace google.golang.org/genproto => google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8
```
beraldoleal, line 179: \o/

beraldoleal (commented):
Hi @stevenhorsman thanks again. Left a few comments, mostly nits asking for squashes.

I tested most of the things here, and are working fine. I will do one more test with the kustomization for the cri socket endpoint and report back.

src/cloud-api-adaptor/install/overlays/aws/cri_runtime_endpoint.yaml
stevenhorsman: Yeah - I was waiting for confirmation that we are fine to delete the cri_runtime_endpoint and then I'd re-order, so that that change happens first.

src/cloud-api-adaptor/Makefile
stevenhorsman: Are you proposing that docs changes, test changes and code changes for related changes are all merged in the same commit?

src/cloud-api-adaptor/pkg/adaptor/proxy/proxy.go
stevenhorsman: If we agree to remove the cri_runtime_endpoint then this commit will be restructured as we don't ned the criapi at all anymore.

src/cloud-api-adaptor/Makefile
stevenhorsman: I don't necessarily disagree - but I've already merged more commits together than the CoCo (and Kata) guidelines suggest: https://github.com/confidential-containers/confidential-containers/blob/main/CONTRIBUTING.md#commit-format 
> The "subsystem" describes the area of the code that the change applies to. It does not have to match a particular directory name in the source tree because it is a "hint" to the reader. The subsystem is generally a single word.

> If the commit touches many different subsystems, you might want to split it into multiple commits. In general more smaller commits are preferred to fewer larger ones.

src/csi-wrapper/pkg/wrapper/nodeservice.go
stevenhorsman: I've updated the commit message to explain this. If we can get the package generated then that would be even better, but I know I had great fun getting this far at the end of last year.

src/cloud-api-adaptor/Makefile
beraldoleal: > Are you proposing that docs changes, test changes and code changes for related changes are all merged in the same commit?

Yes, this would improve our bisect ability, prevent missed commits during reverts/cherry-picks, and make it easier to drop commits that don't make sense, since we don't have to look for related ones.

I know this could be controversial, but I also don't think you would be breaking the guidelines here:

> **It does not have to match** a particular directory name in the source tree because it is a "hint" to the reader.

> If the commit touches many different subsystems, you **might want** to split it into multiple commits.

The guideline *suggests* splitting commits that affect multiple subsystems. This is particularly relevant if each subsystem involves distinct changes. However, if the modifications are closely related or minor, it might be more practical to keep them together in a single commit. 

However, I understand you've put a lot of effort into this. So, feel free to ignore the nit ones. It would be great if we could adopt this in future PRs, tho.

stevenhorsman: Given that you are the only reviewer, I've updated the changes to match what you would like.

src/cloud-api-adaptor/test/provisioner/provision.go
stevenhorsman: Squashed

src/cloud-api-adaptor/install/README.md
stevenhorsman: Squashed

src/cloud-api-adaptor/docker/README.md
stevenhorsman: Done

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: Squashed

src/cloud-api-adaptor/Makefile
beraldoleal: Thanks @stevenhorsman, and sorry if I wasn't clear before, there was no *real* need for that, But appreciated.

beraldoleal:
Hi @stevenhorsman thanks again.

I could confirm yesterday that the cri patch it seems to be unnecessary indeed. Everything worked without it. So on my side this is a big LGTM.

We just need to decide what to do with the tls customization.

stevenhorsman:
> Hi @stevenhorsman thanks again.
> 
> I could confirm yesterday that the cri patch it seems to be unnecessary indeed. Everything worked without it. So on my side this is a big LGTM.
> 
> We just need to decide what to do with the tls customization.

Great, I'll tidy up the commits to remove CRI endpoint at the start now and we'll discuss the tls (and other kustomization) with @bpradipt next week I guess as the next step?

src/cloud-api-adaptor/pkg/adaptor/proxy/proxy.go
stevenhorsman: This has been removed now as I've squashed the criclient removal (which used this) into this commit.

src/cloud-api-adaptor/install/overlays/aws/cri_runtime_endpoint.yaml
stevenhorsman: The cri_runtime_endpoint kustomize delete has been moved above this commit in the stack, so this code is gone now.

src/cloud-api-adaptor/install/README.md
```diff
@@ -96,21 +112,31 @@ Alternatively the manual approach, if you want to pick a specific CoCo release/r
...
     ```
 
 * View cloud-api-adaptor logs
 
     ```
-    kubectl logs pod/cloud-api-adaptor-daemonset-mk8ln -n confidential-containers-system
+    kubectl logs ds/peerpodconfig-ctrl-caa-daemon -n confidential-containers-system
     ```
 
+## Workaround: Clean up the images on the worker
```
mkulke, line 130: we might not need this anymore due to https://github.com/containerd/nydus-snapshotter/pull/593

src/cloud-api-adaptor/test/provisioner/provision.go
```diff
@@ -559,27 +587,72 @@ func (p *CloudAPIAdaptor) Deploy(ctx context.Context, cfg *envconf.Config, props
...
-		wait.WithTimeout(time.Minute*5)); err != nil {
+	return nil
+}
+
+// This is a temporary workaround due to an issue with containerd not caching images with snapshotter labels
+// and peer-pods using the nydus-snapshotter to implement the image pull on guest. We first need to clean up images
+// that have already been cached locally. This is done with the
+// [convenience debug container](https://github.com/stevenhorsman/containerd-image-cleanup)
+// which deletes some common images (and can be expanded)
+func cleanUpContainerdImages(ctx context.Context, cfg *envconf.Config, client klient.Client) error {
```
mkulke, line 604: see other comment, maybe we don't need that cleanup logic any more.

mkulke (approved):
glanced over the code and build + deployed the branch for azure using:
- a [patched nydus](https://github.com/containerd/nydus-snapshotter/pull/593)
- a prebuilt podvm image

It worked well, great work!

src/cloud-api-adaptor/install/README.md
stevenhorsman: Yep - once the nydus-snapshotter changes are bumped and included in the operator we'll remove this. Whether we want to wait to merge this until that happens I'm not sure yet.

src/cloud-api-adaptor/install/overlays/aws/kustomization.yaml
```diff
@@ -63,7 +63,6 @@ secretGenerator:
 ##TLS_SETTINGS
 
 patchesStrategicMerge:
-  #- cri_runtime_endpoint.yaml # set (modify host's runtime cri socket path in the file, default is /run/containerd/containerd.sock)
```
wainersm, line 66: Hi @stevenhorsman ! This entire commit can be squashed on the previous (that removes the cri runtime endpoint) so in case of revert, we revert it atomically. Or do you prefer it separated?

src/cloud-api-adaptor/pkg/adaptor/proxy/proxy.go
```diff
@@ -50,32 +43,30 @@ type AgentProxy interface {
...
-	stopOnce      sync.Once
+	tlsConfig    *tlsutil.TLSConfig
+	caService    tlsutil.CAService
+	readyCh      chan struct{}
+	stopCh       chan struct{}
+	serverName   string
+	socketPath   string
+	pauseImage   string
+	proxyTimeout time.Duration
+	criTimeout   time.Duration
```
wainersm, line 54: I don't think it needs `criTimeout` anymore as it was used on `initCriClient` (function removed) to initialize the cri client.

I failed to find a user parameter to set the cri timeout so I assume it always used the `defaultCriTimeout`.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -140,50 +81,8 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
-			if err != nil {
-				logger.Printf("PullImage fails: %v", err)
-				return nil, err
-			}
-			// kata-agent uses this annotation to fix the image bundle path
-			// https://github.com/kata-containers/kata-containers/blob/8ad86e2ec9d26d2ef07f3bf794352a3fda7597e5/src/agent/src/rpc.rs#L694-L696
-			req.OCI.Annotations[cri.ImageName] = imageName
-		}
+	if !pullImageInGuest {
+		logger.Printf("Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.")
```
wainersm, line 85: hmmm... what happens if CRI-O? `pullImageInGuest` will be `false` IIUC.

src/cloud-api-adaptor/install/overlays/aws/kustomization.yaml
stevenhorsman: Yeah - as you can see from the discussions I had above with Beraldo I really struggled in this one to balance between having separate commits for different areas, like we guide in CoCo and mandate in kata-containers whilst making atomic commits (I already reduced from 50 commits to 25 as we have now). Realistically I don't see a scenario where we are likely to back out to the middle of this commit sequence, but I'm not against bending our rules to make something like a bisect easier to do. It probably means we should push back on lots of kata PRs for the same reasons though!

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -181,12 +174,9 @@ func IsPulledWithNydusSnapshotter(ctx context.Context, t *testing.T, client klie
 				if nydusSnapshotterPullRegex.MatchString(line) {
```
wainersm, line 175: @stevenhorsman were you able to pass a test that uses this `IsPulledWithNydusSnapshotter()` method? Just want to double check the `nydusSnapshotterPullRegex` didn't change on main.

src/cloud-api-adaptor/test/provisioner/provision.go
```diff
@@ -559,27 +559,6 @@ func (p *CloudAPIAdaptor) Deploy(ctx context.Context, cfg *envconf.Config, props
...
-	cmd.Dir = p.rootSrcDir
-	// Set the KUBECONFIG env var
-	cmd.Env = append(os.Environ(), fmt.Sprintf("KUBECONFIG="+cfg.KubeconfigFile()))
-	stdoutStderr, err = cmd.CombinedOutput()
-	log.Tracef("%v, output: %s", cmd, stdoutStderr)
-	if err != nil {
-		return err
-	}
-
-	// Wait for the peerpod-ctrl deployment to be ready
```
wainersm, line 574: Shouldn't it still wait the peerpod-ctrl deployment be ready? Sorry, never tried that new install method so I don't know if any other previous condition already ensure peerpod-ctrl is up and running at this point.

src/peerpod-ctrl/README.md
```diff
@@ -45,8 +44,8 @@ For any changes in the CRD/controller make sure it doesn't break the k8s api cal
 ### How it works
 This project aims to follow the Kubernetes [Operator pattern](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/)
 
-It uses [Controllers](https://kubernetes.io/docs/concepts/architecture/controller/) 
-which provides a reconcile function responsible for synchronizing resources untile the desired state is reached on the cluster 
+It uses [Controllers](https://kubernetes.io/docs/concepts/architecture/controller/)
+which provides a reconcile function responsible for synchronizing resources untile the desired state is reached on the cluster
```
wainersm, line 48: Bonus points if you fix the typo `untile` :D

src/cloud-api-adaptor/install/overlays/aws/kustomization.yaml
stevenhorsman: Done

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
stevenhorsman: So my understanding is the CRI-O still sets the storage driver to `image_guest_pull` as that is what the kata-agent is processing, so I think it should still set `pullImageInGuest` to be `true`. @littlejawa - can you confirm?

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: Yes, it seems to be the same log messages, which is what we check

src/cloud-api-adaptor/test/provisioner/provision.go
stevenhorsman: That's a fair question. My understanding (which could be wrong) is that as the peerpod-ctrl is deployed in the operator stage, not at the end, we would have expected it to be running by this point, but I don't think there is any harm to wait in case it hasn't, so I'll try and re-add that.

stevenhorsman: Thanks for raising this - after looking into it deeper I've found that the operator isn't creating the peerpod-ctrl properly with the error:
```
2024-05-14T08:12:27Z	ERROR	Reconciler error	{"controller": "peerpod", "controllerGroup": "confidentialcontainers.org", "controllerKind": "PeerPod", "PeerPod": {"name":"nginx-resource-xlctp","namespace":"default"}, "namespace": "default", "name": "nginx-resource-xlctp", "reconcileID": "399c7b13-cc9d-4aba-ade8-d332ac01d62a", "error": "libvirt cloud provider not supported"}
```
so that's something else that needs debugging

stevenhorsman: I will push the re-addition of this code with a WIP label as it shows it's important to test, but I think the deployment name we are looking for will be likely change.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
littlejawa: Yes,  `image_guest_pull` is set by CRI-O, so we should be good here

src/cloud-api-adaptor/hack/update-proto.sh
```diff
@@ -2,10 +2,10 @@
...
+PODMVINFO_PATH="proto/podvminfo"
 
-cd "$basedir/proto"
-
-protoc  --gogottrpc_out=. \
-        --gogottrpc_opt=plugins=ttrpc+fieldpath,paths=source_relative \
-        podvminfo/podvminfo.proto
+protoc \
+    --proto_path=$PODMVINFO_PATH \
+    --go_out=$GOPATH/src \
```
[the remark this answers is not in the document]
stevenhorsman: I think so (I made these changes in November/December so can't remember fully). I'm not a good person to chat protobuf as I did all the development of this by trial-and-error if I'm honest!

src/cloud-api-adaptor/test/provisioner/provision.go
stevenhorsman: I'm not sure if there is anyway we can reasonably test for this with the e2e tests though?

src/cloud-api-adaptor/install/yamls/caa-pod.yaml
```diff
@@ -1,91 +0,0 @@
...
-          readOnly: true
-        - mountPath: /root/.ssh/
-          name: ssh
-          readOnly: true
-        - mountPath: /run/peerpod
-          name: pods-dir
-        - mountPath: /run/netns
-          mountPropagation: HostToContainer
-          name: netns
-        # # setting for cloud provider external plugin
```
wainersm, line 62: Hi @stevenhorsman !

Dynamically loading an external cloud provider was introduced by  @liudalibj in commit 549673b5.

I couldn't find a similar configuration on peer-pods's new controller in the operator, but I assume we can still set that settings via `kubectl edit`. Then I started worrying about this dynamically loading feature; in how many ways the new controller breaks it? :)

We might need to update the docs in src/cloud-api-adaptor/docs/addnewprovider.md too.

stevenhorsman: Very good point - this is linked in with the concerns that Beraldo raised about the tls volume mounts - it's a bunch of stuff that with the switch to use the operator (so nothing to do with moving to kata `main`) has been broken and the user experience isn't properly thought about/defined :(

src/cloud-api-adaptor/libvirt/install_operator.sh
```diff
@@ -15,6 +15,7 @@ LIBVIRT_USER="${LIBVIRT_USER:-$USER}"
 LIBVIRT_NET="${LIBVIRT_NET:-default}"
 LIBVIRT_POOL="${LIBVIRT_POOL:-default}"
 SSH_KEY_FILE="${SSH_KEY_FILE:-id_rsa}"
+CAA_IMAGE="${CAA_IMAGE:-}"
```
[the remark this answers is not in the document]
stevenhorsman: Sorry, just to clarify does `CAA_IMAGE	?= "quay.io/confidential-containers/cloud-api-adaptor:latest"` not work as I expected? I was hoping that if `CAA_IMAGE="${CAA_IMAGE:-}"` was not set in `install_operator.sh` then the make would put in the default, but it sounds like `CAA_IMAGE="${CAA_IMAGE:-}"` prevents that? That's annoying to need the default in multiple places, but I don't have a better idea at the moment

src/cloud-api-adaptor/test/provisioner/provision.go
```diff
@@ -536,6 +536,30 @@ func (p *CloudAPIAdaptor) Deploy(ctx context.Context, cfg *envconf.Config, props
 		return err
 	}
 
+	// TODO can we do this better with go code?
```
wainersm, line 543: I think so. I just found this example https://github.com/kubernetes/client-go/blob/master/examples/create-update-delete-deployment/main.go#L111 of using client-go to update a deployment.

src/cloud-api-adaptor/podvm/Makefile.inc
```diff
@@ -135,6 +135,7 @@ $(KATA_AGENT): $(FORCE_TARGET) | $(STATIC_LIBSECCOMP)
 	install -D --compare "$(KATA_AGENT_SRC)/target/$(RUST_TARGET)/$(KATA_AGENT_BUILD_TYPE)/$(@F)" "$@"
 
 $(STATIC_LIBSECCOMP): $(KATA_CONTAINERS_SRC)
+	INSTALL_IN_GOPATH=false $(KATA_CONTAINERS_SRC)/ci/install_yq.sh
```
wainersm, line 138: @stevenhorsman `src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder` installs yq 4.35.1 whereas kata's 3.4.1. Will it overwrite 4.35.1 one?

Yeah, we talked about yq on peer-pods many times but my brain cleaned up these completely.

stevenhorsman: It will, but Magnus wrote https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/hack/yq-shim.sh so we can handle both versions 😃

wainersm: oh, cool!

mkulke (approved):

liudalibj (approved):
Thanks @stevenhorsman I tested this PR in my x86/s390x cluster, they worked as expected, thanks!
Here the images I built out:
```
quay.io/liudalibj/podvm-generic-ubuntu-s390x:bcc4c405f4aa5ac4526f508a4bdb3e8a86f4d6500d4eb56337944b1be83423a1
quay.io/liudalibj/podvm-generic-ubuntu-amd64:bcc4c405f4aa5ac4526f508a4bdb3e8a86f4d6500d4eb56337944b1be83423a1
quay.io/liudalibj/podvm-binaries-ubuntu-s390x:bcc4c405f4aa5ac4526f508a4bdb3e8a86f4d6500d4eb56337944b1be83423a1
quay.io/liudalibj/podvm-binaries-ubuntu-amd64:bcc4c405f4aa5ac4526f508a4bdb3e8a86f4d6500d4eb56337944b1be83423a1
quay.io/liudalibj/podvm-builder-ubuntu:bcc4c405f4aa5ac4526f508a4bdb3e8a86f4d6500d4eb56337944b1be83423a1
quay.io/liudalibj/cloud-api-adaptor:dev-6c395be8927835b6e5e0a59dfab891abedb4893b
```

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
wainersm: Thanks for the confirmation @littlejawa !

wainersm (approved):
HI @stevenhorsman ! Re-reviewed, my eyes I couldn't spot any error. Great work, thanks once again!

PR #1759, opened 2024-03-25
pin packer plugins versions

snir911:
Reasons:
* avoid unexpected bugs (there's one in latest azure plugin)
* avoid license changes
* packer usage for podvm creation is expected to be deprecated soon


bpradipt (approved):
/lgtm

beraldoleal (commented):
Hi @snir911 those lgtm.

Do you mind just copying the PR description into the git log message? So we can easily understand why we are pinning by just looking at the git messages?

huoqifeng (approved):
LGTM

Issue #1760, opened 2024-03-27
Handling the podFailing scenarios


chathuryaadapa:
In Few test case scenarios the desired output is the pod to fail, In such scenarios there is no code in [src/cloud-api-adaptor/test/e2e/assessment_runner.go](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/test/e2e/assessment_runner.go#L402) to cross check the reason for pod failing

PR #1761, opened 2024-03-27
PodFailing scenario: updated assessment_runner.go

chathuryaadapa:
Added code to handle the pod fail scenario by cross checking the error message we pass in .WithFailReason(reason string)

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/1760

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/assessment_runner.go
+++ b/src/cloud-api-adaptor/test/e2e/assessment_runner.go
@@ -6,6 +6,7 @@ package e2e
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -78,6 +79,7 @@ type TestCase struct {
 	deletionWithin       time.Duration
 	testInstanceTypes    InstanceValidatorFunctions
 	isNydusSnapshotter   bool
+	FailReason           string
 }
 
 func (tc *TestCase) WithConfigMap(configMap *v1.ConfigMap) *TestCase {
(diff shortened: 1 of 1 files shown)


src/cloud-api-adaptor/test/e2e/assessment_runner.go
```diff
@@ -416,6 +423,25 @@ func (tc *TestCase) Run() {
...
+						}
+						if strings.Contains(string(jsonData), tc.FailReason) {
+							log.Infof("failed due to expected reason %s", tc.FailReason)
+						} else {
+							t.Fatal(err)
+						}
+					} else {
+						log.Infof("Pod Failed If you want to cross check please give .WithFailReason(error string)")
+					}
+					tc.assert.HasPodVM(t, tc.pod.Name)
```
genjuro214, line 442: If pod is not running, the Pod VM may be not created in some case. Is it necessary to check Pod VM existence here?

stevenhorsman:
The commit message check is failing as your commit header should start with `tests/e2e` and be 50 chars long.
I think you can also squash the commits together as they are all tightly related.
Let me know if you hit any problems with this. Thanks!

stevenhorsman (approved):
I've not tested it locally, and don't know the specific cases that failed, but I ran the libvirt e2e tests and it passes still and the code looks ok

liudalibj (approved):
LGTM

liudalibj:
@chathuryaadapa would you like squash the commits together? after that we are ready to merge this PR.

PR #1762, opened 2024-03-27
podvm-mkosi: fix Architecture cannot be used in [Match]

genjuro214:
When build x86 podvm image with mkosi, we may hit following error:
```
Architecture cannot be used in [Match]
```

We depend on nix to setup mkosi build environment. Current mkosi version in nix is `17.1`:
```
# nix develop ..#podvm-mkosi --command mkosi --version
warning: Git tree '/root/cloud-api-adaptor' is dirty
mkosi 17.1
```

According to the v17.1 doc, Architecture in [Match] has not been supported in this version:
https://github.com/systemd/mkosi/blob/v17.1/mkosi/resources/mkosi.md

If upgrading mkosi version in nix, we may hit some other new errors. So I'd like to keep mkosi version and use flag files instead to fix the Architecture error.

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Makefile
+++ b/src/cloud-api-adaptor/podvm-mkosi/Makefile
@@ -45,26 +45,31 @@ binaries:
 PHONY: image
 image:
 	@echo "Enabling production preset..."
-	rm -rf resources/buildDebugImage
+	rm -rf resources/build*Image
 	rm -rf ./build
 	@echo "Building image..."
 ifeq ($(ARCH),s390x)
+	touch resources/buildS390xImage
 	mkosi --profile production.conf
 	../hack/build-s390x-image.sh
 else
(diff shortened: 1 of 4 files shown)


huoqifeng:
As we discussed on the meeting yesterday, we'll try to get this PR reviewed and merged for 0.8.1 release to fix the x86 problem. And we'll create a new issue to align the approach for s390x and x86. @liudalibj @genjuro214 @stevenhorsman @bpradipt

huoqifeng (approved):
LGTM, thanks @genjuro214 
I don't have a env to test it but I think it's OK to use:
`PathExists=...` instead `Architecture` in [March] because `Architecture` is not supported in mkosi introduced in nix 17.1 while the image built via latest nix does not work either.

liudalibj (approved):
LGTM

Issue #1792, opened 2024-04-12
Revert #1774


stevenhorsman:
Pr #1774 was delivered during release freeze and after a lot of debugging on the release, we found that it breaks the image publish and the push doesn't contain the tags, so it tried to push latest:
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/8664530199/job/23765073592
```
Using default tag: latest
The push refers to repository [quay.io/confidential-containers/cloud-api-adaptor]
tag does not exist: quay.io/confidential-containers/cloud-api-adaptor:latest
```

Revert it and then cut a new release.

liudalibj:
@stevenhorsman ah, sorry for I forget that we are in release freeze window, won't merge any PRs after we finish the release process.

PR #1793, opened 2024-04-12
Revert "Build: Update build.sh"

stevenhorsman:
This reverts commit f1fd4d9da8d1a6353d208ed6ed5f44649ff5a8b1.

Fixes: #1792


beraldoleal (approved):
LGTM

portersrc:
lgtm

wainersm (approved):
LGTM. It seems that we still need to push to `${registry}/${name}`

stevenhorsman:
> LGTM. It seems that we still need to push to `${registry}/${name}`

I think it might be part of the `tag_string` (with the `-t ` removed) rather than `${registry}/${name}` as that wasn't created by the buildx step before

PR #1801, opened 2024-04-15
gh: adding a freeze window workflow

beraldoleal:
Maintainers can enable/disable this workflow via web interface. When enabled all PRs will have at least one test failing.

### DIFF
--- a/.github/workflows/freeze-window.yaml
+++ b/.github/workflows/freeze-window.yaml
@@ -1,0 +1,16 @@
+name: Freeze window
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  freeze:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Blocking PRs
+      run: |
+        echo "We are in a freeze-window, this PR should not be merged."
+        echo "If you believe we are not, please contact the maintainers "
+        echo "and ask to disable this workflow."
+        exit 1


stevenhorsman:
Hey Beraldo, I just want to fully understand the proposal so if it works out we can add it to the release process docs and there are a couple of questions/concerns I have:
- For getting the release related PRs (version bumps etc) through is the intention that a maintainer would disable, merge that PR and then re-enable this workflow?
- Is this supposed to be enforcement, or just a reminder? If we don't make this check required then it won't block PRs merging, but if we do make it required then it blocks merging when it's enabled and fails, but also blocks when it doesn't run. A reminder if better than nothing, but I just wanted to check what you had envisaged.

stevenhorsman:
The other approach I was thinking about is whether we could have a freeze label, and then a permanent job that checks if freeze, or WIP is labelled and that could be required (kata has this step up). I think using the gh api we could probably easily write workflows to label & unlabel all current PRs with this, but I don't have a good idea for how we would set it up so a new PR created during the freeze period would get labelled. I thought I would share in case others have thoughts though.

stevenhorsman:
> We could use this same PR as example. I guess one test is failing but I'm assuming you could merge it anyway, right?

This is the point I was trying to explain - I don't think it will block merging unless it's a required job. At the moment we can merge this due to not having 2 reviews, not the failing test:

image

beraldoleal:
> > We could use this same PR as example. I guess one test is failing but I'm assuming you could merge it anyway, right?
> 
> This is the point I was trying to explain - I don't think it will block merging unless it's a required job. At the moment we can merge this due to not having 2 reviews, not the failing test:
> image

Humm I see now. Yes GH removed the required checks feature, and looks like we can do that via project/repository settings using rulesets, but I don't have permissions to see that here.

stevenhorsman:
> Humm I see now. Yes GH removed the required checks feature, and looks like we can do that via project/repository settings using rulesets, but I don't have permissions to see that here.

Yes - I can make it a required check, but that means it will have to run even when we aren't in freeze and block otherwise (we have similar issues in kata-containers around this)

stevenhorsman (approved):

PR #1805, opened 2024-04-17
Bump google.golang.org/protobuf 1.33

stevenhorsman:
We have multiple security warnings for:
> Golang protojson.Unmarshal function infinite loop when unmarshaling certain forms of invalid JSON CVE-2024-24786

So we should try and bump them to see if we can resolve these without causing any other issues


liudalibj (approved):
LGTM, thanks @stevenhorsman

beraldoleal (approved):

PR #1806, opened 2024-04-17
libvirt: fix default IP logic

beraldoleal:
If we have a default IP address the logic that will calculate the IP based on the libvirt network will never be executed.

### DIFF
--- a/src/cloud-api-adaptor/libvirt/README.md
+++ b/src/cloud-api-adaptor/libvirt/README.md
@@ -100,11 +100,14 @@ Allocation:     631.52 MiB
 
 ## Install and configure Confidential Containers and cloud-api-adaptor in the cluster
 
-The easiest way to install the cloud-api-adaptor along with Confidential Containers in the cluster is through the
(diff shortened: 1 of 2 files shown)


stevenhorsman:
Does this also mean we can remove the line in the doc: https://github.com/confidential-containers/cloud-api-adaptor/blob/fd1cc099903328065aea6ebf91ea0ca6807b391e/src/cloud-api-adaptor/libvirt/README.md?plain=1#L99 that sets it and let the code figure it out itself?

beraldoleal:
> Does this also mean we can remove the line in the doc:
> 
> https://github.com/confidential-containers/cloud-api-adaptor/blob/fd1cc099903328065aea6ebf91ea0ca6807b391e/src/cloud-api-adaptor/libvirt/README.md?plain=1#L99
> that sets it and let the code figure it out itself?

Yes, just fixed with the force-push.

stevenhorsman (approved):
LGTM. Thanks

wainersm (approved):
Thanks @beraldoleal !

wainersm:
re-running the tests

PR #1814, opened 2024-04-22
peerpodconfig-ctrl: Handle updating the CAA DS spec

bpradipt:
The CAA daemonset spec was not getting updated
as the code never updated the spec post creation


stevenhorsman (approved):
I tested this change out with Pradipta's built operator that pointed to this and it worked correctly. Thanks @bpradipt!

huoqifeng (approved):
LGTM, thanks @bpradipt and thanks @stevenhorsman tried it out.

PR #1822, opened 2024-04-26
libvirt: update script to config libvirt on RHEL.

wyuany:
Update script to config libvirt not only work with Ubuntu but also RHEL 

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/1821

PR #1826, opened 2024-04-29
libvirt: Fix config_libvirt to not exit

stevenhorsman:
https://github.com/confidential-containers/cloud-api-adaptor/pull/1822 broke the e2e libvirt tests jobs (https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/8873277040/job/24359701508). We should have added the test_e2e_libvirt label, so sorry for missing that during the review.

config_libvirt.sh has `set -o errexit`, so we can't use a grep for rhel and test if it fails as this will exit the script. Instead use the ID field from `/etc/os-release` or `/usr/lib/os-release`

### DIFF
--- a/src/cloud-api-adaptor/libvirt/config_libvirt.sh
+++ b/src/cloud-api-adaptor/libvirt/config_libvirt.sh
@@ -9,11 +9,11 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source /etc/os-release || source /usr/lib/os-release
 ARCH=$(uname -m)
 TARGET_ARCH=${ARCH/x86_64/amd64}
 OS_DISTRO=ubuntu
-cat /etc/os-release | grep "^ID=" | grep rhel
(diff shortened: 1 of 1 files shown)


huoqifeng:
Tried it on ubuntu, looks good to me. Thanks @stevenhorsman figure out the problem!
```
# source /etc/os-release
# echo $ID
ubuntu
```

huoqifeng (approved):
LGTM

wainersm:
> /etc/os-release

And in fedora:
```
$ echo $(source /etc/os-release; echo $ID)
fedora
```

Many scripts in kata will look for two locations like `source /etc/os-release || source /usr/lib/os-release`. Apparently some distributions rely on `/usr/lib/os-release` only. Anyway, I don't think we should be paranoid here, saying just in case it rings any bell.

wainersm (approved):
LGTM. Thanks @stevenhorsman !

PR #1828, opened 2024-05-01
workflows: e2e: Add the CAA image setup

stevenhorsman:
- In preparation for the switch to use the latest operator we need to add code that will pass the CAA Image in to the e2e test properties file and try and work with both the new and old operator so that the workflow tests can pass pulled from main by the `pull_request_target` trigger
- In a future PR we will remove the old kustomize versions

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -10,6 +10,9 @@ on:
       podvm_image:
         required: true
         type: string
+      caa_image:
+        required: true
+        type: string
       install_directory_artifact:
         description: The archive name of the install directory
         default: ''
@@ -79,6 +82,7 @@ jobs:
       - name: Config Libvirt
         run: |
           ./libvirt/config_libvirt.sh
+          echo "CAA_IMAGE=\"${{ inputs.caa_image }}\"" >> libvirt.properties
           # For debugging
           cat libvirt.properties
 
@@ -105,6 +109,9 @@ jobs:
 
           echo "::group::cloud-api-adaptor logs"
           kubectl logs -l app=cloud-api-adaptor -n confidential-containers-system
+          echo "::endgroup::"
+          echo "::group::peerpodconfig-ctrl-caa-daemon logs"
+          kubectl logs ds/peerpodconfig-ctrl-caa-daemon -n confidential-containers-system
           echo "::endgroup::"
 
           for pod in $(kubectl get pods -o name 2>/dev/null); do
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -125,7 +125,6 @@ jobs:
           providers=(${{ env.PROVIDERS }})
           # If there aren't providers then something is wrong
           [[ ${#providers[@]} -gt 0 ]] || exit 1
-
           for provider in ${providers[@]}; do
             img="${E2E_IMG_REGISTRY}/cloud-api-adaptor"
             tag="${E2E_IMG_RELEASE_TAG}"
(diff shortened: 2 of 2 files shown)


stevenhorsman:
Unfortunately this can't actually we tested until it has merged as we use the `pull_request_target` trigger and I can't run it on my fork as it needs the azure runner from garm, but I've tried to make it backwards and forwards compatible and isolated it from the other changes to make it as easy to revert/update as possible if it does have issues.

.github/workflows/e2e_libvirt.yaml
```diff
@@ -79,6 +82,7 @@ jobs:
       - name: Config Libvirt
         run: |
           ./libvirt/config_libvirt.sh
+          echo "CAA_IMAGE=\"${{ inputs.caa_image }}\"" >> libvirt.properties
```
beraldoleal, line 85: Hi @stevenhorsman, maybe a dump question, but as I'm learning the project, I'm finding hard to understand why `libvirt.properties`?

Will this work only for libvirt? Should we use something like `$TEST_PROVISION_FILE` for a more agnostic approach?

stevenhorsman: So this is specifically the libvirt e2e tests, so `libvirt.properties` is the file that config_libvirt.sh sets some info too and I'm just adding to it: https://github.com/confidential-containers/cloud-api-adaptor/blob/0db87d03d6887884d2c27a0b86dccc1a51bfa8f7/src/cloud-api-adaptor/libvirt/config_libvirt.sh#L115-L117
We then set `export TEST_PROVISION_FILE="$PWD/libvirt.properties"` later in this workflow.

beraldoleal: Got it. thanks!

wainersm, line 85: As `inputs.caa_image` is non-required input then it might end up writing `CAA_IMAGE=` (empty). I don't think we will be testing a case where CAA_IMAGE is empty, will we? If not, I suggest to make `inputs.caa_image` required.

.github/workflows/e2e_libvirt.yaml
```diff
@@ -105,6 +109,7 @@ jobs:
 
           echo "::group::cloud-api-adaptor logs"
           kubectl logs -l app=cloud-api-adaptor -n confidential-containers-system
+          kubectl logs ds/peerpodconfig-ctrl-caa-daemon -n confidential-containers-system
```
wainersm, line 112: could you put this in a new `::group::`?

wainersm:
Hi @stevenhorsman !

As I said on slack I prefer this approach better to test the swap to main. I left only two small comments, overall looks good.

.github/workflows/e2e_libvirt.yaml
stevenhorsman: Yep, that makes sense.

.github/workflows/e2e_libvirt.yaml
stevenhorsman: Updating now.

stevenhorsman:
> Hi @stevenhorsman !
> 
> As I said on slack I prefer this approach better to test the swap to main. I left only two small comments, overall looks good.

Thanks for the review. I've updated based on your changed. Obviously it's a bit tricky to test these workflow changes, but hopefully this straddles the lines between the new and old worlds so we can get both tested. It might take more than one attempt though!

wainersm:
> > Hi @stevenhorsman !
> > As I said on slack I prefer this approach better to test the swap to main. I left only two small comments, overall looks good.
> 
> Thanks for the review. I've updated based on your changed. Obviously it's a bit tricky to test these workflow changes, but hopefully this straddles the lines between the new and old worlds so we can get both tested. It might take more than one attempt though!

yep, worst case nightly will be broken a few days before we manage to merge a fix. My bet is it will work though!

wainersm (approved):
Thanks @stevenhorsman !

stevenhorsman:
The nightly libvirt tests still passed after this merged (https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/9073715897), so part 1 of the success is done at least!

Issue #1825, opened 2024-04-28
Add test case for libvirt secret retrieve for attestation based on sample TEE


huoqifeng:
Add test case for libvirt secret retrieve for attestation based on sample TEE
- [x] Add test cases for libvirt https://github.com/confidential-containers/cloud-api-adaptor/pull/1824/
- [x] Use cc_kbc when build binaries
- [ ] Add github action for libvirt

PR #1840, opened 2024-05-14
libvirt: enable e2e test for attestation case in gha

huoqifeng:
Fixes: #1825

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -51,6 +51,9 @@ jobs:
       - name: Read properties from versions.yaml
         run: |
           sudo snap install yq
+          echo "KBS_REPO=$(yq -e '.git.kbs.url' versions.yaml)" >> "$GITHUB_ENV"
+          echo "KBS_VERSION=$(yq -e '.git.kbs.reference' versions.yaml)" >> "$GITHUB_ENV"
+          echo "RUST_VERSION=$(yq -e '.tools.rust' versions.yaml)" >> "$GITHUB_ENV"
           go_version="$(yq '.tools.golang' versions.yaml)"
           [ -n "$go_version" ]
           echo "GO_VERSION=${go_version}" >> "$GITHUB_ENV"
@@ -86,10 +89,30 @@ jobs:
           # For debugging
           cat libvirt.properties
 
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+          override: true
+          profile: minimal
+
+      - name: Checkout kbs Repository and build kbs-client
+        run: |
+          git clone "${KBS_REPO}" test/trustee
+          pushd test/trustee
+          git checkout "${KBS_VERSION}"
+          pushd kbs
+          make CLI_FEATURES=sample_only cli
+          popd
+          # For debugging
+          ls ./target/release
+          popd
+
       - name: run tests
         id: runTests
         run: |
           export CLOUD_PROVIDER=libvirt
+          export DEPLOY_KBS=true
           export TEST_PROVISION="yes"
           export TEST_TEARDOWN="no"
           export TEST_PROVISION_FILE="$PWD/libvirt.properties"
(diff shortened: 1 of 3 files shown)


huoqifeng:
A test is running in: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/9075730155

liudalibj (approved):
LGTM

.github/workflows/e2e_libvirt.yaml
```diff
@@ -51,6 +51,8 @@ jobs:
       - name: Read properties from versions.yaml
         run: |
           sudo snap install yq
+          echo "KBS_VERSION=$(yq -e '.git.kbs.reference' versions.yaml)" >> "$GITHUB_ENV"
+          echo "RUST_VERSION=$(yq -e '.tools.rust' versions.yaml)" >> "$GITHUB_ENV"
```
stevenhorsman, line 55: Should we also add:
```
echo "KBS_REPO=$(yq -e '.git.kbs. url' versions.yaml)" >> "$GITHUB_ENV"
```
and then use that below rather than hard-code the url?

.github/workflows/e2e_libvirt.yaml
```diff
@@ -86,10 +88,30 @@ jobs:
...
+
+      - name: Checkout kbs Repository and build kbs-client
+        run: |
+          git clone https://github.com/confidential-containers/trustee test/trustee
+          pushd test/trustee
+          git checkout "${KBS_VERSION}"
+          pushd kbs
+          make CLI_FEATURES=sample_only cli
+          popd
+          ls ./target/release
```
stevenhorsman, line 107: I guess this is just for debug?

src/cloud-api-adaptor/versions.yaml
```diff
@@ -21,7 +21,7 @@ cloudimg:
 tools:
   bats: 1.10.0
   golang: 1.21.10
-  rust: 1.72.0
+  rust: 1.74.0
```
stevenhorsman, line 24: This is a slighty tricky one as we now have rust 1.74 for building the KBS client and 1.72 for building the kata-agent. I don't really like the discrepancy between the version that we use and upstream kata does, but as long as it works I'm ok with it as we plan to use the kata-agent cache soon which should remove that.

.github/workflows/e2e_libvirt.yaml
huoqifeng: Yes, I'll add a comments here.

.github/workflows/e2e_libvirt.yaml
huoqifeng: Agreed.

stevenhorsman (approved):
The code looks good to me. I guess we have an issue with testing it without merging, so we might just need to merge then monitor the tests and fix/revert if there are any issues?

huoqifeng:
> The code looks good to me. I guess we have an issue with testing it without merging, so we might just need to merge then monitor the tests and fix/revert if there are any issues?

I'm using the branch here to test the image built from this PR https://github.com/confidential-containers/cloud-api-adaptor/blob/attestation-gha-test/.github/workflows/e2e_libvirt.yaml, I'll monitor the result before merge.

stevenhorsman:
> > The code looks good to me. I guess we have an issue with testing it without merging, so we might just need to merge then monitor the tests and fix/revert if there are any issues?
> 
> I'm using the branch here to test the image built from this PR https://github.com/confidential-containers/cloud-api-adaptor/blob/attestation-gha-test/.github/workflows/e2e_libvirt.yaml, I'll monitor the result before merge.

Nice - I didn't realise that we had a workload dispatch trigger on the job to pick the branch.

huoqifeng:
The job failed, I downloaded this podvm image from the OCI image ghcr.io/confidential-containers/podvm-generic-ubuntu-amd64:ci-pr1840 and tried locally, got error like below.
```
May 14 10:15:58 podvm-busybox-dd5d43ba kata-agent[847]: [2024-05-14T10:15:58Z ERROR attestation_agent::rpc::getresource::ttrpc] Call AA-KBC to get resource failed: AA does not support the given KBC module! Module: cc_kbc
May 14 10:16:00 podvm-busybox-dd5d43ba kata-agent[847]: [2024-05-14T10:16:00Z ERROR attestation_agent::rpc::getresource::ttrpc] Call AA-KBC to get resource failed: AA does not support the given KBC module! Module: cc_kbc
```
So, something wrong when build the kata-agent and binaries when build podvm image in this PR, will check it...

PR #1850, opened 2024-05-31
cloud-config: streamline credentials provisioning

mkulke:
Instead of nesting it into a daemon.json property for the forwarder the auth file is being moved to its own entry, which will simplify the logic and allow cloud-init based podvms to use authenticated registries without running process-user-data alongside cloud-init.

### DIFF
--- a/src/cloud-api-adaptor/pkg/userdata/provision.go
+++ b/src/cloud-api-adaptor/pkg/userdata/provision.go
@@ -2,7 +2,6 @@ package userdata
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"log"
 	"os"
@@ -10,7 +9,6 @@ import (
 	"time"
 
 	"github.com/avast/retry-go/v4"
-	daemon "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/forwarder"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers/aws"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers/azure"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers/docker"
@@ -21,10 +19,10 @@ var logger = log.New(log.Writer(), "[userdata/provision] ", log.LstdFlags|log.Lm
 
 type paths struct {
 	aaConfig     string
+	agentConfig  string
 	authJson     string
 	cdhConfig    string
 	daemonConfig string
-	agentConfig  string
 }
 
 type Config struct {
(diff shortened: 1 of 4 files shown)


mkulke:
according to the discussion on the linked guest-component issue, we might end up having to add an `image_registry_auth_file` param back to kata to make this work, since there are provisions to change the path in image-rs's `ImageConfig` and unless we want to go around and sneak in an override path via env or something, we'd have to specify it at the owner of the `ImageConfig`, which is kata agent.

src/cloud-api-adaptor/pkg/agent/update.go
```diff
@@ -130,25 +129,6 @@ func UpdateConfig(cfg *Config) error {
...
-		if _, err := os.Stat(cfg.authJsonPath); err != nil && os.IsNotExist(err) {
-			// Write the authJson to the defaultAuthJsonFilePath
-			err = os.WriteFile(cfg.authJsonPath, []byte(config.AuthJson), 0644)
-			if err != nil {
-				return fmt.Errorf("failed to write auth.json file: %s", err)
-			}
-		}
-
-		// Update the file path in the agent config
-		agentConfig.ImageRegistryAuthFile = "file://" + cfg.authJsonPath
```
stevenhorsman, line 148: Hey Magnus, whilst I appreciate that the `image_registry_auth_file` isn't in kata-containers `main` branch (yet?). I was wondering if you had an alternative approach to get to the config auth json set as the config.file_path in image-rs without this?

stevenhorsman (commented):
I'm happy with moving the destination path for auth.json and removing it out of daemon.json, but I have a question on how we might be able to use it in image-rs

src/cloud-api-adaptor/pkg/agent/update.go
mkulke: Ah, if we are able to this to kata main, I'd re-add a similar approach. I just removed it because at the moment this doesn't work and it's potentially confusing.

mkulke:
some changes need to be applied once #1933 has been merged

stevenhorsman (approved):
Code LGTM. Thanks!

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -33,9 +31,17 @@ type Config struct {
 }
 
 func NewConfig(aaConfigPath, agentConfig, authJsonPath, daemonConfigPath, cdhConfig string, fetchTimeout int) *Config {
-	cfgPaths := paths{aaConfig: aaConfigPath, agentConfig: agentConfig, authJson: authJsonPath,
-		cdhConfig: cdhConfig, daemonConfig: daemonConfigPath}
-	return &Config{fetchTimeout, cfgPaths}
+	ps := paths{
```
liudalibj, line 34: can we make the ps paths fields order same as the type paths struct order?

bpradipt (approved):
/lgtm

huoqifeng:
@mkulke I'm wondering how was this file get provisioned?
```
SrcAuthfilePath = "/root/containers/auth.json"
```
Is this something can be introduced in initdata? Like in PR https://github.com/confidential-containers/cloud-api-adaptor/pull/1912

mkulke:
> @mkulke I'm wondering how was this file get provisioned?
> 
> ```
> SrcAuthfilePath = "/root/containers/auth.json"
> ```
> 
> Is this something can be introduced in initdata? Like in PR #1912

this is part of the kustomize install routine, an auth.json gets mounted to the CAA daemonset container at this path

Issue #1852, opened 2024-06-06
aa-kbc-params is not customized in agent-config.toml for libvirt provider in fedora


huoqifeng:
When creating libvirt PeerPod based on the fedora image on a s390x machine, which is built from mkosi.
The field "aa-kbc-params" in agent-config.toml under /run/peerpod was not customized correctly.

Which should be caused by the process-user-data. Logs looks like:

```
#journalctl -t process-user-data
Jun 06 00:11:23 podvm-busybox-b4e79cb5 process-user-data[676]: failed to send request: Get "http://169.254.169.254/metadata/instance/compute?api-version=2021-01-01": dial tcp 169.254.169.254:80: connect: network
is unreachable
Jun 06 00:11:23 podvm-busybox-b4e79cb5 process-user-data[676]: failed to send request: Get "http://169.254.169.254/latest/dynamic/instance-identity/document": dial tcp 169.254.169.254:80: connect: network is unre
achable
Jun 06 00:11:23 podvm-busybox-b4e79cb5 process-user-data[676]: Error: failed to create UserData provider: unsupported user data provider
```

After disable the "ExecStartPre" https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/files/etc/systemd/system/process-user-data.service#L12
Error like this:
```
# journalctl -t process-user-data
Jun 06 03:03:14 podvm-busybox-66cded21 process-user-data[675]: 2024/06/06 03:03:14 [agent/update] failed to read daemon config file: open /run/peerpod/daemon.json: no such file or directory
Jun 06 03:03:14 podvm-busybox-66cded21 process-user-data[675]: Error: failed to get daemon config from local file
```

The problem is on fedora, the failure in `ExecStartPre` in `process-user-data` causes `ExecStart` skipped because libvirt provider does not implement the provision API.

## Option 1
I tried and broken it into 2 services. `process-user-data-provision` and `process-user-data-update` while `process-user-data-update` depends on `cloud-final.service` because libvirt and other providers like ibmcloud uses `cloud-init` to provision user-data.
It works for libvirt provider on ubuntu because: 
1. `ExecStartPre` in `process-user-data` won't cause `ExecStart` skip on ubuntu
2. Which is `/etc/agent-config.toml` rather than `/run/peerpod/ agent-config.toml` on ubuntu
[...]

PR #1853, opened 2024-06-06
Bug: fix the aa-kbc-params error in fedora

huoqifeng:
Fixes: #1852

### DIFF
--- a/src/cloud-api-adaptor/cmd/process-user-data/main.go
+++ b/src/cloud-api-adaptor/cmd/process-user-data/main.go
@@ -19,7 +19,7 @@ const (
 	providerAzure = "azure"
 	providerAws   = "aws"
 
-	defaultAgentConfigPath = "/etc/agent-config.toml"
+	defaultAgentConfigPath = "/run/peerpod/agent-config.toml"
 	defaultAuthJsonPath    = "/run/peerpod/auth.json"
 )
 
--- a/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/initrd/mkosi.conf.d/fedora.conf
+++ b/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/initrd/mkosi.conf.d/fedora.conf
@@ -3,7 +3,7 @@ Distribution=fedora
 
 [Distribution]
 Distribution=fedora
-Release=38
+Release=39
 
 [Content]
 CleanPackageMetadata=true
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
@@ -7,9 +7,9 @@ After=netns@podns.service process-user-data-update.service
 [Service]
 Environment=CDH_CONFIG_PATH=/run/confidential-containers/cdh.toml
 ExecStartPre=mkdir -p /run/kata-containers
-ExecStart=/usr/local/bin/kata-agent --config /etc/agent-config.toml
+ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
 ExecStartPre=-umount /sys/fs/cgroup/misc
-ExecStopPost=/usr/local/bin/kata-agent-clean --config /etc/agent-config.toml
+ExecStopPost=/usr/local/bin/kata-agent-clean --config /run/peerpod/agent-config.toml
 # Now specified in the agent-config.toml Environment="KATA_AGENT_SERVER_ADDR=unix:///run/kata-containers/agent.sock"
 SyslogIdentifier=kata-agent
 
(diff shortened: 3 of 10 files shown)


huoqifeng:
@mkulke @bpradipt not sure whether it's reasonable for azure and aws provider, may you please check?

huoqifeng:
The problem is on fedora, the failure in `ExecStartPre` in `process-user-data` blocks  `ExecStart`.
I broken it into 2 services. `process-user-data-provision` and `process-user-data-update` while `process-user-data-update` depends on cloud-final.service because libvirt and other providers like ibmcloud uses cloud-init to provision user-data.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/process-user-data-provision.service
```diff
@@ -2,15 +2,14 @@
...
 [Unit]
-Description=Process user data
+Description=Process user data and provision
 After=network.target
 DefaultDependencies=no
 
 
 [Service]
 Type=oneshot
 ExecStartPre=/usr/local/bin/process-user-data provision-files
```
liudalibj, line 12: Maybe `ExecStartPre` -> `ExecStart` here? It's a single service now.

bpradipt:
@huoqifeng how was the libvirt provider working all this while with the packer created image?

src/cloud-api-adaptor/cmd/process-user-data/main.go
```diff
@@ -19,7 +19,7 @@ const (
 	providerAzure = "azure"
 	providerAws   = "aws"
 
-	defaultAgentConfigPath = "/etc/agent-config.toml"
+	defaultAgentConfigPath = "/run/peerpod/agent-config.toml"
```
liudalibj, line 22: the defaultAgentConfigPath is changed here but
`src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service` didn't use the new changed path.

huoqifeng:
> @huoqifeng how was the libvirt provider working all this while with the packer created image?

I'm not 100% sure, but I guess `ExecStartPre` failure won't cause `ExecStart` skipping on ubuntu.
Another thing I'm digging is why after the service change I made in this PR. the file `/etc/agent-config.toml` changed but not `/run/peerpod/agent-config.toml` without the main.go change. -- It worked on ubuntu image. @bpradipt

src/cloud-api-adaptor/cmd/process-user-data/main.go
huoqifeng: @liudalibj thanks for finding it, I think that's another reason why it works on ubuntu but not on fedora, fedora altered the path to `/run/peerpod/agent-config.toml` here https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/kata-agent.service.d/10-override.conf

huoqifeng: I guess we should change to use `/run/peerpod/agent-config.toml` on ubuntu also.

huoqifeng: I'll give a try after update the path...

liudalibj: Agree, so for ubuntu this file need be updated in this pr, 

https://github.com/confidential-containers/cloud-api-adaptor/pull/1853/files#diff-7a67c51609d012acb98c931da3f33e5494e6d5722e99cd4d9e294ebcaf15b8faL10

src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/initrd/mkosi.conf.d/fedora.conf
```diff
@@ -3,7 +3,7 @@ Distribution=fedora
 
 [Distribution]
 Distribution=fedora
-Release=38
+Release=39
```
mkulke, line 6: if we want to bump fedora, can we do this in a separate PR?

src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
```diff
@@ -1,15 +1,15 @@
 [Unit]
 Description=Kata Agent
 BindsTo=netns@podns.service
-Wants=process-user-data.service
-After=netns@podns.service process-user-data.service
+Wants=process-user-data-update.service
+After=netns@podns.service process-user-data-update.service
```
mkulke, line 5: regarding the idea discussed in https://github.com/confidential-containers/cloud-api-adaptor/issues/1852#issuecomment-2152135964, could we instead change the logic to provision the `agent-config.file` via cloud-config entry, so we don't have to perform any "updates"?

huoqifeng:
I guess this has been fixed by PR https://github.com/confidential-containers/cloud-api-adaptor/pull/1858

mkulke:
> I guess this has been fixed by PR #1858 @mkulke

I wouldn't expect this (at least intentionally). the above PR should not change behaviour in this regard.

huoqifeng:
Close it as it has been fixed in https://github.com/confidential-containers/cloud-api-adaptor/pull/1868

PR #1854, opened 2024-06-10
azure: Delete network interface synchronously

kartikjoshi21:
Delete network interface synchronously if vm creation is failed

Fixes: #1842


kartikjoshi21:
> there was a reason for turning this function into an async function (you can check git blame to verify), I think it also had to do with accidental resource leakage.
> 
> So we might have to have both: async delete for normal podvm termination, sync delete for cleanups after failed creation.

Let me check that why it was kept like this

mkulke (approved):
tested it w/ a non-existing image-id, nics are being removed. thx

bpradipt (approved):
/lgtm

PR #1858, opened 2024-06-11
podvm: launch guest-components as systemd units

mkulke:
Prior to this change the kata-agent was launching the guest-components programatically as child process. For CAA it is preferable to launch the processes as systemd units for improved reliability and to remove redundancy (due to the network ns requirement api-server-rest was spawned in the root ns by kata and via systemd in the podns ns by systemd)

We need to make the "update" functionality aware of the `guest_components_procs` agent setting, so it will be copied to the destination agent configuration file.

The attestation-agent unit is spawned after process-user-data, because it will consume the templated kata-agent config. The confidential-data-hub and api-server-rest units are activated by the presence of the unix sockets they connect to.

attestation-agent will keep defaulting to the kata agent config for the `aa_kbc_params` config, that's why we need to configure the agent path in the mkosi image.

The process tree on a podvm will look like this after in change:

```bash
$ pstree
systemd-+-agent-protocol----6*[{agent-protocol-}]
        |-api-server-rest---2*[{api-server-rest}]
        |-attestation-age---2*[{attestation-age}]
        |-confidential-da---2*[{confidential-da}]
        |-dbus-broker-lau---dbus-broker
        |-kata-agent-+-nginx---2*[nginx]
        |            |-pause
        |            `-5*[{kata-agent}]
        |-2*[login---bash]
        |-opa---6*[{opa}]
        |-sshd---sshd---sshd---bash---pstree
        |-systemd---(sd-pam)
        |-systemd-homed
        |-systemd-journal
        |-systemd-logind
        |-systemd-network
        |-systemd-oomd
        |-systemd-resolve
        |-systemd-udevd
        `-systemd-userdbd---3*[systemd-userwor]
 ```
 
note: at the moment the kata entry in versions.yaml is pointing at a fork, this will be toggled once the [respective PR](https://github.com/kata-containers/kata-containers/pull/9811) is merged

note 2: at the moment both kata-agent and attestation-agent start after the process-user-data oneshot unit, we'll have to see whether there will be race conditions with encrypted images and tweak the order of those units.

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/etc/agent-config.toml
+++ b/src/cloud-api-adaptor/podvm/files/etc/agent-config.toml
@@ -9,6 +9,9 @@ server_addr="unix:///run/kata-containers/agent.sock"
 # This is replaced in the makefile steps so do not set it manually
 aa_kbc_params = ""
 
+# prevent the agent from launching coco guest-components
+guest_components_procs = "none"
+
 # temp workaround for kata-containers/kata-containers#5590
 [endpoints]
 allowed = [
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Confidential Data Hub TTRPC API Server
+After=network.target process-user-data.service
+
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml
+RestartSec=1
+Restart=always
+
+[Install]
+WantedBy=multi-user.target
--- a/src/cloud-api-adaptor/versions.yaml
+++ b/src/cloud-api-adaptor/versions.yaml
@@ -31,7 +31,7 @@ git:
     reference: 277617af60c32661819c1132ffbf3db8dc6e1b9f
   kata-containers:
     url: https://github.com/kata-containers/kata-containers
-    reference: 3.5.0
+    reference: 59ff40f05484da2a462fa44f18fe95e7c8484546
   umoci:
     url: https://github.com/opencontainers/umoci
     reference: v0.4.7
--- a/src/cloud-api-adaptor/pkg/agent/test-data/sample-agent-config.toml
+++ b/src/cloud-api-adaptor/pkg/agent/test-data/sample-agent-config.toml
@@ -12,6 +12,9 @@ aa_kbc_params = ""
 # This field sets up the container registry auth 
 image_registry_auth_file="file:///etc/attestation-agent/auth.json"
 
+# Do not spawn guest components in kata agent
+guest_components_procs = "none"
+
 # temp workaround for kata-containers/kata-containers#5590
 [endpoints]
 allowed = [
--- a/src/cloud-api-adaptor/pkg/agent/update.go
+++ b/src/cloud-api-adaptor/pkg/agent/update.go
@@ -33,6 +33,7 @@ type AgentConfig struct {
 	AaKbcParams                 string    `toml:"aa_kbc_params"`
 	ImageRegistryAuthFile       string    `toml:"image_registry_auth_file"`
 	Endpoints                   Endpoints `toml:"endpoints"`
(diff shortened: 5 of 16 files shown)

### CODE REVIEW

File: src/cloud-api-adaptor/podvm/files/etc/agent-config.toml [hunk 1]
```diff
@@ -7,7 +7,10 @@ server_addr="unix:///run/kata-containers/agent.sock"
 
 # This field sets up the KBC that attestation agent uses
 # This is replaced in the makefile steps so do not set it manually
-aa_kbc_params = ""
+aa_kbc_params = "cc_kbc::null"
```
stevenhorsman, line 10: What's the reason for setting this to `cc_kbc::null` here?

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Confidential Data Hub TTRPC API Server
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml
```
stevenhorsman, line 7: Does this mean we can remove `Environment=CDH_CONFIG_PATH=/run/confidential-containers/cdh.toml` from the kata-agent.service (and potentially the code that reads it in the guest-components)?

src/cloud-api-adaptor/versions.yaml
```diff
@@ -30,8 +30,8 @@ git:
     url: https://github.com/confidential-containers/guest-components
     reference: 277617af60c32661819c1132ffbf3db8dc6e1b9f
   kata-containers:
-    url: https://github.com/kata-containers/kata-containers
-    reference: 3.5.0
+    url: https://github.com/mkulke/kata-containers
+    reference: abc704a72034248e08e60b9b3a1309b575a95014
```
stevenhorsman, line 34: Are you intending to merge with this fork? Is there anything beyond https://github.com/kata-containers/kata-containers/pull/9811 needed?

stevenhorsman (commented):
Generally looks good, just a couple of questions.

src/cloud-api-adaptor/versions.yaml
mkulke: no, i'd wait til the kata PR is merged and change it here.

src/cloud-api-adaptor/podvm/files/etc/agent-config.toml
mkulke: ah, that wasn't intentional. happens at build-time I think.

src/cloud-api-adaptor/versions.yaml
stevenhorsman: Great - I'll not wait for the power static checks to be sorted and merge it now :)

stevenhorsman (approved):
LGTM. Thanks!

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
bpradipt, line 7: @mkulke is cdh.toml provided via process-user-data ?

mkulke: yes (but this has been the case prior to this pr)

bpradipt: Earlier cdh was started by kata-agent and kata-agent had an explicit dep on process-user-data. Now that cdh is a separate service, should this unit be explicitly started after process-user-data?

bpradipt: or the attestation-agent socket monitor is enough?

mkulke: yeah, it's implicitly required since AA requires process-user-data, but yes, it doesn't hurt to add it explicitly I guess.

bpradipt (approved):
/lgtm
Thanks @mkulke

PR #1860, opened 2024-06-12
versions: Bump fedora base image versions

stevenhorsman:
- Fedora 38 has passed EoL 3 weeks ago, so we should try and bump to the next version (or possibly even 40?).
- Once `golang-fedora:1.21.11-39` is built, this will be followed up with a bump of the builder_base we use as well,
but do to the chance of breaking things I'll keep it separate for now.

### DIFF
--- a/hack/Dockerfile.golang
+++ b/hack/Dockerfile.golang
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.5-labs
 
-ARG BASE_IMAGE=registry.fedoraproject.org/fedora:38
+ARG BASE_IMAGE=registry.fedoraproject.org/fedora:39
 FROM --platform=$TARGETPLATFORM ${BASE_IMAGE} as base
 
 # DO NOT UPDATE THIS BY HAND !!
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -1,6 +1,6 @@
 ARG BUILD_TYPE=dev
 ARG BUILDER_BASE=quay.io/confidential-containers/golang-fedora:1.21.11-38
-ARG BASE=registry.fedoraproject.org/fedora:38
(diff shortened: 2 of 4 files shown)


wainersm:
Hi @stevenhorsman !

> * Fedora 38 has passed EoL 3 weeks ago, so we should try and bump to the next version (or possibly even 40?).
> 

I'm glad that you noticed that this time. But I was wondering how we could automatically detect distros reaching EOL and act on. I've found this https://github.com/marketplace/actions/xeol-end-of-life-eol-scan gh action that leverages this https://github.com/marketplace/actions/xeol-end-of-life-eol-scan tool. I liked both action and tool, I believe it can help on solving our problem. Worth an issue to track that idea?

Other than that, I check that mkosi based Fedora podvm is already on 39 (great!):

src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf 

>     * Once `golang-fedora:1.21.11-39` is built, this will be followed up with a bump of the builder_base we use as well,
>       but do to the chance of breaking things I'll keep it separate for now.

It won't make sense to run the libvirt CI as caa binary will be still built on 38 and attempted to run on 39. right?

wainersm (approved):
Thanks @stevenhorsman !

huoqifeng (approved):
LGTM
Thanks @stevenhorsman

mkulke (approved):

stevenhorsman:
> But I was wondering how we could automatically detect distros reaching EOL and act on. I've found this https://github.com/marketplace/actions/xeol-end-of-life-eol-scan gh action that leverages this https://github.com/marketplace/actions/xeol-end-of-life-eol-scan tool. I liked both action and tool, I believe it can help on solving our problem. Worth an issue to track that idea?

Definitely worth an issue to track this. Thanks for the suggestion!

PR #1868, opened 2024-06-17
attestation-agent-config: generate attestation-agent config when createVM instance

huoqifeng:
Fixes: #1852 

- Generate the attestation-agent toml file aa.toml when aaKBCParams provided
- Use the cfg file to start attestation agent service when it exists
- Start attestation agent service directly when no cfg file exists
- remove aa_kbc_params in agent-config so that cdh won't read from it (this is tricky issue in CDH)
- rename package `agent` to `aa` to reflect it's for attestation agent config
- certs field will be added in aa config in https://github.com/confidential-containers/cloud-api-adaptor/pull/1875 after we tried it out
    
    Signed-off-by: Qi Feng Huo <huoqif@cn.ibm.com>


huoqifeng:
@mkulke @stevenhorsman @liudalibj @bpradipt This is still in draft, I need some time to verify the code. Idea is to generate the agent-config.toml similar as cdh.toml. described in https://github.com/confidential-containers/cloud-api-adaptor/issues/1852

src/cloud-api-adaptor/pkg/agent/update_test.go
```diff
@@ -1,227 +0,0 @@
-package agent
```
stevenhorsman, line 1: Would any of the unit tests from here be useful to test agent/config.go - not to do the update but to check the config file is created as expected?

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -270,6 +271,13 @@ func (s *cloudService) CreateVM(ctx context.Context, req *pb.CreateVMRequest) (r
 		},
 	}
 
+	if daemonConfig.AuthJson != "" {
```
mkulke, line 274: I think  we also want to drop the authjson entry from the daemon.json I think, if we want to propagate it directly.

mkulke (commented):
looks good in general. however, we were discussing in slack that we probably don't need to template the agent-config file anymore, after we have a config file for AA, since all the fields but `AAKBCParams` (which we wouldn't need anymore) are static and can be set at buildtime. what do you think?

huoqifeng:
> looks good in general. however, we were discussing in slack that we probably don't need to template the agent-config file anymore, after we have a config file for AA, since all the fields but `AAKBCParams` (which we wouldn't need anymore) are static and can be set at buildtime. what do you think?

@mkulke because `AAKBCParams` is a dynamic endpoint, which might be changed per cluster. means we need a PeerPod VM image per cluster if we build it in PeerPod VM image. Shall we make `AAKBCParams` configurable in configmap?

mkulke:
> @mkulke because `AAKBCParams` is a dynamic endpoint, which might be changed per cluster. means we need a PeerPod VM image per cluster if we build it in PeerPod VM image. Shall we make `AAKBCParams` configurable in configmap?

`AAKBCParams` is already a parameter in configmap (and a cli param for the CAA binary), so we don't need to do anything here.

My reasoning was:

- If we have a CDH and an AA config file, we don't need AAKBCParams the agent config anymore, because it's not used. 
- If we don't have an AAKbcParam in the agent config file it is completely static and the same in each deployment.
- If the file is static, we don't need a cloud-config directive for it but can include it at build time.

huoqifeng:
> > @mkulke because `AAKBCParams` is a dynamic endpoint, which might be changed per cluster. means we need a PeerPod VM image per cluster if we build it in PeerPod VM image. Shall we make `AAKBCParams` configurable in configmap?
> 
> `AAKBCParams` is already a parameter in configmap (and a cli param for the CAA binary), so we don't need to do anything here.
> 
> My reasoning was:
> 
>     * If we have a CDH and an AA config file, we don't need AAKBCParams the agent config anymore, because it's not used.
> 
>     * If we don't have an AAKbcParam in the agent config file it is completely static and the same in each deployment.
> 
>     * If the file is static, we don't need a cloud-config directive for it but can include it at build time.

@mkulke which is reasonable to me, I like this approach. How does the CDH and AA use `AAKBCParams` which is set in k8s configmap? Is some changes happening in guest-components?

mkulke:
> @mkulke which is reasonable to me, I like this approach. How does the CDH and AA use `AAKBCParams` which is set in k8s configmap? Is some changes happening in guest-components?

we already do that for `cdh.toml`, for AA atm, we use `agent-config.toml`, which is a workaround for when there was no AA config file yet. now AA supports a config file, and we also want to use it because the `${kbc}::${uri}` schema of AAKBCParams is not sufficient anymore, we also need to include a kbs certificate in the config.

huoqifeng:
> > @mkulke which is reasonable to me, I like this approach. How does the CDH and AA use `AAKBCParams` which is set in k8s configmap? Is some changes happening in guest-components?
> 
> we already do that for `cdh.toml`, for AA atm, we use `agent-config.toml`, which is a workaround for when there was no AA config file yet. now AA supports a config file, and we also want to use it because the `${kbc}::${uri}` schema of AAKBCParams is not sufficient anymore, we also need to include a kbs certificate in the config.

OK, my understanding is we need generate an `attestation-agent.toml` rather than a `kata-agent.toml` (named agent-config.toml today) in cloud-api-adaptor side.

mkulke:
> OK, my understanding is we need generate an `attestation-agent.toml` rather than a `kata-agent.toml` (named agent-config.toml today) in cloud-api-adaptor side.

yup, exactly

note: there is a nuisance when using a aa.toml file for CAA currently: an aa config file is only valid if it has a `token_config` entry, which only makes sense for `cc_kbc`, at the moment. for `offline_fs_kbc` you cannot specifiy `--config`, which means we need some machinery to do that:

```bash
/usr/bin/attestation-agent "${AA_CONFIG_FILE+--config $AA_CONFIG_FILE}"
```

src/cloud-api-adaptor/pkg/agent/config.go
```diff
@@ -0,0 +1,50 @@
+package agent
```
mkulke, line 1: should we maybe rename the package to aa or attestation-agent?

huoqifeng:
@mkulke I found we still need set `cc_kbc::http://IP:Port` in file `/etc/agent-config.toml` because confidential-data-hub is still using it. If `aa_kbc_params = "offline_fs_kbc::null"` set in file `/etc/agent-config.toml`, cdh will report error like:
```
Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]: [2024-06-20T07:52:24Z ERROR ttrpc_cdh::ttrpc_server] [ttRPC CDH] GetResource :
Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]:     get resource failed
Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]:     Caused by: Kbs client error: offline-fs-kbc: resource not found one/two/key
```
Anything we have not completed yet in `confidential-data-hub` configuration?

src/cloud-api-adaptor/pkg/agent/config.go
```diff
@@ -0,0 +1,50 @@
...
+	config := TokenConfigs{}
+	config.CocoAs.URL = ""
+	config.Kbs.URL = url
+
+	bytes, err := toml.Marshal(config)
+	if err != nil {
+		return "", err
+	}
+	tomlString := strings.ReplaceAll(string(bytes), "['", "[")
+	tomlString = strings.ReplaceAll(tomlString, "']", "]")
```
mkulke, line 48: Why is this required?

mkulke (commented):
looks good! couple of

huoqifeng:
> etc/agent-config

Sounds like CDH lost a function - Never use `/etc/agent-config` if a config file provided by `-c` like `/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml`

mkulke:
> @mkulke I found we still need set `cc_kbc::http://IP:Port` in file `/etc/agent-config.toml` because confidential-data-hub is still using it. If `aa_kbc_params = "offline_fs_kbc::null"` set in file `/etc/agent-config.toml`, cdh will report error like:
> 
> ```
> Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]: [2024-06-20T07:52:24Z ERROR ttrpc_cdh::ttrpc_server] [ttRPC CDH] GetResource :
> Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]:     get resource failed
> Jun 20 07:52:24 podvm-nginx-nydus-54c6f8c6cc-nm6td-e21b40ee confidential-data-hub[2825]:     Caused by: Kbs client error: offline-fs-kbc: resource not found one/two/key
> ```
> 
> While it's OK if files `/etc/agent-config` does not exist at all. Anything we have not completed yet in `confidential-data-hub` configuration?

does CDH use the config file? In CAA we've been using the config file for CDH for a while now and it shouldn't rely on the agent-config.toml any more. This is the respective logic in [CDH](https://github.com/confidential-containers/guest-components/blob/fe9560bb99c0e89c0d81f24f08ac4b63d3e22821/confidential-data-hub/hub/src/bin/ttrpc-cdh.rs#L71).

mkulke:
> Sounds like CDH lost a function - Never use `/etc/agent-config` if a config file provided by `-c` like `/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml`

where do you see that CDH is falling back to kata-agent-config?

src/cloud-api-adaptor/pkg/agent/config.go
huoqifeng: Yeah, I agree.

src/cloud-api-adaptor/pkg/agent/config.go
huoqifeng: Because toml marshall generates string like this:
`['token_configs.kbs']`, I dropped the `'` characters in the table.

huoqifeng:
> > Sounds like CDH lost a function - Never use `/etc/agent-config` if a config file provided by `-c` like `/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml`
> 
> where do you see that CDH is falling back to kata-agent-config?

I did not find the code yet, it's confusion because I saw CDH logs like this and the function does not work if `/etc/agent-config.toml` exists.
```
[2024-06-20T09:08:13Z INFO  ttrpc_cdh] Use configuration file /run/confidential-containers/cdh.toml
[2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] get aa_kbc_params from file
[2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] reading agent config from /etc/agent-config.toml
[2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] get aa_kbc_params from file
[2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] reading agent config from /etc/agent-config.toml
[2024-06-20T09:08:13Z INFO  ttrpc_cdh] [ttRPC] Confidential Data Hub starts to listen to request: unix:///run/confidential-containers/cdh.sock

```

huoqifeng:
> > > Sounds like CDH lost a function - Never use `/etc/agent-config` if a config file provided by `-c` like `/usr/local/bin/confidential-data-hub -c /run/confidential-containers/cdh.toml`
> > 
> > 
> > where do you see that CDH is falling back to kata-agent-config?
> 
> I did not find the code yet, it's confusion because I saw CDH logs like this and the function does not work if `/etc/agent-config.toml` exists.
> 
> ```
> [2024-06-20T09:08:13Z INFO  ttrpc_cdh] Use configuration file /run/confidential-containers/cdh.toml
> [2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] get aa_kbc_params from file
> [2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] reading agent config from /etc/agent-config.toml
> [2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] get aa_kbc_params from file
> [2024-06-20T09:08:13Z DEBUG attestation_agent::config::aa_kbc_params] reading agent config from /etc/agent-config.toml
> [2024-06-20T09:08:13Z INFO  ttrpc_cdh] [ttRPC] Confidential Data Hub starts to listen to request: unix:///run/confidential-containers/cdh.sock
> ```

Sounds like https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/config/mod.rs#L130
https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L57

src/cloud-api-adaptor/pkg/agent/config.go
mkulke: I think it's quoting it because the annotation instructs it to use `.` as a reserved character for nesting.

Why don't we nest the configuration?

```go
type AAConfig struct {
	TokenCfg struct {
		CocoAs struct {
			URL string `toml:"url"`
		} `toml:"coco_as"`
		Kbs struct {
			URL string `toml:"url"`
		} `toml:"kbs"`
	} `toml:"token_configs"`
}
```

this way we don't have to patch the serialized string

mkulke:
> Sounds like https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/config/mod.rs#L130 https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L57

Thanks, for searching, I didn't spot it myself. I suspect this function is dead code, but I'm not 100% sure.

this fn sets AA_KBC_PARAMS + KBS_CERT envs as a side-effect and will do nothing else:

```rust
pub fn set_configuration_envs(&self) {...}
```

I was suspecting that image-rs is using it for encrypted images, but it doesn't seem to be used in guest-components:

```bash
)$ git grep AA_KBC_PARAMS
attestation-agent/attestation-agent/src/config/aa_kbc_params.rs:    if let Ok(params) = env::var("AA_KBC_PARAMS") { confidential-data-hub/README.md:* **AA_KBC_PARAMS** environment variable
confidential-data-hub/hub/src/bin/config/mod.rs:                "AA_KBC_PARAMS",
```

mkulke:
> > Sounds like https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/config/mod.rs#L130 https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L57
> 
> Thanks, for searching, I didn't spot it myself. I suspect this function is dead code, but I'm not 100% sure.
> 
> this fn sets AA_KBC_PARAMS + KBS_CERT envs as a side-effect and will do nothing else:
> 
> ```rust
> pub fn set_configuration_envs(&self) {...}
> ```
> 
> I was suspecting that image-rs is using it for encrypted images, but it doesn't seem to be used in guest-components:
> 
> ```shell
> /dev/guest-components (main)$ git grep AA_KBC_PARAMS
> attestation-agent/attestation-agent/src/config/aa_kbc_params.rs:    if let Ok(params) = env::var("AA_KBC_PARAMS") {
> confidential-data-hub/README.md:* **AA_KBC_PARAMS** environment variable
> confidential-data-hub/hub/src/bin/config/mod.rs:                "AA_KBC_PARAMS",
> ```

Note: `AA_KBC_PARAMS` env does seem to be required in the [constructor](https://github.com/confidential-containers/guest-components/blob/2f1380fe67b5dfe188052f85f5c11ff9aecb7351/confidential-data-hub/kms/src/plugins/kbs/mod.rs#L36) of CDH's kbs kms plugin.

huoqifeng:
Yeah, I think which comes from the CDH config if agent-config.toml does not provide it, https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/config/mod.rs#L133-L136, that's why I dropped it in this commit https://github.com/confidential-containers/cloud-api-adaptor/pull/1868/commits/fc0a04ef34bf93cee1fcd06a5625cbe684174967. @mkulke

huoqifeng:
I tried it on s390x by building the image following as below:
1. Build caa image
```
cd src/cloud-api-adaptor
ARCHES=linux/s390x make image
```
2. Build PodVM image
```
cd podvm-mkosi
make fedora-binaries-builder
ATTESTER=se-attester make binaries
make image-debug
```
3. Deploy a KBS service following https://github.com/confidential-containers/trustee/blob/main/attestation-service/verifier/src/se/README.md
4. Try it in a k8s cluster by using the caa and PodVM image
5. Create a PeerPod instance and retrieve a secret
```
# kubectl exec -it nginx-nydus-54c6f8c6cc-xzxzg -- /bin/sh
# 
# 
# 
# curl http://127.0.0.1:8006/cdh/resource/one/two/key
some random characters....
Sat Jun 15 03:06:01 UTC 2024
# exit
```

huoqifeng:
@mkulke @stevenhorsman I think this PR can be reviewed now as I have tried it. May you please have a look?

huoqifeng:
> CDH's kbs kms plugin.

oh, I think this is a case not covered indeed -- CDH's kbs kms plugin. We need a fix in CDH side.

huoqifeng:
I created an issue in cdh here https://github.com/confidential-containers/guest-components/issues/593 to track this. Sounds it's not big problem so far if we don't use KBS/KMS separately.

liudalibj:
> I tried it on s390x by building the image following as below:
> 
> 1. Build caa image
> 
> ```
> cd src/cloud-api-adaptor
> ARCHES=linux/s390x make image
> ```
> 
> 2. Build PodVM image
> 
> ```
> cd podvm-mkosi
> make fedora-binaries-builder
> ATTESTER=se-attester make binaries
> make image-debug
> ```
> 
We should run `make image` here or before run the `make image-debug` we should update the Makefile, to call `../hack/build-s390x-se-image.sh` for image-debug, so that we can get the SE-enabled podvm image.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/attestation-agent.service
```diff
@@ -1,11 +1,11 @@
 [Unit]
 Description=Attestation Agent TTRPC API Server
-After=network.target process-user-data.service
+After=network.target cloud-final.service process-user-data.service
```
liudalibj, line 3: the `cloud-final.service` seems only enabled for s390x from https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.finalize.chroot#L6

for amd64 build `cloud-final.service` is disable, so maybe we need another way to add this dependency?

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Confidential Data Hub TTRPC API Server
-After=network.target process-user-data.service
+After=network.target cloud-final.service process-user-data.service
```
liudalibj, line 3: the `cloud-final.service` seems only enabled for s390x from https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.finalize.chroot#L6

for amd64 build `cloud-final.service` is disable, so maybe we need another way to add this dependency?

src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
```diff
@@ -2,7 +2,7 @@
 Description=Kata Agent
 BindsTo=netns@podns.service
 Wants=process-user-data.service
-After=netns@podns.service process-user-data.service
+After=netns@podns.service cloud-final.service process-user-data.service
```
liudalibj, line 5: the `cloud-final.service` seems only enabled for s390x from https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.finalize.chroot#L6

for amd64 build `cloud-final.service` is disable, so maybe we need another way to add this dependency?

src/cloud-api-adaptor/podvm/files/etc/systemd/system/attestation-agent.service
mkulke: I think this can be done using a drop-in file? we can have a specific `mkosi.skeleton-s390x` folder?

huoqifeng: I think this is provider specific bot not architecture specific. We'll still need cloud-final.service for libvirt provider on x86. But don't need it for azure and aws providers because these providers are using process-user-data instead of cloud-init.

mkulke: Ok, I understand libvirt is a non-tee environment for testing only? then we could have `mkosi.skeleton-libvirt` and `mkosi.skeleton-ibmse`?

huoqifeng: Libvirt could be a tee environment  on s390x, this is also a supported case.

huoqifeng: But, yes, I think we might use mkosi.skeleton to overwrite the services file. Let me try...

mkulke: I think we have to make up our mind what kind of images we want to build. So far, we had x86_64 images [debug, production] and an exception clause for s390x:

https://github.com/confidential-containers/cloud-api-adaptor/blob/e1712af119297dd37872b795bc155de0f72b8e6c/src/cloud-api-adaptor/podvm-mkosi/Makefile#L54-L65

libvirt x86_64 doesn't seem to be supported on mkosi currently, and we should maybe address it in a later PR?

huoqifeng: There is no providers specific code in mkosi Makefile. 
For x86_64, there is `[debug, release]` image.
For s390x, there is `[debug, se-release] image. I just added a new type `se-debug` in PR https://github.com/confidential-containers/cloud-api-adaptor/pull/1877.
Before the remote attestation support, libvirt should worked well both on s390x and x86_64. With attestation, it breaks on mkosi image for libvirt because mkosi image is fedora based, and the process-user-data failure causes the attestation configuration breaks. Ideally, we should not add provider specific logic in mkosi, but honestly I do not think about a ideal solution yet. If we use mkosi.skeleton.s390x, will cause libvirt provider won't work on x86_64 for fedora. Still thinking about that...

huoqifeng: Per systemd's document, if service A defined as After= service B, it will still start if service B is disabled.
So I think systemd's `After=` constraints will not block the `kata-agent`, `attestation-agent`, `cdh` service's startup if the cloud-init service is not enabled. Maybe we can have a try on azure with the code change in this PR. @mkulke

mkulke: Indeed, that might be an option. it's not pretty, but we can add a comment that cloud-init is not present on read-only rootfs's. I'll test

mkulke (commented):
`process-user-data` will not start on mkosi x86 images currently (and hence there is no `daemon.json` provisioned, because this overlay issues an `update-agent-config` command, that is not available any more:

```bash
mkosi.skeleton/usr/lib/systemd/system/process-user-data.service.d/10-override.conf
```

huoqifeng:
> ```shell
> process-user-data.service.d
> ```

My fault, `update-agent-config` is not used any more because `agent-config.toml` is static and `aa.toml`, `cdh.toml` can be handled by `provision-files`. Dropped a fix.

src/cloud-api-adaptor/cmd/process-user-data/main.go
```diff
@@ -37,7 +35,7 @@ var rootCmd = &cobra.Command{
 }
 
 func init() {
-	var agentConfigPath, cdhConfigPath, daemonConfigPath string
+	var cdhConfigPath, daemonConfigPath string
```
mkulke, line 38: we have to add `aaConfigPath`. process-user-data will not provision any files, but only a selected list of filenames set in `userdata.NewConfig()`

huoqifeng: @mkulke looks userdata.NewConfig does not have `aaConfigPath` here  https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/pkg/userdata/provision.go#L33, `aaConfigPath` is used only by `updateAgentConfigCmd` in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/cmd/process-user-data/main.go#L59?

mkulke: yes, I think we need this.

```patch
From eb2e5174b01bdba299a792a70b004082a44b0f50 Mon Sep 17 00:00:00 2001
From: Magnus Kulke <magnuskulke@microsoft.com>
Date: Mon, 24 Jun 2024 16:13:39 +0200
Subject: [PATCH] process-user-data: ohai!

---
 src/cloud-api-adaptor/cmd/process-user-data/main.go |  6 ++++--
 src/cloud-api-adaptor/pkg/userdata/provision.go     | 11 +++++++++--
 2 files changed, 13 insertions(+), 4 deletions(-)

diff --git a/src/cloud-api-adaptor/cmd/process-user-data/main.go b/src/cloud-api-adaptor/cmd/process-user-data/main.go
index d5ac81e..477c133 100644
--- a/src/cloud-api-adaptor/cmd/process-user-data/main.go
+++ b/src/cloud-api-adaptor/cmd/process-user-data/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	cmdUtil "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/cmd"
+	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/aa"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/cdh"
 	daemon "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/forwarder"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/userdata"
@@ -35,18 +36,19 @@ var rootCmd = &cobra.Command{
 }
 
 func init() {
-	var cdhConfigPath, daemonConfigPath string
+	var aaConfigPath, cdhConfigPath, daemonConfigPath string
 	var fetchTimeout int
 
 	rootCmd.PersistentFlags().BoolVarP(&versionFlag, "version", "v", false, "Print the version")
 	rootCmd.PersistentFlags().StringVarP(&daemonConfigPath, "daemon-config-path", "d", daemon.DefaultConfigPath, "Path to a daemon config file")
+	rootCmd.PersistentFlags().StringVarP(&aaConfigPath, "aa-config-path", "a", aa.DefaultAaConfigPath, "Path to a AA config file")
 	rootCmd.PersistentFlags().StringVarP(&cdhConfigPath, "cdh-config-path", "c", cdh.ConfigFilePath, "Path to a CDH config file")
 
 	var provisionFilesCmd = &cobra.Command{
 		Use:   "provision-files",
 		Short: "Provision required files based on user data",
 		RunE: func(_ *cobra.Command, _ []string) error {
-			cfg := userdata.NewConfig(defaultAuthJsonPath, daemonConfigPath, cdhConfigPath, fetchTimeout)
+			cfg := userdata.NewConfig(aaConfigPath, defaultAuthJsonPath, daemonConfigPath, cdhConfigPath, fetchTimeout)
 			return userdata.ProvisionFiles(cfg)
 		},
 		SilenceUsage: true, // Silence usage on error
diff --git a/src/cloud-api-adaptor/pkg/userdata/provision.go b/src/cloud-api-adaptor/pkg/userdata/provision.go
index 8a045b2..df56a88 100644
--- a/src/cloud-api-adaptor/pkg/userdata/provision.go
+++ b/src/cloud-api-adaptor/pkg/userdata/provision.go
@@ -20,6 +20,7 @@ import (
 var logger = log.New(log.Writer(), "[userdata/provision] ", log.LstdFlags|log.Lmsgprefix)
 
 type paths struct {
+	aaConfig     string
 	authJson     string
 	cdhConfig    string
 	daemonConfig string
@@ -30,8 +31,8 @@ type Config struct {
 	paths        paths
 }
 
-func NewConfig(authJsonPath, daemonConfigPath, cdhConfig string, fetchTimeout int) *Config {
-	cfgPaths := paths{authJsonPath, cdhConfig, daemonConfigPath}
+func NewConfig(aaConfigPath, authJsonPath, daemonConfigPath, cdhConfig string, fetchTimeout int) *Config {
+	cfgPaths := paths{aaConfigPath, authJsonPath, cdhConfig, daemonConfigPath}
 	return &Config{fetchTimeout, cfgPaths}
 }
 
@@ -187,6 +188,12 @@ func processCloudConfig(cfg *Config, cc *CloudConfig) error {
 		return fmt.Errorf("failed to write daemon config file: %w", err)
 	}
 
+	if bytes := findConfigEntry(cfg.paths.aaConfig, cc); bytes != nil {
+		if err = writeFile(cfg.paths.aaConfig, bytes); err != nil {
+			return fmt.Errorf("failed to write aa config file: %w", err)
+		}
+	}
+
 	if bytes := findConfigEntry(cfg.paths.cdhConfig, cc); bytes != nil {
 		if err = writeFile(cfg.paths.cdhConfig, bytes); err != nil {
 			return fmt.Errorf("failed to write cdh config file: %w", err)
-- 
2.34.1
```

huoqifeng: Thanks @mkulke , applied the patch https://github.com/confidential-containers/cloud-api-adaptor/pull/1868/commits/7aacd90447c66db23ef6ba842dd5beb586814de7

mkulke (commented):
this is what we provision when we specify `--aa-kbc-params offline_fs_kbc::null`

```bash
cat /run/peerpod/aa.toml
[token_configs]
[token_configs.coco_as]
url = ''

[token_configs.kbs]
url = 'null'
```

not sure if that makes sense, it probably doesn't, but it might also not hurt at the moment. I think what we could do is just provision that file if `aa-kbc-params` is `cc_kbc::*` from the CAA side, because the config file doesn't cover anything else yet.

huoqifeng:
> this is what we provision when we specify `--aa-kbc-params offline_fs_kbc::null`
> 
> ```shell
> cat /run/peerpod/aa.toml
> [token_configs]
> [token_configs.coco_as]
> url = ''
> 
> [token_configs.kbs]
> url = 'null'
> ```
> 
> not sure if that makes sense, it probably doesn't, but it might also not hurt at the moment. I think what we could do is just provision that file if `aa-kbc-params` is `cc_kbc::*` from the CAA side, because the config file doesn't cover anything else yet.

Indeed, aa.toml, not sure cdh.toml should follow same pattern?

src/cloud-api-adaptor/pkg/aa/config.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func CreateConfigFile(aaKBCParams string) (string, error) {
+	url, err := parseAAKBCParams(aaKBCParams)
+	if err != nil {
+		return "", err
+	}
+
+	config := AAConfig{}
+	config.TokenCfg.CocoAs.URL = ""
```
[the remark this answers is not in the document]
huoqifeng: CocoAs.URL has different port, we don't know it here.

mkulke:
You meant url? I think by default it is the same url, we currently only support an all-in-one KBS deployment in CAA. So I think the value has to be the sameAm 24.06.2024 um 17:16 schrieb Qi Feng Huo ***@***.***>:﻿
@huoqifeng commented on this pull request.



In src/cloud-api-adaptor/pkg/aa/config.go:
> +	parts := strings.SplitN(aaKBCParams, "::", 2)
+	if len(parts) != 2 {
+		return "", fmt.Errorf("Invalid aa-kbs-params input: %s", aaKBCParams)
+	}
+	_, url := parts[0], parts[1]
+	return url, nil
+}
+
+func CreateConfigFile(aaKBCParams string) (string, error) {
+	url, err := parseAAKBCParams(aaKBCParams)
+	if err != nil {
+		return "", err
+	}
+
+	config := AAConfig{}
+	config.TokenCfg.CocoAs.URL = ""

CocoAs.URL has different port, we don't know it here.

src/cloud-api-adaptor/pkg/aa/config.go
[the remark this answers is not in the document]
mkulke: You meant url? I think by default it is the same url, we currently only support an all-in-one KBS deployment in CAA. So I think the value has to be the same

huoqifeng:
> cc_kbc::*

With the fix in https://github.com/confidential-containers/guest-components/pull/599, I think we can leave the code as is, I added a log for s.aaKBCParams.

mkulke (approved):
lgtm. thanks!

Tested the following combinations on azure SNP & TDX cvms:

- empty aa_kbc_params
- aa_kbc_params=offline_fs_kbc
- aa_kbc_params=cc_kbc::$some_kbs

logged into the podvm to verify that AA, CDH and ASR services run properly and retrieved a confidential resource from KBS.

src/cloud-api-adaptor/cmd/process-user-data/main.go
stevenhorsman, line 1: Sorry to be picky, but can we expand and fix the commit message here too add some more explanation about what and why for future reference? "process-user-data add ageng-config" is especially confusing as I think it's the attestation-agent config, not kata-agent config that is being added?

stevenhorsman (commented):
The code is general fine, but I'd like to see the lbivirt e2e tests run before merging.

liudalibj (approved):
LGTM, thanks @huoqifeng 
Tested this pr with s390x fedora image.
Both `sample` and `ibm.se` attestation drivers work as expected.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
huoqifeng: Thanks Dali, Per systemd's document, if service A defined as After= service B, it will still start if service B is disabled.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
huoqifeng: Thanks Dali, Per systemd's document, if service A defined as After= service B, it will still start if service B is disabled.

src/cloud-api-adaptor/cmd/process-user-data/main.go
huoqifeng: Thanks Steve, updated the commit message, lets check the test result...

liudalibj:
I tried to build podvm ubuntu amd64 image, from this PR. 
- `quay.io/liudalibj/podvm-generic-ubuntu-amd64:61a7406743c9ad6915b7dfc1fd3217cb5e581235b8791d791aaf9f669255fd52`
```
make podvm-builder
make podvm-binaries
make podvm-image
```
with the built out podvm-image I can create a peerpod succeed. 
But when I try to use this podvm-image with the sample kbs, `cc_kbc::http://ip:8080`, the `attestation-agent.service` is failed to start in the created peerpod vm.

The related log from  journalctl 
```
...
Jun 25 08:40:28 ubuntu systemd[1]: Installed transient /etc/machine-id file.
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found ordering cycle on attestation-agent.service/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found dependency on cloud-final.service/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found dependency on multi-user.target/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Job attestation-agent.service/start deleted to break ordering cycle starting with multi-user.target/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found ordering cycle on kata-agent.service/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found dependency on cloud-final.service/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Found dependency on multi-user.target/start
Jun 25 08:40:28 ubuntu systemd[1]: multi-user.target: Job kata-agent.service/start deleted to break ordering cycle starting with multi-user.target/start
Jun 25 08:40:28 ubuntu systemd[1]: Created slice system-modprobe.slice.
...
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd-timesyncd[561]: Network configuration changed, trying to establish connection.
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd[1]: multi-user.target: Found ordering cycle on kata-agent.service/start
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd[1]: multi-user.target: Found dependency on cloud-final.service/start
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd[1]: multi-user.target: Found dependency on multi-user.target/start
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd[1]: multi-user.target: Job kata-agent.service/start deleted to break ordering cycle starting with multi-user.target/start
Jun 25 08:40:33 podvm-nginx-nydus-54c6f8c6cc-dz74n-6f4c65e4 systemd[1]: agent-protocol-forwarder.service: Scheduled restart job, restart counter is at 1.
...
```
@huoqifeng it seems that there are cycle decency on  `multi-user.target`, `attestation-agent.service` and `attestation-agent.service` maybe we need do a fix, the e2e-test case `TestLibvirtKbsKeyRelease` is failed because this.

mkulke:
oh, I only tested mkosi on x86_64 (i.e. without cloud-config)

So assuming you tested mkosi + cloud-config, and that worked, we probably should look into the delta of packer vs mkosi + cloud-config

src/cloud-api-adaptor/podvm/files/etc/systemd/system/attestation-agent.service
```diff
@@ -1,11 +1,15 @@
...
```
liudalibj, line 3: to fix issue at https://github.com/confidential-containers/cloud-api-adaptor/pull/1868#issuecomment-2188451116
`cloud-final.service` -> `cloud-init.service`

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
```diff
@@ -1,10 +1,14 @@
...
```
liudalibj, line 3: `cloud-final.service` -> `cloud-init.service`

src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
liudalibj, line 5: `cloud-final.service` -> `cloud-init.service`

liudalibj:
> @huoqifeng it seems that there are cycle decency on `multi-user.target`, `attestation-agent.service` and `kata-agent.service` maybe we need do a fix, the e2e-test case `TestLibvirtKbsKeyRelease` is failed because this.

Verified: replace the `cloud-final.service` to `cloud-init.service`, can fix the cycle dependency issue @huoqifeng @mkulke FYI.

With this replace fix, packer ubuntu amd64 peerpod + cloud-init + libvirt + sample kbc, works fine.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
huoqifeng: Thanks @liudalibj looks we can drop the cloud-final.service in kata-agent as the kata-agent config is static now.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/confidential-data-hub.service
huoqifeng: Thanks @liudalibj let me try whether the cdh toml file placed rightly with this change.

src/cloud-api-adaptor/podvm/files/etc/systemd/system/attestation-agent.service
huoqifeng: Thanks @liudalibj let me try whether the toml file placed rightly with this change.

huoqifeng: looks not good, cloud-init.service can not guarantee the cloud-init files placed correctly,  cloud-final.service can. This might be the difference between ubuntu and fedora.

huoqifeng: Let me have a try to overwrite it in fedora and keep it in ubuntu.

huoqifeng: Tried again, I misused the image in my env, looks `cloud-init.service` is a good approach.

mkulke:
> @mkulke not sure whether worth for you to have another try with the last fix (hopefully latest :-) ) here [1fbc636](https://github.com/confidential-containers/cloud-api-adaptor/commit/1fbc63646559834565be06db6c098f7270dcdf73).

I think that shouldn't impact mkosi on x86_64, let's wait for the libvirt test

src/cloud-api-adaptor/podvm/files/etc/systemd/system/attestation-agent.service
huoqifeng: Double checked, we can use `cloud-init.service` instead cloud-final.service. the IBM SE attestation works well on fedora image, the sample tee works for ubuntu image in e2e.

PR #1875, opened 2024-06-20
enable kbs cert for cdh

wyuany:


src/cloud-api-adaptor/pkg/adaptor/server.go
```diff
@@ -40,6 +40,7 @@ type ServerConfig struct {
 	ForwarderPort           string
 	ProxyTimeout            time.Duration
 	AAKBCParams             string
+	KBSCERT                 string
```
mkulke, line 43:
```suggestion
	KBSCert                 string
```

src/cloud-api-adaptor/pkg/forwarder/forwarder.go
```diff
@@ -44,8 +44,8 @@ type Config struct {
 	TLSClientCA   string `json:"tls-client-ca,omitempty"`
 
 	AAKBCParams string `json:"aa-kbc-params,omitempty"`
-
-	AuthJson string `json:"auth-json,omitempty"`
+	KBSCERT     string `json:"kbs-cert,omitempty"`
```
mkulke, line 47: I don't think we need to add this to daemon.json (we should also drop AAKBCParams, it's unused afaik)

src/cloud-api-adaptor/pkg/cdh/config.go
```diff
@@ -21,8 +21,9 @@ type Config struct {
 }
 
 type KBCConfig struct {
-	Name string `toml:"name"`
-	URL  string `toml:"url"`
+	Name    string `toml:"name"`
+	URL     string `toml:"url"`
+	KBSCERT string `toml:"kbs_cert"`
```
mkulke, line 26:
```suggestion
	KBSCert string `toml:"kbs_cert"`
```

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -242,6 +243,10 @@ func (s *cloudService) CreateVM(ctx context.Context, req *pb.CreateVMRequest) (r
 		daemonConfig.AAKBCParams = s.aaKBCParams
 	}
 
+	if s.kbsCert != "" {
+		daemonConfig.KBSCERT = s.kbsCert
+	}
```
mkulke, line 248: i don't think we need to add this to daemon.json, see below

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -77,7 +77,7 @@ func (s *cloudService) removeSandbox(id sandboxID) error {
 }
 
 func NewService(provider provider.Provider, proxyFactory proxy.Factory, workerNode podnetwork.WorkerNode,
-	secureComms bool, secureCommsInbounds, secureCommsOutbounds, kbsAddress, podsDir, daemonPort, aaKBCParams, sshport string) Service {
+	secureComms bool, secureCommsInbounds, secureCommsOutbounds, kbsAddress, podsDir, daemonPort, aaKBCParams, kbsCert, sshport string) Service {
```
mkulke, line 80: should we only allow `kbsCert` if `aaKBCParams ~ cc_kbc::*`?

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -123,6 +123,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		flags.IntVar(&cfg.networkConfig.VXLANPort, "vxlan-port", vxlan.DefaultVXLANPort, "VXLAN UDP port number (VXLAN tunnel mode only")
 		flags.IntVar(&cfg.networkConfig.VXLANMinID, "vxlan-min-id", vxlan.DefaultVXLANMinID, "Minimum VXLAN ID (VXLAN tunnel mode only")
 		flags.StringVar(&cfg.serverConfig.AAKBCParams, "aa-kbc-params", "", "attestation-agent KBC parameters")
+		flags.StringVar(&cfg.serverConfig.KBSCERT, "kbs-cert", "", "cert of secure KBS")
```
mkulke, line 126:
```suggestion
		flags.StringVar(&cfg.serverConfig.KBSCert, "kbs-cert", "", "certificate of KBS")
```

mkulke (commented):
if we want to introduce this field, we probably want test coverage. Can we add it as part of a kbs test?

src/cloud-api-adaptor/pkg/cdh/config.go
mkulke: I think we want to add `omitempty` to the cert field annotatuion so that an empty string will not be serialized.

wyuany:
Hi @mkulke, thank you very much! All the comments are addressed.

mkulke:
> Hi @mkulke, thank you very much! All the comments are addressed.

thanks for accommodating. I think we are missing the `cert` property in the `aa.toml` file for the kbs tokens, too.

In terms of testing it would be good to have this tested as part of the existing e2e test suite.

I have been testing w/ self-signed certs for IPs, because in the current tests, the kbs deployment is exposed via node-port. Sadly self-signed IP certificates do not work well in the rustls stack, I only got those working when building AA + CDH w/ openssl.

mkulke:
The problem with self-signed certs is that rustls is a bit more strict than openssl and others, so if we want to specify the kbs cert directly, we have to mark it as CA:false

```bash
openssl req -x509 -out kbs.crt -keyout kbs.key -newkey rsa:2048 -nodes -sha256 \
      -subj '/CN=kbs.coco' \
      --addext "subjectAltName=IP:10.200.24.4" \
      --addext "basicConstraints=CA:FALSE"
```

mkulke:
> The problem with self-signed certs is that rustls is a bit more strict than openssl and others, so if we want to specify the kbs cert directly, we have to mark it as CA:false
> 
> ```shell
> openssl req -x509 -out kbs.crt -keyout kbs.key -newkey rsa:2048 -nodes -sha256 \
>       -subj '/CN=kbs.coco' \
>       --addext "subjectAltName=IP:10.200.24.4" \
>       --addext "basicConstraints=CA:FALSE"
> ```

At the moment, the kbs deployment manifests hardcode http-only, so we can do the e2e tests in a follow-up PR. Please add a description to the PR, I'll trigger e2e tests.

wyuany:
@mkulke Thank you very much for the details. I'll add description and address the e2e test in a new PR. 
There are lots of commits of this PR. Let me merge these commits into one.

src/cloud-api-adaptor/pkg/adaptor/server.go
```diff
@@ -71,7 +72,7 @@ func NewServer(provider provider.Provider, cfg *ServerConfig, workerNode podnetw
 
 	agentFactory := proxy.NewFactory(cfg.PauseImage, cfg.TLSConfig, cfg.ProxyTimeout)
 	cloudService := cloud.NewService(provider, agentFactory, workerNode,
-		cfg.SecureComms, cfg.SecureCommsInbounds, cfg.SecureCommsOutbounds, cfg.SecureCommsKbsAddress, cfg.PodsDir, cfg.ForwarderPort, cfg.AAKBCParams, sshutil.SSHPORT)
+		cfg.SecureComms, cfg.SecureCommsInbounds, cfg.SecureCommsOutbounds, cfg.SecureCommsKbsAddress, cfg.PodsDir, cfg.ForwarderPort, cfg.AAKBCParams, cfg.KBSCERT, sshutil.SSHPORT)
```
mkulke, line 75: we print out this string verbatim at startup, while not a secret it's very noisy. maybe find a way to truncate or mask the string, similar to secret values.

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -19,6 +19,7 @@ optionals+=""
 [[ "${TLS_SKIP_VERIFY}" ]] && optionals+="-tls-skip-verify "
 [[ "${PROXY_TIMEOUT}" ]] && optionals+="-proxy-timeout ${PROXY_TIMEOUT} "
 [[ "${AA_KBC_PARAMS}" ]] && optionals+="-aa-kbc-params ${AA_KBC_PARAMS} "
+[[ "${KBS_CERT}" ]] && optionals+="-kbs-cert ${KBS_CERT} "
```
mkulke, line 22: I don't think this will work, with "$KBS_CERT" being a multi-line or even just containing whitespaces, did you test that?

mkulke (commented):
a couple of additional comments.

In general I'm not convinced that it's the best approach to specify the plain text of the cert as `--kbc-cert "----BEGIN CERTIFICATE-----\n ..."` maybe others can chime in

mkulke:
> a couple of additional comments.
> 
> In general I'm not convinced that it's the best approach to specify the plain text of the cert as `--kbc-cert "----BEGIN CERTIFICATE-----\n ..."` maybe others can chime in

I tried something like this, so we can still specify a block of text in the peerpod configmap. we can write it into a file earlier and then pass the filename. In main.go we then need read the file into a `[]byte` and pass it around. sounds like ceremonial back-and-forth, but it's probably better than having to deal with a large unreadable cmdline for the process.

```bash
[[ "${KBS_CERT}" ]] && echo -n "${KBS_CERT}" > /run/kbs.crt
[[ -f /run/kbs.crt ]] && optionals+="-kbs-cert-file /run/kbs.crt "
```

wyuany:
Hold on this PR due to the feature of pass config file through init-data annotation is almost done. https://github.com/confidential-containers/cloud-api-adaptor/pull/1912#issuecomment-2224172159. Once that PR is merged, this current PR is not required.

bpradipt:
@wyuany @huoqifeng are you folks planning to incorporate adding `kbs_cert` in cdh.toml and `cert` in aa.toml as part of this PR or the initData PR ?

huoqifeng:
> @wyuany @huoqifeng are you folks planning to incorporate adding `kbs_cert` in cdh.toml and `cert` in aa.toml as part of this PR or the initData PR ?

@bpradipt yes, we want leverage initdata addressed in https://github.com/confidential-containers/cloud-api-adaptor/pull/1912, CC @liudalibj @stevenhorsman @mkulke

huoqifeng:
@wyuany I think we can use initdata to pass in the certs now after https://github.com/confidential-containers/cloud-api-adaptor/pull/1912 get merged. Not sure shall we close this PR?

wyuany:
@huoqifeng I think so.

PR #1883, opened 2024-06-26
azure: set northeurope as default replication loc

mkulke:
westeurope may not offer free k8s tiers all the time


kartikjoshi21 (approved):

bpradipt (approved):
/lgtm

PR #1888, opened 2024-06-28
config_libvirt.sh: detect also rhel-like distros

esposem:
Extend the script to also detect fedora and centos.

I personally tested this on centos and it works fine.


bpradipt (approved):
/lgtm

beraldoleal (approved):
LGTM, thanks

PR #1892, opened 2024-07-01
release: Update go modules for pre-release

stevenhorsman:
We are working on the `v0.9.0-alpha1` pre-release, so bump the go modules to use the new tag we will create once they are merged.

### DIFF
--- a/src/cloud-api-adaptor/go.mod
+++ b/src/cloud-api-adaptor/go.mod
@@ -47,8 +47,8 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/eks v1.29.5
 	github.com/aws/aws-sdk-go-v2/service/iam v1.22.5
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.38.5
-	github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers v0.8.2
-	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl v0.8.2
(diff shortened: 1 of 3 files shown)


src/cloud-api-adaptor/go.mod
```diff
@@ -47,8 +47,8 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/eks v1.29.5
 	github.com/aws/aws-sdk-go-v2/service/iam v1.22.5
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.38.5
-	github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers v0.8.2
-	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl v0.8.2
+	github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers v0.9.0-alpha0
```
wainersm, line 50: This the "alpha" usage problem that you mentioned on last meeting? i.e., caa always tagged `<version>-alphaN` but now `<version>` is actually an alpha release too!

If so, would make sense to call it `v0.9.0-alpha1-alpha0`? I confess this seems ugly and I never saw such as versioning before...but just in case, let me ask :)

stevenhorsman: I considered that myself, but thought it was ugly. If it occured to you then it's probably worth switching :)

wainersm (approved):
I hope never see those alpha of alpha again :)

bpradipt (approved):
/lgtm

PR #1897, opened 2024-07-02
Fixes vulnerabilities identified via dependabot

bpradipt:


stevenhorsman (approved):
LGTM assuming that there is no negative impact on the Azure tests

bpradipt:
> LGTM assuming that there is no negative impact on the Azure tests

@mkulke can this PR be tested via Azure e2e ci ?

mkulke (approved):

mkulke:
> > LGTM assuming that there is no negative impact on the Azure tests
> 
> @mkulke can this PR be tested via Azure e2e ci ?

we don't have a label toggle for that yet. so, the tests only run on main. If the code compiles but the tests break due to a bump in dependencies, I'd be surprised.

stevenhorsman:
As mentioned on slack. I think we need to decide what to do with respect to the release:
1. Ignore it and continue even with a go vuln complaint (probably not ideal)
2. Merge it before the release, but don't worry about re-creating and re-tagging the pre-release on the assumption it doesn't significantly impact testing (if possible this is the least disruptive)
3. Merge it and re-start the pre-release to create a new pre-release build

mkulke:
> As mentioned on slack. I think we need to decide what to do with this PR with respect to the release:
> 
> 1. Ignore it and continue even with a go vuln complaint (probably not ideal)
> 2. Merge it before the release, but don't worry about re-creating and re-tagging the pre-release on the assumption it doesn't significantly impact testing (if possible this is the least disruptive)
> 3. Merge it and re-start the pre-release to create a new pre-release build

it being an alpha release I would go with 1) or 2) we can add an checklist item to the release doc that we want to verify vulns and bump revisions early in the process

stevenhorsman:
Let's go for 2. and merge this now. I will re-run the libvirt e2e tests on this once merged too and I guess the Azure ones will as well then, so that should be enough.

PR #1845, opened 2024-05-26
Add e2e tests for docker

bpradipt:
Also makes docker network and podvm image configurable to help with e2e, and some minor fixes

PR #1899, opened 2024-07-02
test/e2e: misc improvements

wainersm:
While testing https://github.com/confidential-containers/cloud-api-adaptor/pull/1845 I wanted to run only a single test case (the simple pod test) because all tests were failing and I didn't want to wait all to run (or ctrl+C). So here is a small improvement to allow us to run just a sub-set of e2e tests.

Also small improvement on e2e tests README.md

### DIFF
--- a/src/cloud-api-adaptor/Makefile
+++ b/src/cloud-api-adaptor/Makefile
@@ -21,6 +21,8 @@ SOURCES     := $(shell find $(SOURCEDIRS) -name '*.go' -print)
 ATTESTER    ?= none
 # End-to-end tests overall run timeout.
 TEST_E2E_TIMEOUT ?= 60m
+# To run a sub-set of tests, set an unanchored regex matching the test's name as in go's test -run option.
+RUN_TESTS ?= ''
 
 RESOURCE_CTRL ?= true
 YQ_CHECKSUM_${ARCH} ?= $(YQ_CHECKSUM)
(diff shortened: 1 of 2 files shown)


wainersm:
on hold waiting the 0.9.0-alpha1 release.

src/cloud-api-adaptor/Makefile
```diff
@@ -93,7 +95,7 @@ test: ## Run tests.
 .PHONY: test-e2e
 test-e2e: ## Run end-to-end tests for single provider.
 ifneq ($(CLOUD_PROVIDER),)
-	go test -v -tags=$(CLOUD_PROVIDER) -timeout $(TEST_E2E_TIMEOUT) -count=1 ./test/e2e
+	go test -v -tags=$(CLOUD_PROVIDER) -timeout $(TEST_E2E_TIMEOUT) -count=1 -run $(RUN_TESTS) ./test/e2e
```
stevenhorsman, line 98: Nice one - I've been making this change exact change manually every so often for over a year, and keep meaning to push it, so thanks for not being as lazy as me!

stevenhorsman (approved):
LGTM. Thanks!

src/cloud-api-adaptor/Makefile
wainersm: Every morning I say "I need to send this way of running just a sub-set of tests" and that passed years too. Today is a new era, today the cycle was broken! :)

stevenhorsman: Unfortunately you have just demonstrated that if I put off an issue long enough it will go away!

bpradipt (approved):
/lgtm
Nice. Thanks @wainersm

PR #1891, opened 2024-07-01
versions: add entry for CoCo operator

wainersm:
Added 'git.coco-operator' entry to the project's versions.yaml with information about the repository URL and reference (commit SHA1, tag...etc) as a way to have a single source truth and avoiding the need to update the references in many places at release time (then roll-back after the release).

Now `make deploy` will read the operator's URL and reference from versions.yaml. Likewise, the e2e test's provision.

PR #1900, opened 2024-07-03
Docker hub mirror switch

stevenhorsman:
In some e2e tests, particularly the docker-provider ones, when you do multiple
runs we are likely to hit the rate limits of docker hub, so try switching to the
official mirrors of nginx and curl container images instead to reduce our usage of docker.io.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/common_suite.go
+++ b/src/cloud-api-adaptor/test/e2e/common_suite.go
@@ -26,7 +26,7 @@ var E2eNamespace = envconf.RandomName("coco-pp-e2e-test", 25)
 
 // DoTestCreateSimplePod tests a simple peer-pod can be created.
 func DoTestCreateSimplePod(t *testing.T, e env.Environment, assert CloudAssert) {
-	pod := NewBusyboxPodWithName(E2eNamespace, "simple-test")
+	pod := NewBusyboxPodWithName(E2eNamespace, "simple-test").GetPodOrFatal(t)
 	if isTestOnCrio() {
 		NewTestCase(t, e, "SimplePeerPod", assert, "PodVM is created").WithPod(pod).Run()
 	} else {
@@ -35,7 +35,7 @@ func DoTestCreateSimplePod(t *testing.T, e env.Environment, assert CloudAssert) 
 }
 
 func DoTestDeleteSimplePod(t *testing.T, e env.Environment, assert CloudAssert) {
-	pod := NewBusyboxPodWithName(E2eNamespace, "deletion-test")
+	pod := NewBusyboxPodWithName(E2eNamespace, "deletion-test").GetPodOrFatal(t)
 	duration := assert.DefaultTimeout()
 	NewTestCase(t, e, "DeletePod", assert, "Deletion complete").WithPod(pod).WithDeleteAssertion(&duration).Run()
 }
@@ -43,7 +43,7 @@ func DoTestDeleteSimplePod(t *testing.T, e env.Environment, assert CloudAssert) 
 func DoTestCreatePodWithConfigMap(t *testing.T, e env.Environment, assert CloudAssert) {
 	podName := "busybox-configmap-pod"
 	containerName := "busybox-configmap-container"
-	imageName := BUSYBOX_IMAGE
+	imageName := getBusyboxTestImage(t)
 	configMapName := "busybox-configmap"
 	configMapFileName := "example.txt"
 	podKubeConfigmapDir := "/etc/config/"
@@ -76,7 +76,7 @@ func DoTestCreatePodWithSecret(t *testing.T, e env.Environment, assert CloudAsse
 	//DoTestCreatePod(t, assert, "Secret is created and contains data", pod)
 	podName := "busybox-secret-pod"
 	containerName := "busybox-secret-container"
-	imageName := BUSYBOX_IMAGE
+	imageName := getBusyboxTestImage(t)
 	secretName := "busybox-secret"
 	podKubeSecretsDir := "/etc/secret/"
 	usernameFileName := "username"
@@ -128,7 +128,7 @@ func DoTestCreatePeerPodContainerWithExternalIPAccess(t *testing.T, e env.Enviro
 	// / # ping 8.8.8.8
 	// PING 8.8.8.8 (8.8.8.8): 56 data bytes
 	// ping: permission denied (are you root?)
-	pod := NewPrivPod(E2eNamespace, "busybox-priv")
+	pod := NewPrivPod(E2eNamespace, "busybox-priv").GetPodOrFatal(t)
 	testCommands := []TestCommand{
 		{
 			Command:       []string{"ping", "-c", "1", "www.google.com"},
@@ -169,7 +169,7 @@ func DoTestCreatePeerPodAndCheckUserLogs(t *testing.T, e env.Environment, assert
 
 // DoTestCreateConfidentialPod verify a confidential peer-pod can be created.
 func DoTestCreateConfidentialPod(t *testing.T, e env.Environment, assert CloudAssert, testCommands []TestCommand) {
-	pod := NewBusyboxPodWithName(E2eNamespace, "confidential-pod-busybox")
+	pod := NewBusyboxPodWithName(E2eNamespace, "confidential-pod-busybox").GetPodOrFatal(t)
 	for i := 0; i < len(testCommands); i++ {
 		testCommands[i].ContainerName = pod.Spec.Containers[0].Name
 	}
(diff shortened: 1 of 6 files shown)


wainersm (approved):
yeah, it's 2024 and docker.io is still hurting us :(

Thanks @stevenhorsman !

wainersm:
Re-triggered CI to see if it passes this time.

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -472,10 +472,10 @@ func DoTestPodToServiceCommunication(t *testing.T, e env.Environment, assert Clo
 func DoTestPodsMTLSCommunication(t *testing.T, e env.Environment, assert CloudAssert) {
 	clientPodName := "curl"
 	clientContainerName := "curl"
-	clientImageName := "docker.io/curlimages/curl:8.4.0"
+	clientImageName := "quay.io/curl/curl:8.4.0"
```
bpradipt, line 475: In some other places in the code we use `quay.io/curl/curl:latest`.  Can we use the same curl image ? Better is to use a version tag than latest to avoid issues with change of command line options etc.

Also can all these images be moved to separate file as constants ? It'll help to override the images if needed.
I understand the scope of this PR is to just switch out of docker hub mirror, but since you are anyway touching these files thought of suggesting it :-) 
This can be done later on as well.

stevenhorsman: Sure, that makes sense. I was trying to keep versions the same incase that was important, but happy to pin them and separating out as constants is good too. Let me mark this as draft until I get a chance to get around to this.

wainersm: @stevenhorsman  are you thinking in add the version in versions.yaml or just a simple constant in code? 

I'm okay with a constant in code, but if you want the other approach then you want to use https://github.com/confidential-containers/cloud-api-adaptor/pull/1891/commits/5169aec42757d17c571181a0fd3dd8a38688563a#diff-a3bd85dcf59fe8d1cf225b0edbbd31d0390a9588a53b44941ad49d14d9f562eb

stevenhorsman: I'm on the fence with this. I can see the merit of having the versions defined in a single place, but it also doesn't feel quite right to me to mix up test dependencies with the built dependencies in the versions file (which is weird as lots of toolchains do it!).

bpradipt: May be have a separate versions file for test deps and placed inside the test folder ?

stevenhorsman: Yeah, that sounds like a reasonable plan.

wainersm: yup, I also don't like the mix up as in Kata. Good idea @bpradipt !

wainersm:
The failed test:
```
=== RUN   TestLibvirtPodsMTLSCommunication
=== RUN   TestLibvirtPodsMTLSCommunication/TestPodsMTLSCommunication_test
    assessment_runner.go:259: pods "nginx" already exists
--- FAIL: TestLibvirtPodsMTLSCommunication (0.94s)
    --- FAIL: TestLibvirtPodsMTLSCommunication/TestPodsMTLSCommunication_test (0.93s)
```

has been unstable? I looked the last 4 nightly jobs that failed, that test passed. So no conclusion on that matter ;)

stevenhorsman:
> The failed test:
> 
> ```
> === RUN   TestLibvirtPodsMTLSCommunication
> === RUN   TestLibvirtPodsMTLSCommunication/TestPodsMTLSCommunication_test
>     assessment_runner.go:259: pods "nginx" already exists
> --- FAIL: TestLibvirtPodsMTLSCommunication (0.94s)
>     --- FAIL: TestLibvirtPodsMTLSCommunication/TestPodsMTLSCommunication_test (0.93s)
> ```
> 
> has been unstable? I looked the last 4 nightly jobs that failed, that test passed. So no conclusion on that matter ;)

Maybe the problem is with the image change then?

src/cloud-api-adaptor/test/e2e/common_suite.go
stevenhorsman: So I've defrosted this PR and updated it to add a test versions.yaml with the test images and code to read from there. It feels like quite a bit of extra code, for maybe not a tons better user experience. 

The main issue I have with this is that reading from yaml seems to inherently require the need to throw errors and some of our code in common.go for specifying the busybox image doesn't have a testing.T object or any existing errors thrown, so I don't know if you have any suggestions of how to handle this apart from added return errors and error handling to a whole bunch of methods?

src/cloud-api-adaptor/test/e2e/versions.yaml
```diff
@@ -0,0 +1,10 @@
+test_images:
+  nginx:
+    registry: "quay.io/sjenning/nginx"
+    tag: "1.15-alpine"
+  busybox:
+    registry: "quay.io/prometheus/busybox"
+    tag: "latest"
+  curl:
+    registry: "quay.io/curl/curl"
```
bpradipt, line 9: You can use the following for nginx  - quay.io/confidential-containers/test/nginx:latest
I created the multi-arch manifest including amd64 and s390x. We can keep it updated as needed for tests.

For unprivileged nginx you can use - "quay.io/nginx/nginx-unprivileged:latest"

stevenhorsman: Updated. Thanks, sjeninng's nginx is getting old now!

bpradipt:
@stevenhorsman any plans to merge this ? My runs keep getting throttled due to pull rate limits.

stevenhorsman:
> @stevenhorsman any plans to merge this ? My runs keep getting throttled due to pull rate limits.

So I didn't get any response to the questions I raised in https://github.com/confidential-containers/cloud-api-adaptor/pull/1900#discussion_r1715541601, due to problems with the yaml parsing, so it's blocked due to that as the original approach was rejected.

Admittedly I haven't re-visited it to see if anything has changed to improve it recently, so I will rebase and review and see if anything has changed.

bpradipt:
> > @stevenhorsman any plans to merge this ? My runs keep getting throttled due to pull rate limits.
> 
> So I didn't get any response to the questions I raised in [#1900 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/pull/1900#discussion_r1715541601), due to problems with the yaml parsing, so it's blocked due to that as the original approach was rejected.

Based on your practical experience do you recommend any alternative approach ?

stevenhorsman:
> Based on your practical experience do you recommend any alternative approach ?

So in the third commit I've tried adding a new PodOrError type to help us be able to handle errors here without lots of
```
if err != nil {
	t.Fatal(err)
}
```
everywhere. I did try and find an existing strategy for this, but my googling is poor, so I failed. Assuming it works, then please take a look and let me know if you think it's worth adopting here?

mkulke:
> > Based on your practical experience do you recommend any alternative approach ?
> 
> So in the third commit I've tried adding a new PodOrError type to help us be able to handle errors here without lots of
> 
> ```
> if err != nil {
> 	t.Fatal(err)
> }
> ```

that's just golang, I think 🤷

mkulke (approved):

bpradipt (approved):
/lgtm
Thanks @stevenhorsman. 
A question on the versions.yaml file. Would it make sense to use a different name for versions.yaml to indicate it's for test images? Or you envisage additional test artifact versions to be added to it?

stevenhorsman:
> /lgtm Thanks @stevenhorsman. A question on the versions.yaml file. Would it make sense to use a different name for versions.yaml to indicate it's for test images? Or you envisage additional test artifact versions to be added to it?

Yeah, my thinking with the general name and top-level `test_images:` was that we could use it for a broader range of things later. e.g. there is an argument that the current `kbs` versions we have in the main `versions.yaml` is a test dependency, rather than a development one, but I didn't want to move anything over and confuse people without further discussion. I'm happy to re-name it to test-images though if preferred in the short term?

bpradipt:
> > /lgtm Thanks @stevenhorsman. A question on the versions.yaml file. Would it make sense to use a different name for versions.yaml to indicate it's for test images? Or you envisage additional test artifact versions to be added to it?
> 
> Yeah, my thinking with the general name and top-level `test_images:` was that we could use it for a broader range of things later. e.g. there is an argument that the current `kbs` versions we have in the main `versions.yaml` is a test dependency, rather than a development one, but I didn't want to move anything over and confuse people without further discussion. I'm happy to re-name it to test-images though if preferred in the short term?

Since we may add kbs versions here, I think it's ok to keep the name as-is.

PR #1902, opened 2024-07-03
ci: delete dangling vms in azure e2e test

mkulke:
## Why

Sometime resources aren't properly deleted in the azure e2e test. It impacts VM + aux resources, but also AKS cluster cleanup b/c the VM is still attached to a cluster's node subnet. Matrix tests are short-circuited if one of them fails, so it's not unlikely that we're not able to clean up properly.

## How

* resource tags are being used to mark resources created by specific test run.
* also tag nics
* removed deprovisioner step (it adds a couple of minutes to a test run, and the cluster is discarded anyway)
* cleanup will always run

## Test 

Created a VM in the RG of the test using the tags and confirmed it was removed by the cleanup, in both a successful and a failed run.

image

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -11,6 +11,7 @@ env:
   BUILTIN_CLOUD_PROVIDERS: "azure"
   TEST_E2E_CREATE_RG: "no"
   ACR_URL: "${{ vars.AZURE_ACR_URL }}"
+  TEST_TAGS: "owner=github-actions,run=${{ github.run_id }}-${{ github.run_attempt }}"
 
 on:
   schedule:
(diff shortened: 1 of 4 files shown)


.github/workflows/azure-e2e-test.yml
```diff
@@ -110,14 +111,15 @@ jobs:
     needs:
     - generate-podvm-image-version
     - build-caa-container-image
+    # when none of of required steps failed, skipped is ok
```
kartikjoshi21, line 114: typo -> none of the

kartikjoshi21 (approved):
LGTM, Thanks @mkulke, this will be good to ensure no resource leakage.

bpradipt (approved):
/lgtm

Issue #1895, opened 2024-07-02
Set initdata in PeerPod 


huoqifeng:
This story follows up https://github.com/kata-containers/kata-containers/issues/9468  in cloud-api-adaptor for PeerPod

The initdata schema was discussed, agreed and documented in https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md, a possible approach is documented in https://github.com/confidential-containers/confidential-containers/issues/171#issuecomment-1922821565

I drew a new diagram to explain more details for https://github.com/confidential-containers/confidential-containers/issues/171#issuecomment-1922821565 In PeerPod, as below:
<img width="840" alt="image" src="https://github.com/confidential-containers/cloud-api-adaptor/assets/15607193/55efd70a-1bc9-4e51-9c8b-f7cc07f38383">

The flow:
1. initdata added in pod descriptor as annotation, example:
```
{
  "algorithm": "sha384",
  "version": "0.1.0",
  "data": {
        "aa.toml": "xxx",
        "cdh.toml": "xxx",
        "certificate.crt": "xxxx",
        "policy.rego": "xxxx",
  }
}
```
2. The annotation passed from Containerd/CRI-O to kata-runtime
3. Generate the initdata files for `kata-cc` and `kata-remote-cc` separately
    - 3.1 for `kata-cc`, the initdata is wrote in pod's specific folder on host
    - 3.2 for `kata-remote-cc`, the initdata passed from kata-runtime to cloud-api-adaptor, cloud-api-adaptor write it to iso or meta-data service. 
4.  Mount the initdata files for `kata-cc` and `kata-remote-cc` separately
    - 4.1 for `kata-cc`, the initdata file will be mounted like via 9p.
    - 4.2 for `kata-remote-cc`, the initdata can be mounted as iso device like libvirt provider
    - 4.3 for `kata-remote-cc`, the initdata can also be retrieved from meta-data service like azure provider
5. initdata format after mount
    Initdata have formats similar as cloud-init yaml but with `"algorithm": "sha384"` and `"version": "0.1.0"` added, so that we can calculate hash based on specific algorithm
6. Provision initdata
    - for `kata-remote-cc`, use `process-user-data` to provision the configurations, like aa.toml, cdh.toml
    - for `kata-cc`, a similar process can be launched by init to do similar but it's not necessary to be a systemd service, it could be a program/scripts in initrd and launched by init 
7. Launch aa, cdh and asr etc
[...]

PR #1912, opened 2024-07-08
initdata: use annotation to provision config files

huoqifeng:
Fixes: #1895

- [X] depends on kata-containers change in https://github.com/kata-containers/kata-containers/pull/9988


huoqifeng:
Tried it on s390x for libvirt provider, in created PeerPod, check the files under `/run/peerpod`:
```
# ls /run/peerpod/
aa.toml  agent-config.toml  cdh.toml  checksum.txt  daemon.json  initdata.json  policy.rego
```
Make it ready to review and I'll add unit test code according to review comments.

huoqifeng:
Use pod yaml:
```
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: busybox
  name: busybox
  annotations:
    io.katacontainers.config.runtime.cc_init_data: YWxnb3JpdGhtID0gInNoYTM4NCIKdmVyc2lvbiA9ICIwLjEuMCIKCltkYXRhXQoiYWEudG9tbCIgPSAnJycKW3Rva2VuX2NvbmZpZ3NdClt0b2tlbl9jb25maWdzLmNvY29fYXNdCnVybCA9ICdodHRwOi8vMTI3LjAuMC4xOjgwODAnCgpbdG9rZW5fY29uZmlncy5rYnNdCnVybCA9ICdodHRwOi8vMTI3LjAuMC4xOjgwODAnCicnJwoKImNkaC50b21sIiAgPSAnJycKc29ja2V0ID0gCmNyZWRlbnRpYWxzID0gW10KW2tiY10KbmFtZSA9ICdjY19rYmMnCnVybCA9ICdodHRwOi8vMS4yLjMuNDo4MDgwJywKJycnCgoicG9saWN5LnJlZ28iID0gJycnCnBhY2thZ2UgYWdlbnRfcG9saWN5CgppbXBvcnQgZnV0dXJlLmtleXdvcmRzLmluCmltcG9ydCBmdXR1cmUua2V5d29yZHMuZXZlcnkKCmltcG9ydCBpbnB1dAoKIyBEZWZhdWx0IHZhbHVlcywgcmV0dXJuZWQgYnkgT1BBIHdoZW4gcnVsZXMgY2Fubm90IGJlIGV2YWx1YXRlZCB0byB0cnVlLgpkZWZhdWx0IENvcHlGaWxlUmVxdWVzdCA6PSBmYWxzZQpkZWZhdWx0IENyZWF0ZUNvbnRhaW5lclJlcXVlc3QgOj0gZmFsc2UKZGVmYXVsdCBDcmVhdGVTYW5kYm94UmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgRGVzdHJveVNhbmRib3hSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBFeGVjUHJvY2Vzc1JlcXVlc3QgOj0gZmFsc2UKZGVmYXVsdCBHZXRPT01FdmVudFJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IEd1ZXN0RGV0YWlsc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IE9ubGluZUNQVU1lbVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFB1bGxJbWFnZVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFJlYWRTdHJlYW1SZXF1ZXN0IDo9IGZhbHNlCmRlZmF1bHQgUmVtb3ZlQ29udGFpbmVyUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgUmVtb3ZlU3RhbGVWaXJ0aW9mc1NoYXJlTW91bnRzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU2lnbmFsUHJvY2Vzc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFN0YXJ0Q29udGFpbmVyUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU3RhdHNDb250YWluZXJSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBUdHlXaW5SZXNpemVSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBVcGRhdGVFcGhlbWVyYWxNb3VudHNSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBVcGRhdGVJbnRlcmZhY2VSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBVcGRhdGVSb3V0ZXNSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBXYWl0UHJvY2Vzc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFdyaXRlU3RyZWFtUmVxdWVzdCA6PSBmYWxzZQonJyc=
spec:
  containers:
  - image: quay.io/prometheus/busybox
    name: busybox
    resources: {}
  dnsPolicy: ClusterFirst
  restartPolicy: Never
  runtimeClassName: kata-remote
```
Where `io.katacontainers.config.runtime.cc_init_data` comes from base64 string of toml:
```
algorithm = "sha384"
version = "0.1.0"

[data]
"aa.toml" = '''
[token_configs]
[token_configs.coco_as]
url = 'http://127.0.0.1:8080'

[token_configs.kbs]
url = 'http://127.0.0.1:8080'
'''

"cdh.toml"  = '''
socket = 
credentials = []
[kbc]
name = 'cc_kbc'
url = 'http://1.2.3.4:8080',
'''

"policy.rego" = '''
package agent_policy

import future.keywords.in
import future.keywords.every

import input

# Default values, returned by OPA when rules cannot be evaluated to true.
default CopyFileRequest := false
default CreateContainerRequest := false
default CreateSandboxRequest := true
default DestroySandboxRequest := true
default ExecProcessRequest := false
default GetOOMEventRequest := true
default GuestDetailsRequest := true
default OnlineCPUMemRequest := true
default PullImageRequest := true
default ReadStreamRequest := false
default RemoveContainerRequest := true
default RemoveStaleVirtiofsShareMountsRequest := true
default SignalProcessRequest := true
default StartContainerRequest := true
default StatsContainerRequest := true
default TtyWinResizeRequest := true
default UpdateEphemeralMountsRequest := true
default UpdateInterfaceRequest := true
default UpdateRoutesRequest := true
default WaitProcessRequest := true
default WriteStreamRequest := false
'''
```
Check PeerPod:
```
# ls /run/peerpod/
aa.toml  agent-config.toml  cdh.toml  checksum.txt  daemon.json  initdata.meta  policy.rego
```

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -12,11 +15,27 @@ import (
...
+	ConfigParent = "/run/peerpod"
+
+	AaCfgPath     = "/run/peerpod/aa.toml"
+	AgentCfgPath  = "/run/peerpod/agent-config.toml"
+	AuthJsonPath  = "/run/peerpod/auth.json"
+	CdhCfgPath    = "/run/peerpod/cdh.toml"
+	DaemonCfgPath = "/run/peerpod/daemon.json"
+
+	InitdataMeta = "/run/peerpod/initdata.meta"
+	CheckSumPath = "/run/peerpod/checksum.txt"
```
mkulke, line 32: what is the purpose of checksum path and initdata.meta? is this documented somewhere (how) would it be different from what's documented in the trustee init-data spec?

mkulke (commented):
I think we want to bump the kata containers revision a commit that works?

I think we should remove AA_KBC_PARAMS support together with this change, otherwise it would be weird to have per-deployment and per-pod aa/cdh configuration overlapping.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: Thanks @mkulke , Spec here https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md mentioned initdata digest calculation but not cover this indeed, I'll open a PR to enhance the spec.

huoqifeng:
> I think we want to bump the kata containers revision a commit that works?
> 
> I think we should remove AA_KBC_PARAMS support together with this change, otherwise it would be weird to have per-deployment and per-pod aa/cdh configuration overlapping.

Thanks @mkulke for remind it, I just realized KBS is deployed dynamically in e2e test and using `AA_KBC_PARAMS` here https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/test/e2e/main_test.go#L189, Maybe we need keep it before change the KBS deployment in e2e test.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: https://github.com/confidential-containers/trustee/pull/450

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -286,7 +288,46 @@ func (s *cloudService) CreateVM(ctx context.Context, req *pb.CreateVMRequest) (r
...
+		cloned.Data = nil
+		clonedToml, err := toml.Marshal(cloned) // Data field omitted
+		if err != nil {
+			return nil, fmt.Errorf("Error Marshal cloned initdata: %w", err)
+		}
+		cloudConfig.WriteFiles = append(cloudConfig.WriteFiles, cloudinit.WriteFile{
+			Path:    userdata.InitdataMeta,
+			Content: string(clonedToml),
+		})
+	} else if s.aaKBCParams != "" { // Keep AA_KBC_PARAMS support as it is used by e2e test, KBS is dynamic k8s service in e2e test
```
liudalibj, line 330: It's fine to keep the AA_KBC_PARAMS in this PR, but we should have follow up TODO story to fully remove AA_KBC_PARAMS support with e2e-test cases update.
maybe update the comment add a TODO/FIXME keyword?

liudalibj (approved):
Overall LGTM, thanks @huoqifeng

src/cloud-api-adaptor/docs/initdata.md
```diff
@@ -0,0 +1,112 @@
...
+## Provision initdata files.
+`/run/peerpod/daemon.json`, `/run/peerpod/aa.toml`, `/run/peerpod/cdh.toml` and `/run/peerpod/policy.rego` will be provisioned.
+
+It also generates a meta file like `/run/peerpod/initdata.meta`:
+```toml
+algorithm = "sha384"
+version = "0.1.0"
+```
+
+Then it calculates the digest `/run/peerpod/initdata.digest` based on the `algorithm` in `/run/peerpod/initdata.meta` and the contents of static files `/run/peerpod/aa.toml`, `/run/peerpod/cdh.toml` and `/run/peerpod/policy.rego`. While `/run/peerpod/daemon.json` will be skipped when calculating the digest because it's dynamical for each instance.
```
mkulke, line 110: auth.json is missing from this list.

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -53,6 +85,12 @@ type CloudConfig struct {
 	WriteFiles []WriteFile `yaml:"write_files"`
 }
 
+type InitData struct {
+	Algorithom string            `toml:"algorithm"`
```
mkulke, line 89:
```suggestion
	Algorithm string            `toml:"algorithm"`
```

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -196,13 +230,78 @@ func processCloudConfig(cfg *Config, cc *CloudConfig) error {
...
+		hash := sha256.Sum256(byteData)
+		checksumStr = hex.EncodeToString(hash[:])
+	case "sha384":
+		hash := sha512.Sum384(byteData)
+		checksumStr = hex.EncodeToString(hash[:])
+	case "sha512":
+		hash := sha512.Sum512(byteData)
+		checksumStr = hex.EncodeToString(hash[:])
+	default:
+		return fmt.Errorf("Error creating initdata hash, the algorothom %s not supported", initdata.Algorithom)
```
mkulke, line 297:
```suggestion
		return fmt.Errorf("Error creating initdata hash, the algorithm %s not supported", initdata.Algorithm)
```

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -196,13 +230,78 @@ func processCloudConfig(cfg *Config, cc *CloudConfig) error {
...
+	checksumStr := ""
+	var byteData []byte
+	for _, file := range cfg.staticFiles {
+		if _, err := os.Stat(file); err == nil {
+			logger.Printf("calculateUserDataHash and reading file %s\n", file)
+			bytes, err := os.ReadFile(file)
+			if err != nil {
+				return fmt.Errorf("Error reading file %s: %v", file, err)
+			}
+			byteData = append(byteData, bytes...)
```
mkulke, line 282: can we use a hash extension algorithm (like in a TEE launch measurements or TPM registers) if we just concat all bytes the digest of file1 "abc" + file2 "def" will be the same as the digest of file1 "ab" and file2 "cdef".

it could also define the order in which the digest is calculated, if we use a hash extension algorithm:

```
zeros = [0;32];
previous_digest = zeros
for file in files:
  previous_digest = sha256_digest(previous_digest + bytes(file))
```

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -212,18 +311,25 @@ func ProvisionFiles(cfg *Config) error {
 	ctx, cancel := context.WithTimeout(bg, duration)
 	defer cancel()
 
-	provider, err := newProvider(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to create UserData provider: %w", err)
-	}
+	// some providers provision config files via process-user-data
+	// some providers rely on cloud-init provisin config files
```
mkulke, line 315:
```suggestion
	// some providers rely on cloud-init to provision config files
```

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -212,18 +311,25 @@ func ProvisionFiles(cfg *Config) error {
...
+	// all providers need calculate the hash value for attesters usage
```
mkulke, line 316: Can we make the digest calculation its own subcommand, instead of choosing a code path?

genjuro214 (approved):

mkulke (commented):
A couple of comments/questions:

- I suppose we ignore the problem of large policy bodies that cannot be provisioned via IMDS user-data? that's fine with me, but we should document that restriction in the initdata.md (as a TODO: we should verify that boundary in the cloud-provider code of CAA, I don't know what happens with larger user-data bodies, if they are truncated and then things get weird).

- I assume with `/run/peerpod/initdata.digest` we need to couple guest-components/attester code to a specific initdata implementation in Peerpods. So, a peerpods-only attester would have to read that file in the `attestation-agent::check_init_data()`?

For platforms without `HOSTDATA` (or similar at-launch fields for other TEEs) `check_init_data()` doesn't work, but we need `attestation-agent::extend_runtime_measurement()`. So instead of writing to `initdata.digest()` we need to call this ttrpc fn with the hash. which creates the same AA bootstrapping problem that currently exists with `AA::SetInitData()`: we need AA before/while we write the config file

mkulke (commented):
another question:

will the produced hash be different from non peerpod runtime classes? Can we have a pod with an initdata annotation and change the runtime-class from "kata-remote" to "qemu-tdx" and it would work in the same way?

Edit: 

There might be an option to re-construct the original initdata toml from the annotation 1:1 in process-user-data and hash it the full body with filenames, not just the content of the files:

```
digest = hash(`
"aa.toml" = '''
${aa_toml_str}
'''
"cdh.toml" = '''
...
`)
```

this would also solve the problem of ambiguous hashes.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: I'm thinking we'll remove `cloud-init` incrementally, the process-user-data need add other providers implementation step by step. And the hash calculation should be just one step of the whole process. So I prefer keep it in same command, wdyt?

huoqifeng:
> another question:
> 
> will the produced hash be different from non peerpod runtime classes? Can we have a pod with an initdata annotation and change the runtime-class from "kata-remote" to "qemu-tdx" and it would work in the same way?
> 
> Edit:
> 
> There might be an option to re-construct the original initdata toml from the annotation 1:1 in process-user-data and hash it the full body with filenames, not just the content of the files:
> 
> ```
> digest = hash(`
> "aa.toml" = '''
> ${aa_toml_str}
> '''
> "cdh.toml" = '''
> ...
> `)
> ```
> 
> this would also solve the problem of ambiguous hashes.

Sounds like this approach is more generic. I'll drop a commit to try it.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: I agree `re-construct the original initdata toml from the annotation 1:1 in process-user-data and hash it` might be a better approach.

huoqifeng:
> * I assume with `/run/peerpod/initdata.digest` we need to couple guest-components/attester code to a specific initdata implementation in Peerpods. So, a peerpods-only attester would have to read that file in the `attestation-agent::check_init_data()`?
> 
> 
> For platforms without `HOSTDATA` (or similar at-launch fields for other TEEs) `check_init_data()` doesn't work, but we need `attestation-agent::extend_runtime_measurement()`. So instead of writing to `initdata.digest()` we need to call this ttrpc fn with the hash. which creates the same AA bootstrapping problem that currently exists with `AA::SetInitData()`: we need AA before/while we write the config file

For IBM SE, it's pretty simple, we can add the digest in evidence filed like https://github.com/confidential-containers/guest-components/pull/616/files and verify it in AS side by adding the value in attestation policy before hand. I'm not very familiar with the other TEE, @mkulke would you mind give some suggestions? I can add them in the md file.

src/cloud-api-adaptor/pkg/userdata/provision.go
mkulke: yes, if we reconstruct the full initdata toml, there is no need for hash extension, because there is no ambiguity.

src/cloud-api-adaptor/pkg/userdata/provision.go
mkulke: we have to see how the final picture looks like, but we can add a dedicated command later, if required

mkulke:
> For IBM SE, it's pretty simple, we can add the digest in evidence filed like https://github.com/confidential-containers/guest-components/pull/616/files and verify it in AS side by adding the value in attestation policy before hand. I'm not very familiar with the other TEE, @mkulke would you mind give some suggestions? I can add them in the md file.

I don't have a good idea at the moment, we will need additional code in any case.

I am leaning towards ignoring attestation-agent at this point and extending PCR manually in a systemd unit after launching process_user_data. this will only work for TEEs supporting TPMs, but that's maybe ok for now until we need to support more.

mkulke (commented):
another thought that came up:

TOML does not have a canonical representation (there's many ways to write a legit toml). so, I don't think we can "reconstruct" it 1:1. I think we cannot split the initdata on the CAA, but we have to do it on the guest (in proces-user-data)

huoqifeng:
> another thought that came up:
> 
> TOML does not have a canonical representation (there's many ways to write a legit toml). so, I don't think we can "reconstruct" it 1:1. I think we cannot split the initdata on the CAA, but we have to do it on the guest (in proces-user-data)

Right, this is what I just met.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: @mkulke I'm keeping existing algorithm for 2 reasons: 1. The toml marshall can not guarantee the 1:1 format when reconstruct the initdata as you also found. 2. previous-digest+contents makes it difficult to calculate the hash value manually. So I updated the md file to guide users how to calculate the hash value manually in https://github.com/confidential-containers/cloud-api-adaptor/pull/1912/commits/687fb90160c467008271fc85a442d9532cce8d14

mkulke: I am not sure it's a good idea to have a special hash algorithm for Peerpods. This will be confusing to users (and future initdata tooling). As far as I can see, we don't gain anything from splitting the initdata.toml into individual files on the node, we could just split the initdata on the guest (in process-user-data) and calculate the hash of the full body.

huoqifeng: @mkulke do you mean create a file like
```yaml
write_files:
- path: /run/peerpod/daemon.json
  content: 
- path: /run/peerpod/initdata.toml
  content:
 ```
And then use `process-user-data` to parse `/run/peerpod/initdata.toml`? I guess it's doable for providers which still depend on cloud-init.

Comment (huoqifeng) [reply src/cloud-api-adaptor/pkg/userdata/provision.go]: @mkulke I dropped a commit to use the annotation for initdata directly here https://github.com/confidential-containers/cloud-api-adaptor/pull/1912/commits/96d3466809c93711b07f7eac19837a581af074c7, wdyt this approach?

File: src/cloud-api-adaptor/docs/initdata.md [hunk 9]
```diff
@@ -0,0 +1,111 @@
...
+url = 'http://127.0.0.1:8080'
+'''
+
+"cdh.toml"  = '''
+socket = 'unix:///run/confidential-containers/cdh.sock'
+credentials = []
+
+[kbc]
+name = 'cc_kbc'
+url = 'http://1.2.3.4:8080'
```
bpradipt, line 30: @huoqifeng what about the "cert" property when using https KBS?

huoqifeng: Yes, `cert` can be added directly in `aa.toml` section in initdata. @bpradipt

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -149,60 +164,82 @@ func parseUserData(userData []byte) (*CloudConfig, error) {
...
+	initdata := InitData{}
+	err = toml.Unmarshal(decodedBytes, &initdata)
+	if err != nil {
+		return fmt.Errorf("Error unmarshalling initdata: %w", err)
 	}
 
-	for _, e := range entries {
-		err := e.writeFile(cc)
+	for key, value := range initdata.Data {
+		err := writeFile(filepath.Join(cfg.parentPath, key), []byte(value))
```
mkulke, line 217: what if `key` is `/run/coco/aa.toml`? I suppose it will fail, if `/run/coco` doesn't exist. Should we maybe specify a clear allowlist of valid `key` values and ignore all other keys? I think `StaticFiles` was supposed to be this, but it's not used as far as I can see?

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -149,60 +164,82 @@ func parseUserData(userData []byte) (*CloudConfig, error) {
...
 	}
+	logger.Printf("Wrote %s\n", path)
 	return nil
 }
 
-type entry struct {
-	path     string
-	optional bool
+func processCloudConfig(cfg *Config, cc *CloudConfig) error {
+	for _, wf := range cc.WriteFiles {
```
mkulke, line 183: I would not write arbitrary files in cloud-config, but just the ones that we expect, so a specific allow-list with path `/run/peerpod/daemon.json`, `/run/peerpod/auth.json`, ...

mkulke (commented):
looks good! a couple of nits

bpradipt:
Few things are unclear to me (most likely I missed) and hence asking question here
1. Will initData support via annotation be the only way to configure aa and cdh? 
2. If the answer to 1 is yes, will AA_KBC_PARAMS still be used to set the annotation in the pod spec ?

Also, I think if some commits are squashed and cleaned up it'll be easier to review :-)

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: Yes, added allowlist for items can be in `initdata` and `write_files`.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: Yes, added allowlist for items can be in initdata and write_files.

huoqifeng:
> 2\. AA_KBC_PARAMS

@bpradipt `AA_KBC_PARAMS` is not removed is because e2e test deploy KBS dynamically and the test cases leverage it. I'll considering drop `AA_KBC_PARAMS` after revise the related test cases. And I'll squash the commits later.

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -149,60 +162,102 @@ func parseUserData(userData []byte) (*CloudConfig, error) {
...
 
-type entry struct {
-	path     string
-	optional bool
+func isAllowed(path string, filesList []string) bool {
+	for _, listedFile := range filesList {
+		if listedFile == path {
+			return true
+		}
+		if strings.HasSuffix(path, listedFile) {
```
mkulke, line 185: is there a reason why we don't check for the full path, but only the suffix? Afaik we hardcode the pathes when generating the `write_files` directive, so we could also use the full path here.

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -2,45 +2,52 @@ package userdata
...
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers/docker"
+	toml "github.com/pelletier/go-toml/v2"
 	"gopkg.in/yaml.v2"
 )
 
-var logger = log.New(log.Writer(), "[userdata/provision] ", log.LstdFlags|log.Lmsgprefix)
+const (
+	ConfigParent = "/run/peerpod"
+	DigestPath   = "/run/peerpod/initdata.digest"
+	InitdataPath = "/run/peerpod/initdata"
```
mkulke, line 27: since coco initdata is specified to be a in toml format, do we want to call it `/run/peerpod/initdata.toml`?

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -2,45 +2,52 @@ package userdata
...
 
-type paths struct {
-	aaConfig     string
-	agentConfig  string
-	authJson     string
-	cdhConfig    string
-	daemonConfig string
-}
+var logger = log.New(log.Writer(), "[userdata/provision] ", log.LstdFlags|log.Lmsgprefix)
+var WriteFilesList = []string{"agent-config.toml", "daemon.json", "auth.json", "initdata"}
```
mkulke, line 31: Maybe we can import those paths from the adapator/cloud (or the other way around, export the paths consts from here and use them in adaptor/cloud). I would also suggest using the full path (see comment below)

mkulke:
> @bpradipt `AA_KBC_PARAMS` is not removed is because e2e test deploy KBS dynamically and the test cases leverage it. I'll considering drop `AA_KBC_PARAMS` after revise the related test cases. And I'll squash the commits later.

I think there are 2 issues we have to consider before merging init-data

1) It is a kind of undefined/undocumented breaking change. If I deploy a pod with initdata annotation my AA_KBC_PARAMs will be overwritten (because write_files is processed first, then initdata overwrites the same files). I would avoid that.
2) There is no e2e tests covering this feature, which generally is something that we should avoid.
3) The initdata feature is not 100% ready, we still need to write code that is consuming initdata.digest.

So, one idea to still get this merged in this state would be:
- Comment out the initdata file-provisioning code, only calculate/write initdata.digest. We can start to write code that is working with initdata.digest. 
- The AA_KBC_PARAMS mechanism will stay in place and work as expected in the meantime.
- In a (atomic) successor PR we:
  - Remove the AA_KBC_PARAMS logic
  - Adjust documentation
  - Enable file provisioning from initdata
  - Convert the kbs e2e tests to use initdata. Ideally those would use KBS cert + measurements/runtime_data checks for initdata.digest.
  
btw, I don't know how the remote attestation verification of initdata would look like in the sample attester, it probably won't work unless we extend the sample attester to keep some in-memory log that is attached to the evidence?

huoqifeng:
> > @bpradipt `AA_KBC_PARAMS` is not removed is because e2e test deploy KBS dynamically and the test cases leverage it. I'll considering drop `AA_KBC_PARAMS` after revise the related test cases. And I'll squash the commits later.
> 
> I think there are 2 issues we have to consider before merging init-data
> 
>     1. It is a kind of undefined/undocumented breaking change. If I deploy a pod with initdata annotation my AA_KBC_PARAMs will be overwritten (because write_files is processed first, then initdata overwrites the same files). I would avoid that.
> 
>     2. There is no e2e tests covering this feature, which generally is something that we should avoid.
> 
>     3. The initdata feature is not 100% ready, we still need to write code that is consuming initdata.digest.
> 
> 
> So, one idea to still get this merged in this state would be:
> 
>     * Comment out the initdata file-provisioning code, only calculate/write initdata.digest. We can start to write code that is working with initdata.digest.
> 
>     * The AA_KBC_PARAMS mechanism will stay in place and work as expected in the meantime.
> 
>     * In a (atomic) successor PR we:
>       
>       * Remove the AA_KBC_PARAMS logic
>       * Adjust documentation
>       * Enable file provisioning from initdata
>       * Convert the kbs e2e tests to use initdata. Ideally those would use KBS cert + measurements/runtime_data checks for initdata.digest.
> 
> 
> btw, I don't know how the remote attestation verification of initdata would look like in the sample attester, it probably won't work unless we extend the sample attester to keep some in-memory log that is attached to the evidence?

The key problem to remove `AA_KBC_PARAMs` and not be able to add e2e test cases for `initdata` is because the KBS deployment in our e2e test, which is dynamic and requires to update initdata in test case, it's doable but need more times. It's simpler in a real deployment or production env. Maybe, we can keep `AA_KBC_PARAMs` as-is and use initdata to provision all the configures only if the annotation provided to make sure not breaking existing algorithm. It's `experimental phase`. We then add test cases for initdata and drop `AA_KBC_PARAMs` in successor PRs. @bpradipt @mkulke what do you think?

mkulke:
> Maybe, we can keep AA_KBC_PARAMs as-is and use initdata to provision all the configures only if the annotation provided to make sure not breaking existing algorithm

Not sure I understand fully. Do you mean if `AA_KBC_PARAMS` is set we ignore the initdata annotation? if that's the case: yes, that would work, I think. It would remove the ambiguities of who is actually writing cdh.toml + aa.toml. I woulld add a note in the log that we ignore initdata annotation, though.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: Yes, we can let userdata depends on cloud.

src/cloud-api-adaptor/pkg/userdata/provision.go
huoqifeng: Agreed, we can match the full path.

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -149,60 +162,102 @@ func parseUserData(userData []byte) (*CloudConfig, error) {
...
-	bytes := findConfigEntry(f.path, cc)
-	if bytes == nil {
-		if !f.optional {
-			return fmt.Errorf("failed to find %s entry in cloud config", f.path)
+func processCloudConfig(cfg *Config, cc *CloudConfig) error {
+	for _, wf := range cc.WriteFiles {
+		path := wf.Path
+		bytes := []byte(wf.Content)
+		if isAllowed(path, cfg.writeFiles) {
+			if bytes != nil {
```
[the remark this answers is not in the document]
huoqifeng: Right, bytes should never be nil as it's wrote by caa together with its path.

src/cloud-api-adaptor/docs/initdata.md
```diff
@@ -0,0 +1,113 @@
...
+default StatsContainerRequest := true
+default TtyWinResizeRequest := true
+default UpdateEphemeralMountsRequest := true
+default UpdateInterfaceRequest := true
+default UpdateRoutesRequest := true
+default WaitProcessRequest := true
+default WriteStreamRequest := false
+'''
+```
+
```
liudalibj, line 67: Maybe add **Note** item: `if the kbs server is https enabled`
- in aa.toml need added:
```
cert = '''
...
'''
```
refer to https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/config.example.toml

- in cdh.toml need added
```
kbs_cert = '''
...
'''
```
refer to https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.toml

bpradipt:
> > Maybe, we can keep AA_KBC_PARAMs as-is and use initdata to provision all the configures only if the annotation provided to make sure not breaking existing algorithm
> 
> Not sure I understand fully. Do you mean if `AA_KBC_PARAMS` is set we ignore the initdata annotation? if that's the case: yes, that would work, I think. It would remove the ambiguities of who is actually writing cdh.toml + aa.toml. I woulld add a note in the log that we ignore initdata annotation, though.

This makes sense. If AA_KBC_PARAMS is set CAA can ignore the initdata annotation and log it.

src/cloud-api-adaptor/docs/initdata.md
huoqifeng: Added certs in example and updated the test cases, thanks Dali.

mkulke (approved):
thanks for taking this on - and accommodating all the suggestions!

Issue #1882, opened 2024-06-24
ibmcloud: Peer pods fail during CreateContainer


stevenhorsman:
When creating an ibmcloud set up on with a self-managed cluster with both s390x and amd64 architectures, the tests fail.

The pod describe looks like:
```
Events:
  Type     Reason   Age                     From     Message
  ----     ------   ----                    ----     -------
  Normal   Pulling  49m (x19 over 134m)     kubelet  Pulling image "quay.io/prometheus/busybox:latest"
  Warning  Failed   23m (x23 over 133m)     kubelet  Error: failed to create containerd task: failed to create shim task: context deadline exceeded
  Warning  BackOff  4m25s (x482 over 132m)  kubelet  Back-off restarting failed container busybox in pod simple-test_coco-pp-e2e-test-94b410d8(171202c1-07d7-4f95-b541-b9dadc10dbbe)
```

and CAA log shows and error during the CreateContainer (which includes the pull image step):
```
2024/06/24 16:26:45 [adaptor/proxy]     storages:
2024/06/24 16:26:45 [adaptor/proxy]         mount_point:/run/kata-containers/702a9450e5570a71633834ec4c5f6f407100921862a9015d96038de8518df2f2/rootfs source:quay.io/prometheus/busybox:latest fstype:overlay driver:image_guest_pull
2024/06/24 16:26:49 [adaptor/proxy] CreateContainer fails: context deadline exceeded
time="2024-06-24T16:26:49Z" level=error msg="ttrpc: received message on inactive stream" stream=3603
```

I need to dig into the kata-agent logs and see if there is any more information about this.

stevenhorsman:
Looking in the kata-agent log it has the `info` message
```
{"msg":"pull image \"docker.io/library/nginx@sha256:9700d098d545f9d2ee0660dfb155fe64f4447720a0a763a93f2cf08997227279\", bundle path \"/run/kata-containers/3a9d18335128ca98c7d1f9d86aaad6922c063eeff135ab977ea164fa5ff60dcf/images\"","level":"INFO","ts":"2024-06-26T13:08:03.03219308Z","name":"kata-agent","subsystem":"image","source":"agent","pid":"810","version":"0.1.0"}
```
from https://github.com/kata-containers/kata-containers/blob/893fd2b59cc31518f8a127c9611e3e8265d9bdfd/src/agent/src/image.rs#L160

[...]

PR #1914, opened 2024-07-09
ibmcloud: revise ibmcloud README

genjuro214:
I followed the guide to [Deploy the Confidential-containers operator](https://github.com/confidential-containers/cloud-api-adaptor/tree/main/src/cloud-api-adaptor/ibmcloud#deploy-the-confidential-containers-operator) on a self-managed cluster in IBM Cloud VPC. However, it doesn't work as expected. I try to fix 2 problems:

- Fix the default VPC service url, revert it to the value in v0.8.0: https://github.com/confidential-containers/cloud-api-adaptor/blob/v0.8.0/pkg/adaptor/cloud/ibmcloud/provider.go#L79

- Add missing settings in selfmanaged_cluster.properties.  Without those settings, CAA will fail to create instance with errors like below:
```
failed to create an instance : Expected only one oneOf fields to be set: got 0 and the response is {

"code": "validation_failed_oneof",
                "message": "Expected only one oneOf fields to be set: got 0",
                "target": {
                    "name": "InstancePrototype",
                    "type": "field"
                }

"message": "InstanceByImage is invalid: Expected only one oneOf fields to be set: got 0",
```

### DIFF
--- a/src/cloud-api-adaptor/ibmcloud/README.md
+++ b/src/cloud-api-adaptor/ibmcloud/README.md
@@ -134,6 +134,12 @@ PODVM_IMAGE_ID="$PODVM_IMAGE_ID"
 INSTANCE_PROFILE_NAME="$PODVM_INSTANCE_PROFILE"
 CAA_IMAGE_TAG="$CAA_IMAGE_TAG"
 SSH_KEY_ID="$(terraform output --raw ssh_key_id)"
+REGION="$(terraform output --raw region)"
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):
The changes look fine. Just to note that these won't fix the issues I'm seeing with #1882 as I've been providing the extra properties in self_managed.properties and the VpcServiceUrls is overwritten in our e2e test code to the correct value.

liudalibj (approved):
LGTM

PR #1916, opened 2024-07-10
docker: Switch to fedora and mkosi based binaries

bpradipt:
Switch to fedora base image with systemd and reuse the binaries generated for mkosi based builds.
The docker image is inspired from Kind node image


bpradipt:
@stevenhorsman this should help with switching to a single distro..

bpradipt:
Does anyone have access to the license compliance report to figure out what's being flagged ?

stevenhorsman:
> Does anyone have access to the license compliance report to figure out what's being flagged ?

I think it was @dcmiddle who had the FOSSA login details? We used to only get this triggered when PRs were created from branches not forks, but maybe some config has changed.

stevenhorsman (approved):
This looks okay to me. I tried the mkosi `make container` and `make image-container` options and they seemed to work as expected. Thanks!

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm
```diff
@@ -0,0 +1,44 @@
+# Adapted from https://github.com/kubernetes-sigs/kind/blob/main/images/base/Dockerfile
```
huoqifeng, line 1: It's better move this Dockerfile to podvm/Dockerfile_podvm_docker to align with other distro.
I guess we'll move all Dockerfile_xx_xx from podvm to podvm-mkosi after sunset pakcer, we can do it for all distro at same time to avoid confusion.

bpradipt: It does make sense to have all dockerfiles in podvm, but I was in two minds.
Let's do it once we switch to single distro + method to avoid confusion.
@huoqifeng wdyt ?

huoqifeng: Sure @bpradipt , we can make the change later.

huoqifeng (approved):
LGTM

PR #1920, opened 2024-07-12
podnetwork: Support CNI plugins like PTP and GKE 

yoheiueda:
CNI plugins like PTP and GKE remove a route that is automatically added by kernel for eth0, and then add another route for the same destination.

This patch changes the code to manipulates routes to support such CNI plugins.

Fixes #1909

Issue #1280, opened 2023-07-31
Add ability to assign custom tags to the pod vm cloud instances


bpradipt:
In many environments, there is a need to add specific tags to the VM instances for different queries.
Currently there is no way to add custom tags to the vm instances. This issue is to track the work required to add custom tags.

PR #1926, opened 2024-07-17
Adding initial support for GCP

beraldoleal:
This is basically Cfir' work with me changing paths and small things, to accommodate repository layout changes a few months ago.

We had this on hold because I was working on GKE support as well. However, GKE right now depends on https://github.com/confidential-containers/cloud-api-adaptor/pull/1920. So my proposal is to merge this, since it will not affect other providers, and when #1920 is merged I will send another PR with the GKE support.


wainersm:
Hi @beraldoleal !

I've already started my review but I'd like to send a few words before making comments in code:
 * I'm not an expert on GCP, actually I barely know it. So always be suspicious about my comments .
 * I'm largely comparing this implementation with Azure and AWS providers.
 * I assume this implementation is a MVP, so some features (e.g. confidential VMs) are missing on purpose. Even though, I may make some comments so that we start building a gap list (if you haven't such as list already). These comments shouldn't block the merge of this work.
 * I may have some difficult to test it entirely as I don´t  have a GCP account (still looking for one). I´ll try my best.

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+						Value: proto.String("base64"),
+					},
+				},
+			},
+			MachineType: proto.String(fmt.Sprintf("zones/%s/machineTypes/%s", p.serviceConfig.Zone, p.serviceConfig.MachineType)),
+			NetworkInterfaces: []*computepb.NetworkInterface{
+				{
+					AccessConfigs: []*computepb.AccessConfig{
+						{
+							Name:        proto.String("External NAT"),
```
wainersm, line 126: IIUC this will create a VM with public IP address. The AWS implementation creates a VM with private IP by default but it allow users to choose to use public IP (see commit 39d69d494b982f008e0b1767469cf914b1fa835f). Other providers I don't know how it was implemented.

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+					},
+					StackType: proto.String("IPV4_Only"),
+					Name:      proto.String(p.serviceConfig.Network),
+				},
+			},
+		},
+	}
+	op, err := p.instancesClient.Insert(ctx, insertReq)
+	if err != nil {
+		return nil, fmt.Errorf("Instances.Insert error: %s. req: %v", err, insertReq)
```
wainersm, line 138: In case `Insert()` fails, will it automatically delete any resources (e.g. disks and networks) that were created? Asking because Azure does need to delete those resources explicitly, whereas other providers don't.

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	var podNodeIPs []netip.Addr
+	for _, nic := range instance.GetNetworkInterfaces() {
+		for _, access := range nic.GetAccessConfigs() {
+			ipStr := access.GetNatIP()
+			ip, err := netip.ParseAddr(ipStr)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse pod node IP %q: %w", ipStr, err)
+			}
+			podNodeIPs = append(podNodeIPs, ip)
+			logger.Printf("%s", ip.String())
```
wainersm, line 72: Suggestion: a meaningful log message, e.g., "Found pod node IP: ...."

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+					},
+				},
+			},
+			MachineType: proto.String(fmt.Sprintf("zones/%s/machineTypes/%s", p.serviceConfig.Zone, p.serviceConfig.MachineType)),
+			NetworkInterfaces: []*computepb.NetworkInterface{
+				{
+					AccessConfigs: []*computepb.AccessConfig{
+						{
+							Name:        proto.String("External NAT"),
+							NetworkTier: proto.String("PREMIUM"),
```
wainersm, line 127: When I read "premium" my brain automatically associates with "expensive". So it would be nice to have a option for users to tune this.

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	userData, err := cloudConfig.Generate()
+	if err != nil {
+		return nil, err
+	}
+
+	//Convert userData to base64
+	userDataEnc := base64.StdEncoding.EncodeToString([]byte(userData))
+	logger.Printf("userDataEnc:  %s", userDataEnc)
+
+	insertReq := &computepb.InsertInstanceRequest{
```
wainersm, line 92: You will need to improve the InstanceRequest to allow users to customize the VM. One concept that almost all providers implemented is of "flexible instance sizes" (please see https://github.com/confidential-containers/cloud-api-adaptor/issues/321).

Enable/disable confidential VM is also something desirable that I mentioned elsewhere.

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	err = op.Wait(ctx)
+	if err != nil {
+		return fmt.Errorf("waiting for Instances.Delete error: %s. req: %v", err, req)
+	}
+	logger.Printf("deleted an instance %s", instanceID)
+	return nil
+}
+
+func (p *gcpProvider) Teardown() error {
```
wainersm, line 189: Don't know what's the use of `Teardown()`, do you? Only Vsphere provider implements this...

src/peerpod-ctrl/Makefile
```diff
@@ -63,7 +63,7 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-BUILTIN_CLOUD_PROVIDERS ?= aws azure ibmcloud vsphere libvirt docker
+BUILTIN_CLOUD_PROVIDERS ?= aws azure gcp ibmcloud vsphere libvirt docker
```
wainersm, line 66: Bonus if you re-order alphabetically :)

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	//Convert userData to base64
+	userDataEnc := base64.StdEncoding.EncodeToString([]byte(userData))
+	logger.Printf("userDataEnc:  %s", userDataEnc)
+
+	insertReq := &computepb.InsertInstanceRequest{
+		Project: p.serviceConfig.ProjectId,
+		Zone:    p.serviceConfig.Zone,
+		InstanceResource: &computepb.Instance{
```
wainersm, line 95: Another missing feature is the ability for users to specify custom tags (see https://github.com/confidential-containers/cloud-api-adaptor/issues/1280)

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -85,6 +85,22 @@ azure() {
         ${optionals}
 }
 
+gcp() {
+test_vars GCP_CREDENTIALS
+
+[[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-gcp-image-name ${PODVM_IMAGE_NAME} "
+[[ "${GCP_PROJECT_ID}" ]] && optionals+="-gcp-project-id ${GCP_PROJECT_ID} "
+[[ "${GCP_ZONE}" ]] && optionals+="-gcp-zone ${GCP_ZONE} " # if not set retrieved from IMDS
```
wainersm, line 93: Does GCP has an equivalent to AWS's IMDS?

AWS implements https://github.com/confidential-containers/cloud-api-adaptor/issues/315 which will get information like the subnetId from the IMDS. This means that the subnetId parameter is optional.

Is GCP_PROJECT_ID and GCP_ZONE automatically filled if the user doesn't set? If not, then these variables should be listed in `test_vars` because they become mandatory.

`PODVM_IMAGE_NAME` should be mandatory, thus in `test_vars`.

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -85,6 +85,22 @@ azure() {
...
+[[ "${GCP_MACHINE_TYPE}" ]] && optionals+="-gcp-machine-type ${GCP_MACHINE_TYPE} " # default e2-medium
+[[ "${GCP_NETWORK}" ]] && optionals+="-gcp-network ${GCP_NETWORK} " # defaults to 'default'
+
+set -x
+exec cloud-api-adaptor gcp \
+	-pods-dir /run/peerpod/pods \
+	${optionals} \
+	-socket /run/peerpod/hypervisor.sock
```
wainersm, line 101: I think -socket isn't needed because it defaults to /run/peerpod/hypervisor.sock already.

src/cloud-api-adaptor/entrypoint.sh
beraldoleal: It does, and we are using in other parts, like: https://github.com/beraldoleal/cloud-api-adaptor/blob/d7578e6beb1e884be63888036a2748ef6347c857/src/cloud-providers/gcp/misc.go#L47.

This is just a basic implementation, for tech-preview, which will have lots of rooms for improvements. I will note this one as an issue once this gets merged, so we can improve latter. 

About the variables, you are right, sending the fix with the next force-push.

src/cloud-api-adaptor/entrypoint.sh
beraldoleal: I was just playing safe, since others are using it as well. But I will remove with the next force-push.

src/cloud-providers/gcp/provider.go
beraldoleal: Sure, fixed on next force-push.

src/cloud-providers/gcp/provider.go
beraldoleal: Makes sense, noted for a following PR, if you don't mind.

src/cloud-providers/gcp/provider.go
beraldoleal: lol, makes sense. I changed the default to STANDARD. But I will create an issue for future fine tuning.

src/cloud-providers/gcp/provider.go
beraldoleal: I tried to find any reference or documentation and could not find yet. I will keep looking, however:

Network is not created here, only attached, we are using the default one, for now. (To be improved).

With disks I tried simulating several possible situations that would fail, and the boot disk was not there. So I would say it is safe for now.

src/cloud-providers/gcp/provider.go
beraldoleal: Is part of the Provider Interface.

src/peerpod-ctrl/Makefile
beraldoleal: accepting only beers. Fixed.

src/cloud-providers/gcp/provider.go
beraldoleal: I would add this to the list, too. If you don't mind.

src/cloud-providers/gcp/provider.go
beraldoleal: @wainersm, Will the PodVM be able to communicate with external clusters within a private network? This is a genuine question.

Given that GKE is not supported at this point, I am concerned it may not add value currently.

beraldoleal:
Hi @wainersm, thanks for your review! Really appreciated!

Let me know if you are ok with the open threads, and even the closed ones, please feel free to reopen.

About the account, I will send you on slack the credentials to a GCP cluster, if you need.

beraldoleal:
@wainersm let me know if you are ok with the changes.

@bpradipt please consider review this when you find some spare cycles.

src/cloud-providers/gcp/provider.go
wainersm: Hi @beraldoleal ! Not expert speaking, but I guess CAA running on an external cluster won't be able to talk to the VM within a private network unless you provide the correct routing.

I'm not saying that you should implement the private IP on this first version. I just wanted to caught your attention to the fact that some providers like AWS provides that flexibility, and you might be interested on doing the same on the follow on PRs.

src/cloud-providers/gcp/provider.go
wainersm: OK. Can be improved and/or fixed later.

wainersm (approved):
Hi @beraldoleal !

I see that you followed the instructions in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md  which makes me more comfortable to review.

All the comments I made you addressed and/or will on following changes. Unfortunately I don't have GCP account to give it a try but I know that you tested it, so we are fine.

thanks!

wainersm:
@bpradipt @snir911 mind to review too?

bpradipt:
> @bpradipt @snir911 mind to review too?

Yeah, this is in pipeline. Should be doing soon.

src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+secretGenerator:
+- name: auth-json-secret
+  namespace: confidential-containers-system
+  files:
+  #- auth.json # set - path to auth.json pull credentials file
+- name: peer-pods-secret
+  namespace: confidential-containers-system
+  files:
+  - GCP_CREDENTIALS # make sure this file has the application credentials. You can reuse the Packer creds created in "Build Pod VM Image"
```
bpradipt, line 42: Can env file be used to provide the secrets followed by a local .gitignore file under `overlays/gcp`, similar to aws and azure? This will make it consistent.

https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/overlays/aws/kustomization.yaml#L55
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/overlays/azure/kustomization.yaml#L62

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
+// (C) Copyright Confidential Containers Contributors
```
bpradipt, line 1: Any plans to add unit tests and e2e tests as follow-on PRs ?

.gitignore
```diff
@@ -36,3 +36,6 @@ src/cloud-api-adaptor/*.qcow2
 .git-commit
 .git-commit.tmp
 *.tar.gz
+
+# Secrets
+install/overlays/gcp/GCP_CREDENTIALS
```
bpradipt, line 41: Adding the comment on the commit message here.  You can remove the statement on GKE now since 1909 is merged.

beraldoleal: Are you proposing we bring GKE support back?

If it's okay with you, can we proceed with this version to establish basic GCP support quickly? I can add GKE in a follow-up PR?

src/cloud-providers/gcp/provider.go
beraldoleal: Tests were part of the provisioning code, and we removed from this PR to wait for the GKE fix. But is currently implemented here:

https://github.com/confidential-containers/cloud-api-adaptor/compare/main...beraldoleal:cloud-api-adaptor:gcp-plus-ueda

I just need to check again and enable more test.

So, my plan was to add a follow up PR adding the provisioning + tests after this PR gets merged. Let me know if you agree with that.

.gitignore
bpradipt: I don't see anything which is specific to GKE and the same code will work for GKE (similar to what we have for aws/eks and azure/aks). Hence suggested to drop the statement from the commit message.
Or for GKE there is a need for separate code?

src/cloud-providers/gcp/provider.go
bpradipt: Makes sense to me. The tests can be added as a follow-on PR.

src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
beraldoleal: Yes.. I considered using env variables for GCP credentials. But GCP typically uses the credentials JSON files, which seems to be the standard among tools, docs and users. I think it might be simpler for users to stick with that. Env files make sense for Azure, but not sure if the same is valid for GCP. Converting each key (specially things like `private_key`) from the GCP credentials to individual variables could be a bit cumbersome.

Let’s discuss if you see it differently.

.gitignore
beraldoleal: Ohh understood now, fixed the commit message. thanks. :)

src/cloud-api-adaptor/gcp/README.md
```diff
@@ -0,0 +1,205 @@
...
+  - GCP_ZONE="" # set
+  - GCP_MACHINE_TYPE="e2-medium" # defaults to e2-medium
+  - GCP_NETWORK="global/networks/default" # leave as is.
+  ...
+secretGenerator:
+  ...
+- name: peer-pods-secret
+  namespace: confidential-containers-system
+  files:
+  - GCP_CREDENTIALS # Make sure this file has the application creadentials. You can use the Peerpods creds: copy the file from ${HOME}/.config/gcloud/peerpods_application_key.json
```
bpradipt, line 182: s/creadentials/credentials

src/cloud-providers/gcp/provider.go
```diff
@@ -0,0 +1,191 @@
...
+	if config.GcpCredentials != "" {
+		creds, err := google.CredentialsFromJSON(context.TODO(), []byte(config.GcpCredentials), computeScope)
+		if err != nil {
+			return nil, fmt.Errorf("configuration error when using creds: %s", err)
+		}
+		provider.instancesClient, err = compute.NewInstancesRESTClient(context.TODO(), option.WithCredentials(creds))
+		if err != nil {
+			return nil, fmt.Errorf("NewInstancesRESTClient with credentials error: %s", err)
+		}
+	} else {
```
bpradipt, line 52: Is having an empty GcpCredentials a valid case?

src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
bpradipt: Hmm, I see that the gcp code here expects the creds to be a json file. So I think it's ok for now. We can think about this later.

src/cloud-providers/gcp/provider.go
beraldoleal: Yes, depending on the deployment environment, the library can use ADC. I haven't deeply explored it, though.

src/cloud-api-adaptor/gcp/README.md
beraldoleal: Fixed.

bpradipt (approved):
/lgtm
Thanks @beraldoleal

beraldoleal:
Thank you @wainersm  and @bpradipt

PR #1927, opened 2024-07-18
addons: rhel kernel version matching

snir911:
According to https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/precompiled/ the correct way to install pre-compiled drivers is to update kernel and opt-into a precompiled modularity stream, however, it may take some time between latest kernel version is released until the precompiled pkgs are available; which may cause driver failure.
This commit adds a function which tries to update the kernel to the latest version supported by the nvidia precompiled drivers, if fails it upgrades to the latest kernel version available.

### DIFF
--- a/src/cloud-api-adaptor/podvm/addons/nvidia_gpu/setup.sh
+++ b/src/cloud-api-adaptor/podvm/addons/nvidia_gpu/setup.sh
@@ -20,6 +20,29 @@ END
 chmod +x /usr/share/oci/hooks/prestart/nvidia-container-toolkit.sh
 
 
+# according to https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/precompiled/
+# the correct way to install pre-compiled drivers is to update kernel and opt-into a precompiled modularity stream,
+# however, it may take some time between latest kernel version is released until the precompiled pkgs are available;
+# which may cause driver failure.
+# This function tries to update the kernel to the latest version supported by the nvidia precompiled drivers, if fails
+# it upgrades to the latest kernel version available.
+rhel_kernel_version_matching() {
+    dnf -q -y module enable nvidia-driver:${NVIDIA_DRIVER_VERSION}
+    local latest_nv_version # latest nvidia driver available version
+    local latest_kmod_pkg_name # latest version of kmod-nvidia pkg name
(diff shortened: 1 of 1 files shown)


src/cloud-api-adaptor/podvm/addons/nvidia_gpu/setup.sh
```diff
@@ -34,11 +55,11 @@ if  [[ "$PODVM_DISTRO" == "ubuntu" ]]; then
 fi
 if  [[ "$PODVM_DISTRO" == "rhel" ]]; then
     dnf config-manager --add-repo http://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
```
beraldoleal, line 57: The repository is added twice. Here and at line 30, right?

beraldoleal (commented):
Hi @snir911, thanks for this. I let one small comment for you.

src/cloud-api-adaptor/podvm/addons/nvidia_gpu/setup.sh
beraldoleal: "It doesn't hurt", much. But I would still vote to remove it.

If for some reason, we decide to update one and forget about the other we might have conflicts. Also this is fetching the repository file from the internet twice.

I'm missing the reasoning for keep duplicated code. Do you mind to expand on that?

snir911: FWIW my reasoning was that leaving only one repo configuration will result in sensitivity to the calls order, I'll change it as you're also not the first one mentioning it :)

littlejawa (approved):
lgtm
Thanks @snir911 !

beraldoleal (approved):
Except the CI errors, LGTM, thanks.

snir911:
I find it safe to merge this one as this commit affects RHEL podvm image builds which uses the GPU addon only, hence, it has  limited impact on any other image builds.

stevenhorsman:
Snir has informed me that this PR is urgent for downstream, and given the limited scope of it as mentioned as we have a bug to resolve in peer pods before release already (#1941), I think it's okay to merge now.

PR #1929, opened 2024-07-18
Kata 3.7.0 release bump

stevenhorsman:
- Bump the kata-runtime go module to the 3.7.0 version in the csi-wrapper and cloud-api-adaptor modules
- Bump dependencies to match the versions used in the [Kata 3.7.0 release](https://github.com/kata-containers/kata-containers/blob/3.7.0/versions.yaml) to minimise the chances of incompatibilities and unstable code

PR #1930, opened 2024-07-18
workflow: Install packages for KBS client

stevenhorsman:
- The new version of the KBS client is failing the build in https://github.com/confidential-containers/cloud-api-adaptor/pull/1929
```
The file `openssl.pc` needs to be installed and the
PKG_CONFIG_PATH environment variable must contain its parent directory.
```
so install packages to get it building


bpradipt (approved):
/lgtm

PR #1933, opened 2024-07-18
auth: allow specifying auth.json for guest-pull

mkulke:
Kata introduced a flag to specify `image-registry-auth` which we can in guest pull. Since our agent-configuration is mostly static and we want to avoid brittle templating we provision the full kata-agent config from the CAA application like other config files.

If an auth secret is present in a CAA deployment, CAA will embed it in the agent config and provision it to the guest. The static agent-configuration has been removed and the configuration is pointing to `/run/peerpod/agent-config.toml` now, since it is a resource that changes at runtime.

### DIFF
--- a/src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
+++ b/src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
@@ -22,6 +22,7 @@ import (
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/aa"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/adaptor/k8sops"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/adaptor/proxy"
+	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/agent"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/cdh"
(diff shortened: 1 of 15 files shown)


stevenhorsman (approved):
The code LGTM. I'm struggling to get it working locally, but I think it's just my environment that is going wrong as after I restarted from scratch I can't create any peer pods, so I'm happy to see this merged and then I'll test it out with the proper builds. Thanks for the work @mkulke!

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -34,7 +35,10 @@ import (
 )
 
 const (
-	Version = "0.0.0"
+	SrcAuthfilePath = "/root/containers/auth.json"
+	AgentConfigPath = "/run/peerpod/agent-config.toml"
+	AuthFilePath    = "/run/peerpod/auth.json"
```
bpradipt, line 40: nit: Should these vars be prefixed with `Def` ?

bpradipt:
I'm able to access the link that's failing in the CI. Looks like an infra issue

bpradipt (approved):
/lgtm

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
mkulke: I opted to not call them DefaultSomething if they are hardcoded values that cannot be overridden.

mkulke:
Since it passed the e2e checks, there shouldn't be a regression. i'll merge, let's hope it'll work in automated tests 🤞

Issue #1937, opened 2024-07-22
podvm-mkosi: `make binaries` is failed when I try to build fedora image on the latest main branch.


liudalibj:
The error looks like:

From amd64 build machine:
```
62.04 error[E0445]: crate-private trait `MapFieldAccessor` in public interface
62.04    --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/protobuf-3.5.0/src/reflect/acc/v2/map.rs:115:1
62.04     |
62.04 17  |   pub(crate) trait MapFieldAccessor: Send + Sync + 'static {
62.04     |   -------------------------------------------------------- `MapFieldAccessor` declared as crate-private
62.04 ...
62.04 115 | / pub fn make_map_simpler_accessor_new<M, T>(
62.04 116 | |     name: &'static str,
62.04 117 | |     get_field: for<'a> fn(&'a M) -> &'a T,
62.04 118 | |     mut_field: for<'a> fn(&'a mut M) -> &'a mut T,
62.04 ...   |
62.04 121 | |     M: MessageFull,
62.04 122 | |     MapFieldAccessorImpl<M, T>: MapFieldAccessor,
62.04     | |_________________________________________________^ can't leak crate-private trait
62.04
62.04 error[E0446]: private type `MapFieldAccessorImpl<M, T>` in public interface
62.04    --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/protobuf-3.5.0/src/reflect/acc/v2/map.rs:115:1
62.04     |
62.04 33  |   struct MapFieldAccessorImpl<M, T>
62.04     |   --------------------------------- `MapFieldAccessorImpl<M, T>` declared as private
62.04 ...
62.04 115 | / pub fn make_map_simpler_accessor_new<M, T>(
62.04 116 | |     name: &'static str,
62.04 117 | |     get_field: for<'a> fn(&'a M) -> &'a T,
62.04 118 | |     mut_field: for<'a> fn(&'a mut M) -> &'a mut T,
62.04 ...   |
62.04 121 | |     M: MessageFull,
62.04 122 | |     MapFieldAccessorImpl<M, T>: MapFieldAccessor,
62.04     | |_________________________________________________^ can't leak private type
62.04
62.24 Some errors have detailed explanations: E0445, E0446.
62.24 For more information about an error, try `rustc --explain E0445`.
62.26 error: could not compile `protobuf` (lib) due to 2 previous errors
[...]

Issue #1890, opened 2024-07-01
podvm: Understand and reduce podvm permutations


stevenhorsman:
At the moment we have a matrix of 4 possible options for podvm (mkosi/packer) x (cloud-init/process-user-data).
We want to reduce this, ideally to one so we can minimise differences and duplicated code. One possible plan is:
- Identify all the variants of podvm builds we have and track who is using what
- Try and switch all podvm builds to use mkosi
- Switch the CI to mkosi and then deprecate packer
- Later investigate removing cloud-init and getting process-user-data to read the data from there if possible?

stevenhorsman:
### Part 1 Identify podvm builds

**Note: I think we decided a while ago that mkosi didn't work so well with ubuntu, so we wanted to switch to a fedora-like stack and deprecate the Ubuntu based podvm builds upstream?

| Base OS | Architecture | Cloud provider(s) | mkosi/packer | cloud-init/process-user-data | Being used? | Notes |
| --- | --- | --- | --- | --- | --- | --- |
| Ubuntu | amd64 | aws | packer? | <cloud-init/process-user-data> | ❓  |  |
| Ubuntu | amd64 | azure | packer? | <cloud-init/process-user-data> | ❓ |  |
| Ubuntu | amd64 | docker | packer? | <cloud-init/process-user-data> | ❓ |  |
| Ubuntu | amd64 | ibmcloud | packer | cloud-init? | ✅  |  |
| Ubuntu | amd64 | libvirt | packer | cloud-init? | ✅  |  |
| Ubuntu | amd64 | powervs | <mkosi/packer> | <cloud-init/process-user-data> | ❓ |  |
| Ubuntu | amd64 | vsphere | packer | cloud-init | ❓ | Deprecate this? |
| Ubuntu | s390x | ibmcloud/libvirt | packer? | cloud-init? | ✅ | |
| Fedora | amd64 | aws | mkosi? | <cloud-init/process-user-data> | ✅ ? |  |
| Fedora | amd64 | azure | mkosi? | <cloud-init/process-user-data> | ✅ ? |  |
| Fedora | amd64 | docker | mkosi? | <cloud-init/process-user-data> | ✅ ? |  |
| Fedora | amd64 | ibmcloud | mkosi? | cloud-init? |❓ |  |
| Fedora | amd64 | libvirt | mkosi? | cloud-init? |❓ |  |
[...]

PR #1939, opened 2024-07-22
podvm-mkosi: update podvm-builder for fedora image

liudalibj:
- pick up verion string from versions.yaml
- kata-agent v3.7.0 need rust v1.75.0

fixes https://github.com/confidential-containers/cloud-api-adaptor/issues/1937


stevenhorsman (approved):
The code LGTM, I want to run the e2e tests just to ensure there isn't any other version clashes

src/cloud-api-adaptor/ibmcloud-powervs/image/prereq.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 GO_VERSION="1.21.12"
-RUST_VERSION="1.72.0"
```
bpradipt, line 4: May be you can add a FIXME to pickup these values from versions.yaml as a future PR

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -24,21 +25,18 @@ container: fedora-binaries-builder binaries image-container
...
 fedora-binaries-builder:
 	@echo "Building $(BUILDER) image..."
-ifeq ($(ARCH),s390x)
 	docker buildx build \
 		-t $(BUILDER) \
-		--build-arg ARCH=s390x \
-		--build-arg PROTOC_ARCH=s390x \
-		--build-arg YQ_ARCH=s390x \
-		--build-arg YQ_CHECKSUM=sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45 \
+		--build-arg GO_VERSION=$(GO_VERSION) \
```
bpradipt, line 30: How was it working earlier without all these new additions ?
These new args doesn't seem to be related to the RUST_VERSION change

bpradipt:
Since there are already two lgtms, hence I'm adding `hold` label to avoid accidental merge.

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: So my understanding when I reviewed is that these updates allow the podvm_builder versions to be updated from their defaults that are hardcoded in Dockerfile.podvm_builder.fedora, and are set in Makefile.defaults which is included here. Previously we'd just fall back to the hard-coded versions in the Dockerfile, so I assumed that DaLi decided to allow all to be overrided from versions.yaml, rather than just RUST_VERSION. 

I'm not sure if there is an argument to also make the `ARG` values in the Dockerfiles not set so that we'd spot error like this quicker and not have to duplicate versions in multiple places, but that might have more knock-on changes that I don't understand?

bpradipt: I see, probably a separate commit with description will be helpful here.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -10,6 +10,7 @@ PODVM_DISTRO ?= fedora
 PODVM_TAG ?= $(VERSIONS_HASH)
 PODVM_NAME ?= $(REGISTRY)/podvm-generic-$(PODVM_DISTRO)-$(ARCH)
 PODVM_CONTAINER_NAME ?= $(REGISTRY)/podvm-docker-image
+YQ_CHECKSUM_amd64 ?= $(YQ_CHECKSUM)
```
bpradipt, line 13: How is the s390x checksum handled ?

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -24,21 +25,18 @@ container: fedora-binaries-builder binaries image-container
...
-		--build-arg ARCH=s390x \
-		--build-arg PROTOC_ARCH=s390x \
-		--build-arg YQ_ARCH=s390x \
-		--build-arg YQ_CHECKSUM=sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45 \
+		--build-arg GO_VERSION=$(GO_VERSION) \
+		--build-arg ARCH=$(ARCH) \
+		--build-arg PROTOC_VERSION=$(PROTOC_VERSION) \
+		--build-arg RUST_VERSION=$(RUST_VERSION) \
+		--build-arg YQ_VERSION=$(YQ_VERSION) \
+		--build-arg YQ_CHECKSUM=${YQ_CHECKSUM_$(ARCH)} \
```
bpradipt, line 35: Why not simply use YQ_CHECKSUM? The caller will need to provide the right checksum similar to other params. Also this param name will be aligned to what is already there in the Dockerfiles

stevenhorsman: In Makefile.defaults we have:
https://github.com/confidential-containers/cloud-api-adaptor/blob/31a32722cfd2ea5fa3173baf1b1dcfbc7949801f/src/cloud-api-adaptor/Makefile.defaults#L9-L10
So I think the two changes you've mentioned are to handle the asymmetry there between amd64 and s390x. There is maybe an argument to change that defaults to have the arch on both and then this file will be easier to understand, but that might have other consequences, so something to do later?

bpradipt: Ok, so we already have arch specific checksum variables :-(
Would it be possible to reuse existing vars ie, YQ_CHECKSUM and YQ_CHECKSUM_s390x without introducing another arch variable ie YQ_CHECKSUM_amd64 ?
Or can YQ_CHECKSUM in the Makefile.defaults file be changed to YQ_CHECKSUM_amd64? (This might need changes to other files that is referring to this variable, so I'm not sure how feasible it is)

I'm not against introducing another arch specific variable; I just want us to do it consciously after ruling out other options. And also to follow a pattern which is easier to remember.

stevenhorsman: > Or can YQ_CHECKSUM in the Makefile.defaults file be changed to YQ_CHECKSUM_amd64? (This might need changes to other files that is referring to this variable, so I'm not sure how feasible it is)

This was the change I was alluding to, but I'm concerned about the other things this might break at this point in time.

bpradipt: Agree.  Let's not do this change at this moment and instead create an issue to avoid forgetting this.

Do we need the additional changes now? or only the RUST bump is good enough to proceed? 
Or we go ahead with the release since packer based image and CI is working fine and do a proper fix for mkosi post the 0.9.0 release ?

stevenhorsman: That's a good set of questions. I think it depends if anyone is relying on the packer build for 0.9.0 release. If so we should get this merged soon (or at least the RUST VERSION hard-coded bump) and generate a new alpha pre-release. If not then maybe we can delay this and the other improvements in this PR and merge it soon after 0.9.0 is released.

bpradipt: s/packer/mkosi right ?

bpradipt: SInce @liudalibj already did the work needed, here is my suggestion 
1. Create a commit to align the `fedora-binaries-builder` target with `podvm-builder` target (this fixes the root cause of the mkosi issue encountered)
2. Create a commit to bump up the default version in the Dockerfile to align with versions.yaml 

If needed, `2` can be ignored for now and instead a separate PR to drop all defaults can be made to avoid these confusions.

wainersm:
I honestly don't know what to do here. I would say, let's merge it for 0.9.0 if we have users for the fedora mkosi podvm, which apparently we do: https://github.com/confidential-containers/cloud-api-adaptor/issues/1890#issuecomment-2213578449

src/cloud-api-adaptor/ibmcloud-powervs/image/prereq.sh
liudalibj: Added FIXME string, so that we can get all versions from versions.yaml.

src/cloud-api-adaptor/podvm-mkosi/Makefile
liudalibj: Thanks @stevenhorsman yeah, I saw that these values in the podvm builder for packer are all [to be overriden from versions.yaml](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Makefile#L170-L179,) so that I make this change to make it same as the packer-ubuntu one.

src/cloud-api-adaptor/podvm-mkosi/Makefile
liudalibj: updated.

src/cloud-api-adaptor/podvm-mkosi/Makefile
liudalibj: remove this item `YQ_CHECKSUM_amd64 `

src/cloud-api-adaptor/podvm-mkosi/Makefile
liudalibj: did a search all other places we used `YQ_CHECKSUM` (amd64) only here we need change the value to YQ_CHECKSUM_s390x when it's s390x arch.
So we can use `YQ_CHECKSUM` directly.
If it's s390x arch then the `YQ_CHECKSUM`=`YQ_CHECKSUM_s390x`,

src/cloud-api-adaptor/podvm-mkosi/Makefile
liudalibj: update the commit title to `align the fedora-binaries-builder with podvm-builder`

bpradipt (approved):
@liudalibj thanks for accommodating.

stevenhorsman:
The question now is whether this is reduced in scope enough to mkosi that we can avoid re-cutting a pre-release build?

PR #1947, opened 2024-07-25
doc: Release process re-work

stevenhorsman:
- Update to reflect that we need to wait for the operator payload change before starting the peer-pods release

### DIFF
--- a/docs/Release-Process.md
+++ b/docs/Release-Process.md
@@ -30,11 +30,11 @@ the kata-containers release version.
         go mod tidy
         ```
         in the [cloud-api-adaptor](../src/cloud-api-adaptor/) directory and [csi-wrapper](../src/csi-wrapper/) directory.
+1. The CoCo operator updates to use references to the other component releases and then releases itself
 1. cloud-api-adaptor releases with the following phases detailed below:
     - Pre-release testing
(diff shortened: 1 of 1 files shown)


docs/Release-Process.md
```diff
@@ -30,11 +30,11 @@ the kata-containers release version.
         go mod tidy
         ```
         in the [cloud-api-adaptor](../src/cloud-api-adaptor/) directory and [csi-wrapper](../src/csi-wrapper/) directory.
+1. The CoCo operator updates to use references to the other component releases and then releases itself
 1. cloud-api-adaptor releases with the following phases detailed below:
```
bpradipt, line 34: nit: s/1/2

bpradipt: Ignore it.. In preview the sequence is proper.

stevenhorsman: Yeah, I'm lazy/smart? and just label them all as `1.` so that when we insert new items, or re-juggle the order, it's less work to update them

bpradipt: The is super smart :-) .. Learnt a new way..

wainersm (approved):
Thanks @stevenhorsman !

PR #1948, opened 2024-07-25
ci: Enable exhaustruct linter

stevenhorsman:
exhaustruct "Checks if all structure fields are initialized". This should help us avoid issues with the order of fields in structures going wrong which has happened a couple of times.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -227,6 +227,7 @@ linters:
     - unused
     # Additional linters.
     - gofmt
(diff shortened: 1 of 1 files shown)


stevenhorsman:
I think this goes beyond what we want and enforces that all fields are set on every struct giving us hundreds of errors: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/10092606837/job/27906584686?pr=1948

PR #1953, opened 2024-07-26
go.mod: Update go-urn to v1.2.4

bpradipt:
older versions of go-urn uses GPL code and it's problematic with CNCF projects. Updating the version to use MIT license code


mkulke (approved):
lgtm

PR #1954, opened 2024-07-26
deps: Remove circular dep from cloud-providers

mkulke:
cloud-providers was referencing the cloud-api-adaptor module in a docker-provder test, which produced a circular dependency leading to multiple versions of kata-containers being referenced in the repo:

cloud-providers -> cloud-api-adaptor 0.8.0 -> kata#CCv0 cloud-api-adaptor -> kata#main

I think it's ok to use the hardcoded value in this test, otherwise we have to outsource such constants to a 3rd module consumed by both CAA and CP modules.


bpradipt (approved):
/lgtm

stevenhorsman (approved):
Thanks for sorting this out. I spotted in a couple of days ago when bumping versions for the pre-release 2, so it was on my list, but I can strike it off now!

PR #1956, opened 2024-07-27
kata-agent: set ocicrypt environment variable

mkulke:
this will (re-)enable support for encrypted images on kata#main

kata uses image-rs, which in turn uses ocicrypt-rs. the latter requires the env OCICRYPT_KEYPROVIDER_CONFIG to point to a file which defines how decryption keys can be retrieved from CDH. (note: it's called attestation-agent here, but that's a misnomer for historical reasons)

kata-agent is only setting this env in process and writing the file to a /tmp location, if it is actively managing the guest-component processes.

We start those processes as service units, so we have to take care of it on the podvm manually. the ocicrypt file is static and writing to /tmp wouldn't work well with a read-only rootfs anyway, so we can just include in the podvm image and referene the path in the env.

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
@@ -5,6 +5,7 @@ Wants=process-user-data.service attestation-agent.service
 After=netns@podns.service process-user-data.service attestation-agent.service
 
 [Service]
+Environment=OCICRYPT_KEYPROVIDER_CONFIG=/etc/ocicrypt_config.json
 ExecStartPre=mkdir -p /run/kata-containers
 ExecStart=/usr/local/bin/kata-agent --config /run/peerpod/agent-config.toml
 ExecStartPre=-umount /sys/fs/cgroup/misc
--- a/src/cloud-api-adaptor/podvm/files/etc/ocicrypt_config.json
+++ b/src/cloud-api-adaptor/podvm/files/etc/ocicrypt_config.json
@@ -1,0 +1,7 @@
+{
+	"key-providers": {
+		"attestation-agent": {
+			"ttrpc": "unix:///run/confidential-containers/cdh.sock"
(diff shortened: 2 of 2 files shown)


src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
```diff
@@ -5,6 +5,7 @@ Wants=process-user-data.service attestation-agent.service
 After=netns@podns.service process-user-data.service attestation-agent.service
 
 [Service]
+Environment=OCICRYPT_KEYPROVIDER_CONFIG=/etc/ocicrypt_config.json
```
bpradipt, line 8: Will encrypted images require bumping the version of kata-agent or others in versions.yaml?

mkulke: I don't think so. it shouldn't affect kata directly, but I'll double check. there is a [bug in ocicrypt-rs](https://github.com/confidential-containers/guest-components/pull/630) which makes debugging of issues w/ enc images impractical. so we might want to bump guest-components eventually, but it shouldn't block anything in the green path atm.

mkulke: yup. I checked. it should work. this image can be used for testing:

```bash
skopeo inspect docker://ghcr.io/mkulke/nginx-encrypted:20240123 | jq -r '.Labels'
{
  "coco-key-b64": "pHSE5N+T/3GGfb/umaWgB8bfHc/dQWvmxdsjoWam0Vs=",
  "coco-key-id": "default/key/nginx-encrypted",
  "maintainer": "NGINX Docker Maintainers <docker-maint@nginx.com>"
}
```

bpradipt (approved):
/lgtm

liudalibj (approved):
LGTM

huoqifeng (approved):
LGTM

PR #1962, opened 2024-07-29
podvm: Disable sshd service

bpradipt:
If using packer based images, ssh service is enabled by default. Coupled this with config-drive based ssh key injection, this allows any infra admin to SSH to the pod VM.
This is unlike the mkosi image where ssh is disabled or uses only user specific public key for login.

This commit removes the SSH package from the podvm image during image generation.

It uses a oneshot service to remove the package during shutdown as part of image generation. This is because SSH is used by packer for configuring the initial image.

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/multi-user.target.wants/remove-ssh.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/multi-user.target.wants/remove-ssh.service
@@ -1,0 +1,1 @@
+../remove-ssh.service
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/remove-ssh.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/remove-ssh.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Remove SSH server on shutdown
+DefaultDependencies=no
+Before=shutdown.target reboot.target halt.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/remove-ssh.sh
+
+[Install]
+WantedBy=halt.target reboot.target shutdown.target
+
--- a/src/cloud-api-adaptor/podvm/files/usr/local/bin/remove-ssh.sh
+++ b/src/cloud-api-adaptor/podvm/files/usr/local/bin/remove-ssh.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Function to remove SSH server on Ubuntu
+remove_ssh_ubuntu() {
+    echo "Detected Ubuntu. Removing openssh-server using apt-get."
+    apt-get remove -y openssh-server
+}
+
+# Function to remove SSH server on RHEL
+remove_ssh_rhel() {
+    echo "Detected RHEL. Removing openssh-server using dnf."
+    dnf remove -y openssh-server
+}
+
+# Detect the operating system
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+    case "$ID" in
+        ubuntu)
+            remove_ssh_ubuntu
+            ;;
+        rhel | centos | fedora)
+            remove_ssh_rhel
+            ;;
+        *)
+            echo "Unsupported OS: $ID"
+            exit 1
+            ;;
+    esac
+else
+    echo "Cannot detect the operating system. /etc/os-release not found."
+    exit 1
+fi
+


bpradipt:
@mkulke @stevenhorsman @snir911 @liudalibj and others, would there be any interest in this PR?
This will of course make debugging harder. I didn't add a DEBUG option since we anyway want to move to mkosi based builds. But till then I want to see if there is an interest to disable SSH service completely from the pod VM image. 
Let me know your thoughts

mkulke:
why not use systemd facilities to disable the service? since you can just re-install the package if you have a (serial) shell, I don't think the removal is required. And if we the package has to be removed, why not use misc-settings.sh? 

at the moment we have SSH_PUBKEY param, that is currently not usable in mkosi (you need to package pubkeys at build time). If we don't have ssh on packer-based images we should remove that option.

and as you indicated, for debugging it will add overhead, you have to make sure there is a way you can enter a pod vm from the serial console.

bpradipt:
> why not use systemd facilities to disable the service?

Unfortunately this didn't work for some reason. I used systemd preset as well as disabled the service, but still the service was running when the VM booted up. I couldn't figure out a logical explanation for the below behaviour. Only guess is something to do with the cloud-init or wagent services which Wants sshd.service
```
[peerpod@podvm-hello-b-ebff6ebf ~]$ sudo systemctl status sshd
● sshd.service - OpenSSH server daemon
     Loaded: loaded (/usr/lib/systemd/system/sshd.service; disabled; preset: disabled)
     Active: active (running) since Mon 2024-07-29 11:48:10 UTC; 2min 12s ago
       Docs: man:sshd(8)
             man:sshd_config(5)
   Main PID: 882 (sshd)
      Tasks: 1 (limit: 45920)
     Memory: 5.6M
        CPU: 51ms
     CGroup: /system.slice/sshd.service
             └─882 "sshd: /usr/sbin/sshd -D [listener] 0 of 10-100 startups"
```

> And if we the package has to be removed, why not use misc-settings.sh?

Since disabling the service didn't work I didn't use misc-settings.sh. I had to create a oneshot service that will run once packer has done all the customisations (before shutdown) and hence created the service unit file and the binary under files dir


> at the moment we have SSH_PUBKEY param, that is currently not usable in mkosi (you need to package pubkeys at build time). If we don't have ssh on packer-based images we should remove that option.

I'm basing my answer with the assumption that you meant the following param - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-providers/azure/provider.go#L169

I agree that it can be removed.  And only use the mkosi way to add a pub key to the image if needed for debug scenarios. 
However I couldn't figure out a way to spin up an Azure instance without providing an SSH key. Looks like the API mandates use of a pub key. It can be a dummy key as well but still it's required. Is this a correct conclusion or am I missing something ?


> and as you indicated, for debugging it will add overhead, you have to make sure there is a way you can enter a pod vm from the serial console.

Agree.

mkulke:
> Since disabling the service didn't work I didn't use misc-settings.sh. I had to create a oneshot service that will run once packer has done all the customisations (before shutdown) and hence created the service unit file and the binary under files dir. Although both these files can be created in misc-settings.sh itself.

but we could uninstall the package in misc-settings.sh? or will the agent or cloud-init actually re-install the package?
 
> I agree that it can be removed. And only use the mkosi way to add a pub key to the image if needed for debug scenarios. However I couldn't figure out a way to spin up an Azure instance without providing an SSH key. Looks like the API mandates use of a pub key. It can be a dummy key as well but still it's required. Is this a correct conclusion or am I missing something ?
> 

yes. it might be that we have to set a password or ssh pubkey. we can just generate a useless pubkey in that case.

> > and as you indicated, for debugging it will add overhead, you have to make sure there is a way you can enter a pod vm from the serial console.
> 
> Agree.

but at the moment we don't a password-less login on the serial console? (and not sure we want that?)

bpradipt:
> > Since disabling the service didn't work I didn't use misc-settings.sh. I had to create a oneshot service that will run once packer has done all the customisations (before shutdown) and hence created the service unit file and the binary under files dir. Although both these files can be created in misc-settings.sh itself.
> 
> but we could uninstall the package in misc-settings.sh? or will the agent or cloud-init actually re-install the package?

Hmm packer is still using ssh for the doing the configs. I didn't check if it would be possible to remove the package when the binary is being used. I'll check.

> 
> > I agree that it can be removed. And only use the mkosi way to add a pub key to the image if needed for debug scenarios. However I couldn't figure out a way to spin up an Azure instance without providing an SSH key. Looks like the API mandates use of a pub key. It can be a dummy key as well but still it's required. Is this a correct conclusion or am I missing something ?
> 
> yes. it might be that we have to set a password or ssh pubkey. we can just generate a useless pubkey in that case.
> 
> > > and as you indicated, for debugging it will add overhead, you have to make sure there is a way you can enter a pod vm from the serial console.
> > 
> > 
> > Agree.
> 
> but at the moment we don't a password-less login on the serial console? (and not sure we want that?)

Better to avoid it. 
I think creating a debug and a non-debug image (like mkosi) is a good approach with clear separation. I'll explore few options for packer as an interim measure. But if it turns out to be a big change, may be better to skip it in favour of mkosi.

liudalibj:
> > 
> > but at the moment we don't a password-less login on the serial console? (and not sure we want that?)
> 
> Better to avoid it. I think creating a debug and a non-debug image (like mkosi) is a good approach with clear separation. I'll explore few options for packer as an interim measure. But if it turns out to be a big change, may be better to skip it in favour of mkosi.

Agree that provider a debug and non-debug image (like mkosi) is a good approach with clear separation.

liudalibj:
if we disabled the sshd service, not sure the Secure-Comms feature still work, need @davidhadas 's comment.

davidhadas:
Some thoughts:
1. Removing ssh means kcli will not work  - right? 
    Are there other tools we use today (as part of the project documentation) that rely on ssh to the podvm? 
    We may need to document clearly how to "get shh to work on the podvm for debugging" (delete the `remove-ssh.service` file ?), probably best to document it as part of this PR. 

> if we disabled the sshd service, not sure the Secure-Comms feature still work, need @davidhadas 's comment.

2. Secure-Comms uses a golang ssh implementation, so it does not directly rely on the system ssh, however, it is possible that there are some common dependencies. This need to be tested.

bpradipt:
This PR is superseded by https://github.com/openshift/sandboxed-containers-operator/pull/443
So there will be a debug option to enable ssh service in the pod vm.
Let's have all discussions in PR 443. I'm closing this to avoid duplication of discussions

PR #1968, opened 2024-07-31
e2e-tests: add option to deploy kbs with custom pccs_url

mkulke:
TDX attestation requires a pccs_url to be set. we'll introduce a flag CUSTOM_PCCS_URL that will configure kbs to use the it.

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
+++ b/src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
@@ -127,6 +127,18 @@ func NewKeyBrokerService(clusterName string, cfg *envconf.Config) (*KeyBrokerSer
 
 	}
 
+	customPCCSURL := os.Getenv("CUSTOM_PCCS_URL")
+	if customPCCSURL != "" {
+		log.Info("CUSTOM_PCCS_URL is provided, write custom PCCS config")
+		configFilePath := filepath.Join(TRUSTEE_REPO_PATH, "/kbs/config/kubernetes/custom_pccs/sgx_default_qcnl.conf")
+		collateralUrl := "https://api.trustedservices.intel.com/sgx/certification/v4/"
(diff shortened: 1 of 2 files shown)


src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
```diff
@@ -127,6 +127,18 @@ func NewKeyBrokerService(clusterName string, cfg *envconf.Config) (*KeyBrokerSer
 
 	}
 
+	customPCCSURL := os.Getenv("CUSTOM_PCCS_URL")
+	if customPCCSURL != "" {
+		log.Info("CUSTOM_PCCS_URL is provided, write custom PCCS config")
+		configFilePath := filepath.Join(TRUSTEE_REPO_PATH, "/kbs/config/kubernetes/custom_pccs/sgx_default_qcnl.conf")
+		collateralUrl := "https://api.trustedservices.intel.com/sgx/certification/v4/"
+		config := fmt.Sprintf(`{ "pccs_url": "%s", "collateral_service": "%s"}`, customPCCSURL, collateralUrl)
```
fidencio, line 135: I don't think the `collateral_service` is required, so it can be dropped.

I am not sure if you'll need ` "use_secure_cert": false`, if the tests pass as they're, you're good to go without that.

fidencio (approved):
lgtm, thanks @mkulke!

I've left two comments, which are more informative than something that has to be addressed (thus, I am approving the PR).

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
mythi: FWIW, for quote verification (which is the only thing we are doing with Trustee here), only `collateral_service` is needed and for that we can use the same [Azure setting](https://learn.microsoft.com/en-us/azure/security/fundamentals/trusted-hardware-identity-management#how-do-i-use-intel-qpl-with-trusted-hardware-identity-management)

kartikjoshi21 (approved):
LGTM . Thanks @mkulke

PR #1969, opened 2024-07-31
e2e-test: assert job condition on job tests

mkulke:
We have elaborate condition checks on a job's pods, but we do not test the job status itself, which will produce wrong test results. A test for job that always fails (cmd: /bin/false) will pass.

This keeps the current assertions in-place, but makes the test fail on a failed job status


bpradipt (approved):
/lgtm

stevenhorsman (approved):

PR #1974, opened 2024-08-02
e2e-test: Remove naming conflicts

mkulke:
Test should not create resources with the same name, because this will introduce race conditions if the tests run in parallel.

image

image


bpradipt (approved):
/lgtm

liudalibj (approved):
LGTM

Issue #1966, opened 2024-07-31
External network connectivity issue with EKS CNI


yoheiueda:
As reported at https://github.com/confidential-containers/cloud-api-adaptor/pull/1920#issuecomment-2252935108, peer pod network has external network connectivity issue with EKS CNI.

The design of the CNI plugin for Kubernetes networking over AWS VPC is described here.
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/cni-proposal.md#solution-components

yoheiueda:
@bpradipt @EmmEff is it possible to collect some diagnostic data on EKS?

Create a regular (runc) pod and execute the following commands in the pod with `kubectl exec`. If you can access the worker node that the pod is running, please execute the same commands.

```
ip address show
ip link show
ip rule show
ip route show table main
ip neigh show
```

According to the documentation, the EKS CNI plugin explicitly sets a static ARP entry. If so, I think we can fix the issue by setting the same ARP entry in a network namespace in a peer pod VM.

bpradipt:
@yoheiueda please find the requested details

O/p from regular runc pod
```
[root@priv-pod /]# ip address show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
3: eth0@if17: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc noqueue state UP group default
    link/ether 56:37:6d:d9:1b:6a brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.0.0.149/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::5437:6dff:fed9:1b6a/64 scope link
       valid_lft forever preferred_lft forever
[root@priv-pod /]#
[root@priv-pod /]# ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
[...]

PR #1983, opened 2024-08-07
Support EKS CNI plugin

yoheiueda:
Permanent ARP entries that are explicitly set in the network namespace for a pod in a worker node are propagated to
the podns network namespace in a peer pod VM.

The EKS CNI plugin sets such a permanent ARP entry, and this patch is necessary for peer pods to work with EKS networks.

Fixes https://github.com/confidential-containers/cloud-api-adaptor/issues/1966

I tested this change with Flannel and Calico.

### DIFF
--- a/src/cloud-api-adaptor/pkg/util/netops/netops.go
+++ b/src/cloud-api-adaptor/pkg/util/netops/netops.go
@@ -14,6 +14,8 @@ import (
 	"runtime"
 	"sort"
 
+	"golang.org/x/exp/maps"
+
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	"golang.org/x/sys/unix"
@@ -33,6 +35,8 @@ type Namespace interface {
 	RuleAdd(rule *Rule) error
 	RuleDel(rule *Rule) error
 	RuleList(rule *Rule) ([]*Rule, error)
+	NeighborAdd(neighbor *Neighbor) error
+	NeighborList(filters ...*Neighbor) ([]*Neighbor, error)
 	Run(fn func() error) error
 }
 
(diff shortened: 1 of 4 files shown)


yoheiueda:
@bpradipt @EmmEff This PR is for the EKS CNI plugin.

Is it possible to test this on EKS?

EmmEff:
I will be able to test today. Thank you very much!

src/cloud-api-adaptor/pkg/util/netops/netops.go
```diff
@@ -776,6 +780,139 @@ func (ns *namespace) RuleList(rule *Rule) ([]*Rule, error) {
...
+
+		msg := netlink.Ndmsg{
+			Family: netlink.FAMILY_V4,
+			State:  uint16(filter.State),
+		}
+
+		if filter.Dev != "" {
+			link, err := ns.handle.LinkByName(filter.Dev)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get a list of neighbors on network namespace %s: %w", ns.Path(), err)
```
wainersm, line 886: Hi @yoheiueda !

I suggest to rephrase this error message to something like `failed to get a dev link named %s on network namespace %s: %w`

src/cloud-api-adaptor/pkg/util/netops/netops.go
```diff
@@ -776,6 +780,139 @@ func (ns *namespace) RuleList(rule *Rule) ([]*Rule, error) {
...
+
+// NeighborList gets a list of neighbors
+func (ns *namespace) NeighborList(filters ...*Neighbor) ([]*Neighbor, error) {
+
+	var neighbors []*Neighbor
+	for _, filter := range filters {
+
+		msg := netlink.Ndmsg{
+			Family: netlink.FAMILY_V4,
+			State:  uint16(filter.State),
```
wainersm, line 880: shouldn't check `filter.State` is not "unknown"?

src/cloud-api-adaptor/pkg/util/netops/netops.go
```diff
@@ -776,6 +780,139 @@ func (ns *namespace) RuleList(rule *Rule) ([]*Rule, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get a link with index %d of neighbor %s: %w", nlNeigh.LinkIndex, nlNeigh.String(), err)
+			}
+
+			dev := link.Attrs().Name
+			n := &Neighbor{
+				IP:           toAddr(nlNeigh.IP),
+				Dev:          dev,
+				HardwareAddr: nlNeigh.HardwareAddr.String(),
+				State:        filter.State,
```
wainersm, line 907: And here probably we want to read from `nlNeigh.State` as `filter.State` can be "unknown"?

src/cloud-api-adaptor/pkg/util/netops/netops.go
```diff
@@ -776,6 +780,139 @@ func (ns *namespace) RuleList(rule *Rule) ([]*Rule, error) {
...
+	v, err := unmarshalJSON("neighbor state", data, neighborStates)
+	if err != nil {
+		return err
+	}
+	*s = v
+	return nil
+}
+
+// NeighborAdd adds a new neighbor entry
+
```
wainersm, line 832: del empty line

src/cloud-api-adaptor/pkg/util/netops/netops.go
```diff
@@ -776,6 +780,139 @@ func (ns *namespace) RuleList(rule *Rule) ([]*Rule, error) {
...
+		return fmt.Errorf("failed to add a new neighbor entry: device %s not found: %w", neighbor.Dev, err)
+	}
+
+	hwAddr, err := net.ParseMAC(neighbor.HardwareAddr)
+	if err != nil {
+		return fmt.Errorf("failed to add a new neighbor entry: invalid hardware address %q", neighbor.HardwareAddr)
+	}
+
+	state := neighbor.State
+	if state.String() == "none" || state.String() == "unknown" {
```
wainersm, line 854: I should have said early: I'm not a networking expert at all. I was looking at the manual https://man7.org/linux/man-pages/man8/ip-neighbour.8.html and was wondering if we should attempt to add neighbors which state seems to indicate a transition like `incomplete` or `failed` (or maybe add the neighbor but let `state` unset).

wainersm:
Hi @yoheiueda, I've finished my review which was essentially to check if the implementation adheres to problem and solution described in https://github.com/confidential-containers/cloud-api-adaptor/issues/1966 .  I'm not knowledgeable  on EKS, CNI nor networking, so I cannot judge the solution is good or not, hence I prefer to not actually explicitly approve the PR.

yoheiueda:
@EmmEff Thank you very much for testing this patch on EKS!

@wainersm Thank you for your comments. I am on vacation, and I'll update my PR after I am back.

src/cloud-api-adaptor/pkg/util/netops/netops.go
yoheiueda: fixed.

src/cloud-api-adaptor/pkg/util/netops/netops.go
yoheiueda: For our use case, I think we only need to identify and set permanent neighbor entries that are set statically. The other states are dynamically set when ARP packets are received. ARP packets are delivered to the podns network namespace in a peer pod VM via an VXLAN tunnel, so agent-protocol-forwarder does not need to manipulate those neighbor entries explicitly.

src/cloud-api-adaptor/pkg/util/netops/netops.go
yoheiueda: fixed.

src/cloud-api-adaptor/pkg/util/netops/netops.go
yoheiueda: Ah, yes. I need to use `NeighborState(nlNeigh.State)` instead of `filter.State` here. Fixed.

src/cloud-api-adaptor/pkg/util/netops/netops.go
yoheiueda: fixed.

bpradipt (approved):
/lgtm
Thanks @yoheiueda 
@wainersm ptal

stevenhorsman (approved):
I'm not really familiar with the networking code, but based on the testing by Mike and Pradipta, I'm okay to approve and unblock this.

PR #1990, opened 2024-08-08
docs: redirect users to use make targets on builds

beraldoleal:
Current dockerfiles are not ready to build directly from the folder. Builder is broken as also depends on .git and .hack folders. Make targets are working, so lets redirect our readers to use those targets.

We can improve the targets as needed and this will be transparant for the readers.

### DIFF
--- a/src/cloud-api-adaptor/podvm/README.md
+++ b/src/cloud-api-adaptor/podvm/README.md
@@ -77,31 +77,23 @@ $ docker build -t podvm_builder \
 
 ## Building the image containing the podvm binaries
 
-In order to build the binaries you should be using the corresponding dockerfile
-of the Linux distro for which the builder image was built.
-For example, if the builder image was built with *Dockerfile.podvm_builder.DISTRO* then
-you should use the *Dockerfile.podvm_binaries.DISTRO* to build the image with podvm binaries.
+We have make targets handy for you. Make sure you are under the
+`src/cloud-api-adaptor` folder.
 
-Assuming you have built the podvm_builder image for Ubuntu as explained in the previous step,
-running the following command will build the image with the podvm binaries.
+> **Note:** The `BUILDER_IMG` environment variable is crucial as it specifies
+> the builder image, which is the result of the previous step build. Ensure you
+> have built the builder image before proceeding.
+
+To build the binaries image, use the following command:
 
 ```bash
-$ docker build -t podvm_binaries \
-         --build-arg BUILDER_IMG=podvm_builder \
-         -f Dockerfile.podvm_binaries .
-
+$ BUILDER_IMG=<your_builder_image> make podvm-binaries
 ```
+
 The build process can take significant time.
 
-The binaries can be built for other architectures than `x86_64` by passing the `ARCH` build
-argument to docker. Currently this is only supported for Ubuntu `s390x` as shown below:
-
-```bash
-$ docker build -t podvm_binaries_s390x \
-         --build-arg BUILDER_IMG=podvm_builder \
-         --build-arg ARCH=s390x \
-         -f Dockerfile.podvm_binaries .
-```
+You can customize the build, by using some environment variables like ARCH,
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm

stevenhorsman (approved):
LGTM. Thanks!

src/cloud-api-adaptor/podvm/README.md
```diff
@@ -77,31 +77,17 @@ $ docker build -t podvm_builder \
 
 ## Building the image containing the podvm binaries
 
-In order to build the binaries you should be using the corresponding dockerfile
```
wainersm, line 80: Hi @beraldoleal !

I like the idea of suggesting the use of make targets wherever possible. IMHO CoCo and kata docs make a poor job on standardizing around Make, several places use call to scripts (or sequence of commands) where you could be using an equivalent make target. Anyway...

However, for this particular change I'm afraid that dropping these lines that explain the relation between builder image and podvm binaries will result on less valuable information for the devs (essentially now we are saying devs to look at the sources). For example, if a dev built the builder image as indicated on the previous section, and he/she run `make podvm-binaries` it won't be pulling that image (instead it's getting  the default quay.io/confidential-containers/podvm-builder-ubuntu). If you see the sections on this doc then you will notice they are chained to show how a dev can build the builder -> binaries -> podvm images.

src/cloud-api-adaptor/podvm/README.md
```diff
@@ -77,31 +77,17 @@ $ docker build -t podvm_builder \
...
-argument to docker. Currently this is only supported for Ubuntu `s390x` as shown below:
+The build process can take significant time.
 
-```bash
-$ docker build -t podvm_binaries_s390x \
-         --build-arg BUILDER_IMG=podvm_builder \
-         --build-arg ARCH=s390x \
-         -f Dockerfile.podvm_binaries .
-```
+You can customize the build, but using some environment variables like ARCH,
```
wainersm, line 89: Introducing a `help` target where those variables are listed would be nice too. See https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Makefile#L39 as an example. Can be done on a follow on PR.

src/cloud-api-adaptor/podvm/README.md
beraldoleal: Yes, agree is a little bit confusing. Ideally all the steps should be pointing to the makefile. I have changed only here to fix a broken instruction, for now.

I will rephrase my last paragraph and/or add note to make it clear. Thanks @wainersm

beraldoleal:
@wainersm thanks for your comment, I fixed the main one with a force-push. let me know if you are ok with it now.

stevenhorsman:
@wainersm @beraldoleal - FYI I'm think of doing the same change as this for the podvm_builder as well (to redirect people to use the `make target` rather than running the dockerfile directly and specifying all the versions.

src/cloud-api-adaptor/podvm/README.md
```diff
@@ -77,31 +77,22 @@ $ docker build -t podvm_builder \
...
+
+To build the binaries image, use the following command:
 
 ```bash
-$ docker build -t podvm_binaries \
-         --build-arg BUILDER_IMG=podvm_builder \
-         -f Dockerfile.podvm_binaries .
+$ BUILDER_IMG=<your_builder_image> make podvm-binaries
 
-```
```
stevenhorsman, line 93: This line needs reinstating to close the code block

beraldoleal: nice catch, thanks. fixed.

bpradipt:
@stevenhorsman @beraldoleal can we merge this ?

stevenhorsman:
> @stevenhorsman @beraldoleal can we merge this ?

Yeah, I happy with that.

PR #1993, opened 2024-08-09
e2e tests: consolidate azure nightly builds

mkulke:
Several changes for the azure CI workflows and tests are introduced:

- A single nightly-build workflow will build a podvm image and then use that image in a test.

- A remote-attestation e2e test has been introduced that will only acquire an attestation token from kbs without requesting a secret.

- Early exit in the job assertion logic.

- Introduced configurable backofflimit for job tests.

- Only run the simple pod and remote attestation tests for the time being.

- Introduced jitter value in gh workflow matrix to avoid contention for cross-matrix resources (user assigned identity).


src/cloud-api-adaptor/test/e2e/azure_test.go
```diff
@@ -6,142 +6,127 @@
...
-
-// Negative test with device annotation
-func TestPodWithIncorrectDeviceAnnotationAzure(t *testing.T) {
-	if !isTestOnCrio() {
-		t.Skip("Skipping test as it is not running on CRI-O")
-	}
-	t.Parallel()
-	DoTestPodWithIncorrectCrioDeviceAnnotation(t, testEnv, assert)
-}
+// func TestCreatePodWithConfigMapAzure(t *testing.T) {
```
bpradipt, line 19: Commenting out the tests will be very restrictive. For example I typically run the tests on vanilla K8s in Azure or OpenShift when testing a PR or working on a feature.

Instead of commenting out the tests, why not rename the tests in a way so that only a subset can be run as part of the github workflow using the RUN_TESTS variable.


For example we can create few groupings like
- basic
- advanced
(feel free to use any group names)

and rename the tests like
DoTestCreateSimplePodAzure

Or if you just want to run only two tests, may be run it explicitly in the workflow (much simpler)
eg
RUN_TESTS=DoTestCreateSimplePodAzure CLOUD_PROVIDER=azure make test-e2e 
RUN_TESTS=TestRemoteAttestationAzure CLOUD_PROVIDER=azure make test-e2e

surajssd (approved):

bpradipt (commented):
/lgtm
Thanks @mkulke

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -151,7 +151,8 @@ func DoTestCreatePeerPodContainerWithExternalIPAccess(t *testing.T, e env.Enviro
 
 func DoTestCreatePeerPodWithJob(t *testing.T, e env.Environment, assert CloudAssert) {
 	jobName := "job-pi"
-	job := NewJob(E2eNamespace, jobName)
+	image := "quay.io/prometheus/busybox:latest"
+	job := NewJob(E2eNamespace, image, 8, jobName)
```
bpradipt, line 155: Ordering of arguments is wrong here. Noticed it due to e2e failure
method signature: NewJob(namespace, name string, backoffLimit int32, image string, command ...string)

bpradipt (approved):
/lgtm

PR #1994, opened 2024-08-09
initdata: don't fail process-user-data if absent

mkulke:
currently the process-user-data unit will fail because process-user-data will exit with a non-zero exit code if there is no initdata file on the file system. this PR will skip the initdata processing with a log message and exit successfully in such a case.


bpradipt (approved):
/lgtm

stevenhorsman (approved):
LGTM. Thanks

PR #1995, opened 2024-08-09
ibmcloud: Fix arch suport for selectImage

### LINKED ISSUE #1795: ibmcloud: selectInstanceProfile and selectImage incorrectly handle "architecture" requirements
We provide both amd64 and s390x instance profiles as well as amd64 and s390x images.
```
IBMCLOUD_PODVM_IMAGE_ID: r018-...(amd64-image-id),r018-...(s390x-image-id)
IBMCLOUD_PODVM_INSTANCE_PROFILE_LIST: bx2-2x8,bz2-2x8
```

If we use the "machine_type" annotation to select `bz2-2x8` `selectInstanceProfile` will choose the correct vm flavour however `selectImage` will incorrectly choose the first (amd64) image as the `spec.Arch` check it does does not have the correct information.

To fix this:
1) `updateInstanceProfileSpecList` should populate the `InstanceTypeSpec`s with `Arch` from the `GetInstanceProfileWithContext` result
2) In CreateInstance after calling `selectInstanceProfile` the incoming `spec` should be merged with the instanceProfile `spec`
3) Call `getImage` with the new merged `spec`

skaegi:
This PR ensures the arch information is captured when populating instance profile specs and then ensures that the architecture information is used when selecting the image. This PR also adds accompanying tests.

Fixes #1795

### DIFF
--- a/src/cloud-providers/ibmcloud/provider.go
+++ b/src/cloud-providers/ibmcloud/provider.go
@@ -249,7 +249,7 @@ func (p *ibmcloudVPCProvider) CreateInstance(ctx context.Context, podName, sandb
 		return nil, err
 	}
 
-	imageID, err := p.selectImage(ctx, spec)
+	imageID, err := p.selectImage(ctx, spec, instanceProfile)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


src/cloud-providers/ibmcloud/provider.go
```diff
@@ -335,7 +335,7 @@ func (p *ibmcloudVPCProvider) updateInstanceProfileSpecList() error {
 }
 
 // Add a method to retrieve cpu, memory, and storage from the profile name
```
bpradipt, line 337: nit the comment will need to be changed
`// Add a method to retrieve cpu, memory and arch from the profile name`

stevenhorsman (approved):
Looks okay to me. Thanks!

bpradipt (approved):
/lgtm

PR #1997, opened 2024-08-11
podvm-mkosi: set IMAGE_NAME instead of cp .git

mkulke:
Copying the .git folder can lead to problems, e.g. if the make process aborts, the extra .git folder is not removed and causes confusion in the working tree.

Since the IMAGE_NAME doesn't matter for the mkosi binary builds, we can set it explicitly to a static variable.


src/cloud-api-adaptor/podvm/Dockerfile.podvm_binaries.fedora
```diff
@@ -36,6 +37,7 @@ ENV GUEST_COMPONENTS_REPO=${GUEST_COMPONENTS_REPO}
 ENV ATTESTER=${ATTESTER}
 ENV ARCH=${ARCH}
 ENV DEFAULT_AGENT_POLICY_FILE=${DEFAULT_AGENT_POLICY_FILE}
+ENV IMAGE_NAME=${IMAGE_NAME}
```
huoqifeng, line 40: If I understand correctly, are we going to drop the scripts in `/podvm/Makefile.inc`?
```
ifndef IMAGE_NAME
COMMIT := $(shell	commit=$$(git describe --match '' --dirty --always) && \
					if [ -n "$$(git ls-files --other --exclude-per-directory .gitignore "$(FILES_DIR)")" ]; then \
						commit="$${commit%-dirty}-dirty"; \
					fi && \
	                echo "$$commit")
ifndef COMMIT
$(error Failed to derive an image name. Explicitly define IMAGE_NAME)
endif
IMAGE_NAME := $(IMAGE_PREFIX)-$(COMMIT)-$(DEB_ARCH)
endif
```

mkulke: well, they will still be there, but we will go around them in an mkosi built, by setting IMAGE_NAME.

huoqifeng (approved):
LGTM

stevenhorsman (commented):
So just to double check I understand this PR:
- We previously had the `.git` folder, so that we could generate the IMAGE_NAME in `Makefile.inc` via `git describe...`. We are now using the hard-coded `IMAGE_NAME` to skip the `git describe` logic instead. 
- IMAGE-NAME isn't needed in the podvm-mkosi flow, and is only user in the packer build? I guess this is where QiFeng's question came in that once packer is gone we'll just remove the IMAGE-NAME code altogether and can then remove this build-arg setting?

stevenhorsman (approved):
LGTM. Thanks

PR #1998, opened 2024-08-11
policy: symlink default policy to /run/peerpod

mkulke:
Since policy is provisioned by initdata to /run/peerpod/policy.rego we need to point the symlink to that file.

In case no init-data is provided allow-all.rego is copied by systemd to /run/peerpod/policy.rego

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/etc/kata-opa/default-policy.rego
+++ b/src/cloud-api-adaptor/podvm/files/etc/kata-opa/default-policy.rego
@@ -1,1 +1,1 @@
-allow-all.rego
+/run/peerpod/policy.rego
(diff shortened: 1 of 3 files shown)


stevenhorsman (approved):

huoqifeng (approved):
LGTM
Thank you @mkulke

stevenhorsman:
Hey @mkulke, FYI Paul and I feel into this trap with the mkosi image. If you don't specific the initdata annotation with the policy then `/run/peerpod/policy.rego` doesn't exist and there the kata-agent fails badly. I'm wondering if we could have some fall back to prevent this and pick up the allow-all.rego in that case?

mkulke:
interesting, isn't the `src/cloud-api-adaptor/podvm/files/etc/tmpfiles.d/policy.conf` directive copying ` /etc/kata-opa/allow-all.rego` to `/run/peerpod/policy.rego` anyway? If not, then there's a bug. i'll take a look

stevenhorsman:
FYI I've opened https://github.com/confidential-containers/cloud-api-adaptor/issues/2041 to track this.

PR #1999, opened 2024-08-12
initdata: measure initdata digest into rt register

mkulke:
This adds two post exec directives for process-user-data in the mkosi to extend PCR 8 (grub, which we don't use it mkosi podvms: https://uapi-group.org/specifications/specs/linux_tpm_pcr_registry/) with the digest of initdata. Sha256 and Sha384 banks are attempted. initdata.digest contains a hex value that will fit only in one of those. A failure of either post exec step will be ignored and do not turn the unit status into a failure.

This is a bit provisional, but since things with init-data and runtime measurement are a bit in flux still, which doesn't warrant to put such logic in code yet.


huoqifeng (approved):
LGTM
as `A failure of either post exec step will be ignored and do not turn the unit status into a failure`, thanks @mkulke

liudalibj (approved):
LGTM

PR #2001, opened 2024-08-12
0.9.0 post release updates

stevenhorsman:


src/cloud-api-adaptor/versions.yaml
```diff
@@ -29,7 +29,7 @@ tools:
 git:
   coco-operator:
     url: https://github.com/confidential-containers/operator
-    reference: v0.9.0
+    reference: latest
```
wainersm, line 32: The operator has a regression on the uninstall (https://github.com/confidential-containers/operator/issues/391) but I don't think we ever try to uninstall it on e2e tests. Do you know of any other situation it might be a problem?

stevenhorsman: I don't think so in the CI tests, but I'm not sure if it would annoy people doing development? - I can drop that commit from this PR until it's resolved, or just add a hold on this whole PR?

wainersm: I vote for dropping the commit from this PR, at least we unleash devs to use the latest caa image.

wainersm (approved):
Thanks @stevenhorsman !

src/cloud-api-adaptor/versions.yaml
stevenhorsman: With https://github.com/confidential-containers/operator/pull/413 being merged I've re-added the operator version change as it should be stable now

fitzthum (approved):
LGTM

PR #2004, opened 2024-08-14
Libvirt install latest update and post-release update part II

stevenhorsman:
For reasons I don't fully understand, our `latest` image manifest includes the dev version(s) of CAA, so we don't produce a `dev-latest`, so update the release process to revert the pinning commit instead of trying to be clever.
Also update the CITATION which was missed.

### DIFF
--- a/docs/Release-Process.md
+++ b/docs/Release-Process.md
@@ -211,9 +211,6 @@ If the `main` branch was not already unlocked, then ask an admin to do this now.
 
 The CoCo operator reference commit in the [versions.yaml](../src/cloud-api-adaptor/versions.yaml) should be reverted to use `main` branch.
 
-The changes on the overlay kustomization files should be reverted to start using the latest cloud-api-adaptor images again:
-```
-./hack/release-helper.sh caa-image-tag latest
-```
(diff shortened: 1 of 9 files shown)


docs/Release-Process.md
```diff
@@ -221,9 +221,6 @@ If the `main` branch was not already unlocked, then ask an admin to do this now.
 
 The CoCo operator reference commit in the [versions.yaml](../src/cloud-api-adaptor/versions.yaml) should be reverted to use `main` branch.
 
-The changes on the overlay kustomization files should be reverted to start using the latest cloud-api-adaptor images again:
-```
-./hack/release-helper.sh caa-image-tag latest
-```
+The commit that pinned the overlay kustomization files should be reverted to start using the latest cloud-api-adaptor images again:
```
wainersm, line 224: Good idea!

wainersm: ah, s/:/.

wainersm (approved):
Thanks @stevenhorsman !

beraldoleal (approved):
LGTM, thanks

PR #2007, opened 2024-08-15
Explicitly set proxy dialer timeout

EmmEff:
Use proxy timeout value to set dialer timeout. Default for TCP connections is documented as ~3 minutes, which has proven to be insufficient on AWS.

The proxy timeout value is set through `peer-pods-cm` ConfigMap as `PROXY_TIMEOUT`.


yoheiueda (approved):
LGTM

snir911 (approved):
Thanks

Issue #2012, opened 2024-08-16
TestCloudServiceWithSecureComms is flaky


yoheiueda:
The unit test `TestCloudServiceWithSecureComms` fails in rare cases.

One example is this one.
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/10415163627/job/28845230870#step:7:188

This issue was also reported here https://github.com/confidential-containers/cloud-api-adaptor/pull/1922#issuecomment-2233863655

```
=== RUN   TestCloudServiceWithSecureComms
2024/08/16 05:32:45 [secure-comms] Using PP SecureComms: InitSshServer version v0.2
2024/08/16 05:32:45 [secure-comms] Inbound listening to port 9019
2024/08/16 05:32:45 [secure-comms] SSH service starting on port: 6001
2024/08/16 05:32:45 [secure-comms] Attestation phase: waiting for client to connect
2024/08/16 05:32:45 [secure-comms] Using PP SecureComms: InitSshClient version v0.2
2024/08/16 05:32:45 [secure-comms] CreateSecret 'sshclient'
2024/08/16 05:32:45 [secure-comms] Updating KBS with secret for: default/sshclient/publicKey
2024/08/16 05:32:45 [adaptor/cloud] failed to create PeerPodService, runtime failure may result in dangling resources NewPeerPodService: failed to get cloudProvider
2024/08/16 05:32:45 [adaptor/cloud] credential file /root/containers/auth.json is not present, skipping image auth config
2024/08/16 05:32:45 [adaptor/cloud] stored /tmp/TestCloudServiceWithSecureComms2170180551/001/123/daemon.json
2024/08/16 05:32:45 [adaptor/cloud] initdata: 
2024/08/16 05:32:45 [adaptor/cloud] create a sandbox 123 for pod mypod in namespace default (netns: )
2024/08/16 05:32:45 [adaptor/cloud] created an instance abc for sandbox 123
2024/08/16 05:32:45 [secure-comms] InitPP read/create PP secret named: pp-123
2024/08/16 05:32:46 [secure-comms] CreateSecret 'pp-123'
2024/08/16 05:32:46 [secure-comms] Updating KBS with secret for: default/pp-123/privateKey
2024/08/16 05:32:46 [secure-comms] Inbound listening to port 34959
2024/08/16 05:32:46 [secure-comms] Attestation phase: starting
[...]

PR #2014, opened 2024-08-18
cloud-api-adaptor: SecureComms fix panic for closing chan twice

davidhadas:
Fixes: #2012
Avoid closing an already closed chan.

### DIFF
--- a/src/cloud-api-adaptor/pkg/securecomms/sshproxy/sshproxy.go
+++ b/src/cloud-api-adaptor/pkg/securecomms/sshproxy/sshproxy.go
@@ -34,7 +34,6 @@ var logger = sshutil.Logger
 type SshPeer struct {
 	sid            string
 	phase          string
-	terminated     string
 	sshConn        ssh.Conn
 	ctx            context.Context
 	done           chan bool
@@ -43,6 +42,7 @@ type SshPeer struct {
 	wg             sync.WaitGroup
 	upgrade        bool
 	outboundsReady chan bool
+	closeOnce      sync.Once
 }
 
 // Inbound side of the Tunnel - incoming tcp connections from local clients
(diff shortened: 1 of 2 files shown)


bpradipt (approved):
/lgtm

src/cloud-api-adaptor/pkg/securecomms/sshproxy/sshproxy.go
```diff
@@ -298,9 +299,15 @@ func (peer *SshPeer) Wait() {
...
 func (peer *SshPeer) Close(who string) {
+	peer.mutex.Lock()
+	defer peer.mutex.Unlock()
+
 	if peer.terminated == "" {
 		logger.Printf("%s phase: peer done by >>> %s <<<", peer.phase, who)
 		peer.terminated = who
+		if peer.terminated == "" {
+			peer.terminated = "(unspecified)"
+		}
```
yoheiueda, line 310: Can we use [`sync.Once`](https://pkg.go.dev/sync#Once) here to simplify the logic?

davidhadas: The function may run multiple times (for different peer objects). So it seems based on what I read that sync.Once does not fit here.

yoheiueda: I think you can add a `sync.Once` object into the `SshPeer` struct instead of `sync.Mutex`, then we can call `Close()` multiple times for different peer objects.

```
func (peer *SshPeer) Close(who string) {
    peer.once.Do(func() {
        logger.Printf("%s phase: peer done by >>> %s <<<", peer.phase, who)
        peer.sshConn.Close()
        close(peer.done)
   })
}
```

I think we can also eliminate the `terminate` field from the struct.

yoheiueda (approved):
LGTM. Thank you very much!

Issue #2015, opened 2024-08-22
Connection tracking of VXLAN UDP packets must be disabled


yoheiueda:
The Linux connection tracking system (conntrack) always tracks UDP packets as connections by default. This is OK for usual UDP communications, but does not work with VXLAN UDP packets, since a destination port of each VXLAN UDP packet is fixed (4789) in both directions, and a source port of each packet is randomly selected. This UDP packet flow does not creates a connection flow, and every UDP packet of a VXLAN tunnel is recognized as a first packet of a new UDP stream.

We can confirm this behavior with the `conntrack` command. We can see a lot of "UNREPLIED" conntrack entries as follows.
```
# conntrack -L | grep "udp" | grep "UNREPLIED" | grep "dport=4789"
conntrack v1.4.6 (conntrack-tools): 50 flow entries have been shown.
udp      17 26 src=192.168.122.185 dst=192.168.122.121 sport=34196 dport=4789 [UNREPLIED] src=192.168.122.121 dst=192.168.122.185 sport=4789 dport=34196 mark=0 use=1
udp      17 27 src=192.168.122.121 dst=192.168.122.185 sport=46394 dport=4789 [UNREPLIED] src=192.168.122.185 dst=192.168.122.121 sport=4789 dport=46394 mark=0 use=1
udp      17 25 src=192.168.122.121 dst=192.168.122.185 sport=52605 dport=4789 [UNREPLIED] src=192.168.122.185 dst=192.168.122.121 sport=4789 dport=52605 mark=0 use=1
udp      17 27 src=192.168.122.121 dst=192.168.122.185 sport=59264 dport=4789 [UNREPLIED] src=192.168.122.185 dst=192.168.122.121 sport=4789 dport=59264 mark=0 use=1
udp      17 25 src=192.168.122.185 dst=192.168.122.121 sport=59949 dport=4789 [UNREPLIED] src=192.168.122.121 dst=192.168.122.185 sport=4789 dport=59949 mark=0 use=1
udp      17 25 src=192.168.122.185 dst=192.168.122.121 sport=57929 dport=4789 [UNREPLIED] src=192.168.122.121 dst=192.168.122.185 sport=4789 dport=57929 mark=0 use=1
udp      17 25 src=192.168.122.121 dst=192.168.122.185 sport=57557 dport=4789 [UNREPLIED] src=192.168.122.185 dst=192.168.122.121 sport=4789 dport=57557 mark=0 use=1
[...]

PR #2016, opened 2024-08-23
Install iptables in cloud-api-adaptor daemonset

yoheiueda:
This PR is a prerequisite to fix https://github.com/confidential-containers/cloud-api-adaptor/issues/2015

To run the iptables command within the cloud-api-adaptor container, we need a wrapper command for iptables, and volumes for accessing a lock file and kernel modules in the host side.

https://github.com/kubernetes-sigs/iptables-wrappers

Network-related projects such as [kube-proxy](https://github.com/kubernetes/release/blob/54644168fde905d452980d0d34984a993e6b24a5/images/build/distroless-iptables/distroless-bullseye/Dockerfile#L45-L47) and [flannel](https://github.com/flannel-io/flannel/blob/bb077808679ff71527d5a9a0ff7ffbf7d108dad4/images/Dockerfile#L20)  use this wrapper to manipulate netfilter on worker nodes.

Another PR that actually fixes the issue using iptables will follow after this PR gets merged.

### DIFF
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -8,10 +8,14 @@ ARG BASE=registry.fedoraproject.org/fedora:39
 # that was specified with --platform. For more details see:
 # https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
 FROM --platform=$BUILDPLATFORM $BUILDER_BASE AS builder-release
+ARG YQ_VERSION
+RUN go install github.com/mikefarah/yq/v4@$YQ_VERSION
+
 # For `dev` builds due to CGO constraints we have to emulate the target platform
 # instead of using Go's cross-compilation
 FROM --platform=$TARGETPLATFORM $BUILDER_BASE AS builder-dev
-
+ARG YQ_VERSION
+RUN go install github.com/mikefarah/yq/v4@$YQ_VERSION
 RUN dnf install -y libvirt-devel && dnf clean all
 
 FROM builder-${BUILD_TYPE} AS builder
@@ -19,9 +23,6 @@ ARG RELEASE_BUILD
 ARG COMMIT
 ARG VERSION
 ARG TARGETARCH
-ARG YQ_VERSION
-
-RUN go install github.com/mikefarah/yq/v4@$YQ_VERSION
 
 WORKDIR /work
 COPY cloud-api-adaptor/go.mod cloud-api-adaptor/go.sum ./cloud-api-adaptor/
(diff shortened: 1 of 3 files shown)


src/cloud-api-adaptor/Dockerfile
```diff
@@ -8,20 +8,21 @@ ARG BASE=registry.fedoraproject.org/fedora:39
...
 # For `dev` builds due to CGO constraints we have to emulate the target platform
 # instead of using Go's cross-compilation
 FROM --platform=$TARGETPLATFORM $BUILDER_BASE AS builder-dev
-
+ARG YQ_VERSION
+RUN go install github.com/mikefarah/yq/v4@$YQ_VERSION
 RUN dnf install -y libvirt-devel && dnf clean all
 
-FROM builder-${BUILD_TYPE} AS builder
+FROM builder-${BUILD_TYPE}  AS builder
```
bpradipt, line 21: looks unintended space?

src/cloud-api-adaptor/Dockerfile
```diff
@@ -37,8 +38,25 @@ COPY cloud-api-adaptor/proto ./proto
 
 RUN CC=gcc make ARCH=$TARGETARCH COMMIT=$COMMIT VERSION=$VERSION RELEASE_BUILD=$RELEASE_BUILD cloud-api-adaptor
 
+FROM builder-release AS iptables
+
+ARG TARGETARCH
+
+WORKDIR /iptables
+RUN --mount=type=bind,target=/versions.yaml,source=cloud-api-adaptor/versions.yaml,readonly \
```
bpradipt, line 46: This format is new to me. Do you have any reference on using `--mount` with RUN? 
Is it docker specific or can work with podman also ?

src/cloud-api-adaptor/Dockerfile
yoheiueda: Oh, yes. Fixed.

src/cloud-api-adaptor/Dockerfile
yoheiueda: We can specify cache mounts and bind mounts to a RUN instruction in a Dockerfile.

https://docs.docker.com/build/guide/mounts/

Podman supports the feature as well.

https://github.com/containers/common/blob/10ced654d57245a78d1b2c3746c5fd99209370bd/docs/Containerfile.5.md?plain=1#L102

bpradipt (approved):
/lgtm

src/cloud-api-adaptor/versions.yaml
```diff
@@ -25,6 +25,7 @@ tools:
   protoc: 3.15.0
   packer: v1.9.4
   kcli: 99.0.202407031308
+  iptables-wrapper: v0.0.0-20240819165702-06cad2ec6cb5
```
stevenhorsman, line 28: Just to double check by understanding - iptable-wrapper is a tool as it isn't itself installed into the container image, but is just used to install iptables?

snir911 (approved):
LGTM thanks

src/cloud-api-adaptor/versions.yaml
bpradipt: Both are added to the container image to manipulate iptables rules

https://github.com/confidential-containers/cloud-api-adaptor/pull/2016/files#diff-18d9e213a3bf84ae4d6d02e3b81c7aadb80a88556e3e14d5dcf09d27069bc152R56

Also - https://github.com/kubernetes-sigs/iptables-wrappers

stevenhorsman: So does that mean it shouldn't go into the `tools` section, but should be in an `external`, or `dependencies` section in version.yaml? Or am I over-thinking this?

bpradipt: Since iptables-wrapper is a tool to manage iptables program, logically it makes sense to be in the tools section. Similar to golang, rust etc..

bpradipt: But I have no preference. If any other section makes more sense that's fine with me as well

yoheiueda: @stevenhorsman @bpradipt 

I have no preference either.

Another possible option is to put it in the `git` section, and build `iptables-wrapper` from a cloned git repo. Currently, the builder image does not include the `git` command, so we need to install it for this option.

What do you think?

yoheiueda: @stevenhorsman Thank you. I agree.

I keep it in `tools` for now, and will update that it if necessary.

BTW, the current code to build the command using `go install` looks somewhat complicated due to cross compilation.

There is a discussion ongoing https://github.com/golang/go/issues/44469, and we will update the code when a simpler option is available.

stevenhorsman:
I think that given we have two approvals and are satisfied with the versions.yaml updates then we can merge this. Thanks @yoheiueda!

PR #2017, opened 2024-08-27
Kata 3.8.0 versions bump

stevenhorsman:
Bump components to sync with kata v3.8.0 release to try and ensure we stay up-to-date (ish)

### DIFF
--- a/src/cloud-api-adaptor/versions.yaml
+++ b/src/cloud-api-adaptor/versions.yaml
@@ -34,10 +34,10 @@ git:
     reference: main
   guest-components:
     url: https://github.com/confidential-containers/guest-components
-    reference: v0.9.0
+    reference: d996c692207a983426ae0043952d15ed18e84f66
   kata-containers:
     url: https://github.com/kata-containers/kata-containers
-    reference: a113fc93c8fe75cf58b9b6edadd07a3ebfd665f3
+    reference: 3.8.0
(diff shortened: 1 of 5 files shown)


stevenhorsman:
It looks like libsecomp from kata-containers is causing the License failure as it's LGPL-2.1. I'm not sure why Magnus' previous ignore for this isn't working - maybe it was for a specific version? I tried to add my own matching ignore rule and I got a permission error.

src/cloud-api-adaptor/versions.yaml
```diff
@@ -45,11 +45,11 @@ git:
     reference: v1.5.0
   kbs:
     url: https://github.com/confidential-containers/trustee
-    reference: e890fc90c384207668fa3a4d6a2f2a2d652797ee
```
wainersm, line 48: Hi @stevenhorsman !

e890f is the trustee version used on kata's 3.8.0  (https://github.com/kata-containers/kata-containers/blob/3.8.0/versions.yaml#L237). Do we need a new version? that's why this update?

stevenhorsman: Yeah sorry, this change was reverted: https://github.com/kata-containers/kata-containers/pull/10178, so I'm not sure why I picked the wrong one. I'll revert this, Thanks!

PR #1965, opened 2024-07-31
CI: run e2e tests for docker provider

wainersm:
This adds an initial version of CI for docker provider.

It was created a new workflow (e2e_docker.yaml) that run the e2e tests for docker on x86_64. In order to keep this implementation simple, it isn't installed KBS. Also the tests will be executed only alongside the [libvirt nightly job](https://github.com/confidential-containers/cloud-api-adaptor/actions/workflows/daily-e2e-tests-libvirt.yaml), so not triggered on pull requests. The final status of libvirt nightly job shouldn't change if the docker tests fail because the jobs were marked continue-on-error, this way we will be able to proper test these changes on "production" (reason why I also didn't rename `daily e2e tests for libvirt` to something generic).

As a pre-req to run the e2e tests for docker, I had to ensure the podvm image for docker is built and pushed to the registry. So the very first commits changes the podvm_builder, podvm_binaries, podvm_mkosi, and podvm workflows to reuse existing code. Here I am trying really hard to not introduce new workflows. However, I will need to change further those workflows (on a respin PR) to make them more flexible to allow, for example, run the docker e2e tests on pull requests.

I managed to tests these changes on my forked tree, being the last execution https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/10171362457 .  From that link you will notice that `Build mkosi image` passed but `test-docker` failed, the error being:

```
# github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/e2e [github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/e2e.test]
Error: test/e2e/docker_test.go:104:23: keyBrokerService.EnableKbsCustomizedPolicy undefined (type *provisioner.KeyBrokerService has no field or method EnableKbsCustomizedPolicy)
Error: test/e2e/docker_test.go:108:23: keyBrokerService.EnableKbsCustomizedPolicy undefined (type *provisioner.KeyBrokerService has no field or method EnableKbsCustomizedPolicy)
``` 

@bpradipt ^^^ I need help to understand what's going on.

Fixes: #1946

PR #2019, opened 2024-08-28
workflows: Add new mkosi e2e flow

stevenhorsman:
- Combine the s390x and amd64 mkosi workflows together and enhance to create a qcow2 output that is cached with oras, push intermediate steps to GHCR and then pull these cached versions down if available to speed up builds
- Re-work the libvirt e2e flow to add support for mkosi podvm image testing and decouple the architect tests
- Enable EFI optional support for libvirt for x86 mkosi
- Enable cloud-init on the mkosi image for libvirt


stevenhorsman:
FYI, I've moved on a lot from this approach to try and build on top of #1965. I'm missed about the best strategy and whether it's worth the complexity to keep separate workflows for the podvm's builder, binaries and image stage, so I've posted on slack to see if others have thoughts.

stevenhorsman:
Ok, I've re-worked this to have a single mkosi workflow and not interlink with the separate stages. It's still WIP though and in test!

.github/workflows/podvm_mkosi.yaml
```diff
@@ -19,81 +46,140 @@ defaults:
...
-      - name: Prepare .git folder
-        working-directory: ./
+      - name: Build builder
+        id: build_builder
+        working-directory: src/cloud-api-adaptor/podvm-mkosi
         run: |
-          cp -rf .git src/.git
+          # If the builder image matching our git sha exists, then skip this build
+          builder_image=${{ inputs.registry }}/podvm-builder-fedora-${{ inputs.arch }}
+          tag=${{ github.sha }}
```
wainersm, line 137: Hi @stevenhorsman !

I'd like to review the use of `github.sha` throughout the workflow. Default case `actions/checkout@v4` checks out the code from `git_ref` (mandatory workflow input). Does  `actions/checkout@v4` update the value of `github.sha`? (I think so, but double-checking...). Even if `github.sha` is updated by `actions/checkout@v4`; there is the case of `pull_request_target` where it `rebase-atop-of-the-latest-target-branch`, is `github.sha` correct?

stevenhorsman: Nice points - do you think that something like `git rev-parse --short HEAD` would be better? Part of me worries about the sha not clearly mapping back to a test commit, which might back the transparency of code slightly tougher, but that shouldn't be a problem for non-PR branches where it's more important and I think having the updated builder image is more important, so it's a worthwhile change.

.github/workflows/podvm_mkosi.yaml
```diff
@@ -19,81 +46,140 @@ defaults:
...
+
+          if ! docker manifest inspect ${builder_image}:${tag}; then
+            PODVM_BUILDER_IMAGE="${builder_image}:${tag}" make fedora-binaries-builder
+          fi
+          echo "image=${builder_image}:${tag}" | tee -a "$GITHUB_OUTPUT"
+
+          # If the input has a different image-tag then also push it with that tag
+          if [ -n "${{ inputs.image_tag }}" ] && [ "${{ inputs.image_tag }}" != ${{ github.sha }} ];then
```
wainersm, line 145: Next commit enables `make fedora-binaries-builder` to also push the image, based on value of `PUSH` envvar. I glanced over all commits, it seems that `PUSH` is always `true`. In future it might be `false` on depending on, for example, a workflow input. So I think this `if` should also check the value of `PUSH`.

On a related matter, as is this workflow it will always push the image. Even on pull requests, transient (disposable) images will be pushed. Today we already push images for pull requests, but these are easily traceable by tag (`pr-number`)  which allow us to run clean up routines like in https://github.com/confidential-containers/cloud-api-adaptor/pull/1443 . When we push transient images tagged by sha it's harder to do removal. I'm not saying we should care, in fact I'm okay with pushing all transient images; I just wanted to raise that for general awareness.

stevenhorsman: Yeah, I deliberately chose to cache the binaries images as they can take quite a long time to build (though with the caching of components maybe this isn't a problem anymore) and In the previous flow there was some logic to pass through a prior binaries build to speed this up, which I didn't think was helpful.

I'm not completely sure if there is a problem with having more and more tags in ghcr (the kata-agent cache for example has ~13.5 tags), but if there is maybe we can turn PUSH off for PR builds if you'd prefer, or prefix the tag with the pr number?

.github/workflows/podvm_mkosi.yaml
```diff
@@ -104,32 +190,30 @@ jobs:
           remove-docker-images: "true"
 
       - name: Second checkout
```
wainersm, line 218: Checking I understood correctly, it needs to checkout again because of the previous "maximize build space" step?

Anyway, I think we should run the checkout magic for when pull_request_target too.

stevenhorsman: I think so, but that change predates me

> Anyway, I think we should run the checkout magic for when pull_request_target too.

Do you mean do the rebase after this step?

wainersm: yes, rebase after this step if pull_request_target.

stevenhorsman: Done. Thanks

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -132,6 +137,7 @@ image-container:
 	docker buildx build \
 		-t $(PODVM_CONTAINER_NAME):$(PODVM_TAG) \
 		-t $(PODVM_CONTAINER_NAME):latest \
```
wainersm, line 124: Maybe remove `-t $(PODVM_CONTAINER_NAME):latest` for consistency.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -60,7 +68,8 @@ endif
 		--build-arg VERIFY_PROVENANCE=$(VERIFY_PROVENANCE) \
 		$(if $(AUTHFILE),--build-arg AUTHFILE=$(AUTHFILE),) \
 		$(if $(DEFAULT_AGENT_POLICY_FILE),--build-arg DEFAULT_AGENT_POLICY_FILE=$(DEFAULT_AGENT_POLICY_FILE),) \
-		-o type=local,dest="./resources/binaries-tree" \
+		$(if $(PUSH),,-o type=local,dest="./resources/binaries-tree") \
```
wainersm, line 71: @stevenhorsman , I didn't get this statement.

Also, does it work as intended?
```
wmoschet@wmoschet-thinkpadp16vgen1:/tmp$ cat Makefile 
PUSH ?= false
all:
	@echo $(if $(PUSH),,-o type=local,dest="./resources/binaries-tree")
wmoschet@wmoschet-thinkpadp16vgen1:/tmp$ make

wmoschet@wmoschet-thinkpadp16vgen1:/tmp$ make PUSH=false

```

.github/workflows/e2e_run_all.yaml
```diff
@@ -66,6 +66,24 @@ jobs:
       git_ref: ${{ inputs.git_ref }}
     secrets: inherit
 
+  podvm_mkosi_amd64:
```
wainersm, line 69: combine the two jobs in a matrix strategy?

src/cloud-api-adaptor/podvm/hack/cross-build-extras.sh
```diff
@@ -17,5 +17,12 @@ libc=$([[ $ARCH =~ s390x ]] && echo "gnu" || echo "musl")
...
 
 rustup target add "$rustTarget"
-apt install -y "qemu-system-$ARCH"
-apt install -y "gcc-$ARCH-linux-$libc"
+
+source /etc/os-release || source /usr/lib/os-release
+if [[ ${ID_LIKE:-} == *"debian"* ]]; then
+    apt install -y "qemu-system-$ARCH"
+    apt install -y "gcc-$ARCH-linux-$libc"
+elif [[ "${ID_LIKE:-}" =~ "fedora" ]]; then
```
wainersm, line 25: In my workstation with Fedora 40 there isn't `ID_LIKE`:
```
wmoschet@wmoschet-thinkpadp16vgen1:/tmp$ grep ID_LIKE /etc/os-release
wmoschet@wmoschet-thinkpadp16vgen1:/tmp$ grep ^ID /etc/os-release
ID=fedora
```

Also, maybe take that opportunity to add an `else echo "Unsupported distro $ID"; exit 1`

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -115,25 +104,17 @@ ifeq ($(SE_BOOT),true)
 	sudo mkosi --profile debug.conf
 	sudo -E ../hack/build-s390x-se-image.sh
 	@echo "Building docker image..."
```
wainersm, line 106: Remove this `@echo ...`

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -115,25 +104,17 @@ ifeq ($(SE_BOOT),true)
...
-	sudo chown -R $(USER): build
-	docker buildx build \
-		-t $(addsuffix -debug,$(PODVM_IMAGE)) \
-		$(DOCKER_OPTS) \
-		-f ../podvm/Dockerfile.podvm.fedora .
 else ifeq ($(ARCH),s390x)
 	touch resources/buildS390xImage
 	sudo mkosi --profile debug.conf
 	sudo -E ../hack/build-s390x-image.sh
 	@echo "Building docker image..."
```
wainersm, line 111: Same here

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -82,25 +82,14 @@ ifeq ($(SE_BOOT),true)
...
-	sudo chown -R $(USER): build
-	docker buildx build \
-		-t $(PODVM_IMAGE) \
-		$(DOCKER_OPTS) \
-		-f ../podvm/Dockerfile.podvm.fedora .
 else
 	touch resources/buildBootableImage
-	nix develop ..#podvm-mkosi --command mkosi --environment=VARIANT_ID=production
+	sudo -E env PATH=$(PATH) nix develop ..#podvm-mkosi --command mkosi --environment=VARIANT_ID=production
+	qemu-img convert -f raw -O qcow2 build/system.raw build/podvm-amd64.qcow2
```
wainersm, line 92: Could we have the qcow2's name close to the image's? For example, `podvm-generic-$(PODVM_DISTRO)-$(ARCH).qcow2`

.github/workflows/podvm_mkosi.yaml
```diff
@@ -242,6 +248,38 @@ jobs:
...
+          cp ../build/podvm-*.qcow2 .
+          tar cJf podvm.tar.xz podvm-*.qcow2
+          image=${{ inputs.registry }}/podvm-generic-fedora
+          if [ "${{inputs.debug}}" = "true" ]; then
+            image=${image}-debug
+          fi
+          tag=${{ steps.image_tag.outputs.image_tag }}
+          arch_tag=${tag}_${{ inputs.arch }}
+          oras push "${image}:${arch_tag}" podvm.tar.xz
+          # We need to create the platform annotations with docker, since oras 1.2 doesn't support
```
wainersm, line 276: Thanks for this comment, I didn't know that ;)

.github/workflows/podvm_mkosi.yaml
```diff
@@ -242,6 +248,38 @@ jobs:
...
+          cd oras
+          cp ../build/podvm-*.qcow2 .
+          tar cJf podvm.tar.xz podvm-*.qcow2
+          image=${{ inputs.registry }}/podvm-generic-fedora
+          if [ "${{inputs.debug}}" = "true" ]; then
+            image=${image}-debug
+          fi
+          tag=${{ steps.image_tag.outputs.image_tag }}
+          arch_tag=${tag}_${{ inputs.arch }}
+          oras push "${image}:${arch_tag}" podvm.tar.xz
```
wainersm, line 275: Suppose `${{ inputs.image_tag }}` isn't passed:

* builder: `${{ inputs.registry }}/podvm-builder-fedora-${{ inputs.arch }}:${{ github.sha }}`
* binaries: `${{ inputs.registry }}/podvm-binaries-fedora-${{ inputs.arch }}:${{ github.sha }}`
* podvm (oras) (not debug): `${{ inputs.registry }}/podvm-generic-fedora:${{ github.sha }}__${{ inputs.arch }}`
* podvm (oras) (debug): `${{ inputs.registry }}/podvm-generic-fedora-debug:${{ github.sha }}__${{ inputs.arch }}`

Is there a reason for podvm (oras) doesn't follow the builder/binaries pattern name? i.e, arch on the image's name instead of the tag.

If we name it as `${{ inputs.registry }}/podvm-generic-fedora-${{ inputs.arch }}:${{ github.sha }}` and push as oras then it may collide with the packer based image (which would have the same name, I think)?

Talking about packer builds... we publish, for example, `${{ inputs.registry }}/podvm-builder-fedora-${{ inputs.arch }}:${{ github.sha }}` and this mkosi workflow has a caching mechanism. It might end up using the built-on-packer-workflow builder image. Would it be a problem? Last time I check, IIRC, this isn't a problem but better ask than regret.

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: Consistency with what? The other mkosi generated image have both the `podvm` tag and `latest` tag?

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: So the original code didn't push the binaries image to a container registry, but just stored it locally, but for performance reasons I wanted to cache it, so it could be re-used to subsequent builds if the binaries hadn't changed. That's why I added `PUSH` which sets the `--push` flag, like we have used elsewhere. I then figured that if we were pushing then we probably didn't need to also output to a local directory, so I tried to only enable this option if `PUSH` was false. I've probably messed up the syntax though, so I'll try and fix it.

Does that help & make sense, or do you think this is unecessary/incorrect in principle as well as execution?

stevenhorsman: Ok, I've fixed the syntax error now:
```
root@sh-libvirt-x86:~# cat Makefile
PUSH ?= false

all:
	echo $(if $(filter $(PUSH),true),,-o type=local,dest="./resources/binaries-tree")
root@sh-libvirt-x86:~# make
echo -o type=local,dest="./resources/binaries-tree"
-o type=local,dest=./resources/binaries-tree
root@sh-libvirt-x86:~# PUSH=true make
echo

```

.github/workflows/e2e_run_all.yaml
stevenhorsman: Which jobs? Do you mean `podvm_mkosi_amd64` and `podvm_mkosi_s390x`? Maybe I'm missing something clever, but as I've explained a few things people wanted the amd64 and s390x e2e flows to be completely decoupled, so we don't get delayed/blocked by specific issues in an architecture.

src/cloud-api-adaptor/podvm/hack/cross-build-extras.sh
stevenhorsman: Ah, I was expected that like ubuntu & debian share the same ID_LIKE, that rhel & fedora would as well as it looks like RHEL follows that pattern:
```
# cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="8.6 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
```

I've updated it to handle ID=fedora and add the unsupported distro. Thanks!

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: I think that's a good point, I don't think `generic` should actually be used anymore, if we are comfortable to drop vsphere support as everything use the "generic" image now, but we should reflect that in the podvm image name as well if everyone is happy.

.github/workflows/podvm_mkosi.yaml
```diff
@@ -242,6 +248,38 @@ jobs:
...
+        run: |
+          mkdir oras
+          cd oras
+          cp ../build/podvm-*.qcow2 .
+          tar cJf podvm.tar.xz podvm-*.qcow2
+          image=${{ inputs.registry }}/podvm-generic-fedora
+          if [ "${{inputs.debug}}" = "true" ]; then
+            image=${image}-debug
+          fi
+          tag=${{ steps.image_tag.outputs.image_tag }}
```
wainersm, line 288: `steps.image_tag.outputs.image_tag` is optional and defaults to empty. Maybe we should make it required to avoid further troubles?

wainersm: Never mind. It's not read from the inputs.

wainersm (approved):
Hey @stevenhorsman ! I tried my best to find any bug but (hopefully) it seems all good. Great work! thx!

.github/workflows/podvm_mkosi.yaml
stevenhorsman: I think you are right - the multi-arch approach isn't really working for decoupled builds, so I will go back to `podvm-fedora[-debug]-<arch>:tag`

.github/workflows/podvm_mkosi.yaml
stevenhorsman: Ok, I think I've address this now

stevenhorsman:
Ok, so the latest run (https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12123843432/job/33800871552) passed all tests and I think I've address Wainer's points. There is still a bit of scenario testing I can do, but I think it's open for reviews now.

.github/workflows/e2e_libvirt.yaml
```diff
@@ -81,35 +97,58 @@ jobs:
...
-        with:
-          version: ${{ env.ORAS_VERSION }}
-
       - name: Install gh cli
         run: |
           sudo apt install -y gh
 
+      - name: Double check that OVMF is installed
+        run: |
+          sudo apt install -y ovmf
```
mkulke, line 122: I think it makes sense to always perform, `sudo apt-get install update` before installing. in theory you only need to do it once, but when rearranging steps that can lead to breakage sometimes.

.github/workflows/e2e_libvirt.yaml
```diff
@@ -81,35 +97,58 @@ jobs:
...
         run: |
           command -v kustomize >/dev/null || \
           curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | \
             sudo bash -s /usr/local/bin
 
+# For the legacy packer approach we don't want to use the default firmware, so comment it out
+      - name: Set blank firmware for packer libvirt tests
+        if: ${{ !inputs.oras }}
+        run: |
+          pushd "install/overlays/libvirt"
```
mkulke, line 134: why pushd/popd? doesn't hurt, but every step will execute in its own shell context, so popd is a noop.

.github/workflows/podvm_mkosi.yaml
```diff
@@ -3,90 +3,209 @@ name: Create a Pod VM image with mkosi
...
+      - name: Install crane with go
+        if: ${{ inputs.arch == 's390x' }}
+        run: |
+          go install github.com/google/go-containerregistry/cmd/crane@latest
+          echo "PATH=${PATH}:$(go env GOPATH)/bin" >> "$GITHUB_ENV"
+
         # This removes the checkout and creates a btrfs volume with maximized
         # build space.
       - name: Maximize build space
+        if: ${{ inputs.arch == 'amd64' }}
```
mkulke, line 208: I don't think we need this step anymore, do we?

.github/workflows/podvm_mkosi.yaml
```diff
@@ -97,32 +216,91 @@ jobs:
...
           crane export \
-            ${{ steps.binaries-image.outputs.image }} \
+            ${{ steps.build_binaries.outputs.image }} \
             podvm-binaries.tar
           mkdir -p podvm-mkosi/resources/binaries-tree
           tar xf podvm-binaries.tar -C podvm-mkosi/resources/binaries-tree
           rm podvm-binaries.tar
 
-      - name: Build image
+      - name: Build mkosi debug image
```
mkulke, line 241: if it's toggleable, we probably shouldn't call it debug image by default (build mkosi image) or maybe we can 2 steps with "if" clauses

.github/workflows/podvm_mkosi.yaml
```diff
@@ -97,32 +216,91 @@ jobs:
...
+          fi
+
+      - name: Upload the qcow2 with oras
+        id: publish_oras_qcow2
+        working-directory: src/cloud-api-adaptor/podvm-mkosi
+        run: |
+          mkdir oras
+          cd oras
+          cp ../build/podvm-*.qcow2 .
+          tar cJf podvm.tar.xz podvm-*.qcow2
```
mkulke, line 259: are we missing a step? I thought make image was creating a raw image, so we need to convert it to qcow2

.github/workflows/podvm_mkosi.yaml
```diff
@@ -3,90 +3,209 @@ name: Create a Pod VM image with mkosi
...
+            PODVM_BUILDER_IMAGE="${builder_image}:${tag}" make fedora-binaries-builder
+          fi
+          echo "image=${builder_image}:${tag}" | tee -a "$GITHUB_OUTPUT"
+
+          # If the input has a different image-tag then also push it with that tag
+          if [ -n "${{ inputs.image_tag }}" ] && [ "${{ inputs.image_tag }}" != "${tag}" ];then
+            docker pull "${builder_image}:${tag}"
+            docker tag "${builder_image}:${tag}" ${builder_image}:${{ inputs.image_tag }}
+            docker push "${builder_image}:${{ inputs.image_tag }}"
+          fi
```
mkulke, line 149: it's not obvious to me what this does. what is the improvement over building the builder image locally? build time on re-runs?

mkulke (approved):
some nits

.github/workflows/podvm_mkosi.yaml
stevenhorsman: Yeah exactly just to cache for re-runs. In future I'd consider a "smarter" change mechanism than the sha of the whole repo as the builder tag (as I think it's probably only influenced by the version.yaml and builder Dockerfile, which would lead to us re-using much more often, but It might be that we merge the builder and binaries into a multi-stage dockerfile first.

.github/workflows/podvm_mkosi.yaml
stevenhorsman: the `make image` and `make image-debug` targets create the qcow2, mostly as the s390x one already did that, so having them all the same was easier, if you'd prefer that to be separated out let me know.

.github/workflows/podvm_mkosi.yaml
stevenhorsman: I'm not sure - on the github runner we did run out of space a lot in the past. I will try without this step tomorrow though.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -60,79 +68,60 @@ endif
...
+		$(if $(filter $(PUSH),true),,-o type=local,dest="./resources/binaries-tree") \
+		$(DOCKER_OPTS) \
 		-f ../podvm/Dockerfile.podvm_binaries.fedora ../../
 
 PHONY: image
 image:
 	@echo "Enabling production preset..."
 	rm -rf resources/build*Image
-	rm -rf ./build
+	sudo rm -rf ./build
```
mkulke, line 79: can we avoid sudo here? it would be annoying for local or in-docker runs. we can maybe call make with sudo in the workflow.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -60,79 +68,60 @@ endif
...
+	qemu-img convert -f raw -O qcow2 build/system.raw build/podvm-$(PODVM_DISTRO)-$(ARCH).qcow2
 endif
 
 PHONY: image-debug
 image-debug:
 	@echo "Enabling debug preset..."
 	rm -rf resources/build*Image
 	touch resources/buildDebugImage
-	rm -rf ./build
+	sudo rm -rf ./build
```
mkulke, line 100: see above

.github/workflows/podvm_mkosi.yaml
mkulke: no, that's fine, since raw images are still available.

.github/workflows/e2e_libvirt.yaml
stevenhorsman: Good idea. I've updated this and added to a couple of the other places it was missing too in this commit.

.github/workflows/e2e_libvirt.yaml
stevenhorsman: Good point - I think I just copied the pattern from somewhere else, but agree that it's not necessary.

.github/workflows/podvm_mkosi.yaml
stevenhorsman: Thanks, I've done the later as I think it's a bit clearer even if there is a little bit of duplication in the workflow now.

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: Yeah, I'll try and revert this commit and call sudo from the workflow. Thanks for the idea!

stevenhorsman:
@wainersm & @mkulke - I think I've address all the comments left and re-tested and the e2e flow: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12137733841 and podvm mkosi builds https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12138531982, https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12138528028 are still passing, so please feel free to check my responses are satifactory

wainersm:
> @wainersm & @mkulke - I think I've address all the comments left and re-tested and the e2e flow: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12137733841 and podvm mkosi builds https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12138531982, https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12138528028 are still passing, so please feel free to check my responses are satisfactory

Just found a quote for this occasion which might be perfectly mistakenly attributed to Oscar Wilde: "Everything is dangerous. If it wasn't so, life wouldn't be worth living". In other words, let's merge it!

PR #2020, opened 2024-08-28
Fix script_dir in import.sh

squarti:


stevenhorsman (approved):
LGTM. Thanks!

stevenhorsman (changes requested):
Sorry, actually I think we'll need a one liner in the commit body and a `ibmcloud:` prefix in the commit header would be good to for the commit message check. Thanks!

stevenhorsman (approved):
LGTM. Thanks!

PR #2018, opened 2024-08-27
Podvm image caching

stevenhorsman:
Re-use the kata-containers caches to save time in podvm building.

PR #2026, opened 2024-09-03
Podvm image caching kata agent

stevenhorsman:
Re-use the kata-containers caches to save time in podvm building.
- Start with the kata-agent

#2018 (still in draft) has other components cached, but it's not working with the CI, so can be followed up later.


stevenhorsman:
Given that we have some challenges with the guest-components libary being included (which might be addressed by misc-settings in mkosi?) I think it might be best to merge caching for just the kata-agent as the slowest built component first and follow up with the other components in later PRs. As such I'll put this as ready for review now

mkulke (approved):

liudalibj (approved):
LGTM

PR #2038, opened 2024-09-12
podvm: Attempt to fix missing policy

stevenhorsman:
WIP: Can we even write to `/run/` during podvm build? Since #1998 the default-policy.rego has been symlinked to point to `/run/peerpod/policy.rego`, but if you aren't using init data then this file doesn't exist. In this we try and make that file point to our default policy file to cover this case.

Note: I don't understand how this works for packer images at the moment, so I'm clearly missing something.


mkulke:
> WIP: Can we even write to /run/ during podvm build?

i don't think so, /run should be tmpfs mount, no?

stevenhorsman:
> > WIP: Can we even write to /run/ during podvm build?
> 
> i don't think so, /run should be tmpfs mount, no?

Yeah, that's what I thought, but it sounds like the other mechanism should do this, so I'll not continue with this path

PR #2040, opened 2024-09-16
workflows: Increase libvirt e2e test timeout

stevenhorsman:
We've seen a few nightly and PR test runs fail due to e2e timeout. Looking into it our timeout for over a year has been fixed at 50m, but in that time we've added a lot of tests and do extra deployment stages, like trustee. Looking at a recently run around half of the 50m allowance was taking up just setting up the cluster before tests run. Whilst I think it would be good to try and refine/filter the tests we run and improve some of the slowness, in the short term increasing the timeout will help us avoid e2e failures and re-runs.


mkulke (approved):

bpradipt (approved):
/lgtm

PR #2042, opened 2024-09-16
ci: set initdata in remote-attestation test

mkulke:
Set initdata for remote attestation. allow kbs endpoint to be specified for manual test runs.

edit: tested on a [fork](https://github.com/mkulke/cloud-api-adaptor/actions/runs/10883561900/job/30197561320)

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/common.go
+++ b/src/cloud-api-adaptor/test/e2e/common.go
@@ -152,6 +152,20 @@ func WithCommand(command []string) PodOption {
 	}
 }
 
+type JobOption func(*batchv1.Job)
+
+func WithJobCommand(command []string) JobOption {
+	return func(j *batchv1.Job) {
+		j.Spec.Template.Spec.Containers[0].Command = command
+	}
+}
+
+func WithJobAnnotations(data map[string]string) JobOption {
+	return func(j *batchv1.Job) {
+		j.Spec.Template.ObjectMeta.Annotations = data
+	}
+}
+
 func WithEnvironmentalVariables(envVar []corev1.EnvVar) PodOption {
 	return func(p *corev1.Pod) {
 		p.Spec.Containers[0].Env = envVar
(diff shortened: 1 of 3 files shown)


src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -310,13 +324,13 @@ func NewSecret(namespace, name string, data map[string][]byte, secretType corev1
 }
 
 // NewJob returns a new job
-func NewJob(namespace, name string, backoffLimit int32, image string, command ...string) *batchv1.Job {
-	if len(command) == 0 {
-		command = []string{"/bin/sh", "-c", "echo 'scale=5; 4*a(1)' | bc -l"}
-	}
+// func NewJob(namespace, name string, backoffLimit int32, image string, command ...string) *batchv1.Job {
```
stevenhorsman, line 327: Do we need the comment of the old signature?

stevenhorsman (commented):
Looks okay apart from one comment

stevenhorsman (approved):
LGTM if the tests still pass :)

mkulke:
> LGTM if the tests still pass :)

the libvirt test suite? that's tricky, b/c at the moment those don't work either, right?

stevenhorsman:
> > LGTM if the tests still pass :)
> 
> the libvirt test suite? that's tricky, b/c at the moment those don't work either, right?

I'm hoping with the timeout increase they will be more stable, but I meant the azure tests, as I think they should be the only thing that significantly changes with this PR? I also can't run/see the results of them myself, which is why I mentioned the comment, but I'm happy to trust that you have them working :)

mkulke:
> > > LGTM if the tests still pass :)
> > 
> > 
> > the libvirt test suite? that's tricky, b/c at the moment those don't work either, right?
> 
> I'm hoping with the timeout increase they will be more stable, but I meant the azure tests, as I think they should be the only thing that significantly changes with this PR? I also can't run/see the results of them myself, which is why I mentioned the comment, but I'm happy to trust that you have them working :)

Ah yes, those should work. I can retrigger them in a fork to double-check though

bpradipt (approved):
/lgtm

PR #2043, opened 2024-09-18
Docker: podvm image: remove systemd tmpfiles setup delete

stevenhorsman:
Remove the systemd "systemd-tmpfiles-setup.service" -delete
and tidy up docs and code to remove missing features

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm
+++ b/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm
@@ -10,7 +10,6 @@ RUN echo "Installing Packages ..." \
       libseccomp pigz fuse-overlayfs \
       nfs-utils which systemd-pam \
       bash ca-certificates curl jq procps \
-    && find /lib/systemd/system/sysinit.target.wants/ -name "systemd-tmpfiles-setup.service" -delete \
     && rm -f /lib/systemd/system/multi-user.target.wants/* \
     && rm -f /etc/systemd/system/*.wants/* \
     && rm -f /lib/systemd/system/local-fs.target.wants/* \


mkulke (commented):
I feel like we should untangle both aspects. One is a bug with the docker provider, the other issue is a regression that we should be able to fix in mkosi one way or the other, possibly with variables in the mkosi or docker builds

stevenhorsman:
> I feel like we should untangle both aspects. One is a bug with the docker provider, the other issue is a regression that we should be able to fix in mkosi one way or the other, possibly with variables in the mkosi or docker builds

So are you proposing that I drop the second commit and ignore it completely, or move it to a different PR?

mkulke:
> > I feel like we should untangle both aspects. One is a bug with the docker provider, the other issue is a regression that we should be able to fix in mkosi one way or the other, possibly with variables in the mkosi or docker builds
> 
> So are you proposing that I drop the second commit and ignore it completely, or move it to a different PR?

I'd suggest to open an issue regarding the second issue. I suspect we want to keep the ability of specifying policies at build-time

mkulke (approved):

stevenhorsman:
> I'd suggest to open an issue regarding the second issue. I suspect we want to keep the ability of specifying policies at build-time

Done: https://github.com/confidential-containers/cloud-api-adaptor/issues/2045

bpradipt (approved):
/lgtm

bpradipt:
@stevenhorsman do we need the e2e libvirt test for this PR ?

stevenhorsman:
> @stevenhorsman do we need the e2e libvirt test for this PR ?

Good point - this only impacts the docker provider, so I think we can ignore that

bpradipt:
Merging this as e2 libvirt is not relevant for this PR.

PR #2049, opened 2024-09-23
.github: dependabot: Check github actions

stevenhorsman:
We have a few warnings in our workflows about actions that run on older node versions:
```
e2e / podvm_builder / Create pod vm builder image (ubuntu, ubuntu-latest)
The following actions use a deprecated Node.js version and will be forced to run on node20
```
We can enable dependabot to check for updates to these each week in order to stay up-to-date more easily


bpradipt (approved):
/lgtm

mkulke (approved):

PR #2056, opened 2024-09-25
azure: implicit nic creation + public ip support

mkulke:
The is a rework of the networking code. Currently we manage NICs explicitly, which requires a lot of code that is prone to race conditions and resource leaks. This PR changes the code to create NICs implicitly when creating a VM, the network resource will be managed as part of the VM lifecycle.

### Public IPs

There's also an opt-in toggle to attach (a implicitly-managed public ip) to the podvm. This option is mostly relevant for development purposes, there are obvious risks to expose a podvm's services to the internet in production environments. A given kubernetes deployment might not allow nodes (and hence the kata runtime) to connect to a podvm on the internet, further network configuration is required in those cases.

### PodVM internet access

In the current Pod VM deployment method, the VMs rely on a feature called [Default Outbound Access](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access#add-an-explicit-outbound-connectivity-method). This will equip a machine with a transparent public ip if no other outbound connectivity means are configured. This feature is [being retired](https://azure.microsoft.com/en-us/updates/default-outbound-access-for-vms-in-azure-will-be-retired-transition-to-a-new-method-of-internet-access/) due to security considerations. Today implicit NIC-management does not support this feature any more.

A PodVM might rely on internet connectivity e.g. for guest-pull from public registries like DockerHub (note: the pod traffic itself is namespaced and routed through the node, so a Pod itself will share the outbound connectivity of the other pods in a cluster)

Outbound access should be explicitly configured, the following would be an example for AKS, where create a dedicated peerpod subnet on the Worker Node's VNet, to which we attach a NAT gateway with a public ip:

```bash
export AKS_RG="$(az aks show -n my-cluster -g my-resorce-group --query nodeResourceGroup -o tsv)"
export AZURE_VNET_NAME="$(az network vnet list -g ${AKS_RG} --query '[].name' -o tsv)"
export AKS_CIDR="$(az network vnet show -n $AZURE_VNET_NAME -g $AKS_RG --query "subnets[?name == 'aks-subnet'].addressPrefix" -o tsv)"
# 10.224.0.0/16
export MASK="${AKS_CIDR#*/}"
# 16
PEERPOD_CIDR="$(sipcalc $AKS_CIDR -n 2 | grep ^Network | grep -v current | cut -d' ' -f2)/${MASK}"
# 10.225.0.0/16
az network public-ip create -g "$AKS_RG" -n peerpod
az network nat gateway create -g "$AKS_RG" -l "$AZURE_REGION" --public-ip-addresses peerpod -n peerpod
az network vnet subnet create -g "$AKS_RG" --vnet-name "$AZURE_VNET_NAME" --nat-gateway peerpod --address-prefixes "$PEERPOD_CIDR" -n peerpod
# SUBNET_ID is to be used in the peerpod configmap
export SUBNET_ID="$(az network vnet subnet show -g "$AKS_RG" --vnet-name "$AZURE_VNET_NAME" -n peerpod --query id -o tsv)"
```


src/cloud-providers/azure/provider.go
```diff
@@ -454,14 +433,8 @@ func (p *azureProvider) getVMParameters(instanceSize, diskName, cloudConfig stri
...
 			NetworkProfile: &armcompute.NetworkProfile{
-				NetworkInterfaces: []*armcompute.NetworkInterfaceReference{
-					{
-						ID: vmNIC.ID,
-						Properties: &armcompute.NetworkInterfaceReferenceProperties{
-							DeleteOption: to.Ptr(armcompute.DeleteOptionsDelete),
-						},
-					},
-				},
+				NetworkAPIVersion:              to.Ptr(armcompute.NetworkAPIVersionTwoThousandTwenty1101),
```
bpradipt, line 436: Is there any reference to the API version ? And do we need to keep a track of updates to the version?

mkulke: no, there shouldn't be a need to update it, it would use the latest API when unspecified. the network apis are generally not very dynamic, but it's a good idea to indicate which version has been used, so you'll get e.g. deprecation warnings if things change.

bpradipt (approved):
/lgtm

src/cloud-providers/azure/provider.go
```diff
@@ -181,42 +182,26 @@ func (p *azureProvider) CreateInstance(ctx context.Context, podName, sandboxID s
...
 	if err != nil {
 		return nil, err
 	}
 
 	logger.Printf("CreateInstance: name: %q", instanceName)
 
-	result, err := p.create(ctx, vmParameters)
+	vm, err := p.create(ctx, vmParameters)
 	if err != nil {
-		if err := p.deleteDisk(context.Background(), diskName); err != nil {
```
snir911, line 201: IIUC disk implicit deletion is already happening no? are these leftovers?

snir911 (approved):

Issue #2059, opened 2024-09-25
Extracting oras failing on s390x for podvm-builder due to ARCH hardcoded to amd64


ajaypvictor:
### Describe the bug

While building the `podvm-builder` image on s390x, due to `ARCH` being hardcoded on the Dockerfile, it is looking for amd64 version while untaring and failing with the following error:
```
0.962 tar (child): Error is not recoverable: exiting now
0.962 tar: Child returned status 2
0.962 tar: Error is not recoverable: exiting now
------
Dockerfile.podvm_builder.rhel:85
--------------------
  83 |
  84 |     ADD https://github.com/oras-project/oras/releases/download/v${ORAS_VERSION}/oras_${ORAS_VERSION}_linux_${ARCH}.tar.gz oras_${ORAS_VERSION}_linux_amd64.tar.gz
  85 | >>> RUN rm -rf /usr/local/bin/oras && tar -C /usr/local/bin -xzf oras_${ORAS_VERSION}_linux_${ARCH}.tar.gz && rm -f oras_${ORAS_VERSION}_linux_${ARCH}.tar.gz
  86 |
  87 |     WORKDIR /src
--------------------
ERROR: failed to solve: process "/bin/sh -c rm -rf /usr/local/bin/oras && tar -C /usr/local/bin -xzf oras_${ORAS_VERSION}_linux_${ARCH}.tar.gz && rm -f oras_${ORAS_VERSION}_linux_${ARCH}.tar.gz" did not complete successfully: exit code: 2
make: *** [Makefile:169: podvm-builder] Error 1
```

### How to reproduce

Try building the podvm-builder image on s390x.

### CoCo version information

CAA main

### What TEE are you seeing the problem on

None

PR #2060, opened 2024-09-25
podvm: Update hardcoded value for oras on podvm-builder

ajaypvictor:
Updating the ARCH value on podvm-builder, so that it can pick up the respective architecture while it's being built.

Fixes https://github.com/confidential-containers/cloud-api-adaptor/issues/2059


stevenhorsman:
Given that we don't really want to dig into native s390x support for the ubuntu builder (as we build the s390x binaries image on top of the amd64 builder) and mkosi will replace it, so I've suggested to Ajay to just update the rhel builder for now (which IBM are building natively due to subscription requirements), so I don't think we need the libvirt e2e tests passing now as we don't build rhel anywhere upstream.

stevenhorsman (approved):
LGTM. Thanks @ajaypvictor!

bpradipt (approved):
/lgtm

PR #2062, opened 2024-09-26
Webhook changes

bpradipt:
Fix handling of fractional cpu units during mutation
Reliably install cert-manager for webhook if needed.
~~Deploy webhook by default during install~~ : Will create a separate PR for this

### DIFF
--- a/src/cloud-api-adaptor/docker/prereqs.sh
+++ b/src/cloud-api-adaptor/docker/prereqs.sh
@@ -86,14 +86,14 @@ install_os_packages() {
 # the function will download the packages, extract them and install them in /usr/local/bin
 # Following are the packages that are installed:
 # yq=https://github.com/mikefarah/yq/releases/download/v4.44.2/yq_linux_amd64
-# kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.27.11/bin/linux/amd64/kubectl
+# kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.29.4/bin/linux/amd64/kubectl
 # kind=https://kind.sigs.k8s.io/dl/v0.23.0/kind-linux-amd64
 
 install_binary_packages() {
     # Define the required binary packages
     REQUIRED_BINARY_PACKAGES=(
         "yq=https://github.com/mikefarah/yq/releases/download/v4.44.2/yq_linux_amd64"
-        "kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.27.11/bin/linux/amd64/kubectl"
+        "kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.29.4/bin/linux/amd64/kubectl"
         "kind=https://kind.sigs.k8s.io/dl/v0.23.0/kind-linux-amd64"
     )
 
--- a/src/webhook/Makefile
+++ b/src/webhook/Makefile
@@ -1,3 +1,4 @@
+ARCH        ?= $(subst x86_64,amd64,$(shell uname -m))
 
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/confidential-containers/peer-pods-webhook:latest
@@ -109,7 +110,6 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 kind-cluster: ## create a kind cluster for testing 
 	kind create cluster --image "kindest/node:v$(ENVTEST_K8S_VERSION)" \
 		--wait 120s --name "$(KIND_CLUSTER_NAME)"
-	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.1/cert-manager.yaml
 
 .PHONY: kind-delete
 kind-delete: ## delete the kind test cluster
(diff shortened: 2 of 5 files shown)


bpradipt:
This should fix the webhook CI errors due to cert-manager issues.

src/cloud-api-adaptor/docker/prereqs.sh
```diff
@@ -86,7 +86,7 @@ install_os_packages() {
 # the function will download the packages, extract them and install them in /usr/local/bin
 # Following are the packages that are installed:
 # yq=https://github.com/mikefarah/yq/releases/download/v4.44.2/yq_linux_amd64
-# kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.27.11/bin/linux/amd64/kubectl
+# kubectl=https://storage.googleapis.com/kubernetes-release/release/v1.29.4/bin/linux/amd64/kubectl
```
wainersm, line 89: Hi @bpradipt ,

am I missing something or this just changed the comment line? Probably you want to update the version listed in `REQUIRED_BINARY_PACKAGES`, right?

src/webhook/pkg/mutating_webhook/mutating-webhook_test.go
```diff
@@ -566,3 +566,226 @@ func TestMutatePod_NoChangeForDifferentRuntimeClass(t *testing.T) {
...
+		t.Errorf("Expected Memory annotation to be 4096, got %s", mutatedPod.Annotations[PEERPODS_MEMORY_ANNOTATION])
+	}
+
+	if _, exists := mutatedPod.Annotations[PEERPODS_GPU_ANNOTATION]; exists {
+		t.Errorf("Expected no GPU annotation, got %s", mutatedPod.Annotations[PEERPODS_GPU_ANNOTATION])
+	}
+
+	// Check resource requirements for all containers are cleared
+	// Except for the first container which has the peer-pod resource added
+	for idx, container := range mutatedPod.Spec.Containers {
```
wainersm, line 690: Likewise

src/webhook/pkg/mutating_webhook/remove-resourcespec.go
```diff
@@ -80,11 +81,12 @@ func adjustResourceSpec(pod *corev1.Pod) *corev1.Pod {
 
 	// Add cpu annotation
 	if !cpuRequest.IsZero() && cpuLimit.Cmp(cpuRequest) >= 0 {
-		logger.Printf("Adding CPU annotation based on cpuLimit: %s", cpuLimit.String())
-		annotations[PEERPODS_CPU_ANNOTATION] = cpuLimit.String()
+		logger.Printf("Adding CPU annotation based on cpuLimit (scaled value): %d", cpuLimit.Value())
```
wainersm, line 84: The documentation of https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity.Value got me confused as it says `Value returns the unscaled value of q rounded up to the nearest integer away from 0`. I'm not expect here so just bringing to your attention to double-check we are using the right function.

src/webhook/Makefile
```diff
@@ -122,6 +122,25 @@ kind-load: ## load the image in the local kind cluster
...
+	chmod +x cmctl
+	# Deploy cert-manager
+	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.15.3/cert-manager.yaml
+	# Wait for service to be up
+	kubectl wait --timeout=30s -n cert-manager endpoints/cert-manager --for=jsonpath='{.subsets[0].addresses[0].ip}'
+	kubectl wait --timeout=30s -n cert-manager endpoints/cert-manager-webhook --for=jsonpath='{.subsets[0].addresses[0].ip}'
+	# Wait for few seconds for the cert-manager API to be ready
+	# otherwise you'll hit the error "x509: certificate signed by unknown authority"
+	# Best is to use cmctl - https://cert-manager.io/docs/installation/kubectl/#2-optional-wait-for-cert-manager-webhook-to-be-ready
+	./cmctl check api --wait=2m
```
wainersm, line 137: I tested this `deploy-cert-manager` target; worked pretty well. The API was up and running very quickly so maybe we can even reduce the wait time to a few seconds so that, on fail, we don't need to wait 2m.

src/webhook/pkg/mutating_webhook/mutating-webhook_test.go
```diff
@@ -566,3 +566,226 @@ func TestMutatePod_NoChangeForDifferentRuntimeClass(t *testing.T) {
...
```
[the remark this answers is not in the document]
bpradipt: I just kept the loop same for all the tests. Since only the first container is expected to have the resource req/limit after mutation it will continue and exit the loop. Either we can check for number of containers as you mentioned or just use the one like in the current code.

src/webhook/pkg/mutating_webhook/remove-resourcespec.go
bpradipt: Basically the my comment may be misleading. I will remove the scale word. The key is rounding up to the nearest integer.

src/webhook/Makefile
bpradipt: I think wait is the max wait. It'll will exit the moment the api is up.

src/cloud-api-adaptor/docker/prereqs.sh
bpradipt: yes ;) .. Nice catch

src/webhook/pkg/mutating_webhook/remove-resourcespec.go
wainersm: ah ok, I was suspicious that we wanted rounding up instead :)

src/webhook/Makefile
bpradipt: You can try re-running cmctl check api again on a ready system and it'll return immediately. 
Let's leave it at 2m as anyway in CI and otherwise things can be flaky due to network issues and it can take time. 
If you still prefer a lower time, let me know a value and I'll update the PR.

wainersm: no, that's ok. let's keep as is

wainersm (approved):
Thanks @bpradipt !

src/webhook/Makefile
bpradipt: Our messages crossed ;) .. Leaving the time as 2m for now

src/webhook/pkg/mutating_webhook/remove-resourcespec.go
bpradipt: kata and caa cannot handle vcpus values as 100m or 0.5, so we need it to be rounded to the neared integer :-)

PR #2064, opened 2024-09-26
azure: switch nightly podvm to mkosi-based image

mkulke:
This change will switch the nightly azure images to an mkosi-based image. The image will be used in the e2e tests and is hence becoming the reference image for azure, deprecating the packer image.

Successful test run with mkosi image [here](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/11070061926)


stevenhorsman (approved):
As long as the azure image builds work for you this LGTM 😄

mkulke:
> As long as the azure image builds work for you this LGTM 😄

if we would do this right, we would probably build 2 images alongside and then toggle the e2e tests, but I don't think we need we need this ceremony.

stevenhorsman:
> > As long as the azure image builds work for you this LGTM 😄
> 
> if we would do this right, we would probably build 2 images alongside and then toggle the e2e tests, but I don't think we need we need this ceremony.

Good point - the parallel is my plan for the libvirt e2e switch to mkosi, but the Azure CI doesn't block PRs in the same way that libvirt does, which is the main motivation for phasing there, so I'm happy for this to go directly.

PR #2067, opened 2024-09-27
APF: Rename KataAgentNamespace

davidhadas:
KataAgentNamespace parameter is controlling the namespace in which the pod runs (not the namespace in which the Kata Agent runs)

To make the code readable, it is renamed to PodNamespace


stevenhorsman (approved):
LGTM. Thanks

PR #2065, opened 2024-09-26
SecureComms: Add support daemonConfig

davidhadas:
See:
 - https://github.com/confidential-containers/cloud-api-adaptor/issues/2096

Support configuring the APF Secure Comms from the CAA side including:
 - WN public Key
 - PP private key
 - Activating Secure Comms
 - inbouns and outbounds of th PP

This is useful for activating Secure Comms from the CAA and without Trustee. It can be used for Testing without producing dedicated podvms which activate Secure Comms and set Inbounds/Outbounds by default. It can also be used for non-Coco peerpods.

PR #2072, opened 2024-09-30
SecureComms: Add Support for activating using InitData

davidhadas:
Add `apj.json` to InitData
`apf.josn` includes an `sc` key used to activate secure comms (if not already activated using an agent-protocol-forwarder.service flag)


davidhadas:
cc: @bpradipt

src/cloud-api-adaptor/cmd/agent-protocol-forwarder/main.go
```diff
@@ -52,6 +54,8 @@ func load(path string, obj interface{}) error {
 		return fmt.Errorf("failed to decode a Agent Protocol Forwarder config file file: %s: %w", path, err)
 	}
 
+	logger.Printf("succesful loading config from %s\n", path)
```
stevenhorsman, line 57: Is this supposed to say "successfully loaded config..."?

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -114,7 +114,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		flags.BoolVar(&secureComms, "secure-comms", false, "Use SSH to secure communication between cluster and peer pods")
 		flags.StringVar(&secureCommsInbounds, "secure-comms-inbounds", "", "Inbound tags for secure communication tunnels")
 		flags.StringVar(&secureCommsOutbounds, "secure-comms-outbounds", "", "Outbound tags for secure communication tunnels")
-		flags.StringVar(&secureCommsKbsAddr, "secure-comms-kbs", "kbs-service.kbs-operator-system:8080", "Address of a KBS Service for Secure-Comms")
+		flags.StringVar(&secureCommsKbsAddr, "secure-comms-kbs", "kbs-service.trustee-operator-system:8080", "Address of a KBS Service for Secure-Comms")
```
stevenhorsman, line 117: The trustee operator namespace changes should be in a separate commit with an explanation that it's triggered by the change in the trustee-operator project

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
 
 Make sure to uncomment the secret generation as recommended for both public and private key (`kbs-auth-public-key` and `kbs-client` secrets). 
 
-Copy the kbs-client secret from the `kbs-operator-system` namespace to the `confidential-containers-system` ns. This can be done using:
+```sh
+kubectl get secrets -n trustee-operator-system
+NAME                  TYPE     DATA   AGE
+kbs-auth-public-key   Opaque   1      28h
+kbs-client            Opaque   1      28h
+```
```
stevenhorsman, line 45: What's the reason for this command being added?

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
...
 
 ### Activate CAA Secure-Comms feature
-Use `kubectl edit cm peer-pods-cm -n confidential-containers-system` to add to the `peer-pods-cm` config map at the `confidential-containers-system` namespace:
+Activate Secure-Comms of CAA by changing the `SECURE_COMMS` parameter of the `peer-pods-cm` configMap in the `confidential-containers-system` namespace to `"true"`.  
+
+```sh
+kubectl -n confidential-containers-system  get cm peer-pods-cm  -o yaml | sed "s/SECURE_COMMS: \"false\"/SECURE_COMMS: \"true\"/"|kubectl apply -f -
+```
+
+Set InitData to point KBC services to IP address 127.0.0.1
```
stevenhorsman, line 91: Should this have a heading like `Build a podvm that enforces Secure-Comms (Optional)` section does as it's an alternative to it?

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
...
+
+For example:
+```sh
+ExecStart=/usr/local/bin/agent-protocol-forwarder -kata-agent-namespace /run/netns/podns -secure-comms -secure-comms-inbounds KUBERNETES_PHASE:mytunnel:6666 -kata-agent-socket /run/kata-containers/agent.sock $TLS_OPTIONS $OPTIONS
+```
 
 Once you changed `podvm/files/etc/systemd/system/agent-protocol-forwarder.service`, you will need to [rebuild the podvm](./../podvm/README.md).
 
 
 ### Activate CAA Secure-Comms feature
```
stevenhorsman, line 84: I'm not sure if this heading is helpful as it's needed in both the podvm and initdata approach and it's slightly different and doesn't have a heading there?

src/cloud-api-adaptor/docs/SecureComms.md
stevenhorsman: Some expansion of the explanation of what this is doing and why would be nice.

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
...
 ```
 
-You may also include additional Inbounds and Outbounds configurations to the Adaptor using the `SECURE_COMMS_INBOUNDS` and `SECURE_COMMS_OUTBOUNDS` config points. See more details regarding Inbounds and Outbounds below.
-
 You may also set the KBS address using the `SECURE_COMMS_KBS_ADDR` config point.
 
 
-### Adding named tunnels to the SSH channel
+
+## Adding named tunnels to the SSH channel
```
stevenhorsman, line 139: What's the reason for the move of the newline?

stevenhorsman (commented):
Some initial comments. It would be good to fix the typos in the commit message too.

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
davidhadas: Yes - #2073 should be merged first

src/cloud-api-adaptor/docs/SecureComms.md
davidhadas: This is a slight improvement to the SecureComms doc which shows the correct result after following the instructions trustee operator and following the recommendation: "Make sure to uncomment the secret generation as recommended for both public and private key (`kbs-auth-public-key` and `kbs-client` secrets). "

We can add it to #2073 if it will make things clearer

davidhadas: I prefer to leave this extra documentation detail  as is, although it also appears in #2065 - unless someone finds this very disturbing.

src/cloud-api-adaptor/docs/SecureComms.md
davidhadas: I will be adapting the documentation based on this comment. Please reconsider the next version.

src/cloud-api-adaptor/docs/SecureComms.md
davidhadas: changed to `### Activate the Secure-Comms feature`
This will be our main section to show the config changes needed. 

The other option of changing the podvm will move to the end. We will use it as an extra step for cases where one doe snot trust the InitData (e.g. InitData is not measured).

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
...
+[token_configs]
+[token_configs.coco_as]
+url = 'http://127.0.0.1:8080'
+
+[token_configs.kbs]
+url = 'http://127.0.0.1:8080'
+'''
+"apf.json" = '''
+{
+    sc: true
```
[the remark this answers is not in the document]
davidhadas: changed to `secure-comms` to align with other config parameters enabling Secure-Comms

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -37,49 +37,106 @@ Deploy Trustee-Operator by following instructions at [trustee Operator Getting S
...
+}
```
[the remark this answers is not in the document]
davidhadas: fixed

mkulke (commented):
I understand that kata is proceeding with an init-data approach that is passed from the runtime to the agent via a `SetInitData()` RPC call for the upcoming CoCo release.

This looks like a CAA-specific extension to InitData. If the project adopts kata's init-data approach, would this still work?

davidhadas:
> I understand that kata is proceeding with an init-data approach that is passed from the runtime to the agent via a `SetInitData()` RPC call for the upcoming CoCo release.
> 
> This looks like a CAA-specific extension to InitData. If the project adopts kata's init-data approach, would this still work?

This PR extends https://github.com/confidential-containers/cloud-api-adaptor/issues/1895 which introduced a CAA-specific extension to InitData. Here we added configuration of the APF to the existing configuration of AA and CDH. 

Regardless, it seems that the `SetInitData()` RPC call cannot be used under peer-pods (even without this PR) - see https://github.com/kata-containers/kata-containers/pull/10163#issuecomment-2424618637

stevenhorsman:
> Can you elaborate on why a SetInitData() RPC wouldn't work for CAA? I think I POC'd that approach when the init-data design was being discussed to make sure it would work and I didn't run into issues.

In your prototype did you keep the guest-components managed by systemd, or the kata-agent? It seems to be that in if the kata-agent isn't managing them and they are started before the kata-agent then relying on config from kata-agent endpoint doesn't really make sense as we then have this weird undefined start until the kata is up and running and receiving the setInitData request.

In fairness I'm still not sure why bare-metal doesn't adopt the approach of systemd managing the processes like the peer pod does anyway to avoid this either, so I'm obviously missing something.

davidhadas:
@mkulke, 
 
> Can you elaborate on why a SetInitData() RPC wouldn't work for CAA?

To add to what @stevenhorsman indicated, we use use attestation when connecting CAA to APF in SecureComms. 
To do attestation, we need the measurements of the configuration. 
To get the measurements, we need InitData.
We connect runtime to kata-agent only after the attestation is done and keys delivered - keys we are using to complete the establishment of the CAA to APF secure communication channel.

mkulke:
> In your prototype did you keep the guest-components managed by systemd, or the kata-agent? It seems to be that in if the kata-agent isn't managing them and they are started before the kata-agent then relying on config from kata-agent endpoint doesn't really make sense as we then have this weird undefined start until the kata is up and running and receiving the setInitData request.
> 
> In fairness I'm still not sure why bare-metal doesn't adopt the approach of systemd managing the processes like the peer pod does anyway to avoid this either, so I'm obviously missing something.

I would have to dig a bit to find the details, but I think in the PoC code kata-agent was talking via ttRPC to attestation-agent to update the configuration, so it was indeed half-initialized. however, that endpoint is about to be removed as I understand, though. somehow AA will now have to perform a binding to the TEE itself (e.g. compare host-data, extend an init-data PCR).

mkulke:
> We connect runtime to kata-agent only after the attestation is done and keys delivered - keys we are using to complete the establishment of the CAA to APF secure communication channel.

I understand this applies if you use the secure communications feature, but in a default CAA installation there is no attestation ceremony before runtime <=> agent communication is established, afaik?

davidhadas:
@mkulke, @stevenhorsman, @bpradipt 

This PR follows on the footsteps of the initData support that already exists in peer pods to-date. 
While we have a discussion if to change this mechanism some time in the future to follow what will be done by Kata RPC etc., is there any impediment to continue supporting the existing mechanism that we have till now? 
Is there any impediment to adjust it to also deliver the secureComms flag?

I would like to suggest that whatever implementation we will have for initData, we will still need to add this flag to enable/disable secureComms using measured initData (or whatever we choose to call it), so adding it now seems like a progress in the right direction. 

The two benefits of adding it now and not waiting is that (1) it will allow e2e testing of the secureComms with a complete attestation, and key retrieval phase, (2) it will allow using/testing secureComms by users/downstream without having a dedicated podvm image as needed today. 

I would prefer to move forward and add the e2e tests regardless of how the future measured initData will be implemented in peer-pods and by following the same mechanism that works today. Does this make sense?

mkulke:
> @mkulke, @stevenhorsman, @bpradipt
> 
> This PR follows on the footsteps of the initData support that already exists in peer pods to-date. While we have a discussion if to change this mechanism some time in the future to follow what will be done by Kata RPC etc., is there any impediment to continue supporting the existing mechanism that we have till now? Is there any impediment to adjust it to also deliver the secureComms flag?

afaiu the init-data feature will be in next release of coco and unless there will be a change of plans the agent will receive init-data via RPC then. That means whatever we are doing in terms of initdata on the daemonset and podvm is being redundant/conflicting (since a CAA release is synced with kata/guest-components/operator).

To a coco user this change should (mostly) be transparent, but internally we would have to remove the related code and AA + CDH would only run after the agent has received init-data.

in this context I would not recommend building anything on top of the initdata impl in peerpods, currently.

huoqifeng:
> @huoqifeng, can you please review this PR as it extends your work in #1895

Sorry my late, I'll have a look...

src/cloud-api-adaptor/docs/SecureComms.md
```diff
@@ -58,28 +58,58 @@ kubectl -n trustee-operator-system exec deployment/trustee-deployment --containe
...
+version = "0.1.0"
 
-Once you changed `podvm/files/etc/systemd/system/agent-protocol-forwarder.service`, you will need to [rebuild the podvm](./../podvm/README.md).
+[data]
+"aa.toml" = '''
+[token_configs]
+[token_configs.coco_as]
+url = 'http://127.0.0.1:8080'
 
+[token_configs.kbs]
```
huoqifeng, line 83: As KBS endpoint is different when enabling SecureComms, shall we also add link from ./initdata.md to this file?

davidhadas:
@stevenhorsman, @bpradipt 
This PR was open to allow us to get to a point where we can use initData to deliver a flag activating SecureComms. 
Such a flag will allow securely reuse the same image for both running with and without SecureComms activated. 
We still have no such a secure mechanism in place. 
Is there a plan to further persue this direction or are we dropping it and closing the respective https://github.com/confidential-containers/cloud-api-adaptor/issues/2094, https://github.com/confidential-containers/cloud-api-adaptor/issues/2095 ?

PR #2074, opened 2024-09-30
podvm: retrieve guest-components via ORAS

mkulke:
~Note: draft until [GC 731](https://github.com/confidential-containers/guest-components/pull/731) and #2064 have been merged~

In this change the artifacts are being retrieved from guest-component's ORAS now. Hence the rust build infrastructure can be removed with this change.

There is an option to verify the provenance of the guest component artifacts that we download as part of the build. It is opt-in, you have to set `VERIFY_PROVENANCE=yes` when building a podvm. There are respective build flags on the `src/cloud-api-adaptor/podvm/Dockerfile.podvm_binaries.fedora` and the `src/cloud-api-adaptor/podvm-mkosi/Makefile`. Currently only the azure-podvm-image-build ci workflow has the provenance checks enabled.

There are some notable changes:
- guest-component exposed the TEE_PLATFORM param on its top level build script, which we use to pull the correct artifact. Since we don't build attestation-agent directly anymore the ATTESTER param has been removed from the projects build scripts
- in versions.yaml kata and guest-components have been moved from the "git" section to the "oci" section, however since the tag is dynamic, we also provide a "reference" field in those entries.


stevenhorsman (commented):
I'll re-review once the guest-component caching is merged and working, but on initial review this looks great and supercedes #2033 nicely. Thanks!

.github/workflows/e2e_libvirt.yaml
```diff
@@ -89,13 +88,6 @@ jobs:
           # For debugging
           cat libvirt.properties
 
-      - name: Install rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ env.RUST_VERSION }}
-          override: true
-          profile: minimal
-
```
stevenhorsman, line 98: Hey Magnus, I've just realised that we still need the rust toolchain to build the kbs-client, so I don't think we can remove it from the workflow yet, unless there is a cached binary there we can use (which was discussed, but I don't think has been done yet)? I'm not sure our `.tools.rust` is required though as it looks like trustee uses a different version, but it's hard-coded in the workflow, so I'm not sure it's easy to get at.

mkulke: oh, you're right. there is a kbs-client being published, though. can we use this?

https://github.com/confidential-containers/trustee/pkgs/container/staged-images%2Fkbs-client

mkulke: looks good:

```
oras pull ghcr.io/confidential-containers/staged-images/kbs-client:sample_only-x86_64-linux-gnu-7df333dba5550571ec2999018ae33e46068ba586
Downloading eebf628a6f2b kbs-client
Downloaded  eebf628a6f2b kbs-client
Pulled [registry] ghcr.io/confidential-containers/staged-images/kbs-client:sample_only-x86_64-linux-gnu-7df333dba5550571ec2999018ae33e46068ba586
Digest: sha256:b4f090dc0af970badc17ed73c6871f3f3b2f2345726745b39b6b0a4d37cb9317
magnuskulke@DESKTOP-2KAPMGM:~/tmp/oras$ ls
kbs-client
magnuskulke@DESKTOP-2KAPMGM:~/tmp/oras$ chmod +x kbs-client
magnuskulke@DESKTOP-2KAPMGM:~/tmp/oras$ ./kbs-client
A command line client tool for KBS APIs.

Usage: kbs-client [OPTIONS] <COMMAND>

Commands:
  config        Set and Config KBS
  get-resource  Get confidential resource
  attest        Attestation and get attestation results token
  help          Print this message or the help of the given subcommand(s)

Options:
      --url <URL>              The KBS server root URL [default: http://127.0.0.1:8080]
      --cert-file <CERT_FILE>  The KBS HTTPS server custom root certificate file path (PEM format)
  -h, --help                   Print help
  -V, --version                Print version
```

stevenhorsman: Nice - it has been done, so I think we switch to that approach then too as part of this. Thanks!

mkulke: I pushed a change to retrieve the kbs-client from oci instead, this means we can also remove the git.kbs entry and should be able to fully de-oxidize the repo.

stevenhorsman: Sorry, I just realised that we also `kustomize` and deploy the KBS from that repo, so I don't think we can get rid of all this code, just the kbs client bit.

mkulke: yup, thx. readded the kbs entry in versions and the cloning step.

looking at versions.yml I'd argue that the split between `KBS_REPO#KBS_VERSION` (kustomize code) and `KBS_IMAGE#KBS_IMAGE_TAG` is probably a bit too contingent. 

there are implicit assumptions that kbs' kustomize scripts w/ version X work with kbs-image-tag version Y and kbs' kustomize "API" doesn't change. there are no guarantees for that. I'd suggest dropping KBS_IMAGE & KBS_IMAGE_TAG including the logic that tweaks those values in the test-provisioner code. CAA should deploy with what is the default in the kbs kustomize manifests (which should be the same revision as KBS_VERSION). In the niche case that someone wants to change that for an experimental build, they would:
- clone the trustee repo
- modify the kustomize manifests
- specify their clone#branch in KBS_REPO#KBS_VERSION

that's not in scope of this PR and probably subjective, but I recommend getting rid of excessive knobs to make the tests more reliable.

stevenhorsman: > looking at versions.yml I'd argue that the split between KBS_REPO#KBS_VERSION (kustomize code) and KBS_IMAGE#KBS_IMAGE_TAG is probably a bit too contingent.

Yeah, I agree and was wondering about this when I realised we deployed from source. It feels like we could just get rid of the test sections like:
```
        log.Infof("Updating kbs image with %q", props["KBS_IMAGE"])
	if err = lio.overlay.SetKustomizeImage("kbs-container-image", "newName", props["KBS_IMAGE"]); err != nil {
		return err
	}

	log.Infof("Updating kbs image tag with %q", props["KBS_IMAGE_TAG"])
	if err = lio.overlay.SetKustomizeImage("kbs-container-image", "newTag", props["KBS_IMAGE_TAG"]); err != nil {
		return err
	}
```
(and all the other related code in the test provisioner) as you suggest and we'd be back to a more stable base.

wainersm:
The required PRs were merged but oras images aren't published yet (https://github.com/confidential-containers/guest-components/actions/workflows/publish-artifacts.yml) due a bug on setup-oras action (https://github.com/oras-project/setup-oras/pull/57); so I could not test this yet.

src/cloud-api-adaptor/Makefile
```diff
@@ -7,18 +7,18 @@ include Makefile.defaults
...
-PACKAGES    := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
-SOURCES     := $(shell find $(SOURCEDIRS) -name '*.go' -print)
-ATTESTER    ?= none
+GOOPTIONS    ?= GOOS=linux GOARCH=$(ARCH) CGO_ENABLED=0
+GOFLAGS      ?=
+BINARIES     := cloud-api-adaptor agent-protocol-forwarder process-user-data
+SOURCEDIRS   := ./cmd ./pkg
+PACKAGES     := $(shell go list $(addsuffix /...,$(SOURCEDIRS)))
+SOURCES      := $(shell find $(SOURCEDIRS) -name '*.go' -print)
+TEE_PLATFORM ?= none
```
wainersm, line 21: I don't think it needs to define `TEE_PLATFORM` here because it will be inherited from Makefile.defaults

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -1,10 +1,10 @@
 include ../Makefile.defaults
 
-ATTESTER  ?= none
-ARCH      ?= $(subst x86_64,amd64,$(shell uname -m))
-BUILDER   = fedora-binaries-builder-$(ARCH)
-SE_BOOT   ?= false
-IS_DEBIAN := $(shell if grep -q 'ID_LIKE=debian' /etc/os-release; then echo "true"; else echo "false"; fi)
+TEE_PLATFORM ?= none
```
wainersm, line 3: also inherited from Makefile.defaults

.github/workflows/e2e_libvirt.yaml
```diff
@@ -27,6 +27,8 @@ on:
 env:
   CLOUD_PROVIDER: libvirt
   DEBIAN_FRONTEND: noninteractive
+  KBS_REGISTRY: ghcr.io/confidential-containers/staged-images/kbs
+  KBS_CLIENT_REGISTRY: ghcr.io/confidential-containers/staged-images/kbs-client
```
stevenhorsman: What are these envs being used for in this commit? Should they be part of a later one?

src/cloud-api-adaptor/podvm/Makefile.inc
```diff
@@ -36,6 +36,20 @@ AGENT_POLICY ?= yes
...
+GUEST_COMPONENTS_REPO := confidential-containers/guest-components
+
+# probe version if the ref does not appear to be a git commit digest
+ifeq ($(shell echo $(GUEST_COMPONENTS_REF) | grep -E "[0-9a-f]{40}"),)
+	GUEST_COMPONENTS_COMMIT := $(shell curl \
+		--silent \
+		--retry 3 \
+		https://api.github.com/repos/$(GUEST_COMPONENTS_REPO)/commits/v$(GUEST_COMPONENTS_REF) \
+		| jq -e -r .sha)
+else
```
stevenhorsman, line 50: Does this effectively mean that we can't use a guest-component build from a fork, unless we use the sha directly? I'm not sure that's a problematic restriction, just wanting to check my understanding.

src/cloud-api-adaptor/podvm/Dockerfile.podvm_binaries.fedora
```diff
@@ -25,7 +25,7 @@ ARG PAUSE_REPO
 ARG PAUSE_VERSION
 ARG PAUSE_BIN
 ARG IMAGE_NAME
-ARG GH_TOKEN
```
stevenhorsman: nit: I guess we could remove the `GH_TOKEN` refs from commit 1 as we just remove them here?

stevenhorsman (approved):
This look good to me, just a couple of minor questions/comments

.github/workflows/e2e_libvirt.yaml
stevenhorsman: nit: I saw in the next commit they were deleted, so they could be cleaned up in commit 1

src/cloud-api-adaptor/podvm/Makefile.inc
mkulke: well, we have this chain of indirections from git reference => git digest => oci tag => oci digest since want to specify a human-friendly reference in versions.yaml.  

This chain needs to also work if you want to consume something from a fork, so you have to push something with the `resolved_to_sha($reference) + "_" + $arch` to the registry.

I think if this turns out to be a nuisance, there can be an "oci.digest" param in version yaml that will short circuit everything.

.github/workflows/e2e_libvirt.yaml
```diff
@@ -89,12 +89,9 @@ jobs:
           # For debugging
           cat libvirt.properties
 
-      - name: Install rust toolchain
-        uses: actions-rs/toolchain@v1
+      - uses: oras-project/setup-oras@v1
         with:
-          toolchain: ${{ env.RUST_VERSION }}
-          override: true
-          profile: minimal
```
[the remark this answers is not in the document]
stevenhorsman: I figured it was just `pull_target_request` meaning we aren't testing the workflow change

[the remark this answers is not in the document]
mkulke: indeed, good catch.

stevenhorsman:
@mkulke - in case you hadn't spotted it, this needs a rebase to pick up the KBS version change from #2099 and resolve the conflict. Thanks and let's hope we can get this merged today 🤞

bpradipt (approved):
/lgtm

Issue #2076, opened 2024-10-02
CI: Remove KBS_IMAGE option


mkulke:
### Describe the bug

in version.yaml we can specify both `KBS_REPO#KBS_VERSION` (kustomize code) and `KBS_IMAGE#KBS_IMAGE_TAG`.

there are implicit assumptions that kbs' kustomize scripts w/ version X work with kbs-image-tag version Y and kbs' kustomize "API" doesn't change. there are no guarantees for that. I'd suggest dropping KBS_IMAGE & KBS_IMAGE_TAG including the logic that tweaks those values in the test-provisioner code. CAA should deploy with what is the default in the kbs kustomize manifests (which should be the same revision as KBS_VERSION). In the niche case that someone wants to change that for an experimental build, they would:
- clone the trustee repo
- modify the kustomize manifests
- specify their clone#branch in KBS_REPO#KBS_VERSION

the reasoning is that reducing the amount of knobs would make the tests more deterministic and reliable.

### How to reproduce

n/a

### CoCo version information

n/a

### What TEE are you seeing the problem on

None

PR #2077, opened 2024-10-02
kbs: Remove oci.kbs version and references

stevenhorsman:
Our KBS logic currently checks out the git.kbs version of code, but then edits the deployment config to override the image based on the oci.kbs version.

This means that there is an assumption that the oci image of the KBS is compatible with the kustomization, which might not not always be the case and in the "always safe" case where the image is built from that exact version of code (e.g. releases), just means that we have to specify it in two places, so there isn't much advantage to this approach.

This PR removes the oci.kbs image and versioning to avoid this potential incompatibility and simplify trustee version updates.

Fixes: #2076


mkulke (approved):
🕳️🏃‍♂️

wainersm:
Hi @stevenhorsman !

I just recently I tried e2e tests with KBS and I ended up with those variables copied in many places and exported, and I didn't even know if it was all needed. So this change simplifies a lot!

The downside is that we cannot point to an arbitrary SHA-1 of trustee's repo. How likely are we of hitting that case?

There is the case of trustee releasing an outdated kustomization file, for example, https://github.com/confidential-containers/trustee/blob/v0.10.0/kbs/config/kubernetes/base/kustomization.yaml#L8 . Well, in this case we would probably ask for another release....the problem is maybe the time we take to realize the file is not synchronized.

Also perhaps drop the [pinning mechanism on kata side](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/confidential_kbs.sh#L314) too, so that if we test with the wrong image here then we it test wrong there too (increasing the odds of finding the problem?).

stevenhorsman:
> The downside is that we cannot point to an arbitrary SHA-1 of trustee's repo. How likely are we of hitting that case?

I think we should be able to point to a sha with the git.kbs.reference like we had in https://github.com/confidential-containers/cloud-api-adaptor/blob/899b8057336fcd7e2ac4491aaba22cdbe2943146/src/cloud-api-adaptor/versions.yaml#L49, or have I misunderstood your point?

> There is the case of trustee releasing an outdated kustomization file, for example, https://github.com/confidential-containers/trustee/blob/v0.10.0/kbs/config/kubernetes/base/kustomization.yaml#L8 . Well, in this case we would probably ask for another release....the problem is maybe the time we take to realize the file is not synchronized.

Yeah, that is a really good point, but it looks like that not being bumped in the v0.10.0 was the reason for the v0.10.1 release: https://github.com/confidential-containers/trustee/releases/tag/v0.10.1, so hopefully it's already included in their release process

> Also perhaps drop the [pinning mechanism on kata side](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/confidential_kbs.sh#L314) too, so that if we test with the wrong image here then we it test wrong there too (increasing the odds of finding the problem?).

Yeah - I think that makes sense. I'll create an issue for this.

stevenhorsman:
I've created https://github.com/kata-containers/kata-containers/issues/10379 for the kata side of this and as part of that checked with Ding that it's a fair expectation for them to keep their images pinned and "up-to-date"

mkulke:
> The downside is that we cannot point to an arbitrary SHA-1 of trustee's repo. How likely are we of hitting that case?

I think that's not unlikely, but then it would be handled like the coco operator dependency: you cannot pick a given kbs revision directly, you'll get whatever is the default in the kbs kustomize manifest at git.trustee.reference. for in-between-release commits that might be :latest

If a developer wants to deploy a specific oci image _and_ also use the provisioner for their test setup, then they have to fork the trustee repo, hardcode their oci image + tag and use that fork in git.trustee.url/reference in versions.yaml.

stevenhorsman:
Tobin made the comment on the kata issue: https://github.com/kata-containers/kata-containers/issues/10379#issuecomment-2391489223 that suggests maybe we do need set the kustomize image tag, but we could read it from the git reference version. That would be re-adding some of this logic back to the e2e test in I think though, or having a pre-step that does the kustomize edit set image at the same time as we clone the repo?

mkulke:
> Tobin made the comment on the kata issue: [kata-containers/kata-containers#10379 (comment)](https://github.com/kata-containers/kata-containers/issues/10379#issuecomment-2391489223) that suggests maybe we do need set the kustomize image tag, but we could read it from the git reference version. That would be re-adding some of this logic back to the e2e test in I think though, or having a pre-step that does the kustomize edit set image at the same time as we clone the repo?

oof. maybe that could be done as part of the trustee clone step in the CI?

stevenhorsman:
> oof. maybe that could be done as part of the trustee clone step in the CI?

Yeah, that was my thinking.

stevenhorsman:
> oof. maybe that could be done as part of the trustee clone step in the CI?

I've added 
```
pushd kbs/config/kubernetes/base/
kustomize edit set image kbs-container-image=*:${KBS_VERSION}
popd
```
after the clone, which should hopefully do the trick

stevenhorsman:
> I've added
> 
> ```
> pushd kbs/config/kubernetes/base/
> kustomize edit set image kbs-container-image=*:${KBS_VERSION}
> popd
> ```
> 
> after the clone, which should hopefully do the trick

I've already spotted an issue with this:

 https://github.com/confidential-containers/trustee/pkgs/container/key-broker-service/versions?filters%5Bversion_type%5D=tagged only contains released versions
and ghcr.io/confidential-containers/staged-images/kbs has the commit version, so we might need to do some logic and switch the image as well?

mkulke:
> I've already spotted an issue with this:
> 
> https://github.com/confidential-containers/trustee/pkgs/container/key-broker-service/versions?filters%5Bversion_type%5D=tagged only contains released versions and ghcr.io/confidential-containers/staged-images/kbs has the commit version, so we might need to do some logic and switch the image as well?

we could use this to resolve releases to a sha:

```bash
gh api repos/confidential-containers/trustee/commits/v0.10.1 -q .sha
68607d4300dda5a8ae948e2562fd06d09cbd7eca
```

stevenhorsman:
> we could use this to resolve releases to a sha:
> 
> ```shell
> gh api repos/confidential-containers/trustee/commits/v0.10.1 -q .sha
> 68607d4300dda5a8ae948e2562fd06d09cbd7eca
> ```

Yeah, I guess we can do that and always use the staged-image, or alternative do a KBS_VERSION pattern match for vx.y.z and then use the release image in that circumstance?

mkulke (approved):

.github/workflows/e2e_libvirt.yaml
```diff
@@ -105,6 +105,12 @@ jobs:
           git checkout "${KBS_VERSION}"
           pushd kbs
           make CLI_FEATURES=sample_only cli
+          pushd config/kubernetes/base/
+          # Trustee only updates their staging image reliably with sha tags,
+          # so switch to use that and convert the version to the sha
+          KBS_SHA=$(gh api repos/confidential-containers/trustee/commits/${KBS_VERSION} -q .sha)
+          kustomize edit set image kbs-container-image=ghcr.io/confidential-containers/staged-images/kbs:${KBS_SHA}
```
wainersm, line 112: Hi @stevenhorsman !

Allow me to ask you 2 changes here:

1) cat the kustomization file, for debugging sake
2) Grep for `ghcr.io/confidential-containers/staged-images/kbs` and `${KBS_SHA}`. In some places in kata we have this safeguard, in others not. The problem is that `kustomize` may silently not set the image (e.g. in case `kbs-container-image` change), and usually it takes ages until it's realized we were using a wrong image.

wainersm:
I've mixed feelings about this piece of code. In one side I'm happy seeing those confusing kbs properties going away from the e2e tests framework, but on the flip side I don't like the half-in-e2e-half-manual setup required to run the KBS based tests as we end up duplicating stuffs on workflows as well as making the execution on dev workstation more obscure.

Anyway, that's going to the right direction IMO.

wainersm (approved):
Thanks @stevenhorsman !

.github/workflows/e2e_libvirt.yaml
stevenhorsman: For part 2 do you mean in the workflow output:
```
$ cat kustomization.yaml | grep -A 3 images:
images:
- name: kbs-container-image
  newName: ghcr.io/confidential-containers/key-broker-service
  newTag: test
```
for debug, or something else?

stevenhorsman:
> I've mixed feelings about this piece of code. In one side I'm happy seeing those confusing kbs properties going away from the e2e tests framework, but on the flip side I don't like the half-in-e2e-half-manual setup required to run the KBS based tests as we end up duplicating stuffs on workflows as well as making the execution on dev workstation more obscure.
> 
> Anyway, that's going to the right direction IMO.

Agreed, maybe we should look at making a script for the trustee checkout commands and then call this in the e2e framework if KBS deploy is set.

PR #2078, opened 2024-10-02
podvm: Make download-image.sh always pull

stevenhorsman:
During some recent testing of bug fixes I spent quite a lot of time trying to understand why the fixes weren't having any effect on peer pods. After a while I found it was because although the registry version of the podvm image had been updated, when running `download-image.sh` it was just using the version I'd previously downloaded.

If we pass `--pull=always`, in the container create command, this should resolve that problem and it looks to be supported on docker and podman clis

### DIFF
--- a/src/cloud-api-adaptor/podvm/hack/download-image.sh
+++ b/src/cloud-api-adaptor/podvm/hack/download-image.sh
@@ -47,7 +47,7 @@ else
 fi
 
 # Create a non-running container to extract image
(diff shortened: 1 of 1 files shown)


mkulke (approved):

wainersm (approved):
Hi @stevenhorsman !

I had the same problem too, thanks for the fix!
I ran with docker on my env, worked as expected.

PR #2089, opened 2024-10-06
SecureComms: E2e Test SecureComms without KBS

davidhadas:
Support testing SecureComms without KBS

See
 - https://github.com/confidential-containers/cloud-api-adaptor/issues/2096

Should be merged after #2065

### DIFF
--- a/src/cloud-api-adaptor/libvirt/config_libvirt.sh
+++ b/src/cloud-api-adaptor/libvirt/config_libvirt.sh
@@ -92,7 +92,7 @@ installK8sclis() {
     fi
 }
 
-TEST_E2E_SECURE_COMMS=${TEST_E2E_SECURE_COMMS:-none}.
+TEST_E2E_SECURE_COMMS=${TEST_E2E_SECURE_COMMS:-none}
 echo "SECURE_COMMS is ${TEST_E2E_SECURE_COMMS}"
 
 echo "Installing Go..."
@@ -123,7 +123,14 @@ echo "libvirt_ssh_key_file=\"id_rsa\"" >> libvirt.properties
 echo "CLUSTER_NAME=\"peer-pods\"" >> libvirt.properties
 
 # switch to the appropriate e2e test and add configs to libvirt.properties as needed
-case $TEST_E2E_SECURE_COMMS in
+case ${TEST_E2E_SECURE_COMMS} in
+
+  withoutKbs)
+    echo "processing withoutKbs"
+    echo "SECURE_COMMS=\"true\"" >> libvirt.properties
+    echo "SECURE_COMMS_NO_TRUSTEE=\"true\"" >> libvirt.properties
+    echo "INITDATA=\"\"" >> libvirt.properties
+    ;;
 
   *)
     echo "processing none"
--- a/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
+++ b/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
@@ -292,6 +292,21 @@ func VerifyAlternateImage(ctx context.Context, t *testing.T, client klient.Clien
 	return nil
 }
 
+func VerifySecureCommsActivated(ctx context.Context, t *testing.T, client klient.Client, pod *v1.Pod) error {
+	nodeName, err := GetNodeNameFromPod(ctx, client, pod)
+	if err != nil {
+		return fmt.Errorf("VerifySecureCommsConnected: GetNodeNameFromPod failed with %v", err)
+	}
+
+	expectedSuccessMessage := "Using PP SecureComms"
+	err = VerifyCaaPodLogContains(ctx, t, client, nodeName, expectedSuccessMessage)
+	if err != nil {
+		return fmt.Errorf("VerifySecureCommsConnected: failed: %v", err)
+	}
+	t.Logf("PodVM was brought up using SecureComms")
+	return nil
+}
+
 func VerifyCaaPodLogContains(ctx context.Context, t *testing.T, client klient.Client, nodeName, expected string) error {
 	caaPod, err := getCaaPod(ctx, client, t, nodeName)
 	if err != nil {
--- a/src/cloud-api-adaptor/test/e2e/common_suite.go
+++ b/src/cloud-api-adaptor/test/e2e/common_suite.go
@@ -35,6 +35,14 @@ func DoTestCreateSimplePod(t *testing.T, e env.Environment, assert CloudAssert) 
 	}
 }
 
+func DoTestLibvirtCreateSimplePodWithSecureCommsIsValid(t *testing.T, e env.Environment, assert CloudAssert) {
+	if os.Getenv("SECURE_COMMS") != "true" {
+		t.Skip("Skip - SecureComms is configured to be inactive - no need to test")
+	}
+	pod := NewBusyboxPodWithName(E2eNamespace, "simple-test-with-security-comms-is-active").GetPodOrFatal(t)
+	NewTestCase(t, e, "SimplePeerPodWithSecureComms", assert, "PodVM is created with secure comms").WithPod(pod).WithSecureCommsIsActive().Run()
+}
+
 func DoTestDeleteSimplePod(t *testing.T, e env.Environment, assert CloudAssert) {
 	pod := NewBusyboxPodWithName(E2eNamespace, "deletion-test").GetPodOrFatal(t)
 	duration := assert.DefaultTimeout()
--- a/src/cloud-api-adaptor/test/e2e/main_test.go
+++ b/src/cloud-api-adaptor/test/e2e/main_test.go
@@ -124,6 +124,11 @@ func TestMain(m *testing.M) {
 		// Get properties
 		props := provisioner.GetProperties(ctx, cfg)
 
+		if props["SECURE_COMMS"] == "true" {
+			os.Setenv("SECURE_COMMS", "true")
+			log.Info("Do setup secureComms is active")
+		}
+
 		// Set CONTAINER_RUNTIME env variable if present in the properties
 		// Default value is containerd.
 		containerRuntime := defaultContainerRuntime
--- a/src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
+++ b/src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
@@ -23,17 +23,21 @@ const AlternateVolumeName = "another-podvm-base.qcow2"
 
 // LibvirtProvisioner implements the CloudProvisioner interface for Libvirt.
 type LibvirtProvisioner struct {
-	conn             *libvirt.Connect // Libvirt connection
-	containerRuntime string           // Name of the container runtime
-	network          string           // Network name
-	ssh_key_file     string           // SSH key file used to connect to Libvirt
-	storage          string           // Storage pool name
-	uri              string           // Libvirt URI
-	wd               string           // libvirt's directory path on this repository
-	volumeName       string           // Podvm volume name
-	clusterName      string           // Cluster name
-	tunnelType       string           // Tunnel Type
-	vxlanPort        string           // VXLAN port number
+	conn                    *libvirt.Connect // Libvirt connection
+	containerRuntime        string           // Name of the container runtime
+	network                 string           // Network name
+	ssh_key_file            string           // SSH key file used to connect to Libvirt
+	storage                 string           // Storage pool name
+	uri                     string           // Libvirt URI
+	wd                      string           // libvirt's directory path on this repository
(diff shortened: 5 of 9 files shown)


src/cloud-api-adaptor/test/e2e/main_test.go
```diff
@@ -58,6 +60,12 @@ func TestMain(m *testing.M) {
 	// unless it is running with an in-cluster configuration.
 	testEnv = env.New()
 
+	// TEST_CAA_LOG is an option variable which specifies whether the CAA Log should be
+	// presented at the end of the test or not
```
stevenhorsman, line 64: What's the reason for adding this? What do you think it is needed for?

davidhadas: It allows knowing that the test was indeed done with secureComms active and see the log

stevenhorsman: In the CI we already print the logs if the test fails. Is that not sufficient?

stevenhorsman:
So this change is just regression testing to check that the no_trustee version of securecomms doesn't cause any failures?

src/cloud-api-adaptor/test/provisioner/provision.go
```diff
@@ -295,6 +295,16 @@ func (p *CloudAPIAdaptor) Deploy(ctx context.Context, cfg *envconf.Config, props
 		}
 	}
 
+	fmt.Printf("CAA ConfigMap:\n")
```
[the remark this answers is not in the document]
davidhadas: It is providing the contextual information needed to understand what was the setup in which the e2e tests succeeded (or failed)

src/cloud-api-adaptor/test/e2e/main_test.go
stevenhorsman: Is there something specific in the logs that you are looking for for this evidence?

src/cloud-api-adaptor/test/provisioner/provision.go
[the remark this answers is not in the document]
stevenhorsman: If it's needed can we follow the pattern elsewhere in this file and use proper log statements e.g.

```
log.Trace("CAA configmap:")
cmd := exec.Command("kubectl", "get", "cm", "peer-pods-cm", "-n", "confidential-containers-system", "-o", "yaml")
cmd.Env = append(os.Environ(), fmt.Sprintf("KUBECONFIG="+cfg.KubeconfigFile()))
stdoutStderr, err = cmd.CombinedOutput()
log.Tracef("%v, output: %s", cmd, stdoutStderr)
```
?

src/cloud-api-adaptor/test/e2e/main_test.go
davidhadas: SecureComms logs showing the SecureComms flow and spelling out the configuration at the forwarder side
This is just so we are not blind when running e2e tests to see that we are indeed testing what we think that we are testing.

stevenhorsman: What I'm looking for is if there any message we can use as an automated test to validate that secure comms is being used. IMHO that is much better than relying on every user to look through logs.

davidhadas: 1. Yes there is and I can point to one in by looking at the logs :)  I will add it in a comment below. 
    Lets say we have a log statement that shows "......... SecureComms is Enabled....."
    What do suggest to do with it?
    Also, where do you suggest to do that?
    Is log parsing an attractive option for e2e tests? Unless we do something very simple, maybe it is better to consider it in a greater context and outside of this PR.

2. I don't expect every user to look through logs. It is more for anyone developing e2e tests to see what is going on in the forwarder in such tests

src/cloud-api-adaptor/test/provisioner/provision.go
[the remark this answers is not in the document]
davidhadas: The following line uses fmt.Printf (which I used as well) 
```
fmt.Printf("Wait for the %s runtimeclass be created\n", p.runtimeClass.GetName())
```
There are many log.Info()
And some log.Trace()

What do we really need to use here?

src/cloud-api-adaptor/test/e2e/main_test.go
stevenhorsman: 1. We already do some CAA log parsing in our e2e tests (see https://github.com/confidential-containers/cloud-api-adaptor/blob/711a54235193da7092daa5e540f7822b76cb5b24/src/cloud-api-adaptor/test/e2e/assessment_helpers.go#L295)
2. If it's just for developers manually running it then they can just use `kubectl logs ds/cloud-api-adaptor-daemonset -n confidential-containers-system` to check themselves?

davidhadas: I don't think running the tests blindly is a good option, 
manual is not what we need, we need to be able to see the logs of the forwarder during e2e tests
I would suggest that we always have these logs recorded
if you think this is a problem or anti-pattern, I can remove them at the last minute after we see that the commit is ready (and will add them manually with every change I will be making to any e2e test) - I prefer not to do that.

stevenhorsman: We can't see the logs of the forwarder unless you log into the VM which this doesn't do?
Regardless my view is that if we need the logs on CI runs then we should use (and improve) our existing mechanism for printing the logs, not create a new one.

My main point is that a test that requires a user to check is not appropriate for the CI, so we should have a check. At the moment this PR isn't really checking anything that I can see and if `SECURE_COMMS` setting was deleted from the code base then this work act in exactly the same way?

davidhadas: Removed CAA_LOG and added a test to check if CAA logs indicate that SecureComms is activated

src/cloud-api-adaptor/test/provisioner/provision.go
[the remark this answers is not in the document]
davidhadas: Fixed

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -58,6 +58,10 @@ func NewTestCase(t *testing.T, e env.Environment, testName string, assert CloudA
 		deletionWithin: assert.DefaultTimeout(),
 	}
 
+	if os.Getenv("TEST_SECURE_COMMS") == "true" {
+		testCase.WithSecureComms()
+	}
```
stevenhorsman, line 63: You already have `TEST_E2E_SECURE_COMMS` set, so can't you use the value of this rather than setting another env?

davidhadas: This mechanism was replaced

src/cloud-api-adaptor/libvirt/config_libvirt.sh
```diff
@@ -123,10 +123,20 @@ echo "libvirt_ssh_key_file=\"id_rsa\"" >> libvirt.properties
...
+    ;;
 
   *)
     echo "processing none"
     echo "SECURE_COMMS=\"false\"" >> libvirt.properties
     ;;
 esac
+
+echo "Content of libvirt.properties"
+cat libvirt.properties
```
stevenhorsman, line 142: We already do this in the CI (https://github.com/confidential-containers/cloud-api-adaptor/blob/1079abfd3e430a08bc657bb85b565a1b0b8be832/.github/workflows/e2e_libvirt.yaml#L123), so it's not needed here

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -35,6 +35,18 @@ func DoTestCreateSimplePod(t *testing.T, e env.Environment, assert CloudAssert)
...
+func DoTestLibvirtCreateSimplePodWithSecureCommsIsValid(t *testing.T, e env.Environment, assert CloudAssert) {
+	if !testCase_secureComms_isActive {
+		t.Skip("Skip - SecureComms is configured to be inactive - no need to test")
+	}
+	pod := NewBusyboxPodWithName(E2eNamespace, "simple-test-with-security-comms-is-active").GetPodOrFatal(t)
+	if isTestOnCrio() {
+		NewTestCase(t, e, "SimplePeerPodWithSecureComms", assert, "PodVM is created with secure comms").WithPod(pod).WithSecureCommsIsActive().Run()
+	} else {
+		NewTestCase(t, e, "SimplePeerPodWithSecureComms", assert, "PodVM is created with secure comms").WithPod(pod).WithNydusSnapshotter().WithSecureCommsIsActive().Run()
+	}
```
stevenhorsman, line 47: I don't see the need to test that nydus is used here, so can you just run a single test case for containerd and crio please.

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -35,6 +35,18 @@ func DoTestCreateSimplePod(t *testing.T, e env.Environment, assert CloudAssert)
 	}
 }
 
+func DoTestLibvirtCreateSimplePodWithSecureCommsIsValid(t *testing.T, e env.Environment, assert CloudAssert) {
+	if !testCase_secureComms_isActive {
+		t.Skip("Skip - SecureComms is configured to be inactive - no need to test")
+	}
```
stevenhorsman, line 41: We typically do the skip in the provider test case, so libvirt_test in this case, so it might be worth following this pattern.

Also our current pattern for whether to skip tests or not is to use environment variables, so can you use `TEST_E2E_SECURE_COMMS` rather than creating this extra go variable?

src/cloud-api-adaptor/test/e2e/main_test.go
```diff
@@ -123,6 +123,10 @@ func TestMain(m *testing.M) {
 
 		// Get properties
 		props := provisioner.GetProperties(ctx, cfg)
+		if props["SECURE_COMMS"] == "true" {
+			testCase_secureComms_isActive = true
+			log.Info("Do setup secureComms is active")
+		}
```
stevenhorsman, line 129: This code would not be needed if you followed the env var approach

src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
```diff
@@ -95,19 +99,47 @@ func NewLibvirtProvisioner(properties map[string]string) (pv.CloudProvisioner, e
...
 	}
 
+	secure_comms := "false"
+	if properties["SECURE_COMMS"] != "" {
+		secure_comms = properties["SECURE_COMMS"]
+	}
+
+	log.Tracef("NewLibvirtProvisioner properties %v", properties)
+
+	log.Tracef("NewLibvirtProvisioner secure_comms %s", secure_comms)
```
stevenhorsman, line 109: Do we need these traces?

src/cloud-api-adaptor/test/provisioner/provision.go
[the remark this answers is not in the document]
stevenhorsman: You are still using a string builder and not printing stderr?

src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
davidhadas: Left overs

src/cloud-api-adaptor/test/e2e/common_suite.go
stevenhorsman: I understand your reasoning, but none of the other providers would call this test, unless you are planning to support testing this elsewhere in future?

davidhadas: I dont see why SecureComms will not be tested with other providers in the future.

davidhadas: TEST_E2E_SECURE_COMMS is a matrix parameter. The right env is SECURE_COMMS. 

However I tried twice and the env variable is not set during the e2e test. Specifically i tried to replace `if !testCase_secureComms_isActive {` with `if os.Getenv("SECURE_COMMS") != "true" { ` and it failed. 
Before that I tried setting up an additional env variable and check if it exists in the test  - this also failed. Maybe this is because the env is different during the test. Anyhow I could not bring it to work the way you prefer it to be done using an env variable and therefore created the global variable instead.

src/cloud-api-adaptor/test/e2e/main_test.go
davidhadas: See above, I was unable to get this to work with an env var approach.

stevenhorsman (approved):
Looks okay to me. Thanks for the updates!

bpradipt (approved):
/lgtm

PR #2090, opened 2024-10-07
Bump artifact versions in caa multi arch build

stevenhorsman:
Re-enable the upload-artifact and download-artifact version bumps and fix the workflow and script changes that meant there were non-unique file names being used.

The workflow running in my fork (with an additional commit to make the CAA build have a workflow dispatch for easier running): https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/11216651939/job/31176425619


wainersm (approved):
Thanks @stevenhorsman !

mkulke (approved):

PR #2093, opened 2024-10-08
WIP: SecureComms: E2e test SecureComms with KBS

davidhadas:
See:
 - https://github.com/confidential-containers/cloud-api-adaptor/issues/2095

Install KBS and test SecureComms with KBS
Based on #2072 which should be merged first

This PR is on hold and possibly need to be reworked once #2072 is merged.
The plan here was to use the existing KBS installed as part of  CAA testing to also deliver the SecureComms keys. 
An alternative is to add a Trustee Operator installation and use it for SecureComms e2e testing (such PR was not yet introduced). 
In any case it apparently make sense o move forward here only after #2072 is resolved.

### DIFF
--- a/src/cloud-api-adaptor/cmd/agent-protocol-forwarder/main.go
+++ b/src/cloud-api-adaptor/cmd/agent-protocol-forwarder/main.go
@@ -34,7 +34,9 @@ var logger = log.New(log.Writer(), "[forwarder] ", log.LstdFlags|log.Lmsgprefix)
 type Config struct {
 	tlsConfig           *tlsutil.TLSConfig
 	daemonConfig        daemon.Config
-	configPath          string
+	apfConfig           daemon.ApfConfig
+	daemonConfigPath    string
+	apfConfigPath       string
 	listenAddr          string
 	kataAgentSocketPath string
 	podNamespace        string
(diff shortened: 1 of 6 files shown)


davidhadas:
/hold
This PR rely on the existing KBS instrumentation in e2e tests. 
SecureComms may require a more complete instrumentation. 
We will continue with #2094 which should introduce a complete Trustee setup before coming back to this PR and considering if we should pursue it.

src/cloud-api-adaptor/cmd/agent-protocol-forwarder/main.go
```diff
@@ -34,7 +34,9 @@ var logger = log.New(log.Writer(), "[forwarder] ", log.LstdFlags|log.Lmsgprefix)
 type Config struct {
 	tlsConfig           *tlsutil.TLSConfig
 	daemonConfig        daemon.Config
```
bpradipt, line 36: @davidhadas can you please split the apf code changes introducing APF config file and the test related code changes into separate PRs.  Also wondering why APF config is required separate and the daemonConfig cannot be used ?

davidhadas: Please elaborate. 
We already have an e2e test merged for daemonConfig in which the keys are delivered using the daemonConfig. 
This test suppose to cover a case where the daemonConfig is untrusted and we use Trustee to deliver the keys. 
 
 This work is on hold until there is a decision to go ahead with this additional e2e test. 
 I have rebased it, but it is currently untested and should not be merged.

PR #2098, opened 2024-10-08
ci: fix azure nightly mkosi build

mkulke:
Due to a rebase fail, the workflow attempted to use the ubuntu mkosi build. Reverted to building using nix to stay close to the documented procedure. If we opt to use the ubuntu-packaged mkosi, more testing is required and we should change it everywhere.

In the mid-term this would make sense, because currently mkosi is marked as broken in nixos-24.05 and nix-unstable. the version we have pinned in the nix flake (v17) is rather old now.

Drive-by fix: sudo is not required for mkosi images on Ubuntu 22.04

### DIFF
--- a/.github/workflows/azure-podvm-image-build.yml
+++ b/.github/workflows/azure-podvm-image-build.yml
@@ -35,7 +35,7 @@ env:
 
 jobs:
   build-podvm-image:
-    runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):

bpradipt (approved):
/lgtm

PR #2099, opened 2024-10-08
test: Update Trustee deployment to match the simplifed deployment overlays

stevenhorsman:
In https://github.com/confidential-containers/trustee/pull/521 I did some work to enable Trustee to work on s390x (without SE) and generally simplify deployments, such that overlays are only needed for ibm-se, not for x86/s390x in general.

This PR adapts us to use this approach.

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
+++ b/src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
@@ -25,8 +25,7 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 )
 
-// trustee repo related base path
-const TRUSTEE_REPO_PATH = "../trustee"
+var trusteeRepoPath string
 
 func getHardwarePlatform() (string, error) {
 	out, err := exec.Command("uname", "-m").Output()
@@ -46,7 +45,19 @@ func getOverlaysPath() (string, error) {
 	return overlaysPath, nil
 }
 
+func getKbsKubernetesFilePath() string {
+	return filepath.Join(trusteeRepoPath, "/kbs/config/kubernetes/")
+}
+
 func NewKeyBrokerService(clusterName string, cfg *envconf.Config) (*KeyBrokerService, error) {
+	e2e_dir, err := os.Getwd()
+	if err != nil {
+		err = fmt.Errorf("getting the current working directory: %w\n", err)
+		log.Errorf("%v", err)
+		return nil, err
+	}
+	trusteeRepoPath = filepath.Join(e2e_dir, "../trustee")
+
 	log.Info("creating key.bin")
 
 	// Create secret
@@ -55,7 +66,8 @@ func NewKeyBrokerService(clusterName string, cfg *envconf.Config) (*KeyBrokerSer
 	if err != nil {
 		return nil, err
 	}
-	filePath := filepath.Join(TRUSTEE_REPO_PATH, "/kbs/config/kubernetes/"+overlaysPath+"/key.bin")
+
+	filePath := filepath.Join(getKbsKubernetesFilePath(), overlaysPath, "key.bin")
 	// Create the file.
 	file, err := os.Create(filePath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mkulke (approved):

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
```diff
@@ -29,20 +29,33 @@ import (
 const TRUSTEE_REPO_PATH = "../trustee"
 
 func getHardwarePlatform() (string, error) {
-	out, err := exec.Command("uname", "-i").Output()
+	out, err := exec.Command("uname", "-m").Output()
 	return strings.TrimSuffix(string(out), "\n"), err
 }
 
+func getOverlaysPath() (string, error) {
+	arch, err := getHardwarePlatform()
```
bpradipt, line 37: nit: s/arch/platform to align with the method name and other usages later in the code

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
```diff
@@ -477,11 +495,11 @@ func (p *KeyBrokerService) EnableKbsCustomizedAttestationPolicy(customizedOpaFil
 func (p *KeyBrokerService) SetSampleSecretKey() error {
 	kbsClientDir := filepath.Join(TRUSTEE_REPO_PATH, "target/release")
 	privateKey := "../../kbs/config/kubernetes/base/kbs.key"
-	platform, err := getHardwarePlatform()
+	overlaysPath, err := getOverlaysPath()
 	if err != nil {
 		return err
 	}
-	keyFilePath := "../../kbs/config/kubernetes/overlays/" + platform + "/key.bin"
+	keyFilePath := "../../kbs/config/kubernetes/" + overlaysPath + "/key.bin"
```
bpradipt, line 502: Is there any way to use absolute path for privateKey and keyFilePath based on the TRUSTEE_REPO_PATH?

stevenhorsman: I started writing this code, but then noticed `const TRUSTEE_REPO_PATH = "../trustee"`, so the TRUSTEE_PATH isn't absolute. I'm happy to continue the refactor to reduce the number of relative paths we have though?

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
stevenhorsman: Done. Thanks

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
stevenhorsman: It looks like the refactor might have broken the tests too, so I'll look into this locally

bpradipt: Based on your comment, it looks like a general refactor is needed to use absolute path. How about you merge this  and handle the refactor as a separate PR ?

stevenhorsman: Yeah, I think this commit does help reduce relative paths of relatives paths (in the old code we'd have something like `../../trustee/../../<file>`) so we just have the single relative path here now to `../trustee` from `test/e2e`, but there is room for improvement later. I think I already have 1/2 refactors in this area already, so we can steadily get it better.

stevenhorsman: I decided to just take on the battle here, so have re-work the commit to remove the relative trustee path too. It relies on `NewKeyBrokerService` to set it up initially, but as all the references to the trustee path are in that func, or methods of `KeyBrokerService` I think we'll be okay.

stevenhorsman:
FYI the linting failure is (I believe) caused by github updating their default runner, so I have try to address this in https://github.com/confidential-containers/cloud-api-adaptor/pull/2120

bpradipt (approved):
/lgtm

PR #2106, opened 2024-10-11
csi-wrapper: add support for creating peerpod volumes with manually created persistent volumes

daniel-weisse:
The peerpod volume custom resource is currently only created during the `CreateVolume` rpc of the controller service.
This endpoint is called by Kubernetes during the creation of a persistent volume claim (PVC) without an already existing persistent volume (PV).
All subsequent controller, node, and podvm csi-wrapper endpoints that interact with peerpod volumes CRs (e.g. `ControllerPublishVolume`) now assume that a peerpod volume already exists.

However, this is not the case if a PVC with a manually created, or already existing, PV is created.
This is required when statically provisioning a volume, for example, by the [azuredisk-csi-driver to consume a pre-existing disk through a PVC](https://learn.microsoft.com/en-us/azure/aks/azure-csi-disk-storage-provision#statically-provision-a-volume).
When a PV already exists, the `CreateVolume` endpoint is never called, and we directly go to `ControllerPublishVolume` instead.

This PR aims to fix this problem by performing an check during `ControllerPublishVolume` if there is no matching peerpod volume found.
Since the information we get during `ControllerPublishVolume` is a lot more limited than what we get during `CreateVolume`, a little workaround is needed.
Mainly, we don't have access to the `Parameters` field of the PVC, which we use during `CreateVolume` to discern if a peerpod volume should be created, by checking if it contains the key `peerpod`.
So instead, we rely on the user to set the `peerpod` key in the `VolumeContext` field of the manually created PV.

A second issue is that we don't have access to the `VolumeName` of the PVC during `ControllerPublishVolume`.
This value is checked for comparison during `NodePublishVolume` to ensure the peerpod volume matches the "real" volume to publish.
We side steps this issue by setting a placeholder name during `ControllerPublishVolume`, and if this placeholder is set during `NodePublishVolume`, replace it with the real value.

### DIFF
--- a/src/csi-wrapper/pkg/wrapper/nodeservice.go
+++ b/src/csi-wrapper/pkg/wrapper/nodeservice.go
@@ -6,6 +6,7 @@ package wrapper
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -125,10 +126,17 @@ func (s *NodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		podUid, volumeName := s.getPodUIDandVolumeName(targetPath)
 		savedPeerpodvolume.Labels["podUid"] = podUid
 		savedPeerpodvolume.Spec.PodUid = podUid
-		if volumeName != savedPeerpodvolume.Labels["volumeName"] {
-			glog.Error("The volume name from target path doesn't match with the CRD")
-			return
+		savedVolumeName := savedPeerpodvolume.Spec.VolumeName
+		if volumeName != savedVolumeName && savedVolumeName != peerpodVolumeNamePlaceholder {
+			glog.Error("The volume name from target path doesn't match with the CR")
+			return nil, errors.New("the volume name from target path doesn't match with the CR")
 		}
+		if savedVolumeName == peerpodVolumeNamePlaceholder {
+			glog.Info("Detected a placeholder volume name in the CR. Updating the CR with the volume name from the target path")
+			savedPeerpodvolume.Labels["volumeName"] = volumeName
+			savedPeerpodvolume.Spec.VolumeName = volumeName
+		}
+
 		savedPeerpodvolume.Spec.WrapperNodePublishVolumeReq = nodePublishVolumeRequest
 		_, err = s.PeerpodvolumeClient.ConfidentialcontainersV1alpha1().PeerpodVolumes(s.Namespace).Update(context.Background(), savedPeerpodvolume, metav1.UpdateOptions{})
 		if err != nil {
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@yoheiueda - are you able to review this when you get a chance as I think you worked on the csi-wrapper feature?

src/csi-wrapper/pkg/wrapper/nodeservice.go
```diff
@@ -125,10 +126,17 @@ func (s *NodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		podUid, volumeName := s.getPodUIDandVolumeName(targetPath)
 		savedPeerpodvolume.Labels["podUid"] = podUid
 		savedPeerpodvolume.Spec.PodUid = podUid
-		if volumeName != savedPeerpodvolume.Labels["volumeName"] {
-			glog.Error("The volume name from target path doesn't match with the CRD")
-			return
+		savedVolumeName := savedPeerpodvolume.Spec.VolumeName
```
yoheiueda, line 129: I think this is OK, but I am wondering why `.Spec.VolumeName` is used instead of `.Labels["volumeName"]`. Is there any specific reason for this change, or just for conciseness?

src/csi-wrapper/pkg/wrapper/nodeservice.go
```diff
@@ -125,10 +126,17 @@ func (s *NodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		if volumeName != savedVolumeName && savedVolumeName != peerpodVolumeNamePlaceholder {
+			glog.Error("The volume name from target path doesn't match with the CR")
+			return nil, errors.New("the volume name from target path doesn't match with the CR")
+		}
```
yoheiueda, line 133: In the original code, this function returns `nil, nil` when `volumeName != savedVolumeName`. The node returns `nil, error`. Do you think this is a bug in the the original code, and do you fix the bug in the new code?

yoheiueda (commented):
Thank you very much for this great contribution! I am glad to hear that peer pods will support Azure CSI with this PR.

Overall LGTM. I only have a few questions for my better understanding.

src/csi-wrapper/pkg/wrapper/nodeservice.go
daniel-weisse: This was done for conciseness. It felt a bit weird to me to rely on a label, when we have the same information already available as part of the CR spec

src/csi-wrapper/pkg/wrapper/nodeservice.go
daniel-weisse: I think this was a bug in the previous version.
Not returning an error here makes it a bit hard to debug this case without checking the csi-wrapper logs, since it will result in simply never updating the peerpod volume CR, causing the Pod to be stuckin the creating phase

yoheiueda (approved):
LGTM

katexochen (approved):

PR #2107, opened 2024-10-11
test: don't use gh cli in kbs checkout

mkulke:
Since we clone the repo we can use git to resolve a tag/branch to a sha. We probably want to avoid using gh cli in places where it's not required, because it involves setting secrets and makes executing tests more cumbersome. There could also be breakage if people specify a fork as KBS_REPO, since we use the gh cli with a hardcoded repository.


stevenhorsman (approved):
This makes sense given some of the challenges with gh

wainersm (approved):
yes, please. gh is starting to remind me of the yq drama :(

PR #2108, opened 2024-10-11
csi-wrapper: pass `PublishContext` from `ControllerPublishVolume` to `NodeStageVolume`

daniel-weisse:
A CSI driver may rely on the output of `ControllerPublishVolume` during the `NodeStageVolume` call.
Since the csi-wrapper simply creates a fake `ControllerPublishVolumeResponse`, information may be lost between the two calls.
Since the peerpod volume caches the real response, we can use this during the wrapper's `NodeStageVolume` to supply the driver with the real values.

### DIFF
--- a/src/csi-wrapper/pkg/wrapper/podvmservice.go
+++ b/src/csi-wrapper/pkg/wrapper/podvmservice.go
@@ -138,18 +138,20 @@ func (s *PodVMNodeService) ReproduceNodeStageVolume(peerPodVolume *peerpodvolume
 	} else {
 		// The cached NodeStageVolumeRequest contains a faked PublishContext from [ControllerService.ControllerPublishVolume].
 		// Since a CSI driver may depend on PublishContext to pass required information from ControllerPublishVolume to NodeStageVolume,
-		// we need to update the cached PublishContext in NodeStageVolumeRequest with the real one from
+		// we need to replace the PublishContext in the cached NodeStageVolumeRequest with the real one from
 		// the cached ControllerPublishVolumeResponse.
+		publishContext := make(map[string]string)
 		controllerPublishVolumeResJSON := peerPodVolume.Spec.WrapperControllerPublishVolumeRes
 		var controllerPublishVolumeRes csi.ControllerPublishVolumeResponse
 		if err := (&jsonpb.Unmarshaler{}).Unmarshal(bytes.NewReader([]byte(controllerPublishVolumeResJSON)), &controllerPublishVolumeRes); err != nil {
 			glog.Errorf("Failed to convert to ControllerPublishVolumeResponse, err: %s", err)
 		}
 		for k, v := range controllerPublishVolumeRes.PublishContext {
-			modifiedRequest.PublishContext[k] = v
+			publishContext[k] = v
 		}
+		publishContext["device-path"] = peerPodVolume.Spec.DevicePath
+		modifiedRequest.PublishContext = publishContext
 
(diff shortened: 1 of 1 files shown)


stevenhorsman:
@yoheiueda - are you able to review this when you get a chance as I think you worked on the csi-wrapper feature?

src/csi-wrapper/pkg/wrapper/podvmservice.go
```diff
@@ -136,6 +136,19 @@ func (s *PodVMNodeService) ReproduceNodeStageVolume(peerPodVolume *peerpodvolume
...
+		// we need to update the cached PublishContext in NodeStageVolumeRequest with the real one from
+		// the cached ControllerPublishVolumeResponse.
+		controllerPublishVolumeResJSON := peerPodVolume.Spec.WrapperControllerPublishVolumeRes
+		var controllerPublishVolumeRes csi.ControllerPublishVolumeResponse
+		if err := (&jsonpb.Unmarshaler{}).Unmarshal(bytes.NewReader([]byte(controllerPublishVolumeResJSON)), &controllerPublishVolumeRes); err != nil {
+			glog.Errorf("Failed to convert to ControllerPublishVolumeResponse, err: %s", err)
+		}
+		for k, v := range controllerPublishVolumeRes.PublishContext {
+			modifiedRequest.PublishContext[k] = v
+		}
```
yoheiueda, line 150: Do you think we need to clear old key-values in PublishContext here instead of just adding real ones?

The current code explicitly sets these fake key values.

https://github.com/confidential-containers/cloud-api-adaptor/blob/e64346cb1bc992acacfff5c9361ee65ce21795f7/src/csi-wrapper/pkg/wrapper/controllerservice.go#L155-L161

I am not sure why these fake key-values are necessary... I guess these key-values are specific to the IBM Cloud CSI driver. Anyway, this problem is not related to this PR.

https://github.com/kubernetes-sigs/ibm-vpc-block-csi-driver/blob/47fa96465c21ff21d06d447e4571751a5e8885bf/pkg/ibmcsidriver/controller.go#L46-L61

I guess these key-values are harmless for other CSI drivers such as Azure driver, so I think it is OK for now to keep it until some issues occur.

yoheiueda (commented):
This is a good catch! Thank you very much.

src/csi-wrapper/pkg/wrapper/podvmservice.go
daniel-weisse: That's a good point. I guess, as long as CSI drivers don't strictly check for unknown keys, this isn't a problem.
But if everything works as expected, this fake response should never be seen by the CSI driver before first passing through the csi-wrapper component.
I agree that it's probably OK to keep it as is for now.

yoheiueda: Thank you for the explanation.

> But if everything works as expected, this fake response should never be seen by the CSI driver before first passing through the csi-wrapper component.

So, I think it is better to clear the fake response here. Could you please update this PR to clear the fake response in `ReproduceNodeStageVolume(peerPodVolume`?

I don't think this change will affect the behavior of the CSI wrapper with Azure and IBM CSI drivers.

I think we should keep the code to set these fake context values in `ControllerPublishVolume` for now.  I guess the CSI wrapper with IBM Cloud CSI driver will not work correctly without the fake values.

yoheiueda (approved):
LGTM. Thank you very much!

katexochen (approved):

PR #2110, opened 2024-10-11
util: verify provenance

mkulke:
This is prep-work to consume binaries with attestation from guest-components (#2074) The tool asserts that the OCI image has been built on the specified repo with a push on the main branch and the specified digest matches the git sha of the source code and of the workflow.

Note: such a verification is only solid when performed for an oci image w/ digest, since the tags are mutable. we want to resolve a tag to a digest uri and then verify and pull that digest uri. (oras resolve image:tag)

```bash
$ cd src/cloud-api-adaptor
$ ./hack/verify-provenance.sh \
  -a ghcr.io/confidential-containers/guest-components/api-server-rest@sha256:0d2f600490caddb024c4e1e4c9d512c38a0d38e20131dd74702e6dfa4c6890b1 \
  -r confidential-containers/guest-components \
  -d d8da69072424e496486dfb5421a26f16ff2a7abf
Verification passed
```

### DIFF
--- a/src/cloud-api-adaptor/hack/verify-provenance.sh
+++ b/src/cloud-api-adaptor/hack/verify-provenance.sh
@@ -1,0 +1,115 @@
+#!/bin/bash
+
+# Verify Github's attestation reports. Meant to verify binaries built
+# by upstream projects (kata-containers and guest-components).
+#
+# GH cli is used to verify.
+#
+# Asserts on the claims are:
+# - Triggered by push on the main branch
+# - Built on the given repository
+# - The gh action workflow is matching the given digest
+# - The code is matching the given digest
+#
+# -g will fetch attestation via gh cli, this requires GH_TOKEN to be
+# set. By default the attestation will be retrieved by walking the OCI
+# manifest
+
+set -euo pipefail
+
+usage() {
+  echo "Usage: $0 "
+  echo "  -a <oci-artifact w/ sha256 digest>"
+  echo "  -d <expected git sha1 from which the artifact was built>"
+  echo "  -r <repository on which the artifact was built>"
+  echo "  [-g] (optional. fetch attestation using github api)"
+  exit 1
+}
+
+oci_artifact=""
+expected_digest=""
+repository=""
+github="0"
+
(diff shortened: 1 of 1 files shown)


src/cloud-api-adaptor/hack/verify-provenance.sh
```diff
@@ -0,0 +1,111 @@
...
+	usage
+fi
+
+if [[ "$oci_artifact" =~ @sha256:[a-fA-F0-9]{32}$ ]]; then
+	echo "The OCI artifact should be specified using its digest: my-repo.io/my-image@sha256:abc..."
+	exit 1
+fi
+
+# Convention by gh cli
+attestation_bundle="${oci_artifact#*@}.jsonl"
```
stevenhorsman, line 62: I'm trying to decide if there is merit to clean this up at the end of the run, but I'm not sure it's worth it as we are likely to be using this inside the builder container.

stevenhorsman (approved):
Looks great. I've tried positive and negative cases locally with and without the -g mode and it seems to work in them. Great job @mkulke!

src/cloud-api-adaptor/hack/verify-provenance.sh
```diff
@@ -0,0 +1,111 @@
...
+#
+# GH cli is used to verify.
+#
+# Asserts on the claims are:
+# - Triggered by push on the main branch
+# - Built on the given repository
+# - The gh action workflow is matching the given digest
+# - The code is matching the given digest
+#
+# -g will fetch attestation via gh cli, this requires GH_TOKEN to be
```
bpradipt, line 14: nit: If the description of the "g" option is added to the usage function, then it'll avoid looking at the code for the first time user.

bpradipt (approved):
/lgtm

src/cloud-api-adaptor/hack/verify-provenance.sh
mkulke: ah, forgot to address that. yeah, we can clean this up. i'll do it another pr.

PR #2111, opened 2024-10-11
workflow: Fix kustomize permission issue

stevenhorsman:
- Use Wainer's suggested approach of `sudo bash -s` rather than the separate install
- Also update Azure's e2e workflow with the fix

libvirt tested with forked workflow: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/11295615233/job/31418883368


mkulke (approved):
curiously the azure job succeeded anyway, I suppose the user on github-runners has access to /usr/local/bin

snir911 (approved):
LGTM thanks!

PR #2120, opened 2024-10-15
ci: Pin the terraform lint runner

stevenhorsman:
Github are updating their default ubuntu-latest runner from 22.04 to 24.04: https://github.com/actions/runner-images/issues/10636 The 24.04 image doesn't have terraform installed, so this breaks the terraform linting. Pin this at the old version to fix this.

The other linting jobs pass on 24.04, so pin these at 24.04, so we don't have a repeat of this in two years time (accidentally)

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,7 +15,7 @@ on:
 jobs:
   vet-and-fmt:
     name: vet and fmt
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     permissions:
       contents: read
     steps:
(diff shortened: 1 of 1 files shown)


mkulke (approved):

stevenhorsman:
FYI: I was tempted to pin all the `ubuntu-latest` runs-on in this PR, but I figure doing them one file at a time was probably safer.

wainersm (approved):
Thanks @stevenhorsman !

.github/workflows/lint.yaml
```diff
@@ -135,7 +135,7 @@ jobs:
 
   terraform:
     name: Terraform check
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
stevenhorsman, line 138: Add a comment to explain the versioning here

PR #2121, opened 2024-10-16
test/e2e: kbs: Use the cached kbs-client 

stevenhorsman:
To save us storage space, time and the requirement
on the rust toolchain, we can pull down the cached
kbs-client rather than building it ourselves.

Update checkout_kbs.sh to use oras to pull the kbs-client
binary from the cache rather than building it.


stevenhorsman:
Draft as this needs rebasing after #2099 is merged

mkulke (approved):

bpradipt (approved):

PR #2124, opened 2024-10-22
SecureComms: Add testing facility for e2e tests

davidhadas:
Modify the .github/workflow files to enable adding e2e tests for secure comms.
The added code is generalized to enable controlling the libvirt e2e testing matrix using a json file

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -23,7 +23,11 @@ on:
         description: Git ref to checkout the cloud-api-adaptor repository. Defaults to main.
         required: false
         type: string
-
+      secure_comms:
+        default: 'none'
+        description: SecureComms configuration. Defaults to none.
+        required: false
+        type: string
 env:
   CLOUD_PROVIDER: libvirt
   DEBIAN_FRONTEND: noninteractive
@@ -86,6 +90,7 @@ jobs:
 
       - name: Config Libvirt
         run: |
+          export TEST_E2E_SECURE_COMMS="${{ inputs.secure_comms }}"
           ./libvirt/config_libvirt.sh
           echo "CAA_IMAGE=\"${{ inputs.caa_image }}\"" >> libvirt.properties
           # For debugging
@@ -124,6 +129,7 @@ jobs:
           export TEST_PROVISION_FILE="$PWD/libvirt.properties"
           export TEST_PODVM_IMAGE="${{ env.PODVM_QCOW2 }}"
           export TEST_E2E_TIMEOUT="75m"
+          export TEST_E2E_SECURE_COMMS="${{ inputs.secure_comms }}"
 
           make test-e2e
 
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -99,6 +99,8 @@ jobs:
   prep_install:
     needs: [image]
     runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.matrix }}
     env:
       PROVIDERS: "libvirt"
     steps:
@@ -148,6 +150,11 @@ jobs:
           path: src/cloud-api-adaptor/install/
           retention-days: 7
 
+      - name: Define Test Matrix
+        id: matrix
+        run: |
+          echo "matrix=$(jq -c . < ./libvirt/e2e_matrix_libvirt.json)" >> "$GITHUB_OUTPUT"
+
   # Run libvirt e2e tests if pull request labeled 'test_e2e_libvirt'
   libvirt:
     name: libvirt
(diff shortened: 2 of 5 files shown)


davidhadas:
This PR is expected to fail using the existing workflow and succeed using the modified workflow.

It fails now since the existing workflow calls config_libvirt.sh without a parameter
The config_libvirt.sh then returns the list of e2e tests instead of actually installing libvirt et al.
This results in make e2e_tests failure later on. 

As an alternative we could create a new script (e.g. config_libvirt_matrix.sh) that will return the list of matrix options. 
Then, as part of the new workflow, we can call this script instead of calling config_libvirt.sh without parameters.  
The only benefit that I see from this is to allow the existing workflow to pass the e2e tests of this PR.
(I am not sure this justifies introducing another script)

.github/workflows/e2e_libvirt.yaml
```diff
@@ -84,7 +88,8 @@ jobs:
 
       - name: Config Libvirt
         run: |
-          ./libvirt/config_libvirt.sh
+          echo "secure_comms=\"${{ inputs.secure_comms }}\""
+          ./libvirt/config_libvirt.sh ${{ inputs.secure_comms }}
```
stevenhorsman, line 92: What is the end purpose of this change and adding secure comms as a parameter of config_libvirt.sh?

.github/workflows/e2e_run_all.yaml
```diff
@@ -148,6 +150,11 @@ jobs:
           path: src/cloud-api-adaptor/install/
           retention-days: 7
 
+      - name: Define Secure Comms Test Configs
+        id: securecomms
+        run: |
+          ./src/cloud-api-adaptor/libvirt/config_libvirt.sh >> "$GITHUB_OUTPUT"
```
stevenhorsman, line 156: I'm a bit confused by this approach, could you not just set the `secure_comms` in the matrix below rather than going via `config_libvirt.sh`?

.github/workflows/e2e_libvirt.yaml
davidhadas: Different tests in the matrix require different parameters to be set in libvirt.properties
See for example: https://github.com/confidential-containers/cloud-api-adaptor/pull/2089/files#diff-5612979f81fb0700a41c09f8fc0c55c15a264590115690474d10b6b46d40c2ffR136

.github/workflows/e2e_run_all.yaml
davidhadas: If we will set secure_comms in the matrix below rather than going via config_libvirt.sh we will need to change the github workflow for every test we add. I am avoiding this by using the matrix values from a file as shown here and as suggested in https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-an-output-to-define-two-matrices.

.github/workflows/e2e_libvirt.yaml
stevenhorsman: okay, I don't love it, but I guess the properties output has become core to config_libvirt.sh now. Did you consider driving it with an env rather than parameter?

.github/workflows/e2e_run_all.yaml
stevenhorsman: I'm just not sure why config_libvirt is doing this matrix generation? Could you do something like add a json file that has a property of the secure_com matrix and then we could read that out in this section (and potentially re-use if for the other matrix sections in future)?

.github/workflows/e2e_libvirt.yaml
davidhadas: Moved to a TEST_E2E_SECURE_COMMS env variable

.github/workflows/e2e_run_all.yaml
davidhadas: Moved to a json file to define the e2e test matrix. 
Now serving all matrix and is easily extendable.

src/cloud-api-adaptor/libvirt/config_libvirt.sh
```diff
@@ -118,3 +121,11 @@ rm -f libvirt.properties
...
 echo "libvirt_ssh_key_file=\"id_rsa\"" >> libvirt.properties
 echo "CLUSTER_NAME=\"peer-pods\"" >> libvirt.properties
+
+# switch to the appropriate e2e test and add configs to libvirt.properties as needed
+case $TEST_E2E_SECURE_COMMS in
+
+  *)
+    echo "processing none"
+    ;;
+esac
```
stevenhorsman, line 131: Do we need this debug here given that it is echo'd above?

stevenhorsman (commented):
This is looking cool - thanks for adding the matrix json, this could be a really good pattern assuming it works fine 😄

src/cloud-api-adaptor/libvirt/config_libvirt.sh
davidhadas: I think it should stay here as it is the basis of what we will do next - e.g.:
```
# switch to the appropriate e2e test and add configs to libvirt.properties as needed
case $TEST_E2E_SECURE_COMMS in

  "withoutKbs")
    echo "processing withoutKbs"
    echo "SECURE_COMMS=\"true\"" >> libvirt.properties
    echo "SECURE_COMMS_KBS_ADDR=\"false\"" >> libvirt.properties
    echo "INITDATA=\"\"" >> libvirt.properties
    ;;

  *)
    echo "processing none"
    ;;
esac
```

It maybe better to remove the earlier `echo "SECURE_COMMS is ${TEST_E2E_SECURE_COMMS}"` if it feels we are putting too much output.

stevenhorsman: Why can't you add that in when you start using it? At the moment it's not doing anything useful?

davidhadas: Myabe the right thing to do is to spell this out more clearly by adding `echo "SECURE_COMMS=\"false\"" >> libvirt.properties`

So we will have:
```
# switch to the appropriate e2e test and add configs to libvirt.properties as needed
case $TEST_E2E_SECURE_COMMS in

  *)
    echo "processing none"
    ;;
esac
```

Will this do?

stevenhorsman: I think that's okay, it won't be used by anything, so I still don't understand the need for it in this PR, but 🤷

stevenhorsman (approved):

bpradipt (approved):
/lgtm

@davidhadas a small request - can you please add another commit updating the README with the TEST_E2E_SECURE_COMMS option?

PR #2127, opened 2024-10-23
ci: set golang cache key as **/go.sum glob

mkulke:
This should speed up all go builds somewhat, since after reorganizing the repo, the cache that is set up by setup-go has not been working anymore:

```
Restore cache failed: Dependencies file is not found in
/home/runner/work/cloud-api-adaptor/cloud-api-adaptor.
```

drive-by fix: remove unnecessary steps from az oci image build

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -41,15 +41,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Extract go version number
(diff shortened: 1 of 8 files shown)


.github/workflows/azure-e2e-test.yml
stevenhorsman, line 1: If I'm being picky, I'd probably prefer to see the. drive-by fix in a separate commit in future, but it's fine for this PR.

stevenhorsman (approved):
Looks fine to me. Thanks!

bpradipt (approved):
/lgtm

PR #2129, opened 2024-10-24
workflows: Cancel previous jobs if PR/branch is updated

stevenhorsman:
During some testing on my fork and sometimes when running PR testing I've found that I want to re-push whilst the existing CI runs are still going and I need to go in and cancel the old jobs first. This PR looks to address this.

Update our (non-callable) jobs to cancel if another push is done to the PR, or branch, or the unlikely event that another cron job is triggered in order to save time and costs running out of date jobs.

Approach based on stack overflow suggestion:
This is based on the following stack overflow suggestion: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre which is the same approach that kata-containers uses.

Some of these feel a bit redundant as the jobs are quite short e.g. the commit-message-check, but I
figured that for consistency's sake there wasn't any harm in updating these anyway.

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -29,6 +29,10 @@ on:
         type: string
         description: prebuilt caa image
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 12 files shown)


mkulke (approved):
good improvement!

PR #2130, opened 2024-10-24
GitHub hosted runner support

stevenhorsman:
Workflow updates to support using the github-hosted runner. Specifically:
- docker image clean up due to reduced disk space on the runner. If we switch to oras for the podvm "publish" in future we might be able to reduce this, but after looking at it it's non-trivial, so doesn't seem worth adding support for this to the packer based podvm build, which we want to remove soon.
- Making the docker install optional to avoid clashes with the gh-provided version

If people feel we are ready to get switch over in this PR I'm happy to try it, but I thought that getting the changes in and stable first might de-risk the switch.

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -70,8 +70,10 @@ jobs:
       - name: Extract qcow2 from ${{ inputs.podvm_image }}
         run: |
            qcow2=$(echo ${{ inputs.podvm_image }} | sed -e "s#.*/\(.*\):.*#\1.qcow2#")
-           ./hack/download-image.sh ${{ inputs.podvm_image }} . -o ${qcow2}
+           ./hack/download-image.sh ${{ inputs.podvm_image }} . -o ${qcow2} --clean-up
            echo "PODVM_QCOW2=$(pwd)/${qcow2}" >> "$GITHUB_ENV"
+           # Clean up docker images to make space
+           docker system prune -a -f
         working-directory: src/cloud-api-adaptor/podvm
 
       - name: Get the install directory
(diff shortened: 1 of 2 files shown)


stevenhorsman:
This was testing on my fork here: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/11499244178/job/32006785078

wainersm:
Hi @stevenhorsman !

> Workflow updates to support using the github-hosted runner. Specifically:
>

Thanks for this! Testing workflow changes is a pain, it's an extra suffering because of the self-hosted GARM runner. Usually I do a lot of tweaks on my forked tree...it's really counterproductive.
 
>     * docker image clean up due to reduced disk space on the runner. If we switch to oras for the podvm "publish" in future we might be able to reduce this, but after looking at it it's non-trivial, so doesn't seem worth adding support for this to the packer based podvm build, which we want to remove soon.
>

Just curious...is oras images be much smaller or it's that oras will pull-and-clean under the hoods? I assume oras extract the artifact locally but it should have some mechanism to optimize disc space and networking.
 
>     * Making the docker install optional to avoid clashes with the gh-provided version
> 
> 
> If people feel we are ready to get switch over in this PR I'm happy to try it, but I thought that getting the changes in and stable first might de-risk the switch.

I'm okay with this more conservative approach.

wainersm (approved):
LGTM. Thanks @stevenhorsman !

stevenhorsman:
> Just curious...is oras images be much smaller or it's that oras will pull-and-clean under the hoods? I assume oras extract the artifact locally but it should have some mechanism to optimize disc space and networking.

In my experience the image doesn't show up the docker image list after an oras pull, but I've never actually confirmed with the doc that it does a pull and clean automatically and there might be a separate oras cache, so I'll have a poke around to see if I can confirm.

mkulke (approved):
I

PR #2132, opened 2024-10-27
Add support for selecting GPU instance based on Kata pod annotations

bpradipt:
Currently support is added only for AWS provider. Couldn't find a way to retrieve GPU details of an instance in Azure.
Depends on https://github.com/kata-containers/kata-containers/pull/10453


stevenhorsman (approved):
LGTM. Thanks!

src/cloud-providers/util.go
```diff
@@ -102,6 +114,21 @@ func GetBestFitInstanceType(sortedInstanceTypeSpecList []InstanceTypeSpec, vcpus
 
 }
 
+// Implement the GetBestFitInstanceTypeWithGPU function
+func GetBestFitInstanceTypeWithGPU(sortedInstanceTypeSpecList []InstanceTypeSpec, gpus, vcpus, memory int64) (string, error) {
```
snir911, line 118: Should the original GetBestFitInstanceType make sure 0 GPUs instance type are returned (in case gpuCount is 0)?

snir911 (approved):
LGTM, Thanks!, one minor comment,
Maybe worth pointing in the doc also to the [peerpod-ctrl resource management](https://github.com/confidential-containers/cloud-api-adaptor/tree/main/src/peerpod-ctrl) as "resource management" is ambiguous term in the peer-pods context..

bpradipt:
The TestSshProxyReverseKBS is flaky. I'm merging this one while we fix the test case

PR #2133, opened 2024-10-28
ci: disable concurrency for child workflows

mkulke:
Those should not be set for callable workflows, since it's not supported to have concurrency groups set on both parent and children workflows.

https://github.com/github/vscode-github-actions/issues/135


wainersm:
> LGTM. Sorry for my lack of knowledge that let this in!

And I tested that latest https://github.com/rhysd/actionlint/ wouldn't have caught it either.

wainersm (approved):

PR #2134, opened 2024-10-28
workflow: Switch libvirt e2e test to gh-runnner

stevenhorsman:
Now the pre-work changes have gone in in #2130,
we should be able to switch over to use the gh-hosted runner for libvirt e2e tests, so save Azure credits and make it easier to do testing on local forks.

Drive by fixes to add KBS debug info & add cluster clean-up support for "bare-metal" runners.

Tested in https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/11558678218/job/32171610739

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -34,7 +34,7 @@ defaults:
 
 jobs:
   test:
-    runs-on: az-ubuntu-2204
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -176,3 +176,7 @@ jobs:
           echo "::endgroup::"
         # Avoid running with `set -e` as command fails should be allowed
         shell: bash {0}
+
+      - name: Clean-up cluster
+        if: ${{ runner.environment == 'self-hosted' }}
+        run: ./libvirt/kcli_cluster.sh delete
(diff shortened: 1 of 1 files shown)


wainersm (approved):
Good stuff. Thanks @stevenhorsman !

mkulke (approved):

mkulke:
do we plan to drop the pull-request-target/authorize settings for those e2e tests then, so that libvirt e2e tests can run in the context of a fork?

stevenhorsman:
> do we plan to drop the pull-request-target/authorize settings for those e2e tests then, so that libvirt e2e tests can run in the context of a fork?

So I'm working on getting the s390x self-hosted runner working for the libvirt tests as well, which would need the `pull-request-target` trigger. What are you looking for in the run from a fork? I am running it from my fork for testing, either running from the daily-libvirt e2e workflow, or just this workflow with a `workflow_dispatch:` trigger option. If you think the standalone run would be useful then I'm happy to add that into this PR?

mkulke:
> > do we plan to drop the pull-request-target/authorize settings for those e2e tests then, so that libvirt e2e tests can run in the context of a fork?
> 
> So I'm working on getting the s390x self-hosted runner working for the libvirt tests as well, which would need the `pull-request-target` trigger. What are you looking for in the run from a fork? I am running it from my fork for testing, either running from the daily-libvirt e2e workflow, or just this workflow with a `workflow_dispatch:` trigger option. If you think the standalone run would be useful then I'm happy to add that into this PR?

ah, ok. yes, in this case we need to keep the pull-request-target setting in place

stevenhorsman:
> ah, ok. yes, in this case we need to keep the pull-request-target setting in place

I've put an item in tomorrow's meeting to discuss the new proposed flow and maybe we can find some issues/improvements to make with it and have a plan for things like this.

PR #2135, opened 2024-10-28
versions: Bump fedora base image to f40

stevenhorsman:
Fedora 39 is EoL in a couple of weeks and I've started seeing a few repo connection errors, so lets bump to F40 as the latest supported version.

> [!NOTE]
> Once `golang-fedora:1.22.7-39` is built, this will be followed up with
a bump of the builder_base as well.

### DIFF
--- a/hack/Dockerfile.golang
+++ b/hack/Dockerfile.golang
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.5-labs
 
-ARG BASE_IMAGE=registry.fedoraproject.org/fedora:39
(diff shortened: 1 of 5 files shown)


wainersm:
Hi @stevenhorsman !

peerpods-ctrl image built fine with F40 but I don't remember if we have a workflow for testing it. I was attempted to trigger the libvirt e2e CI to see if caa runtime on F40 would work with binaries built with F39, but it doesn't matter in the end as you will bump the binaries builder to F40 anyway.

wainersm (approved):
Thanks @stevenhorsman !

mkulke (approved):

PR #2136, opened 2024-10-29
versions: Bump golang-fedora image version

stevenhorsman:
Now that `golang-fedora:1.22.7-40` has been built and pushed to quay, we can update our builds to use this and remove our use of fedora 39 which is EoL in two weeks.


bpradipt (approved):
/lgtm

wainersm (approved):

PR #2137, opened 2024-10-30
podvm_builder: Add s390x gh install support

stevenhorsman:
#2074 introduced the installation of the gh cli into the fedora podvm_builder Dockerfile using dnf. This isn't available for s390x and there isn't a released s390x version of the gh cli, so build it from source on that architecture.


BbolroC (approved):
LGTM as far as the CI goes green. Thanks @stevenhorsman !

mkulke (approved):

PR #2139, opened 2024-10-30
versions: Fix csi-wrapper's base image

stevenhorsman:
In #2136 I completed missed that we'd forgotten to bump the base image for golang updates, so just updated the fedora version and broken things.


bpradipt (approved):
/lgtm

mkulke (approved):

PR #2141, opened 2024-11-04
Misc security related fixes for AWS and Azure

bpradipt:
aws: Avoid integer overflow when processing RootVolumeSize
azure: Verify the SSH key to ensure the file is using a supported  public key format

### DIFF
--- a/src/cloud-providers/util.go
+++ b/src/cloud-providers/util.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers/util"
+	"golang.org/x/crypto/ssh"
 )
 
 var logger = log.New(log.Writer(), "[adaptor/cloud] ", log.LstdFlags|log.Lmsgprefix)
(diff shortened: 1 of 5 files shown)


stevenhorsman (approved):
LGTM

src/cloud-providers/util.go
```diff
@@ -142,3 +143,39 @@ func WriteUserData(instanceName string, userData string, dataDir string) (string
 	// Return the file path
 	return filePath, nil
 }
+
+// Verify SSH public key file
+// Check the permissions and the content of the file to ensure it is a valid SSH public key
+func VerifySSHKeyFile(sshKeyFile string) error {
```
mkulke, line 149: with the change to mkosi images, we won't have the option to provision SSH keys anymore, since this functionality relies on cloud-init and writeable root filesystems

bpradipt: True for CoCo and that's why I added it to the ConfigVerifier method which is disabled by default. For non CoCo (regular peer-pods) this at least gives an option to ensure using valid public key. Is that acceptable ?

mkulke (approved):

PR #2143, opened 2024-11-04
podvm: re-arrange service order for sealed secrets

mkulke:
For sealed secrets to work kata-agent needs to connect to CDH, since it will repace env-variables in the container procoss with an unsealed confidential secret. A systemd path monitor watches for the CDH socket to appear to launch kata-agent.

Also, kata is writing the ocicrypt_config to /run in the agent process so we should not have bundle another in the podvm disk.

~Draft until https://github.com/kata-containers/kata-containers/pull/10481 is merged, need to bump kata to the new revision.~

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
+++ b/src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
@@ -1,11 +1,9 @@
 [Unit]
 Description=Kata Agent
 BindsTo=netns@podns.service
-Wants=process-user-data.service attestation-agent.service
-After=netns@podns.service process-user-data.service attestation-agent.service
+After=netns@podns.service process-user-data.service
 
 [Service]
-Environment=OCICRYPT_KEYPROVIDER_CONFIG=/etc/ocicrypt_config.json
 ExecStartPre=mkdir -p /run/kata-containers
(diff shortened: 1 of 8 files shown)


src/cloud-api-adaptor/podvm/files/etc/systemd/system/kata-agent.service
```diff
@@ -1,11 +1,9 @@
 [Unit]
 Description=Kata Agent
 BindsTo=netns@podns.service
-Wants=process-user-data.service attestation-agent.service
-After=netns@podns.service process-user-data.service attestation-agent.service
+After=netns@podns.service process-user-data.service
```
bpradipt, line 4: attestation-agent.service should be there as well, no ?

mkulke: kata-agent is activated by the CDH socket (in the kata-agent.path directive), there is no direct dependency on attestation-agent (indirectly yes, since CDH is depending on the AA socket). It probably wouldn't hurt to leave it in, but since we have to follow a strict ordering of components for sealed secrets, I think it makes sense to be explicit about the dependencies.

bpradipt (approved):
/lgtm

stevenhorsman (approved):
LGTM. Thanks!

PR #2152, opened 2024-11-13
workflows: Switch to pin ubuntu runners

stevenhorsman:
Switch all our workflows to use the pinned ubuntu-24.04 runner rather than ubuntu-latest


bpradipt (approved):
/lgtm

PR #2155, opened 2024-11-15
Enable pod VM image selection via pod annotation

bpradipt:


snir911 (approved):
LGTM, thanks!
TBD: Should we add an option to disable image thought annotations?

bpradipt:
> LGTM, thanks! TBD: Should we add an option to disable image thought annotations?
This can be done at containerd/crio or in configuration.toml, like any other Kata hypervisor annotations.  I don't think it makes sense to duplicate it.

PR #2156, opened 2024-11-18
ci: install oras specifically for azure e2e tests

mkulke:
When the runners were fixed to 24.04 oras wasn't part of the base installation anymore.


stevenhorsman (approved):
Thanks @mkulke

wainersm (approved):

PR #2157, opened 2024-11-18
podvm-mkosi: bump to fedora 40

mkulke:
We forgot to bump the mkosi version to fedora 40 in the mkosi config.

I triggered an azure e2e-test w/ an [fedora40 podvm image](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/11894045370/job/33140667630#step:6:44) and it passed


wainersm:
Just like Bart Simpsons I should write 100x in a board "mkosi is the future" so I don't keep forgetting about it.

wainersm (approved):

PR #2160, opened 2024-11-19
optimize process-user-data startup time

mkulke:
the binary is currently importing a lot of libraries which have costly init() procedures (around 8-10s at startup) this is due to sharing some consts with the CAA modules. We can avoid that by inlining code in process-user-data and moving shared consts into dedicated modules.

We can use a cpuid field to narrow down a hypervisor, so we don't have to probe an IMDS endpoint while circling through the cloud provider, delaying startup.

this is the result w/ debug enabled, the crypto libraries that are used to hash initdata are still consuming a few, but it's far better now.

```bash
Nov 19 15:51:19 fedora process-user-data[676]: init main @1029 ms, 0.039 ms clock, 2600 bytes, 22 allocs
```


src/cloud-api-adaptor/pkg/userdata/heuristics.go
stevenhorsman, line 1: Is it worth mentioning this file in `addnewprovider.md` as it's a bit less obvious that the cloud-providers changes?

src/cloud-api-adaptor/pkg/paths/paths.go
```diff
@@ -0,0 +1,11 @@
+package paths
+
+const (
+	AACfgPath          = "/run/peerpod/aa.toml"
+	AuthFilePath       = "/run/peerpod/auth.json"
+	CDHCfgPath         = "/run/peerpod/cdh.toml"
+	InitDataPath       = "/run/peerpod/initdata"
+	AgentCfgPath       = "/run/peerpod/agent-config.toml"
+	ForwarderCfgPath   = "/run/peerpod/daemon.json"
+	DockerUserDataPath = "/peerpod/userdata.json"
```
stevenhorsman, line 10: Is there an opportunity to move the existing constants to use this instead e.g. in cloud.go. I'm slightly concerned about us now having two copies that could go out of sync?

src/cloud-providers/docker/provider.go
```diff
@@ -72,11 +72,11 @@ func (p *dockerProvider) CreateInstance(ctx context.Context, podName, sandboxID
 
 	// Create volume binding for the container
 
-	// mount userdata to DockerUserDataUrl=/peerpods/userdata.json
+	// mount userdata to /peerpods/userdata.json
 	// This file will be read by process-user-data and daemon.json will be written to
 	// /run/peerpods/daemon.json at runtime
 	volumeBinding := []string{
-		fmt.Sprintf("%s:%s", instanceUserdataFile, DockerUserDataUrl),
+		fmt.Sprintf("%s:%s", instanceUserdataFile, "/peerpods/userdata.json"),
```
stevenhorsman, line 79: I guess we don't want to pull in the `DockerUserDataPath` from paths and we want have the cloud-providers->CAA cyclic dependency. Maybe a comment here and in paths.go to say they have to stay in sync is the least bad approach then?

src/cloud-api-adaptor/pkg/paths/paths.go
stevenhorsman: I mean these: https://github.com/confidential-containers/cloud-api-adaptor/blob/a22b492ebc709cac8b361e850647e67494d7b973/src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go#L38-L41. They are in the cloud-api-adaptor module, so I don't think there is a dependency problem?

stevenhorsman: If we aren't using them apart from here can we just delete them?

mkulke: that's true. we are probably using them when creating the user-data file, I probably mixed something up, in theory those should be able to use the paths module

src/cloud-api-adaptor/pkg/userdata/heuristics.go
mkulke: actually it's more, you also need to implement a way to retrieve userdata for a new provider

stevenhorsman (approved):
LGTM. Thanks for the updates!

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -78,51 +83,57 @@ func (d DefaultRetry) GetRetryDelay() time.Duration {
...
 }
 
 type AWSUserDataProvider struct{ DefaultRetry }
 
 func (a AWSUserDataProvider) GetUserData(ctx context.Context) ([]byte, error) {
-	url := aws.AWSUserDataImdsUrl
+	url := AWSUserDataImdsUrl
 	logger.Printf("provider: AWS, userDataUrl: %s\n", url)
-	return aws.GetUserData(ctx, url)
+	return imdsGet(ctx, url, false, nil)
```
bpradipt, line 96: shouldn't the bool be true here ?

mkulke: aws doesn't encode the userdata with base64, does it? at least it wasn't in the original code

bpradipt: Correct.. I forgot about it.. Do you mind adding a comment above the code ?

bpradipt (approved):
/lgtm

PR #2162, opened 2024-11-21
Kata 3.11.0 bump

stevenhorsman:
Bump kata-agent and go modules to use the 3.11.0 release version that will be used in CoCo 0.11.0


bpradipt (approved):
/lgtm

PR #2163, opened 2024-11-21
e2e: add encrypted image test

mkulke:
this adds a test for image layer decryption to the libvirt & azure suite. The test currently uses a predefined image with hardcoded key + keyid. In the future we might want to do something more comprehensive, but it's crucial that we have CI coverage for this feature. 

note: we _do_ need the `io.containerd.cri.runtime-handler` annotation set for encrypted images, at least that was the case in my local tests.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/README.md
+++ b/src/cloud-api-adaptor/test/e2e/README.md
@@ -180,7 +180,7 @@ AZURE_IMAGE_ID="unused"
 EOF
 ```
 
-Run the test suite with the respective flags:
+Run the test suite with the respective flags. Enable `TEST_KBS` and provide a `KBS_ENDPOINT` to run KBS-related test.
 
 ```bash
 make test-e2e \
@@ -189,6 +189,9 @@ TEST_TEARDOWN=no \
 TEST_PROVISION=no \
 TEST_INSTALL_CAA=no \
 TEST_PROVISION_FILE="${PWD}/skip-provisioning.properties" \
+TEST_KBS=yes \
+KBS_ENDPOINT=http://10.224.0.5:30362 \
+RUN_TESTS=TestAzureImageDecryption
 ```
 
 ### IBM Cloud
--- a/src/cloud-api-adaptor/test/e2e/assessment_runner.go
+++ b/src/cloud-api-adaptor/test/e2e/assessment_runner.go
@@ -14,9 +14,11 @@ import (
 
 	"gopkg.in/yaml.v2"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/e2e-framework/klient"
 	"sigs.k8s.io/e2e-framework/klient/wait"
 	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
 	"sigs.k8s.io/e2e-framework/pkg/env"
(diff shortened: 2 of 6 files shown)


src/cloud-api-adaptor/test/e2e/README.md
```diff
@@ -189,6 +189,9 @@ TEST_TEARDOWN=no \
 TEST_PROVISION=no \
 TEST_INSTALL_CAA=no \
 TEST_PROVISION_FILE="${PWD}/skip-provisioning.properties" \
+KBS_TEST=yes \
```
stevenhorsman, line 192: Is this the correct variable, or is it azure only? We have the section elsewhere in the doc that says to run
```
export DEPLOY_KBS=yes
```
to run KBS tests - 
We also have `os.Setenv("TEST_KBS", "true")` that is set when ` DEPLOY_KBS=yes`, so was it that you meant to set here as the deployment is a separate step in AZURE that we don't want to conflate with testing?

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -499,6 +500,41 @@ func DoTestPodsMTLSCommunication(t *testing.T, e env.Environment, assert CloudAs
...
+		key := []byte{
+			164, 116, 132, 228, 223, 147, 255, 113, 134, 125,
+			191, 238, 153, 165, 160, 7, 198, 223, 29, 207,
+			221, 65, 107, 230, 197, 219, 35, 161, 102, 166,
+			209, 91}
+
+		_ = kbs.SetImageDecryptionKey(keyID, key)
+		_ = kbs.EnableKbsCustomizedResourcePolicy("allow_all.rego")
+		_ = kbs.EnableKbsCustomizedAttestationPolicy("allow_all.rego")
+		kbsEndpoint, _ = kbs.GetCachedKbsEndpoint()
```
stevenhorsman, line 528: I'm happy to update this in #2055 (if I remember), but in ee76f06936c2df02226230e04b82bd5b3fefc83b I added checks of the errors thrown as I've seen at least one example of where we had issues and ignored them which made debugging a lot harder in the CI, so maybe we should consider that here too even if it is more golang boilerplate?

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
```diff
@@ -495,15 +495,10 @@ func (p *KeyBrokerService) EnableKbsCustomizedAttestationPolicy(customizedOpaFil
 	return nil
 }
 
-func (p *KeyBrokerService) SetSampleSecretKey() error {
+func (p *KeyBrokerService) setSecretKey(resource string, path string) error {
```
stevenhorsman, line 498: This will clash with #2055 again, which is okay I can handle it, but it's frustrating that we've not been able to land that as it's pretty old (which I appreciate is not at all your fault @mkulke!)

stevenhorsman (commented):
A couple of minor comments, but otherwise this looks great. Thanks very much for adding coverage of this crucial feature @mkulke!

src/cloud-api-adaptor/test/e2e/README.md
mkulke: not sure, I too that TEST_KBS variable from the`isTestWithKbs()` source code (test/e2e/common.go). I think there is a difference: if provision=false there won't be a KBS deployment, but you might still want to run the tests against an existing KBS deployment

src/cloud-api-adaptor/test/e2e/common_suite.go
mkulke: sure, just for this test?

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
mkulke: I don't mind rebasing, but I would recommend having this test in 0.11. so you might want to hit the gong to get people to review that PR.

src/cloud-api-adaptor/test/e2e/README.md
stevenhorsman: I under the distinction with `DEPLOY_KBS` and `TEST_KBS`, but my main point (which I probably watered down) is that you have documented `KBS_TEST`, not `TEST_KBS`.

src/cloud-api-adaptor/test/e2e/common_suite.go
stevenhorsman: Yeah, that's fine. I've handled the rest elsewhere.

src/cloud-api-adaptor/test/e2e/README.md
mkulke: ohh, eagle eyes. thx 👍

src/cloud-api-adaptor/test/provisioner/trustee_kbs.go
stevenhorsman: Sorry, my comment was mostly an unproductive moan rather than a helpful suggestion.

src/cloud-api-adaptor/test/e2e/assessment_runner.go
```diff
@@ -167,6 +169,33 @@ func (tc *TestCase) WithNydusSnapshotter() *TestCase {
...
+	err := client.Resources().Get(ctx, DEFAULT_AUTH_SECRET, E2eNamespace, secret)
+	if err == nil {
+		secretExists = true
+	}
+	if secretExists {
+		return nil
+	}
+
+	providerName := os.Getenv("CLOUD_PROVIDER")
+	authfile, err := os.ReadFile("../../install/overlays/" + providerName + "/auth.json")
```
bpradipt, line 184: Can you please add a comment here that this path is relative to `test/e2e`

bpradipt (approved):
/lgtm

stevenhorsman (approved):
LGTM. Thanks for the updates

PR #2165, opened 2024-11-21
ci: fix verify-provenance for agent

mkulke:
The agent attestation can also be created by a workflow_trigger on main, so we have to allow that in the verification. There was also a bug in the routine that resolved a release tag to commit. kata doesn't tag their releases with a preceding 'v'.


bpradipt (approved):
/lgtm

PR #2170, opened 2024-11-27
Update go directive to 1.22.7 to address CVE

beraldoleal:
The encoding/gob package in the Go standard library is affected by CVE-2024-34156. This vulnerability is resolved in Go versions >=1.22.7 or >=1.23.1.

Our current go.mod directive (go 1.22.0) could allow the 1.22.0 toolchain to compile the code, leaving it vulnerable. Updating the directive to 1.22.7 ensures the code is built with a secure version of Go.

### DIFF
--- a/src/cloud-api-adaptor/go.mod
+++ b/src/cloud-api-adaptor/go.mod
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor
 
-go 1.22.0
+go 1.22.7
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.11.1
--- a/src/cloud-api-adaptor/podvm/go.mod
+++ b/src/cloud-api-adaptor/podvm/go.mod
@@ -4,4 +4,4 @@
 
 module github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/podvm
 
-go 1.22.0
+go 1.22.7
--- a/src/cloud-providers/go.mod
+++ b/src/cloud-providers/go.mod
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers
 
-go 1.22.0
+go 1.22.7
 
 require (
 	cloud.google.com/go/compute v1.23.3
--- a/src/csi-wrapper/go.mod
+++ b/src/csi-wrapper/go.mod
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/csi-wrapper
 
-go 1.22.0
+go 1.22.7
 
 require (
 	github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor v0.10.0
--- a/src/peerpod-ctrl/go.mod
+++ b/src/peerpod-ctrl/go.mod
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl
 
-go 1.22.0
+go 1.22.7
 
 require (
 	github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers v0.10.0
--- a/src/webhook/go.mod
+++ b/src/webhook/go.mod
@@ -1,6 +1,6 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/webhook
 
-go 1.22.0
+go 1.22.7
 
 require (
 	k8s.io/api v0.29.6


beraldoleal:
Hi @stevenhorsman, @wainersm I would like to revist our previous discussion about the frequency to update this. Afaict, right now we are affected by this CVE, and I would like to know if we can consider bumping it to reflect our minimal go lang requirement.

stevenhorsman:
I think I would fall back to say that we are using 1.22.7 in our builds and if the [Kubernetes team](https://github.com/kubernetes/kubernetes/blob/master/go.mod#L9) are happy with .0 versions, then it feels like there isn't a significant disadvantage to it? I think in the previous PR we agreed to model ourselves after them?

beraldoleal:
> I think I would fall back to say that we are using 1.22.7 in our builds and if the [Kubernetes team](https://github.com/kubernetes/kubernetes/blob/master/go.mod#L9) are happy with .0 versions, then it feels like there isn't a significant disadvantage to it? I think in the previous PR we agreed to model ourselves after them?

Yes, I record that, but we found this CVE recently and I was wondering if we should do something here too.

While it's true that our builds are "safe" against this CVE, the code itself isn’t. Anyone trying to build this themselves could end up with vulnerable software.

The inconsistency might not seem significant because 1.22.7 used by CI, satisfies the 1.22.0 directive, but leaving 1.22.0 in the go.mod could give the impression that we’re okay with it being built using 1.22.0, which isn’t ideal.

About k8s, it is on 1.23.0 already, but that version is also affected by the same CVE.

beraldoleal:
> IMHO we should disregard the minor version of Go in go.mod, our concern should be about API/ABI compatibility. On CI/build, on the other hand, we tune the minor version for releasing our code free of CVE as much as possible. Likewise, downstream consumers should be concerned about the compiler version they use.

Let me just clarify something... this isn’t about our downstream builds, they’re already fine since our tools use newer Go versions. I’m coming at this with an "upstream hat".

I feel like we have some responsibility to help guide users toward secure practices. One thing to consider: where’s the best place to signal the minimum Go version? Docs, CI, or the go.mod directive? go.mod feels like more a natural way to nudge people toward safer Go versions.

> Updating minor version often would disturb the developers environment too, that would be required to keep updating their compiler version.

I don’t think that’s much of an issue anymore. Since 1.21+, fetching the right version happens automatically based on what’s in go.mod, so it’s pretty seamless.

Also having go.mod and CI consistent keeps CI and dev environments in sync, making life easier.

On the Kubernetes side, I couldn't find any rule about not bumping patch versions in go.mod. I did found [only this example](https://github.com/kubernetes/kubernetes/commit/702d911e22604aa69b42f77804c27caaab7e4def#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R9) with a bump, tho.

My understanding is that Go 1.21+ backward compatibility improvements are helping the Kubernetes community adopt updates more confidently, especially to stay on supported versions for critical fixes.

Anyway, that’s where I’m coming from and I'm def not a go/k8s expert... I get it’s a bit of a balancing act, and I'm happy to go with what you guys choose here.

wainersm:
> IMHO we should disregard the minor version of Go in go.mod, our concern should be about API/ABI compatibility. On CI/build, on the other hand, we tune the minor version for releasing our code free of CVE as much as possible. Likewise, downstream consumers should be concerned about the compiler version they use.
> 
> Updating minor version often would disturb the developers environment too, that would be required to keep updating their compiler version.

Errata to my comment: I meant to say, disregard the **patch** version as in semantic versioning `major.minor.patch` :)

bpradipt:
So if I understand correctly, there are two options being discussed: 

Option-1: Use patch version for golang release in go.mod (eg. go 1.22.7)
Option-2: Don't use patch version for golang release in go.mod (eg. go 1.22 or go 1.22.0). Instead the patch version is used in the build tools used for releasing project artifacts

A good explanation can be found here - https://github.com/cli/cli/issues/9489
Reproducing a snippet from the issue
"Since I'm not aware of any language changes (like range-over-func) coming in patch releases, it makes the most sense in my mind to use the language version.
This has the advantage of not forcing every downstream consumer of your module to bump every time you change the patch version of your go.mod for no real reason"

My reading is that the above statement recommends Option-2.  What do you folks think ?

beraldoleal:
Hi @bpradipt, thanks for jumping in!

Looking at the `cli/cli` example, I think they’re dealing with a similar issue (since Go 1.21 changed how it handles the go directive), but their situation is a bit different because they use the `toolchain` directive, which we don’t. Most of their discussion seems to be about switching from x.y to x.y.z, like what we did here: https://github.com/confidential-containers/cloud-api-adaptor/pull/2080.

From what I understand, they’re decided to stick with the x.y.z approach, *including Z bumps*, as you can see at their [go.mod](https://github.com/cli/cli/blob/trunk/go.mod#L3).

For us, the inconsistency comes from how CI builds don’t fully match the developer or downstream environments. CI is bootstrapping the environment manually, while we’ve been using a “minimal” value for the go directive. Aligning go.mod and CI would definitely help keep things in sync between CI and dev environments.

IMO patch versions are mostly just minor updates or security fixes, so they’re designed to be safe and backward-compatible. Personally, I’m much more comfortable bumping the patch version than minor or major versions. One downside others pointed  is the frequency of updates, but with the new download mechanism, even that feels manageable. I don’t think we need to chase every patch version, though, just bump when there’s a security fix or bug that impacts us.

Besides the example you gave, I noticed a couple of Kubernetes projects that also use patch versions in their go.mod:

* https://github.com/kubernetes/kops/blob/master/go.mod
* https://github.com/kubernetes/ingress-nginx/blob/main/go.mod

It feels like there isn’t a strict rule that works for every project, so it’s just about finding what makes the most sense for us.

stevenhorsman:
Apologies Beraldo as we discussed this in your absence, but there seems to be consensus that we don't want to dictate go versions to local developers, so we are going to close this.

PR #2171, opened 2024-11-28
ci: kustomize CAA_IMAGE in e2e libvirt test

mkulke:
There is no logic to kustomize the image in the provisioner code, so it will always deploy the default :latest from quay.io

Introduced steps to overwrite the image in the kustomization.yaml using kustomize.

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -92,7 +92,6 @@ jobs:
         run: |
           export TEST_E2E_SECURE_COMMS="${{ inputs.secure_comms }}"
           ./libvirt/config_libvirt.sh
-          echo "CAA_IMAGE=\"${{ inputs.caa_image }}\"" >> libvirt.properties
           # For debugging
           cat libvirt.properties
 
@@ -115,6 +114,15 @@ jobs:
           test/utils/checkout_kbs.sh
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Kustomize CAA image
+        working-directory: src/cloud-api-adaptor/install/overlays/libvirt
+        run: |
+          kustomize edit set image cloud-api-adaptor="${{ inputs.caa_image }}"
+          # Print for debugging
+          echo "::group::libvirt kustomization"
+          cat kustomization.yaml
+          echo "::endgroup::"
 
       - name: run tests
         id: runTests


stevenhorsman (approved):
LGTM. In my version of this: https://github.com/confidential-containers/cloud-api-adaptor/pull/2019/commits/a0ac4a835f9db31eced4c7bf222088268acf76ba I printed the contents of the file, which has been very helpful for debugging, so it might be worth considering?

wainersm:
Hi @mkulke @stevenhorsman , we've a duplication here.

In [e2e_run_all.yaml](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml), the [Update kustomization configuration](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml#L125) step is meant to update the installation configuration only once; the `src/cloud-api-adaptor/install/` is archived in [Upload install directory for next test runs](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml#L146) so that downstream provider-specific jobs just download the install directory (eg., see [Get the install directory](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_libvirt.yaml#L84) for e2e_libvirt.yaml ).

This way 1) avoiding the duplication of logic on the downstream jobs (e2e_libvirt, e2e_docker...etc); 2) ensuring downstream jobs see the same installation configurations (i.e. they will be guaranteed to test the same CAA image).

mkulke:
> If we do that I will just revert this as I need to pass through a specific caa_image for the decoupled flow as I mentioned above.

then let's close this and you add what you need in your PR, so we don't end up in a weird in-between state.

stevenhorsman:
> then let's close this and you add what you need in your PR, so we don't end up in a weird in-between state.

Thanks, my PR has more than enough weird in-between state as it is :)

wainersm:
> > Hi @mkulke @stevenhorsman , we've a duplication here.
> > In [e2e_run_all.yaml](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml), the [Update kustomization configuration](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml#L125) step is meant to update the installation configuration only once; the `src/cloud-api-adaptor/install/` is archived in [Upload install directory for next test runs](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_run_all.yaml#L146) so that downstream provider-specific jobs just download the install directory (eg., see [Get the install directory](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_libvirt.yaml#L84) for e2e_libvirt.yaml ).
> > This way 1) avoiding the duplication of logic on the downstream jobs (e2e_libvirt, e2e_docker...etc); 2) ensuring downstream jobs see the same installation configurations (i.e. they will be guaranteed to test the same CAA image).
> 
> Hi Wainer, In [f3b6c44](https://github.com/confidential-containers/cloud-api-adaptor/commit/f3b6c44bb0ac7cb0efd159d4a776c7f686744f2a) I deliberately added the change to switch away from the install directory. The reason being that in my new world the libvirt test flows are decoupled for each architecture, so I'm not sure if relying on a pre-cached installation directory is helpful (as some runs need a `s390x-dev` image, some need `amd64-dev` etc. Given that we also pass throught `caa_image` as an input, I'm not sure I really understand the need for the install cache in general, but didn't necessarily want to have that debate as part of my PR.
>

Ok, at some point I missed those changes. I think I understood your point.

That install directory caching was made before s390x IIRC, at that time I was only concerned about duplication and avoiding inconsistencies in the testing of different providers. Now that we have architecture in the equation, I think it makes sense to re-visit that logic. Possible removing the install directory caching stuff.

The real obstacle is `install/overlays/libvirt` being arch independent. On possibility is to have an `install/overlays/libvirt/s390x`...
 
> I'm happy to receive feedback on this call though if there is a case to be made for keeping it, but maybe we close this PR and handle it in the wider context of re-working the e2e flows in the mkosi PR?

Yes, it's time for me to go back to your mkosi PR, Steve!

PR #2173, opened 2024-11-28
ibmcloud: make image import.sh work on Mac

frankbu:
Running import.sh on Mac required the following changes:

1. login.sh
    - do not use mapfile
2. multipart_upload.sh
    - reorder args for split command
3. import.sh
   - added platform arg to download for different arch
   - use tr instead of ,, variable expansion

### DIFF
--- a/src/cloud-api-adaptor/ibmcloud/image/import.sh
+++ b/src/cloud-api-adaptor/ibmcloud/image/import.sh
@@ -22,6 +22,7 @@ bucket_region=$region
 instance=
 endpoint=
 api=${IBMCLOUD_API_ENDPOINT-https://cloud.ibm.com}
+platform=
 operating_system=
 
 shift 2
@@ -33,6 +34,7 @@ while (( "$#" )); do
         --region) bucket_region=$2 ;;
         --os) operating_system=$2 ;;
         --api) api=$2 ;;
+        --platform) platform=$2 ;;
         --help) usage; exit 0 ;;
         *)      usage 1>&2; exit 1;;
     esac
@@ -77,7 +79,7 @@ if [ -f ${image_file} ]; then
 else
     # Download image
     echo "Downloading file from image ${image_file}"
-    file=$($script_dir/../../podvm/hack/download-image.sh ${image_file} $script_dir) || error "Unable to download ${image_file}"
+    file=$($script_dir/../../podvm/hack/download-image.sh ${image_file} . --platform "${platform}") || error "Unable to download ${image_file}"
 fi
 
 echo "Uploading file ${file}"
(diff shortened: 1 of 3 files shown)


src/cloud-api-adaptor/ibmcloud/image/login.sh
```diff
@@ -26,7 +26,9 @@ if ! ibmcloud iam oauth-tokens &> /dev/null; then
 fi
 
 required_plugins=(vpc-infrastructure cloud-object-storage)
-mapfile -t installed_plugins < <(ibmcloud plugin list --output json | jq -r '.[].Name')
+# mapfile -t installed_plugins < <(ibmcloud plugin list --output json | jq -r '.[].Name')
+installed_plugins=($(ibmcloud plugin list --output json | jq -r '.[].Name'))
+# ^^^ mapfile is not available on mac and is not needed here anyway
```
stevenhorsman, line 31: Hi Frank, you have remove the comments about mapfile here I think - the code fix should be sufficient

src/cloud-api-adaptor/ibmcloud/image/import.sh
```diff
@@ -101,7 +103,9 @@ if [ -z "$operating_system" ]; then
     image_arch="${image_name##*-}"
     operating_system="ubuntu-20-04-${image_arch/x86_64/amd64}"
 fi
-image_name="$(echo ${image_name,,} | sed 's/\./-/g' | sed 's/_/-/g')"
+# image_name="$(echo ${image_name,,} | sed 's/\./-/g' | sed 's/_/-/g')"
+image_name="$(echo ${image_name} | tr '[:upper:]' '[:lower:]' | sed 's/\./-/g' | sed 's/_/-/g')"
+# ^^^ the ,, parameter expansion is not available on mac, so use tr command instead
```
stevenhorsman, line 108: Same here

src/cloud-api-adaptor/ibmcloud/image/import.sh
```diff
@@ -77,7 +79,7 @@ if [ -f ${image_file} ]; then
 else
     # Download image
     echo "Downloading file from image ${image_file}"
-    file=$($script_dir/../../podvm/hack/download-image.sh ${image_file} $script_dir) || error "Unable to download ${image_file}"
+    file=$($script_dir/../../podvm/hack/download-image.sh ${image_file} . --platform "${platform}") || error "Unable to download ${image_file}"
```
stevenhorsman, line 82: What was the reason for changing the download directory from `$script_dir` to `.`? I'm not sure if this will impact some existing scripts, so would like to understand it more.

stevenhorsman (approved):

src/cloud-api-adaptor/ibmcloud/image/import.sh
frankbu: Oh right, I forgot I changed that. It's a minor improvement. Previously, the script only worked if you ran it in the script dir (i.e., `./import.sh ...`), because the later upload used the current directory to find the image file. With this change, you can run the script from whatever directory you want (e.g., cd ~/work; path/to/import.sh ...`).

src/cloud-api-adaptor/ibmcloud/image/import.sh
frankbu: Sure, I can remove the comments if you think that's best. My logic for adding them was to prevent someone from changing it to the "better" way, not realizing we need to do it this way for portability to Mac.

bpradipt:
looks like the shellcheck related commits needs to be squashed ?

bpradipt (approved):
/lgtm
Thanks @frankbu

PR #2176, opened 2024-12-02
Podnetwork: removal of unused tunneler code and clean up

yoheiueda:
This PR contains the two patches below.

* Remove routing tunneler
This patch removes the old tunneler implementation called "routing". This tunneler was implemented at first, but has been superseded by the "vxlan" tunneler.  The routing tunneler has not been maintained and tested with unit tests and e2e test for a long time.  It was first used in onprem environments, but it turned out that there were some limitations on public clouds.  So, it is safe to remove this tunneler now.

* Refactor pod networking
This patch refactors the code for pod networking to support a new tunneler implementations better modularity.

This PR is a prerequisite of https://github.com/confidential-containers/cloud-api-adaptor/issues/688


src/cloud-api-adaptor/pkg/podnetwork/tunneler/config.go
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package tunneler
+
+type TunnelerConfigurator interface {
+	Tunneler
+	Configure(*NetworkConfig, *Config) error
+}
+
+type NetworkConfig struct {
```
stevenhorsman, line 11: What are your thoughts about making this extensible for other potential Tunnel configs? Is this a VXLAN specific implementation of NetworkConfig as it has the `VXLANConfig` as a field, if so should it be renamed to something like `VXLANNetworkConfig`. I'm guessing not as you have split out the VXLANConfig into a separate structs?

stevenhorsman (commented):
Generally it looks good to me, just one question about naming and how that plays into future plans.

src/cloud-api-adaptor/pkg/podnetwork/tunneler/config.go
mkulke: yup, I was wondering the same. should there be a tunneling interface and the VXLAN specifics hidden behind it?

mkulke: yup, makes sense if we map cli params directly into config structs.

mkulke (approved):

PR #2178, opened 2024-12-03
csi: Fix AWS example for CSI wrapper

bpradipt:
Updating the csi driver to latest 1.37 release and RBAC.
Additionally the `peerpod` key is mandaotry in the storageclass definition, without which peerpodVolume CRD will not be created.

Fixes: #2144


yoheiueda (approved):
LGTM. Thank you very much!

snir911 (approved):
Nice, Thanks!

PR #2181, opened 2024-12-04
azure: Install qemu-utils

stevenhorsman:
Now the qcow2 build is part of the mkosi make image target, we need to install `qemu-utils`, so that the `qemu-img` tool is available


huoqifeng (approved):
/lgtm

PR #2185, opened 2024-12-05
workflow: Temporarily skip the s390x e2e tests

stevenhorsman:
There are some potential permissions issues on the s390x runners, so skip these tests until Choi is online and available to help debug them.


stevenhorsman:
Unfortunately I can't test this on my fork as it required an s390x runner to get to the point that it will then skip the test.

mkulke (approved):

wainersm (approved):

PR #2186, opened 2024-12-05
podvm: truncate initdata digest to 32 bytes on az

mkulke:
~depends on #2187 to be merged first~

According to initdata spec the digest needs to be truncated/padded according to the requirements of the TEE. for az tpm we use the sha256 bank of TPM for initdata.

This will fix a bug when a initdata body with alg=sha384+ was used and the PCR8 value in the TEE evidence will not be extended, since you cannot extend sha256 w/ a digest that's bigger than 32 bytes.

An e2e test for azure was added to assert this behaviour.


bpradipt (approved):
/lgtm

stevenhorsman (approved):
This looks reasonable to me.

PR #2187, opened 2024-12-05
e2e-test: Add curl-jq test image

mkulke:
This image is required to perform remote attestation tests which use api-server-rest. It will be built and pushed by the existing CI.


stevenhorsman (approved):
LGTM. Thanks! Let's hope the CI test image build is working as I don't think we've added to it in about 18 months!

bpradipt (approved):
/lgtm

PR #2191, opened 2024-12-05
 workflow: Actually temporarily skip the s390x e2e tests

stevenhorsman:
I realised too late that adding a `false` to a list of or's does nothing. I wanted to preserve the code, but we can just revert, so it was pointless. Sorry!

### DIFF
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -266,12 +266,7 @@ jobs:
   libvirt_s390x:
     name: E2E tests on libvirt for the s390x architecture
     # Skip s390x e2e tests until Choi is available to set-up the s390x runner's pre-action properly. Then revert this.
-    if: |
-      false ||
-      github.event_name == 'schedule' ||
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_run_all.yaml
```diff
@@ -266,12 +266,7 @@ jobs:
   libvirt_s390x:
     name: E2E tests on libvirt for the s390x architecture
     # Skip s390x e2e tests until Choi is available to set-up the s390x runner's pre-action properly. Then revert this.
-    if: |
```
wainersm, line 269: my bad, I reviewed and merged the previous PR without realizing that.

Here you want use 'false &&' to preserve the entire block. Whatever you prefer, I will ack regardless.

wainersm (approved):

.github/workflows/e2e_run_all.yaml
stevenhorsman: Yeah, `false &&` was what I should have done originally, but I'm not taking any chance this time!

mkulke (approved):

PR #2195, opened 2024-12-06
podvm: mount cloud-config disk via ExecStartPre

mkulke:
With the introduction of a mount unit for the config disk we introduced a boot-time regression (+45s) in cases where there is not config disk available.

Moved the mount to a can-fail ExecStartPre statement on the process-user-data service unit instead.


stevenhorsman (approved):
LGTM. Thanks!

bpradipt (approved):
/lgtm

PR #2197, opened 2024-12-06
podvm: remove sudo if not required in Makefile

mkulke:
Some sudos where added which are not required. We shouldn't force sudo unless it is required to not mess up local dev environments.


stevenhorsman (approved):
Thanks - this makes sense. Early on in the native runner work we had lots of permissions issues, so Choi added a bunch of sudo's in a commit to fix them, but in response to your feedback I reverted it and they were all good, but I didn't realise that I had continued that trend do to thing being sudo in other commits, so sorry about that.

wainersm (approved):
Thanks @mkulke !

wainersm:
@mkulke ,

still needing super-privileges? `PermissionError: [Errno 1] Operation not permitted`

mkulke:
> @mkulke ,
> 
> still needing super-privileges? `PermissionError: [Errno 1] Operation not permitted`

yup, there is probably a problem with an earlier step being run as root. let's see...

stevenhorsman:
Ah, maybe that mkosi error we are seeing it what I saw and why I added `sudo` in the first place?

mkulke:
gah, ubuntu 24.04 [blocks userns by default](https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces), so you cannot build rootless unless you do: 

```
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```

this will not get picked up however in this PR, so we need #2199 with just this fix 😭

wainersm:
https://github.com/confidential-containers/cloud-api-adaptor/pull/2199 is merged now \o/

PR #2198, opened 2024-12-06
workflows: mksoi: Remove mid-build caching

stevenhorsman:
Remove the push of the builder and binaries during the mkosi build, but just output them to the local directory always to simplify the flow and remove the need to use crane to pull them down.

### DIFF
--- a/.github/workflows/podvm_mkosi.yaml
+++ b/.github/workflows/podvm_mkosi.yaml
@@ -114,15 +114,7 @@ jobs:
 
       - name: Read properties from versions.yaml
         run: |
-          go_version="$(yq '.tools.golang' versions.yaml)"
-          echo "GO_VERSION=${go_version}" >> "$GITHUB_ENV"
           echo "ORAS_VERSION=$(yq -e '.tools.oras' versions.yaml)" >> "$GITHUB_ENV"
-
-      - name: Setup Golang version ${{ env.GO_VERSION }}
-        if: ${{ inputs.arch == 's390x' }}
-        uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - uses: oras-project/setup-oras@v1
         with:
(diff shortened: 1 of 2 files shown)


stevenhorsman:
The e2e tests works with this code in my fork: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12202102693/job/34042019318 so I think it's okay to go.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -15,10 +15,6 @@ PODVM_IMAGE ?= $(REGISTRY)/podvm-generic-$(PODVM_DISTRO)$(if $(filter $(SE_BOOT)
 PODVM_CONTAINER_NAME ?= $(REGISTRY)/podvm-docker-image-$(ARCH)
 VERIFY_PROVENANCE ?= no
 
-PUSH ?= false
```
wainersm, line 18: I suggest to leave this makefile toggle. We never know, in future we might decide to push the builder/binaries. Or developers can still use it to easily push and share with another dev.

.github/workflows/podvm_mkosi.yaml
```diff
@@ -192,39 +159,6 @@ jobs:
         if: ${{ inputs.arch == 'amd64' }}
         run: nix build .#devShells.x86_64-linux.podvm-mkosi
 
-      - name: Install crane with nix
-        if: ${{ inputs.arch == 'amd64' }}
-        run: nix profile install nixpkgs#crane
-
-      - name: Install crane with go
```
wainersm, line 199: I think we can remove `Setup Golang version ${{ env.GO_VERSION }}` step as it was only needed for building crane? Consequently remove the line were it gets the go version from versions.yaml.

src/cloud-api-adaptor/podvm-mkosi/Makefile
mkulke: wouldn't a developer who is interested in this just push the image instead of using this toggle? I would generally lean towards removal of dead code, even if it could be useful [in the future](https://martinfowler.com/bliki/Yagni.html).


toggles like this incur a cost. coginitive burden for people who want to work on those build-scripts and often we end up propagating them in Dockerfile build-args/envs, workflow inputs, ... (cost of carry)

If we have a use case it should be trivial to re-add.

.github/workflows/podvm_mkosi.yaml
stevenhorsman: Good spot; updated now. Thanks!

stevenhorsman:
I've tested this on my fork: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/12232882664/job/34119638806 and apart from s390x (which I don't have a runner for, so can't test), it is all passing apart from a single test on cri-o, which I don't think it related to this.

src/cloud-api-adaptor/podvm-mkosi/Makefile
wainersm: Good points. I'm convinced this may incur in cost unnecessarily. Thanks @mkulke !

wainersm (approved):
LGTM. Thanks @stevenhorsman !

PR #2199, opened 2024-12-06
podvm-ci: un-restrict user ns on ubuntu 24.04

mkulke:
There is a restriction introduced in new revisions of ubuntu runners. This change allows user ns to be created, so mkosi can be built rootless.


stevenhorsman (approved):
LGTM. Thanks for hunting this down!

wainersm (approved):

PR #2201, opened 2024-12-09
caa: remove option to configure agent via userdata

mkulke:
The agent doesn't require dynamic configuration any more. since userdata is untrusted input, we can reduce the surface and package a static configuration.

In this change image auth support is always enabled. An empty auth file is provisioned by default to /run/peerpod/auth.json, it can be overwritten by userdata.


bpradipt (approved):
/lgtm

stevenhorsman (approved):
The makes sense to me. Thanks!

PR #2203, opened 2024-12-11
S390x runner testing

stevenhorsman:
Revert the changes that skipped s390x testing so we can debug the runners now that @BbolroC is back to help us.
Unfortunately we can't test it on fork runners as in a fresh environment it all passed, so we specifically need to ensure that the runners shared with other CoCo projects are cleaned up to run our workfloads


BbolroC (approved):
LGTM, let's debug anything on the runner! Thanks @stevenhorsman !

mkulke (approved):

PR #2206, opened 2024-12-11
workflows: Switch s390x e2e runner

stevenhorsman:
The get stable libvirt e2e tests we need to just use the large runner, so switch the runner label over


mkulke (approved):

BbolroC (approved):
LGTM, thanks @stevenhorsman !

PR #2209, opened 2024-12-12
ci: re-order sysctl step for podvm_mkosi wf

mkulke:
A default directory is set workflow-wide that will only work after checkout, hence the sysctl call fails.


wainersm (approved):
thanks @mkulke !

PR #2211, opened 2024-12-16
libvirt: Fix shell reload

stevenhorsman:
I experimented with moving the shell reload straight after the pipx ensurepath when discussing this with Magnus, but in the GHA once merged this doesn't seem to be working as the reload means the script ends, so we need to move this back.

### DIFF
--- a/src/cloud-api-adaptor/libvirt/config_libvirt.sh
+++ b/src/cloud-api-adaptor/libvirt/config_libvirt.sh
@@ -86,8 +86,6 @@ installKcli() {
             # export PATH="$PATH:$HOME/.local/bin"
             pipx install kcli==${kcli_version}
             pipx ensurepath
-            # Reload shell so that pipx install PATH is available
-            exec $SHELL
         else
             sudo pip3 install kcli==${kcli_version}
         fi
@@ -140,3 +138,8 @@ case $TEST_E2E_SECURE_COMMS in
     echo "SECURE_COMMS=\"false\"" >> libvirt.properties
     ;;
 esac
+
+if [[ "${OS_DISTRO}" == "ubuntu" ]] && [[ "${CI:-}" != "true" ]]; then
+    # Reload shell so that pipx install PATH is available
+    exec $SHELL
+fi


mkulke:
I thought we concluded that launching a shell at the end of a script will make some sense on a local developer box, but if invoked in GHA, in a step without a tty, the shell would just immediately exit? 

can you point to the failed run maybe?

stevenhorsman:
The latest daily e2e test is the failed run: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/12356330516 and it's passing in this one. Sorry, I thought I'd reverted this after our discussion about it. I should have re-checked before I merged.

mkulke (approved):

src/cloud-api-adaptor/libvirt/config_libvirt.sh
```diff
@@ -86,8 +86,6 @@ installKcli() {
             # export PATH="$PATH:$HOME/.local/bin"
             pipx install kcli==${kcli_version}
             pipx ensurepath
-            # Reload shell so that pipx install PATH is available
```
wainersm, line 89: Running `exec $SHEL` here on gha will definitively stop here and do not run important setup steps (like creating the ssh key-pair).

Sorry, I missed the first discussion about that exec... would make sense to append `$HOME/.local/bin` to system's PATH if in gha action as shown in `https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-system-path`

stevenhorsman: I did that in the previous PR: https://github.com/confidential-containers/cloud-api-adaptor/blob/8f534e3e27f81d4794d8bd548a276a67fad99773/.github/workflows/e2e_libvirt.yaml#L125

mkulke: but yes, @wainersm is right. if we pivot to a new shell mid-script, the rest of the script will be discarded in any case (locally or on the CI). There are a gazillion ways to update the path, but we should probably focus on unblocking the CI.

one of those options would be to have pipx installed in an outside step, the next steps starts a new shell and should have the path set. 

```yaml
name: install pipx
if: ... # onyl on ubuntu
run: |     
  sudo apt-get update
  sudo DEBIAN_FRONTEND=noninteractive apt-get  install pipx -y
  pipx ensurepath
name: run config-libvirt
run: ...  
```

stevenhorsman: I definitely understand the approach is probably nicer, but I would prefer to not move pipx in a separate step as I'd like to move more towards moving things from the workflow into shell scripts for easier PR testing than the opposite.

I've tried adding a new commit with `"${CI}" != "true"` which should mean we don't run the reload at all on GHA. Hopefully that is sufficient to get this PR merged and unblock the CI and I'm happy to discuss further improvements once we are back green again.

mkulke: I think the original approach of opening a new shell at the end of the script should at least do no harm. me, as an interactive user of a script, won't like this, but that's maybe idiosyncrasy

src/cloud-api-adaptor/libvirt/config_libvirt.sh
```diff
@@ -140,3 +138,8 @@ case $TEST_E2E_SECURE_COMMS in
     echo "SECURE_COMMS=\"false\"" >> libvirt.properties
     ;;
 esac
+
+if [[ "${OS_DISTRO}" == "ubuntu" ]] && [[ "${CI}" != "true" ]]; then
```
wainersm, line 142: Hi @stevenhorsman ,

One last fix, use `${CI:-}` to make `set -o nounset` happy.

stevenhorsman: It is interesting that the GHA shellcheck didn't pick up on that, maybe that's something we need to be aware of

wainersm: hmmm... are we shell-checking the scripts on workflows or all scripts?

Issue #1958, opened 2024-07-28
In-guest storage for container images


bpradipt:
With peer pods, we currently have no way to share image layers downloaded on the worker node. Consequently, we only support the in-guest image pull approach. This creates some challenges regarding container rootfs storage inside the guest. The challenge has increased manifold with the size of AI runtime and model images running into multiple GBs (10+).

Let's understand how we handle container rootfs storage inside the guest.

For packer-created guest (pod VM) images, the container rootfs storage is the guest's root disk. We mount the `/kata-containers` folder on the guest's root disk to `/run/kata-containers/image` during boot. 
The container image layers are downloaded and extracted under `/run/kata-containers/image/layers`. The overlay (rw) layer is under `/run/kata-containers/image/overlay`

If using an unencrypted root disk, this approach violates the CoCo threat model, as a privileged infra admin can easily tamper with the root disk contents or read the contents of the container image (which you don't want when using an encrypted container image).

For mkosi-created guest images, the guest root disk is read-only and dm-verity protected. The container rootfs storage is the guest memory. Using memory for container rootfs is safest as the memory is encrypted. However this approach introduces a new problem. Guest memory is limited, and downloading and extracting huge container images in memory may not be the best use of the available memory. Also, there is no easy way to [automatically determine the guest memory](https://github.com/confidential-containers/cloud-api-adaptor/issues/1921) requirement based on the container images used in the pod specs.

So how can we improve the container rootfs storage handling in peer-pods? This issue is to share some ideas and kickstart the discussion.

Proposal-1
- Create a LUKS encrypted guest root disk partition with a dynamically generated key during boot 
- Use the LUKS encrypted guest root disk partition as the container rootfs store

Recently, CDH introduced support for encrypted (ephemeral) block devices. We can explore whether it will be feasible to reuse this: https://github.com/ChengyuZhu6/guest-components/blob/main/confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md.

Proposal-2
- For extremely large  (unencrypted) container images, embed the layers in the guest root disk. Ensure the root disk is ro and dm-verity protected
- Mount the `overlay` layer as a tmpfs volume during boot so any writes are on the memory.

cc @mkulke @stevenhorsman @genjuro214 @huoqifeng @snir911

mkulke:
I have a POC for Proposal 1 working.

[...]

PR #2212, opened 2024-12-16
Add encrypted scratch space for container sandbox

### LINKED ISSUE #1921: podvm-mkosi: need allocate more space for "/run" to store large image data
I attempted to run e2e tests on a **Fedora** s390x Pod VM image built using **mkosi**, and found the case `TestCreatePeerPodWithLargeImage` fails on the Fedora image but succeeds on an **Ubuntu** image built using **Packer**.

The root cause is there is no enough space to store the image data.

The large test image exceeds `2GB` in size and is pulled into the directory "/run/kata-containers/image/layers". However, the mount point "/run" is only allocated `1.6GB` of space via tmpfs:
```
# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/vdb2        98G  7.2G   86G   8% /
devtmpfs        4.0M     0  4.0M   0% /dev
tmpfs           4.0G     0  4.0G   0% /dev/shm
tmpfs           1.6G  162M  1.5G  11% /run
tmpfs           4.0G  4.0K  4.0G   1% /tmp
/dev/vdb1       223M   98M  108M  48% /boot
tmpfs           802M  4.0K  802M   1% /run/user/0
```

So it run out of space when pulling an image larger than `1.6GB`.

There is no such problem for Ubuntu image built using Packer, because the mount service "[run-kata\x2dcontainers.mount](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/files/etc/systemd/system/run-kata%5Cx2dcontainers.mount)" will mount "/run/kata-containers" to hard disk "/kata-containers".

However, this mount service was removed for **mkosi**. I can find the related PR and commit:
https://github.com/confidential-containers/cloud-api-adaptor/pull/1606
https://github.com/confidential-containers/cloud-api-adaptor/commit/c6ac908799244fb7b7874047e71a61eba48d8a29

Since "/" should be read only, should we allocate more space to "/run" for runtime data?

Based on my investigation, Fedora typically allocates 20% of physical RAM to "/run" via tmpfs by default. We can increase this allocation to `50%` or more in `/etc/fstab`:
```
tmpfs           /tmp            tmpfs   rw,nosuid,nodev,size=50%   0 0
```

Moreover, we can allocate an even larger size than physical RAM by enabling **swap**. However, we must prepare the swap file on the hard disk in advance.

I hope we can discuss the following options:

1. increase size percentage to `50%` or `75%`
2. enable swap and allocate a larger size.
3. keep the default size for "/run" but reduce the size of the large test image.

mkulke:
fixes #1921

This is a PoC to address the problem of secure scratch space in which we can download and unpack images to the sandbox, write files at runtime etc.

Currently this is either backed by storage that is not protected by the TEE (packer) or as tmpfs mounts that will consume ram during runtime (mkosi). For larger images like pytorch w/ cuda this implies in theory that we have to set aside 8GB of ram just to be able to download and unpack the image. tmpfs mounts are by default limited to a ratio of the ram, so we might end up having to overprovision the machine to be able to run a workload.

The proposes change introduces a CAA-wide parameter that allows to specify the disk size. A cloud provider can use this as an indicator of the size of the to-be-provisioned podvm disk. A zero value means that we stick to the default size of the image.

if a size has been given (the mkosi image is ~0.5gb) it will extend the image to that size. At launch the podvm will claim that space and create an ad-hoc encrypted volume on it. Depending on the presence of a marker file in user-data the agent unit will mount that to `/run/kata-containers` prior to launching kata-agent.

I'm not super happy about the API (I would prefer to specify it per pod) and the implementation (the marker file) but I wanted to push early to get feedback. (also, it's only implemented for azure at the moment)

FWIW it does what it's supposed to do, see some numbers on #1958


src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -64,6 +64,8 @@ type ServerConfig struct {
 	SecureCommsPpOutbounds  string
 	SecureCommsKbsAddress   string
 	PeerPodsLimitPerNode    int
+	UseEncryptedDisk        bool
```
mkulke, line 67: leftover

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -30,6 +30,7 @@ optionals+=""
 [[ "${SECURE_COMMS_PP_OUTBOUNDS}" ]] && optionals+="-secure-comms-pp-outbounds ${SECURE_COMMS_PP_OUTBOUNDS} "
 [[ "${SECURE_COMMS_KBS_ADDR}" ]] && optionals+="-secure-comms-kbs ${SECURE_COMMS_KBS_ADDR} "
 [[ "${PEERPODS_LIMIT_PER_NODE}" ]] && optionals+="-peerpods-limit-per-node ${PEERPODS_LIMIT_PER_NODE} "
+[[ "${DISK_SIZE}" ]] && optionals+="-disk-size ${DISK_SIZE} "
```
bpradipt, line 33: For AWS provider we use a param "ROOT_VOLUME_SIZE" as a way to expand the VM root volume size if requested explicitly. May be the same can be reused? Or we can settle on another name for all the providers?

src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/repart.d/30-scratch.conf
```diff
@@ -0,0 +1,5 @@
+[Partition]
+Type=linux-generic
+Label=scratch
+Encrypt=key-file
+Format=ext4
```
bpradipt, line 5: @mkulke can you please explain how the size of the partition gets calculated ? Is it automatically done by systemd?

bpradipt:
> 
> if a size has been given (the mkosi image is ~0.5gb) it will extend the image to that size. At launch the podvm will claim that space and create an ad-hoc encrypted volume on it. Depending on the presence of a marker file in user-data the agent unit will mount that to `/run/kata-containers` prior to launching kata-agent.
> 

Would it also make sense to extend this approach to use add-on disk like available with some instance types ? Not needed for initial iteration, but is this something we should consider and lay the foundation now?

> I'm not super happy about the API (I would prefer to specify it per pod) and the implementation (the marker file) but I wanted to push early to get feedback. (also, it's only implemented for azure at the moment)
> 

I think it would be better to also add an annotation (eg `io.katacontainers.config.hypervisor.root_disk_size`) in kata for remote hypervisor to use for per-pod scratch space. Similar to what we use for other such annotations.

src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/repart.d/30-scratch.conf
mkulke: by default, it grabs as much as it can get on the disk.

if the size of the disk is 1GB it's about 0.5gb
if the size of the disk is 10gb it's about 9.5gb

src/cloud-api-adaptor/entrypoint.sh
mkulke: ROOT_VOLUME_SIZE would be good IMO

mkulke:
> Would it also make sense to extend this approach to use add-on disk like available with some instance types ? Not needed for initial iteration, but is this something we should consider and lay the foundation now?

for future extensibility, I think we can freely iterate. We don't currently have a version contract between PodVM and CAA. We would have to look at the user-facing api maybe to make this possible.

> I think it would be better to also add an annotation (eg `io.katacontainers.config.hypervisor.root_disk_size`) in kata for remote hypervisor to use for per-pod scratch space. Similar to what we use for other such annotations.

I agree, I didn't spot any existing fitting annotation in kata at first glance, so we might have to add it.

src/cloud-api-adaptor/entrypoint.sh
mkulke: should it be made a generic param or copied to azure?

bpradipt: Afaik it will not be straight forward to implement disk resize for all the providers. So I think copying to azure should be ok for now.

bpradipt:
> > I think it would be better to also add an annotation (eg `io.katacontainers.config.hypervisor.root_disk_size`) in kata for remote hypervisor to use for per-pod scratch space. Similar to what we use for other such annotations.
> 
> I agree, I didn't spot any existing fitting annotation in kata at first glance, so we might have to add it.

Yes, a new annotation will need to be added.

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -125,7 +125,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		flags.StringVar(&cfg.serverConfig.Initdata, "initdata", "", "Default initdata for all Pods")
 		flags.BoolVar(&cfg.serverConfig.EnableCloudConfigVerify, "cloud-config-verify", false, "Enable cloud config verify - should use it for production")
 		flags.IntVar(&cfg.serverConfig.PeerPodsLimitPerNode, "peerpods-limit-per-node", 10, "peer pods limit per node (default=10)")
-		flags.Uint64Var(&cfg.serverConfig.DiskSize, "disk-size", 0, "Disk size in GB. Default is 0, which implies the default image disk size")
+		flags.Uint64Var(&cfg.serverConfig.RootVolumeSize, "root-volume-size", 0, "Root volume size in GB. Default is 0, which implies the default image disk size")
```
bpradipt, line 128: Looks like these changes need to be made to the commit "caa: add scratch-space logic"

mkulke:
> Yes, a new annotation will need to be added.

We discussed it in the last community call and concluded to start with a caa-wide option. once this is merged, we can start add something to kata-runtime to make it tweakable per-pod

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -76,6 +76,7 @@ else
 	touch resources/buildBootableImage
 	nix develop ..#podvm-mkosi --command mkosi --environment=VARIANT_ID=production
 	qemu-img convert -f raw -O qcow2 build/system.raw build/podvm-$(PODVM_DISTRO)-$(ARCH).qcow2
+	qemu-img resize build/podvm-$(PODVM_DISTRO)-$(ARCH).qcow2 +100M
```
stevenhorsman, line 79: Is there any significance to +100M, or is it just a starter amount of space?

stevenhorsman (commented):
I think commits 1 & 2 need some tidying as there is new code added in 1 that is changed in 2 which was a bit confusing. Additionally it would have been nicer (for me to review) to split the resource refactor out from commit 1, into a previous commit. It's not a blocker, maybe adding a sentence to explain that in the commit message #1 would be helpful for future people looking back at this?

src/cloud-api-adaptor/pkg/util/cloud.go
```diff
@@ -44,48 +45,48 @@ func GetImageFromAnnotation(annotations map[string]string) string {
...
+			gpus = 0
 		}
 	} else {
-		gpuInt = 0
+		gpus = 0
 	}
 
-	// Return vCPU, memory and GPU
-	return vcpuInt, memoryInt, gpuInt
+	storage := int64(0)
```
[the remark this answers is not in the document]
mkulke: yup, since the annotation will not yield a value at the moment, 0 is probably the sensible choice, here.

src/cloud-api-adaptor/podvm-mkosi/Makefile
mkulke: it needs to have a bit of size, so repart will not stumble over it at launch, attempting to create a partition without space. not sure if needs to be +100m, could just also work with +10m, have to test.

src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go
```diff
@@ -221,17 +223,20 @@ func (s *cloudService) CreateVM(ctx context.Context, req *pb.CreateVMRequest) (r
 	instanceType := util.GetInstanceTypeFromAnnotation(req.Annotations)
 
 	// Get Pod VM cpu and memory from annotations
-	vcpus, memory, gpus := util.GetPodvmResourcesFromAnnotation(req.Annotations)
+	resources := util.GetPodVMResourcesFromAnnotation(req.Annotations)
+
+	// Add disk size to resources if set
+	if s.serverConfig.DiskSize > 0 {
+		resources.Storage = int64(s.serverConfig.DiskSize)
+	}
```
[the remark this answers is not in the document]
mkulke: makes sense

src/cloud-api-adaptor/podvm-mkosi/Makefile
stevenhorsman: Ok, I don't have an issue with the 100M, just wanted to check if there was special logic behind it.

stevenhorsman: This is only being done on the x86 path here as only Azure is supporting this scratch space at the moment I assume? Do you see any obstacles in the code for s390x support if/when needed?

mkulke: I don't know much about partitioning schemes on s390x. if systemd-repart works in the same way, it should work ootb. The cloud-provider code would need to consider the `.storage` property when provisioning a VM and then repart would grab whatever it available to allocate it to the scratch space.

bpradipt (approved):
/lgtm

bpradipt:
@mkulke reopening this one..

bpradipt:
@mkulke do you mind rebasing this PR? 
I think we can merge this and iterate.

mkulke:
> @mkulke do you mind rebasing this PR? I think we can merge this and iterate.

will take a look, maybe we need to discuss it at a community call again, I'm not sure what the state of supporting an annotation on the kata side is, atm.

bpradipt:
> > @mkulke do you mind rebasing this PR? I think we can merge this and iterate.
> 
> will take a look, maybe we need to discuss it at a community call again, I'm not sure what the state of supporting an annotation on the kata side is, atm.

We don't have the per pod annotation in kata yet. So we can use it with global config and add the per-pod annotation later on. Probably collating any other per-pod config that we want to expose via kata annotation

mkulke:
@bpradipt in yesterday's coco community call this topic was discussed briefly. it was indicated that this use case is already covered by the encrypted volume feature of CDH (afair). We probably want to avoid adding a redundant peerpod-specific implementation, since the problem should be the same on bare-metal (pulling large images on the guest requires a lot of ram-backed tmpfs).

@fitzthum can you confirm that and maybe provide a link to a doc/pr so we can link it here for reference?

fitzthum:
The docs for secure mount area iittle bit sketchy, but the basic interface is [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/storage/volume_type/mod.rs#L44-L52) and [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/protos/api.proto#L25-L30). It's basically just a utility for mounting stuff with logic for LUKS built in. It isn't responsible for mounting anything into a container, so it can be used for volumes that are attached to a container or things that are exposed at the VM level. 

I do think the storage interfaces could be improved a little bit. For instance we might want to establish a layer abstraction so that we can mount the block device generically and then do LUKs stuff on top of that. Today those two things are coupled together in one storage implementation. Nonetheless this feature might be useful for your purposes as well.

fitzthum:
Yeah, it is using a bash script which is actually packaged via kata. See [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/storage/volume_type/blockdevice/luks/mod.rs#L17). I'm not a fan of this but still would be good to have a unified approach. I think we do support ephemeral block volumes for pulling images on bare metal using this approach, although the feature is not well-documented.

bpradipt:
> ok, I remember seeing a shell script somewhere. If CDH is invoking this one, then it is probably doing what crypttab + systemd do under the hood. I would generally recommend not to handroll such things, but if it's already there and being used, we shouldn't add new method.
> 
> @bpradipt can you confirm that this covers the use case of pulling large images? I have no idea how this works practically, but I suspect there is some local block device that is passed to the kata vm on bare metal. for CAA that would need to be translated into a disk that is attached to the vm.

I haven't used the ephemeral block device on CoCo baremetal. I need to see if it really works. Will try next week. 
Although, I have used the approach in this PR for CoCo peer-pods to pull large images.

mkulke:
converting this to draft for now, until we understand whether this usecase is covered by CDH

mkulke:
closed in favor of #2486

PR #2215, opened 2024-12-18
podvm_mkosi: Remove remaining references to builder

stevenhorsman:
Sorry, I didn't do the full job when merging the builder and binaries phase, so left some references to
fedora-binaries-builder in the code and docs (I was going to run both in parallel, but decided it
made more sense to switch and only did a partial job). This should fix the azure nightly build (and we still need to work
on integrating that with the other workflows I guess).


mkulke (approved):
thx. was about to look at that!

bpradipt (approved):
/lgtm

PR #2216, opened 2024-12-18
test/e2e: Add debug of failed pods

stevenhorsman:
Print logs of failed pods and CAA in case of failures to help with debugging in the case of failures

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
+++ b/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
@@ -225,6 +225,26 @@ func GetPodEventWarningDescriptions(ctx context.Context, client klient.Client, p
 	return descriptionsBuilder.String(), nil
 }
 
+func LogPodDebugInfo(ctx context.Context, t *testing.T, client klient.Client, pod *v1.Pod) {
+	podLogString, err := GetPodLog(ctx, client, pod)
+	if err != nil {
+		t.Error(err)
+	}
+	t.Logf("Pod log: %s\n", podLogString)
+
+	events, err := GetPodEventWarningDescriptions(ctx, client, pod)
+	if err != nil {
+		t.Error(err)
+	}
+	t.Logf("Pod error events: %s\n", events)
+
+	caaPodLog, err := getCaaPodLogForPod(ctx, t, client, pod)
+	if err != nil {
+		t.Error(err)
+	}
+	t.Logf("CAA log: %s\n", caaPodLog)
+}
+
 // This function takes an expected pod event "warning" string (note warning also covers errors) and checks to see if it
 // shows up in the event log of the pod. Some pods error in failed state, so can be immediately checks, others fail
 // in waiting state (e.g. ImagePullBackoff errors), so we need to poll for errors showing up on these pods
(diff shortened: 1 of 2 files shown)


stevenhorsman:
Typically - when I want tests to fail they all pass! Re-running...

stevenhorsman:
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/12415908135/job/34663884546?pr=2216 shows an example (with deliberately failing tests of it selecting the correct bit of the CAA log related to the pod that fails.

stevenhorsman:
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/12417178259/job/34668109916?pr=2216 shows that the CAA log is now trimmed to only show the messages related to the pod that failed.

mkulke:
can you maybe screen shot an example? I fail to see spot anything in the debug step output, but maybe I'm looking in the wrong place

stevenhorsman:
> can you maybe screen shot an example? I fail to see spot anything in the debug step output, but maybe I'm looking in the wrong place

Sorry, it's not in the debug step, but inline with the tests (I'm not sure there is a way to separate and report later, as by the time we debug we've lost the specifics. As an example you can search the test output for `--- FAIL: TestLibvirtCreatePeerPodAndCheckWorkDirLogs (40.12s)` and see that just before before it we have:
image
which is the pod log, the pod events that were in warning (none in this case as I just changed the expected output to force an error) and the CAA logs starting with the request to create the sandbox for the pod

mkulke:
> > can you maybe screen shot an example? I fail to see spot anything in the debug step output, but maybe I'm looking in the wrong place
> 
> Sorry, it's not in the debug step, but inline with the tests (I'm not sure there is a way to separate and report later, as by the time we debug we've lost the specifics)

I guess the logs can be written to file in a ./fail folder to be printed out in the debug step, but having them in the test output also works

mkulke (approved):

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -370,25 +390,36 @@ func GetNodeNameFromPod(ctx context.Context, client klient.Client, customPod *v1
 	return getStringFromPod(ctx, client, customPod, getNodeName)
 }
 
+func GetPodsFromJob(ctx context.Context, t *testing.T, client klient.Client, job *batchv1.Job) (*v1.PodList, error) {
+	clientset, err := kubernetes.NewForConfig(client.RESTConfig())
+	if err != nil {
+		return nil, fmt.Errorf("GetPodFromJob: get Kubernetes clientSet failed: %v", err)
```
wainersm, line 396: Hi @stevenhorsman !

s/GetPodFromJob/GetPodsFromJob/

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -370,25 +390,36 @@ func GetNodeNameFromPod(ctx context.Context, client klient.Client, customPod *v1
...
+	}
+
+	pods, err := clientset.CoreV1().Pods(job.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "job-name=" + job.Name})
+	if err != nil {
+		return nil, fmt.Errorf("GetPodFromJob: get pod list failed: %v", err)
```
wainersm, line 401: same here

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: This returns a PodList, so thats why I called it GetPods as it could return multiple

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -428,6 +459,32 @@ func GetSuccessfulAndErroredPods(ctx context.Context, t *testing.T, client klien
...
+	if err != nil {
+		return "", fmt.Errorf("GetCaaPodLog: failed to getCaaPod: %v", err)
+	}
+	podLogString, err := getStringFromPod(ctx, client, caaPod, GetPodLog)
+	if err != nil {
+		return "", fmt.Errorf("GetCaaPodLog: failed to getStringFromPod: %v", err)
+	}
+
+	// Find the logs starting with the pod
+	pod_matcher := regexp.MustCompile(`[0-9]{4}\/[0-9]{2}\/[0-9]{2} ([0-1]?[0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9] \[adaptor\/cloud\] create a sandbox [0-9|a-f]* for pod ` + pod.Name)
```
wainersm, line 477: This regex is pure evil :)

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: sorry, I didn't look at the line properly!

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: good point, I'll split it up to be clearer

wainersm: I think it's fine. It just caught my attention, I used to enjoy these crazy regex :)

wainersm (approved):
Thanks @stevenhorsman !

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: Should be fixed

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
stevenhorsman: I've provided an example of what we are trying to match against and broken out the date and time matcher, so I think it should be more understandable now. Just re-testing with a deliberate error again as regex is prone to failures and I don't test my manual testing as I'm on a run of mistakes!

stevenhorsman:
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/12434706844/job/34719400000?pr=2216 shows that the CAA pod specific logging is still working, so my regex looks okay

PR #2217, opened 2024-12-18
S390x encrypted image test

stevenhorsman:


stevenhorsman:
The test of just truncating the decryption key to the first 32-bytes didn't work - it was a long shot, so I'm not too surprised

PR #2222, opened 2024-12-23
docker: Fix userdata bind mount path inside the container

bpradipt:
For docker provider we use systemd container which mounts /run as a tmpfs volume.
This mounting of /run happens after docker has bind mounted the userdata file inside the container and consequently the userdata file is no longer accessible.

This commit changes the path of the userdata file to not use `/run` for the docker provider.

### DIFF
--- a/src/cloud-api-adaptor/pkg/paths/paths.go
+++ b/src/cloud-api-adaptor/pkg/paths/paths.go
@@ -7,5 +7,5 @@ const (
 	InitDataPath     = "/run/peerpod/initdata"
 	AgentCfgPath     = "/run/peerpod/agent-config.toml"
 	ForwarderCfgPath = "/run/peerpod/daemon.json"
-	UserDataPath     = "/run/media/cidata/user-data"
+	UserDataPath     = "/media/cidata/user-data"
 )
--- a/src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/process-user-data.service.d/10-override.conf
+++ b/src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/process-user-data.service.d/10-override.conf
@@ -1,6 +1,5 @@
 [Service]
 # mount config disk if available
-ExecStartPre=-/bin/mkdir -p /run/media/cidata
-ExecStartPre=-/bin/mount -t iso9660 -o ro /dev/disk/by-label/cidata /run/media/cidata
+ExecStartPre=-/bin/mount -t iso9660 -o ro /dev/disk/by-label/cidata /media/cidata
 # The digest is a string in hex representation, we truncate it to a 32 bytes hex string
 ExecStartPost=-/bin/bash -c 'tpm2_pcrextend 8:sha256=$(head -c64 /run/peerpod/initdata.digest)'
--- a/src/cloud-api-adaptor/podvm/files/media/cidata/.gitignore
+++ b/src/cloud-api-adaptor/podvm/files/media/cidata/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/src/cloud-providers/docker/provider.go
+++ b/src/cloud-providers/docker/provider.go
@@ -72,12 +72,12 @@ func (p *dockerProvider) CreateInstance(ctx context.Context, podName, sandboxID 
 
 	// Create volume binding for the container
 
-	// mount userdata to /run/media/cidata/user-data
+	// mount userdata to /media/cidata/user-data
 	// This file will be read by process-user-data and daemon.json will be written to
 	// /run/peerpods/daemon.json at runtime
 	volumeBinding := []string{
 		// note: we are not importing that path from the CAA package to avoid circular dependencies
-		fmt.Sprintf("%s:%s", instanceUserdataFile, "/run/media/cidata/user-data"),
+		fmt.Sprintf("%s:%s", instanceUserdataFile, "/media/cidata/user-data"),
 	}
 
 	// Add host bind mount for /run/kata-containers and /run/image to avoid


bpradipt:
I'm not sure if there is a better way to solve the issue. Open for suggestion

mkulke (commented):
it's probably better to change the default mount path for the ci disk rather than adding an exception for docker, no? it doesn't need to be in /run I think, it can be in /mnt or wherever

bpradipt:
> it's probably better to change the default mount path for the ci disk rather than adding an exception for docker, no? it doesn't need to be in /run I think, it can be in /mnt or wherever

Yeah, having a uniform one is better. I was not sure if there is a hard dependency on `/run` for any platform. Let me just use /media as the mount path for the ci disk and check through ci if anything breaks. That should help.

bpradipt:
One challenge is how to make /media mountpath available early in the boot cycle since root (/) is RO. Only the following are available via systemd - https://freedesktop.org/wiki/Software/systemd/APIFileSystems/

bpradipt:
The CI failures may be fixed via https://github.com/confidential-containers/cloud-api-adaptor/pull/2226

wainersm:
> > Hi @bpradipt !
> > I finally managed to test this fix. It seems to fix the original problem: ` /var/tmp/media/cidata/user-data` is present and the process-user-data service is up and running just fine.
> > However, I'm still unable to get a pod running. The problem that I'm seeing is related with image pull on guest:
> > ```
> > === RUN   TestDockerCreateSimplePod/SimplePeerPod_test/PodVM_is_created
> >     assessment_runner.go:389: unexpected warning/error event(s): Error: failed to create containerd container: error unpacking image: content digest sha256:9fa9226be034e47923c0457d916aa68474cdfb23af8d4525e9baeebc4760977a: not foundError: failed to create containerd container: error unpacking image: failed to extract layer sha256:1e604deea57dbda554a168861cff1238f93b8c6c69c863c43aed37d9d99c5fed: failed to get reader from content store: content digest sha256:9fa9226be034e47923c0457d916aa68474cdfb23af8d4525e9baeebc4760977a: not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Above is a snippet of e2e test. I got a similar error when launching another pod from a different container image manually.
> 
> @wainersm you are hitting the known nydus-snapshotter issue. Workaround is documented here - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/troubleshooting/nydus-snapshotter.md#pod-creation-fails-with-error-unpacking-image Alternatively you can use crio runtime which doesn't have this issue.

Interesting that I almost never hit that problem when running in a fresh environment like the one setup by the e2e testing provisioner. Let's see how it goes on CI...

Anyway, with the workaround above I could launch a simple pod \o/

wainersm:
> One challenge is how to make /media mountpath available early in the boot cycle since root (/) is RO. Only the following are available via systemd - https://freedesktop.org/wiki/Software/systemd/APIFileSystems/

I found in https://systemd.io/TEMPORARY_DIRECTORIES/ that use of `/var/tmp` has some security implications that we should be aware of. I read that doc fully and concluded we should be fine, however, I'm not an expert so better a 2nd check would be nice.

wainersm (approved):
I tested it locally and it worked out. It seems that we safe with using /var/tmp (keep in min I'm not an expert on the topic).

bpradipt:
> > > Hi @bpradipt !
> > > I finally managed to test this fix. It seems to fix the original problem: ` /var/tmp/media/cidata/user-data` is present and the process-user-data service is up and running just fine.
> > > However, I'm still unable to get a pod running. The problem that I'm seeing is related with image pull on guest:
> > > ```
> > > === RUN   TestDockerCreateSimplePod/SimplePeerPod_test/PodVM_is_created
> > >     assessment_runner.go:389: unexpected warning/error event(s): Error: failed to create containerd container: error unpacking image: content digest sha256:9fa9226be034e47923c0457d916aa68474cdfb23af8d4525e9baeebc4760977a: not foundError: failed to create containerd container: error unpacking image: failed to extract layer 
> Interesting that I almost never hit that problem when running in a fresh environment like the one setup by the e2e testing provisioner. Let's see how it goes on CI...
> 

This issue is pretty consistent with docker provider.  With other providers it's occasional. 

> Anyway, with the workaround above I could launch a simple pod \o/

bpradipt:
> > One challenge is how to make /media mountpath available early in the boot cycle since root (/) is RO. Only the following are available via systemd - https://freedesktop.org/wiki/Software/systemd/APIFileSystems/
> 
> I found in https://systemd.io/TEMPORARY_DIRECTORIES/ that use of `/var/tmp` has some security implications that we should be aware of. I read that doc fully and concluded we should be fine, however, I'm not an expert so better a 2nd check would be nice.

Seems ok. But would be good to hear comment from @mkulke ..

mkulke:
> Seems ok. But would be good to hear comment from @mkulke ..

I mean generally the path sounds a bit odd, why var? why tmp? Can't we create a /media/cidata path statically on the rootfs at image-build time?

bpradipt:
> > Seems ok. But would be good to hear comment from @mkulke ..
> 
> I mean generally the path sounds a bit odd, why var? why tmp? Can't we create a /media/cidata path statically on the rootfs at image-build time?

Yeah, that's a possibility as well. I wanted to avoid changing the rootfs partition if there was an alternative. And it turned out systemd already creates those in tmpfs. Let me know what you prefer?

mkulke:
> > > Seems ok. But would be good to hear comment from @mkulke ..
> > 
> > 
> > I mean generally the path sounds a bit odd, why var? why tmp? Can't we create a /media/cidata path statically on the rootfs at image-build time?
> 
> Yeah, that's a possibility as well. I wanted to avoid changing the rootfs partition if there was an alternative. And it turned out systemd already creates those in tmpfs. Let me know what you prefer?

personally, I'd prefer a `/media/cidata` mount point.

bpradipt:
> > > > Seems ok. But would be good to hear comment from @mkulke ..
> > > 
> > > 
> > > I mean generally the path sounds a bit odd, why var? why tmp? Can't we create a /media/cidata path statically on the rootfs at image-build time?
> > 
> > 
> > Yeah, that's a possibility as well. I wanted to avoid changing the rootfs partition if there was an alternative. And it turned out systemd already creates those in tmpfs. Let me know what you prefer?
> 
> personally, I'd prefer a `/media/cidata` mount point.

@mkulke I added a new test commit, can you please check and let me know if the approach looks fine?

src/cloud-api-adaptor/pkg/paths/paths.go
```diff
@@ -7,5 +7,7 @@ const (
 	InitDataPath     = "/run/peerpod/initdata"
 	AgentCfgPath     = "/run/peerpod/agent-config.toml"
 	ForwarderCfgPath = "/run/peerpod/daemon.json"
-	UserDataPath     = "/run/media/cidata/user-data"
+	// This is not mounted under /run to avoid conflicting with
```
mkulke, line 10: I think we can scratch need that comment. /run is arguably not proper either

src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.repart/30-media.conf
```diff
@@ -0,0 +1,5 @@
+[Partition]
```
mkulke, line 1: Why do we create a partition? we just need a folder on the existing root partition, no? we can't create that adhoc at runtime, because it's a read-only fs, but a folder in ./podvm/files should work, I guess

```
[root@fedora ~]# mount /dev/sr0 /media
mount: /media: WARNING: source write-protected, mounted read-only.
[root@fedora ~]# ls -l /media/
total 5
-rw-r--r-- 1 1000 1000    0 Nov 22 14:10 meta-data
-rw-r--r-- 1 1000 1000 4584 Jan  9 09:57 user-data
```

src/cloud-api-adaptor/podvm-mkosi/mkosi.skeleton/usr/lib/systemd/system/process-user-data.service.d/10-override.conf
```diff
@@ -1,6 +1,7 @@
 [Service]
 # mount config disk if available
-ExecStartPre=-/bin/mkdir -p /run/media/cidata
-ExecStartPre=-/bin/mount -t iso9660 -o ro /dev/disk/by-label/cidata /run/media/cidata
+ExecStartPre=-/bin/mount /dev/disk/by-label/media /media
```
mkulke, line 3: see above, I don't think we need to mount a volume

src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.repart/30-media.conf
bpradipt: Ah yes :-) ..

mkulke (approved):

Issue #1831, opened 2024-05-03
tests/e2e: `TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly` test unstable


stevenhorsman:
We see occasional (anecdotally <20% of the time) failures on the libvirt nightly CI, which seems to always (so far) pass on re-run and now we've seen in on a PR test, so it's becoming more of an obstacle, so we should investigate it when we get the chance
```
=== RUN   TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly
=== RUN   TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly/EnvVariablePeerPodWithImageOnly_test
    assessment_runner.go:262: timed out waiting for the condition
--- FAIL: TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly (600.10s)
    --- FAIL: TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly/EnvVariablePeerPodWithImageOnly_test (600.10s)
```

stevenhorsman:
This is getting worse and we are hitting it multiple times on each PR now. I've tried running this test locally and in about 8 re-runs it worked every time, so I'm not sure of the cause of the failure. In the short term I think we need to skip it in the CI to stop it blocking PRs.

stevenhorsman:
It is possible that this is related to the image-pull changes as Chengyu is touch the config merge code in https://github.com/kata-containers/kata-containers/pull/9695, so after this, we should try re-testing this.

stevenhorsman:
Hmm, this is suspicious, now the e2e tests related to env are skipped I've seen:
```
=== RUN   TestLibvirtCreatePeerPodAndCheckWorkDirLogs
=== RUN   TestLibvirtCreatePeerPodAndCheckWorkDirLogs/WorkDirPeerPod_test
    assessment_runner.go:262: timed out waiting for the condition
--- FAIL: TestLibvirtCreatePeerPodAndCheckWorkDirLogs (600.16s)
    --- FAIL: TestLibvirtCreatePeerPodAndCheckWorkDirLogs/WorkDirPeerPod_test (600.16s)
```
start failing, so maybe it's related to something before now being cleaned up, or the workdir has the same issue?

stevenhorsman:
> start failing, so maybe it's related to something before now being cleaned up, or the workdir has the same issue?

[...]

PR #2226, opened 2025-01-03
tests/e2e: Skip tests after analysis

stevenhorsman:
Based in the test analysis done for 18 days of nighty tests in https://github.com/confidential-containers/cloud-api-adaptor/issues/1831#issuecomment-2541712756 the only containerd test failures we saw were:
- `TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithImageOnly` - three times
- `TestLibvirtCreatePeerPodAndCheckWorkDirLogs` - four times
- `TestLibvirtCreatePeerPodAndCheckEnvVariableLogsWithDeploymentOnly` - twice

Although the chances of failure for each of these tests is < 25%, we want to reduce the re-runs required, so if we skip these we should have more stable CI tests. It should also be noted that most of the failures were seen on the packer built images. This is probably just chance, but might indicate that the peer pod boot speed is related and we should re-evaluate again once we can remove the packer podvm images.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/libvirt_test.go
+++ b/src/cloud-api-adaptor/test/e2e/libvirt_test.go
@@ -66,16 +66,25 @@ func TestLibvirtCreatePeerPodAndCheckUserLogs(t *testing.T) {
 }
 
 func TestLibvirtCreatePeerPodAndCheckWorkDirLogs(t *testing.T) {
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm

wainersm (approved):
LGTM!

mkulke (approved):

PR #2229, opened 2025-01-08
podvm-mkosi: fixes nix error during make image

beraldoleal:
Currently, the `make image` target fails with the following error:

"experimental Nix feature 'nix-command' is disabled; add '--extra-experimental-features nix-command' to enable it."

This change explicitly enables the 'nix-command' and 'flakes' experimental features in the Makefile target to resolve this issue.

Otherwise, users would need to modify their local nix.conf configuration.

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Makefile
+++ b/src/cloud-api-adaptor/podvm-mkosi/Makefile
@@ -74,7 +74,7 @@ else ifeq ($(ARCH),s390x)
 	sudo -E ../hack/build-s390x-image.sh
 else
 	touch resources/buildBootableImage
-	nix develop ..#podvm-mkosi --command mkosi --environment=VARIANT_ID=production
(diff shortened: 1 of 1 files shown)


beraldoleal:
Hi @liudalibj, @stevenhorsman, @mkulke, happy new year!

I'm wondering if this happen with you guys too, or just on my env.

Maybe I'm missing something.

mkulke:
there are multiple ways to set this. I think the most common way to do it is this:

```bash
cat .config/nix/nix.conf
experimental-features = nix-command flakes
```

that's probably what the nix github action step does, too. We had a discussion around using the nix flake in the last community call and concluded that we might align amd64 with s390x and just download the python wheel file instead of using nix.

wainersm:
With https://github.com/confidential-containers/cloud-api-adaptor/pull/2233 merged, we no longer depend on nix. So closing this PR...

PR #2232, opened 2025-01-09
CAA: add support to look up imagePullSecrets for pods

squarti:
This PR initializes auth.json with the imagePullSecrets listed on the pod and service account.

Fixes: #2231

### DIFF
--- a/src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go
+++ b/src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go
@@ -81,6 +81,98 @@ func RemoveExtendedResources() error {
 	return nil
 }
 
+// Auths contains Registries with credentials
+type Auths struct {
+	Registries Registries `json:"auths"`
+}
+
+// Registries contains credentials for hosts
+type Registries map[string]Auth
+
+// Auth contains credentials for a given host
+type Auth struct {
+	Auth string `json:"auth"`
+}
+
+// GetImagePullSecrets gets image pull secrets for the specified pod
+func GetImagePullSecrets(podName string, namespace string) ([]byte, error) {
+
+	config, err := getKubeConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k8s config: %v", err)
+	}
+
+	cli, err := getClient(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k8s client: %v", err)
+	}
+
+	pod, err := cli.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	accountName := pod.Spec.ServiceAccountName
+	if accountName == "" {
+		accountName = "default"
+	}
+	serviceaAccount, err := cli.CoreV1().ServiceAccounts(namespace).Get(context.TODO(), accountName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	auths := Auths{}
+	auths.Registries = make(map[string]Auth)
+	for _, secret := range serviceaAccount.ImagePullSecrets {
+		err := addAuths(cli, namespace, secret.Name, &auths)
+		if err != nil {
+			return nil, err
+		}
+	}
+	for _, secret := range pod.Spec.ImagePullSecrets {
+		err := addAuths(cli, namespace, secret.Name, &auths)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if len(auths.Registries) > 0 {
+		authJSON, err := json.Marshal(auths)
+		if err != nil {
+			return nil, err
+		}
+		return authJSON, nil
+	}
+	return nil, nil
+}
+
+// getAuths get auth credentials from specified docker secret
+func addAuths(cli *k8sclient.Clientset, namespace string, secretName string, auths *Auths) error {
+	secret, err := cli.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	if err != nil {
+		// Ignore errors getting secrets to match K8S behavior
+		return nil
+	}
+	registries := Registries{}
+	if secretData, ok := secret.Data[".dockerconfigjson"]; ok {
+		auths := Auths{}
+		err := json.Unmarshal(secretData, &auths)
+		if err != nil {
+			return err
+		}
+		registries = auths.Registries
+	} else if secretData, ok := secret.Data[".dockercfg"]; ok {
+		err = json.Unmarshal(secretData, &registries)
+		if err != nil {
+			return err
+		}
+	}
+	for registry, creds := range registries {
+		auths.Registries[registry] = creds
+	}
+	return nil
+}
+
 // patchNodeStatus patches the status of a node
 func patchNodeStatus(c *k8sclient.Clientset, nodeName string, patches []jsonPatch) error {
 	if len(patches) > 0 {
--- a/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
+++ b/src/cloud-api-adaptor/test/e2e/assessment_helpers.go
@@ -708,12 +708,33 @@ func DeleteAndWaitForNamespace(ctx context.Context, client klient.Client, namesp
 	return nil
 }
 
-func AddImagePullSecretToDefaultServiceAccount(ctx context.Context, client klient.Client, secretName string) error {
+func getDefaultServiceAccount(ctx context.Context, client klient.Client) (*v1.ServiceAccount, error) {
+	clientSet, err := kubernetes.NewForConfig(client.RESTConfig())
+	if err != nil {
+		return nil, err
+	}
+	serviceAccount, err := clientSet.CoreV1().ServiceAccounts(E2eNamespace).Get(context.TODO(), "default", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return serviceAccount, nil
+}
+
+func setImagePullSecretsOnServiceAccount(ctx context.Context, client klient.Client, serviceAccount *v1.ServiceAccount, imagePullSecrets []v1.LocalObjectReference) error {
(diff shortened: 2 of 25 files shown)


mkulke (commented):
thx, looks good! the override semantics are interesting here, I'd intuitively expect that a cluster-wide auth.json is overridden by a pod-specific auth.json, but we're doing the opposite here. some questions:

does it make sense to retire the existing caa-wide "auth.json" mechanism in favor of passing this from kubernetes? the caa-wide param is roughly comparable to a node-wide auth.json in the kubelet config, which is discouraged (and which we don't consider it here, either).

I would suggest to at least deprecate the parameter and mark it for later removal. 

If we don't want to remove the caa-wide auth.json, would we be able to "merge" both sources?

if we don't want to remove the caa-wide auth.json, would it make sense for pod-specific auth.json to take precedence?

What are the implications for auth.json that come from KBS?

bpradipt:
> @bpradipt @snir911 - I think you added the CAA wide auth.json, so wanted to check if you had any concerned with this. I'm inclined to agree with Magnus, that it would be good to replace the CAA-wide approach with a pod specific one, but am unsure if this would cause challenges with the pause image pull on OpenShift?

Afaicr, the pause image is embedded now. 
Overall I like this approach, but I was thinking that imagePullSecrets, if present, would replace the default cluster wide auth.json as Magnus had mentioned. 
In the current of using serviceAccount and imagePullSecrets may be we can have a flexibility to specify which service account to retrieve from. That could be a CAA config with default being the `default` SA. This will avoid any need for cluster wide auth.json to be used.

mkulke:
> > thx, looks good! the override semantics are interesting here, I'd intuitively expect that a cluster-wide auth.json is overridden by a pod-specific auth.json, but we're doing the opposite here. some questions:
> 
> I left the cluster wide auth.json overwriting the pod specific secrets because this way current code that uses cluster wide auth.json would not be broken. I think merging the secrets would also be ok as long as the pod secrets do not conflict with the cluster wide secret.
> 
> Let me know which approach to take and I will adjust the changes.

I would opt for removing the option to have a cluster-wide auth.json in this case. If I understood it correctly there is no use case for that anymore. Propagating it from the cluster is the right call, IMO. since the manifest will still be pulled on the node different maintaining different sets of registry credentials in k8s and caa config can be pretty confusing.

snir911:
I like the idea, from the pause image pulling POV i don't think it would be an issue in Openshift..
Another point is that there're node/runtime level pull credentials also IIRC, if keeping auth.json is the only option to (roughly) mimic this behavior i'd keep it.

src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go
```diff
@@ -81,6 +81,97 @@ func RemoveExtendedResources() error {
...
+	}
+
+	pod, err := cli.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	accountName := pod.Spec.ServiceAccountName
+	if accountName == "" {
+		accountName = "default"
```
bpradipt, line 117: Making the SA name configurable will be good. In some environment specific SAs are created and usage of "default" is discouraged.

squarti: This is following Kubernetes [documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server):

`If you do not specify a ServiceAccount when you create a Pod, Kubernetes automatically assigns the ServiceAccount named default in that namespace.`

bpradipt: Thanks @squarti. We will have to follow the K8s convention :-) ..

bpradipt (approved):
/lgtm
Thanks @squarti

stevenhorsman:
@squarti - are you able to rebase this please to pick up the latest changes. Also I'd like to understand where we ended up with on the user experience here: I think the consensus was having support for node wide secret that is overridden with a deployment specific imagePullSecret?

stevenhorsman:
> I think the consensus was to remove the option to have a cluster-wide auth.json.

Ok, that's not what was discussed on the community call, which I appreciate you weren't present at. From a quick scan above I think it looks like mostly @mkulke was for removing the cluster-wide option, but was supporting it on the call?

I don't really mind either way, but I'd like some community agreement here,

bpradipt:
> > I think the consensus was to remove the option to have a cluster-wide auth.json.
> 
> Ok, that's not what was discussed on the community call, which I appreciate you weren't present at. From a quick scan above I think it looks like mostly @mkulke was for removing the cluster-wide option, but was supporting it on the call?
> 
> I don't really mind either way, but I'd like some community agreement here,

Fwiw, I'm ok to remove the cluster-wide auth option since imagepullSecret can be provided via the SA. Also with this change it decouples the pod image pull secret from node specific secret. 
Also to be clear, I don't have any hard preference either way..

mkulke:
afair, in the discussion the rationale for the keeping a cluster-wide option was: we currently don't know how to propagate credentials that are specified "out-of-band" on the node to the podvm, e.g. as a kubelet option or via other container runtime means.

an alternative that I can imagine would be to provide a simple docs explaining how an "auth.json" can be embedded in in the image.

I think in both cases we need logic to merge auth.json and a specified behaviour for overlaps / preference. maybe we can leverage existing [libraries](https://pkg.go.dev/github.com/docker/docker/api/types/registry).

src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go
```diff
@@ -81,6 +81,97 @@ func RemoveExtendedResources() error {
...
+		}
+	}
+	for _, secret := range pod.Spec.ImagePullSecrets {
+		err := getAuths(cli, namespace, secret.Name, &auths)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if len(auths.Registries) > 0 {
```
huoqifeng, line 139: I like the idea to retrieve auth.json from imagepullsecret. :-)

squarti:
Coming back to this.  is this the consensus?

- keep the cluster wide `auth-json-secret`

stevenhorsman:
> Coming back to this. is this the consensus?
> 
>     * keep the cluster wide `auth-json-secret`
> 
>     * merge the cluster wide secret with the pod image pull secrets
> 
>     * give precedence to the pod image pull secrets

I think when we last discussed this, we decided that whatever of the discussed options we were talking about here were picked, it would be an improvement on the current situation, so let's go ahead with this and re-visit later if necessary?

stevenhorsman:
Hi Silenio, just to give you an update on the PR - I'm currently working on updating/re-writing the automated tests for this PR as I think that imagePullsecrets via the pod/service account should be the primary way to provide these rather than our old `auth-json-secret` approach that we used to have. It would be great if you could update the doc: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/registries-authentication.md to reflect this new approach though (and I think that removing the auth-json-secret approach probably makes sense even if it might still technically work). Does that sound okay?

stevenhorsman:
I've created https://github.com/squarti/cloud-api-adaptor/pull/1 with the e2e test updates for this feature and it seems to be working in my fork :)

stevenhorsman:
@squarti - sorry - I think I mess this up creating a PR against your branch rather than asking you to cherry-pick, so we've got a merge commit now - if you can rebase to just keep the two code commits that would be great. Thanks

squarti:
sorry, I think I should have squash merged our PR instead.  Hopefully, the rebase is ok.

stevenhorsman:
@mkulke - I wonder if you are able to review this - as I wrote the e2e test commit then it's probably not great for me to do it. It's worth noting that the crio tests are not stable in the daily either, so I don't think that failure is related to this change. Thanks!

src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go
```diff
@@ -81,6 +81,98 @@ func RemoveExtendedResources() error {
...
+		if err != nil {
+			return nil, err
+		}
+		return authJSON, nil
+	}
+	return nil, nil
+}
+
+// getAuths get auth credentials from specified docker secret
+func getAuths(cli *k8sclient.Clientset, namespace string, secretName string, auths *Auths) error {
```
mkulke, line 150: nit: we're actually setting the auths param here, maybe `setAuths()` or `populateAuths()`?

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -708,12 +708,34 @@ func DeleteAndWaitForNamespace(ctx context.Context, client klient.Client, namesp
...
+	clientSet, err := kubernetes.NewForConfig(client.RESTConfig())
+	if err != nil {
+		return err
+	}
+	serviceAccount.ImagePullSecrets = imagePullSecrets
+	_, err = clientSet.CoreV1().ServiceAccounts(E2eNamespace).Update(context.TODO(), serviceAccount, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	log.Infof("ServiceAccount %s updated successfully.", "default")
```
mkulke, line 733: as this fn is generic, I would move the log statement to the caller of the fn (`Add...ToDefaultServiceAccount`)

we don't need string templating if we use 2 constant strings.

mkulke (approved):
lgtm, a few nits

stevenhorsman:
@squarti - are you able to squash commit 3 into 1 and then hopefully we can get this passing the tests and merged 🤞

stevenhorsman:
> @squarti - are you able to squash commit 3 into 1 and then hopefully we can get this passing the tests and merged 🤞

I just wanted commit 3 into 1, not 2 as well, but lets just try and get this in

squarti:
Sorry, I misread your comment.  Thanks for getting this in.

PR #2233, opened 2025-01-10
ci: install mkosi from repository

mkulke:
As discussed in the last community call, we can adopt the installation method of the s390x builds and use it also for the amd64 builds. this allows us to use the profile feature available in more recent versions of mkosi and simplify some things.

[here](https://github.com/mkulke/cloud-api-adaptor/actions/runs/12708815306/job/35426525187) would be a successful build on a fork. due to the pull-request-target logic there will be a mismatch between Makefile (on PR branch) and github-actions workflow (on main), so it doesn't make sense to run e2e tests here.


src/cloud-api-adaptor/podvm-mkosi/README.md
```diff
@@ -77,7 +79,7 @@ reduce complexity of configuration and CI and shall not be seen as open to-dos.
   from IMDS via the project's `process-user-data` tool.
 
 ## Build s390x image
-Since the [nix OS](https://nixos.org/download/#download-nix) does not support s390x, we can use the mkosi **ToolsTree** feature defined in `mkosi.conf` to download latest tools automatically:
+We can use the mkosi **ToolsTree** feature defined in `mkosi.conf` to download latest tools automatically:
```
wainersm, line 82: Hi @mkulke , yesterday I was playing with mkosi and I solved dependency problems by:

```
diff --git a/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf b/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf
index 08891dda..1921a848 100644
--- a/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf
+++ b/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.conf.d/fedora.conf
@@ -32,3 +32,5 @@ RemoveFiles=/var/log
 RemoveFiles=/var/cache
 RemoveFiles=/etc/pki/ca-trust/extracted/java/cacerts
             /usr/lib/sysimage/libdnf5/transaction_history.sqlite*
+[Host]
+ToolsTree=default
```

For s390x we do the same in mkosi.presets/system/mkosi.conf.d/fedora-s390x.conf

mkosi newbie's question: why not set ToolsTree=default on fedora.conf too? Any downside that I might be missing?

mkulke: it wasn't 100% reliable in the past, but since s390x is using it, it'll probably work here too, let's try...

it's a bit intransparent when things don't work, like when you use an exotic distro. but then people probably know how to figure out missing deps on their own, I guess.

mkulke: nope, that doesn't work w/ v22:

image

it's not adding bubblewrap to the toolstree, maybe that's solved in later version. but I'm not super keen to figure out manually which dep we need from the distro and which ones are included in the toolstree.

wainersm: hmm Interesting... started a fresh fedora 40 VM where I installed the packaged mkosi (v22) and basic C development tools; it built the image just fine. But as you said, it wasn't 100% reliable while ago -- it seems still not. Ok, let's ignore that.

wainersm (approved):
Tested locally on Fedora 40. Thanks @mkulke !

mkulke:
@stevenhorsman the e2e tests will fail since the workflow and the Makefile were edited

.github/workflows/podvm_mkosi.yaml
```diff
@@ -192,7 +189,6 @@ jobs:
           subject-digest: ${{ steps.publish_oras_qcow2.outputs.digest }}
           push-to-registry: true
 
-
       - name: Clean up some space for the docker provider build
         working-directory: src/cloud-api-adaptor/podvm-mkosi
         run: |
```
stevenhorsman, line 194: Can we remove the
```
          sudo rm -rf /nix
```
(line 197) now too?

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -86,15 +83,12 @@ image-debug:
 	rm -rf ./build
```
stevenhorsman, line 83: Can we remove `touch resources/buildDebugImage` (line 82) now as well?

stevenhorsman (commented):
A couple of code tidying suggestions, but otherwise looks good assuming the tests pass

stevenhorsman:
> @stevenhorsman the e2e tests will fail since the workflow and the Makefile were edited

Good point - can we test (some of it) on a fork repo?

mkulke:
> > @stevenhorsman the e2e tests will fail since the workflow and the Makefile were edited
> 
> Good point - can we test (some of it) on a fork repo?

yup, see PR description

stevenhorsman:
> > > @stevenhorsman the e2e tests will fail since the workflow and the Makefile were edited
> > 
> > 
> > Good point - can we test (some of it) on a fork repo?
> 
> yup, see PR description

Sorry!

stevenhorsman (approved):
LGTM. Thanks!

PR #2234, opened 2025-01-13
docs: Remove packer docs and files for azure

mkulke:
The azrue build-image instructions have been updated for mkosi podvms.

Since we don't test packer images for azure anymore, we should remove it here, it's likely that there is a version drift and the resulting images won't work as expected anymore.


stevenhorsman (approved):
Makes sense to me. Thanks!

snir911:
Would it make sense to add a note indicates this is about to be deprecated and plan to deprecate it in the next release? Or perhaps start by removing the documentation while keeping the code for an additional release? it’s still being used by OCP.

mkulke:
> Would it make sense to add a note indicates this is about to be deprecated and plan to deprecate it in the next release? Or perhaps start by removing the documentation while keeping the code for an additional release? it’s still being used by OCP.

I see, we can add a note and leave the existing packer code in place for reference.

PR #2235, opened 2025-01-13
CI: don't install uefi deps on s390x podvm builds

mkulke:
There are no UKI tools for s390x packaged in ubuntu at the moment, thus we have to split this out. s390x will not attempt to build UEFI images, so we don't need those tools for an mkosi build on that platform.

### DIFF
--- a/.github/workflows/podvm_mkosi.yaml
+++ b/.github/workflows/podvm_mkosi.yaml
@@ -118,11 +118,16 @@ jobs:
             alien \
             bubblewrap \
             dnf \
-            mtools \
             qemu-utils \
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
LGTM - sorry I didn't spot these added in the previous PR.

mkulke:
I took the liberty to the updated workflow on this repo to leverage to the s390x runner. it seems to work: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/12745991124/job/35521109158

BbolroC (approved):
LGTM thanks @mkulke !

PR #2242, opened 2025-01-16
CI: revert mkosi-amd64 back to v17

mkulke:
The images that are built on the runners directly with updated mkosi do not boot on libvirt VMs, so we need to revert and revisit this at a later point.

Revert "CI: don't install uefi deps on s390x podvm builds"

This reverts commit c1b26f1803c179bf8589b557318f4f77dd315854.

Revert "ci: bump azure podvm build workflow to 24.04"

This reverts commit 3ebb741a6766bf9a797e39accf839ba7e7b7dc76.

Revert "ci: use profile instead of marker files"

This reverts commit 6e25be399498f9353e14cb826dba72cd877bdb89.

Revert "ci: install mkosi from repository for x86_64"

This reverts commit e7bd8bada2846de92d4298fa1e5298f1ab9f8967.

### DIFF
--- a/.github/workflows/azure-podvm-image-build.yml
+++ b/.github/workflows/azure-podvm-image-build.yml
@@ -35,7 +35,7 @@ env:
 
 jobs:
   build-podvm-image:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     defaults:
       run:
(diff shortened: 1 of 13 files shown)


mkulke:
this won't help with the CI breakage, this is most likely a kernel problem that was introduced in-flight on fedora 40

PR #2243, opened 2025-01-16
CI: build mkosi in docker + downgrade podvm to f39

mkulke:
There is a problem when running mkosi on a github ubuntu runner, the images will build successfully, but the EFI partition will not be setup in a way that libvirt can boot from it.

The packages on a generic github runner aren't really predictable, to make an mkosi build more reproducible, we can launch mkosi in a fedora container with `--tools-tree=default`.

Since this implies running a user ns within a container, we need to configure a docker buildx builder that allows this and run mkosi with a special flag in the Dockerfile.

Another problem has been introduced in f40 in recent days. With a bump from 6.12.8-100.fc40 => 6.12.9-100.fc40 in fedora40, there is a regression for peerpod, tests will not pass. We need further investigation about that, but to unblock the CI we can downgrade to fedora39, which is stuck on kernel 6.11.9-100.fc39.

Finally, there are some minor fixes to the Azure podvm build.

[Here](https://github.com/mkulke/cloud-api-adaptor/actions/runs/12809685381/job/35715029998) is a successful run w/ mkosi22 + fedora39

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi
+++ b/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi
@@ -1,0 +1,30 @@
+# syntax=docker/dockerfile:1.5.0-labs
+FROM fedora:40 AS builder
+
+ARG MKOSI_VERSION="v22"
+ARG PROFILE="debug"
+
+RUN dnf install -y \
+	bubblewrap \
+	git
+
+RUN mkdir -p /build
+WORKDIR /build
+RUN git clone -b "$MKOSI_VERSION" https://github.com/systemd/mkosi
+ENV PATH="/build/mkosi/bin:$PATH"
+RUN mkosi --version
+
+RUN mkdir /image
+WORKDIR /image
+COPY mkosi.postinst /image/mkosi.postinst
+COPY mkosi.presets /image/mkosi.presets
+COPY mkosi.profiles /image/mkosi.profiles
+COPY mkosi.skeleton /image/mkosi.skeleton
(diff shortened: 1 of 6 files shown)


src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi
```diff
@@ -0,0 +1,33 @@
...
+COPY mkosi.postinst /image/mkosi.postinst
+COPY mkosi.presets /image/mkosi.presets
+COPY mkosi.profiles /image/mkosi.profiles
+COPY mkosi.skeleton /image/mkosi.skeleton
+COPY mkosi.skeleton-debug /image/mkosi.skeleton-debug
+COPY mkosi.workspace /image/mkosi.workspace
+COPY resources /image/resources
+COPY mkosi.conf /image/mkosi.conf
+RUN touch /image/resources/buildBootableImage
+RUN touch /image/resources/buildDebugImage
```
stevenhorsman, line 28: Do we still need these files now we have profiles, or are you thinking of a phased approach for more stability?

stevenhorsman (approved):

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi
mkulke: no, those shouldn't have any effect anymore.

beraldoleal (approved):
I was affected by the issues you listed... tested on GCP and worked. Images are booting and services are up. Thanks.

PR #2244, opened 2025-01-16
Use logger instead of fmt.Printf()

EmmEff:
Use logger for consistent output, instead of fmt.Printf()


mkulke (approved):

stevenhorsman (approved):
LGTM. Thanks!

PR #2247, opened 2025-01-16
ci: add podvm-smoketest workflow

mkulke:
This workflow will build a podvm image, launch a vm and then attempt to address it via the kata agent API.

This should give earlier signals of podvm problems and can run on PRs.

nb: the wf is running on 22.04 b/c libvirt struggles with OVMF_CODE_4M firmware that is default on ub 24.04.

### DIFF
--- a/.github/workflows/podvm_smoketest.yaml
+++ b/.github/workflows/podvm_smoketest.yaml
@@ -1,0 +1,160 @@
+name: smoke test
+
+on:
+  pull_request:
+
+jobs:
+  podvm-mkosi:
+    # We're pinning the runner to 22.04 b/c libvirt struggles with the
+    # OVMF_CODE_4M firmware that is default on 24.04.
+    runs-on: 'ubuntu-22.04'
+
+    defaults:
+      run:
+        working-directory: src/cloud-api-adaptor/podvm-mkosi
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # Required by rootless mkosi on Ubuntu 24.04
+      # - name: Un-restrict user namespaces
+      #   run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            genisoimage \
+            qemu-utils \
+            socat \
+            virt-manager
+          sudo snap install yq
+
+      - name: Read properties from versions.yaml
+        working-directory: src/cloud-api-adaptor
+        run: |
+          {
+            echo "MKOSI_VERSION=$(yq -e '.tools.mkosi' versions.yaml)"; 
+            echo "ORAS_VERSION=$(yq -e '.tools.oras' versions.yaml)";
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
nice ..

.github/workflows/podvm_smoketest.yaml
```diff
@@ -0,0 +1,157 @@
...
+on:
+  pull_request:
+
+defaults:
+  run:
+    working-directory: src/cloud-api-adaptor/podvm-mkosi
+
+jobs:
+  smoke-test:
+    runs-on: 'ubuntu-22.04'
```
stevenhorsman, line 12: Is it worth adding this comment:
> nb: the wf is running on 22.04 b/c libvirt struggles with OVMF_CODE_4M firmware that is default on ub 24.04.

as a workflow comment, so that someone (probably me) doesn't come and mistakenly try and update the runner in the future?

.github/workflows/podvm_smoketest.yaml
```diff
@@ -0,0 +1,157 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      # Required by rootless mkosi on Ubuntu 24.04
+      # - name: Un-restrict user namespaces
+      #   run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+
```
stevenhorsman, line 19: Is this just a reference for developers trying to do this manually, or potentially for future when we can migrate to 24.04 after the OVMF firmware issues are resolved?

.github/workflows/podvm_smoketest.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build binaries
+        run: make binaries
+
+      - name: Disable TLS for agent-protocol-forwarder
+        run: |
+          mkdir -p ./resources/binaries-tree/etc/default
+          echo "TLS_OPTIONS=-disable-tls" > ./resources/binaries-tree/etc/default/agent-protocol-forwarder
```
stevenhorsman, line 53: Do you think this is a candidate to also enable in the "developer mode" I'm thinking of in https://github.com/confidential-containers/cloud-api-adaptor/pull/2227
I'm trying to work out if there would be value in reusing the current podvm_mkosi workflow to help do this image build section, but we'd need to add support for that in. This isn't a blocker for this PR, just thinking about it whilst reviewing.

stevenhorsman (approved):
This looks like a great addition. Thanks a lot for actioning it so quickly!

.github/workflows/podvm_smoketest.yaml
mkulke: the latter, because we'd probably forget about this

.github/workflows/podvm_smoketest.yaml
mkulke: yeah, although I'm not convinced we require a build-time flag for TLS. An empty tls config in daemon.json should just indicate that we're not using TLS.

Issue #2254, opened 2025-01-21
Failure to Extract Layers in Containerd


beraldoleal:
### Describe the bug

In a GKE cluster with containerd and Ubuntu (1.7.24-0ubuntu0~22.04.1), containerd fails to extract image layers during container creation.

This occurs even when image is successfully pulled and present in the content store. Running `ctr content fetch ...` manually on node fixes the issue, but this requires manual intervention.

Container creation could seamlessly pull and unpack all image layers without requiring manual fetch commands.

_** Note: This might not be specific to GCP and could be a combination of containerd and Nydus version?_

### How to reproduce

1. Create a GKE cluster using the following command:

```
gcloud container clusters create my-cluster \
  --zone ${GCP_REGION}-a \
  --machine-type "e2-standard-4" \
  --image-type UBUNTU_CONTAINERD \
  --num-nodes 1
```

2. Deploy a pod using an image from a public registry (e.g., quay.io/prometheus/busybox) and `kata-remote`
3. Observe the container creation logs for failures during image unpacking.

### CoCo version information

caa HEAD, containerd 1.7.24

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell
failed to extract layer sha256:<layer-digest>: failed to get reader from content store: content digest sha256:<digest>: not found
```

PR #2257, opened 2025-01-22
gcp: add note on discard_unpacked_layers for GKE

beraldoleal:
GKE uses `discard_unpacked_layers=true` for optimizations, starting with version 1.27. This can cause `failed to extract layer sha256:xxx: not found` errors with PeerPods. The note explains disabling this flag in containerd to avoid issues.

### DIFF
--- a/src/cloud-api-adaptor/gcp/README.md
+++ b/src/cloud-api-adaptor/gcp/README.md
@@ -101,6 +101,14 @@ gcloud container clusters create my-cluster \
   --num-nodes 3
 ```
 
+> [!NOTE]
+> Starting with GKE version 1.27, GCP configures containerd with the
+> `discard_unpacked_layers=true` flag to optimize disk usage by removing
+> compressed image layers after they are unpacked. However, this can cause
+> issues with PeerPods, as the workload may fail to locate required layers. To
+> avoid this, disable the `discard_unpacked_layers` setting in the containerd
+> configuration.
+
 Regardless of the method used, at the end you should have a KUBECONFIG pointing
 to the auth file and a cluster up and running:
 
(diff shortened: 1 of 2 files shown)


beraldoleal:
@wainersm, @stevenhorsman I did some tests with peerpods + GCP using devel branch and the only issue I found is related to #2254, also documented [here](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/troubleshooting/nydus-snapshotter.md#pod-creation-fails-with-error-unpacking-image).

My local tests shows the issue is caused by `discard_unpacked_layers=true` in containerd's configuration, which is set by default in GKE (and maybe kind) to save some space. According with the implementation:

> DiscardUnpackedLayers is a boolean flag to specify whether to allow GC to
> remove layers from the content store after successfully unpacking these
> layers to the snapshotter.

Disabling this flag resolved the problem in my tests. So I think it might be worth adding a note to the documentation.

stevenhorsman (approved):
Looks okay to me. Thanks!

src/cloud-api-adaptor/gcp/README.md
```diff
@@ -101,6 +101,14 @@ gcloud container clusters create my-cluster \
...
 ```
 
+> [!NOTE]
+> Starting with GKE version 1.27, GCP configures containerd with the
+> `discard_unpacked_layers=true` flag to optimize disk usage by removing
+> compressed image layers after they are unpacked. However, this can cause
+> issues with PeerPods, as the workload may fail to locate required layers. To
+> avoid this, disable the `discard_unpacked_layers` setting in the containerd
+> configuration.
+
```
bpradipt, line 111: Nice find @beraldoleal.
Do you mind adding a mention about the discard_unpacked_layers in nydus troubleshooting doc as well - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/troubleshooting/nydus-snapshotter.md

beraldoleal: Done.

bpradipt (approved):
/lgtm
Thanks @beraldoleal

beraldoleal:
> Hi @beraldoleal !
> 
> I know that you discussed that `discard_unpacked_layers=true` configuration on Slack but I didn't follow closely. All in all, does it solve the containerd/nydus-snapshotter problem or was just a false-positive?

Yes, this solved the problem on my side. GCP was using this option by default. I checked a basic run of `kind` and it was using it too.

PR #2260, opened 2025-01-23
podvm: Fix debug image build issue on s390x

yoheiueda:
This patch fixes the debug image build issue on s390x. 

Currently, I cannot build a debug podvm image on s390x. The following command just build a production podvm image.

```
sudo make image-debug
```

I guess this issue was revealed by this recent change.

https://github.com/confidential-containers/cloud-api-adaptor/commit/6e25be399498f9353e14cb826dba72cd877bdb89#diff-0578158244ea494f1ad5e1970ee5560d1cf6bfc62fd70243bd3a46280a237a14


mkulke (approved):
Oh right, that’s a typo and on an unknown profile it‘ll default to production, thx for catching it.

PR #2267, opened 2025-01-29
GCP: handling DISABLECVM and allow users to set podvm disk-type

beraldoleal:
Currently, DISABLECVM is ignored on GCP. This PR adds initial support for handling DISABLECVM and allows users (especially those using TDX) to set the disk type instead of relying on hardcoded values, as the default values are not supported by TDX instances.


src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -94,11 +94,11 @@ azure() {
 gcp() {
     test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
 
-    [[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-gcp-image-name ${PODVM_IMAGE_NAME} "
+    [[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-imageid ${PODVM_IMAGE_NAME} "
```
bpradipt, line 97: Personally I think it's more intuitive to keep the options aligned to the cloud provider nuances to the extent possible. Like instance-type (aws), instance-size( azure).  
I see all public GCP docs refers to machine type (for cloud instances) so may be it's ok to use `machine-type` for GCP. Wdyt ?

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -94,11 +94,11 @@ azure() {
...
     [[ "${GCP_PROJECT_ID}" ]] && optionals+="-gcp-project-id ${GCP_PROJECT_ID} "
     [[ "${GCP_ZONE}" ]] && optionals+="-gcp-zone ${GCP_ZONE} "                         # if not set retrieved from IMDS
-    [[ "${GCP_MACHINE_TYPE}" ]] && optionals+="-gcp-machine-type ${GCP_MACHINE_TYPE} " # default e2-medium
-    [[ "${GCP_NETWORK}" ]] && optionals+="-gcp-network ${GCP_NETWORK} "                # defaults to 'default'
+    [[ "${GCP_MACHINE_TYPE}" ]] && optionals+="-instance-type ${GCP_MACHINE_TYPE} "    # default e2-medium
+    [[ "${GCP_NETWORK}" ]] && optionals+="-subnetid ${GCP_NETWORK} "                   # defaults to 'default'
```
bpradipt, line 101: I see public doc mention subnets so this is better. What about the var? Should it be renamed to GCP_SUBNET_ID?

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -94,11 +94,11 @@ azure() {
 gcp() {
     test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
 
-    [[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-gcp-image-name ${PODVM_IMAGE_NAME} "
+    [[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-imageid ${PODVM_IMAGE_NAME} "
     [[ "${GCP_PROJECT_ID}" ]] && optionals+="-gcp-project-id ${GCP_PROJECT_ID} "
     [[ "${GCP_ZONE}" ]] && optionals+="-gcp-zone ${GCP_ZONE} "                         # if not set retrieved from IMDS
-    [[ "${GCP_MACHINE_TYPE}" ]] && optionals+="-gcp-machine-type ${GCP_MACHINE_TYPE} " # default e2-medium
-    [[ "${GCP_NETWORK}" ]] && optionals+="-gcp-network ${GCP_NETWORK} "                # defaults to 'default'
+    [[ "${GCP_MACHINE_TYPE}" ]] && optionals+="-instance-type ${GCP_MACHINE_TYPE} "    # default e2-medium
```
bpradipt, line 100: Personally I think it's more intuitive to keep the options aligned to the cloud provider nuances to the extent possible. Like instance-type (aws), instance-size( azure).  
I see all public GCP docs refers to machine type (for cloud instances) so may be it's ok to use `machine-type` for GCP. Wdyt ?

src/cloud-api-adaptor/entrypoint.sh
bpradipt, line 97: What about the env variable? s/PODVM_IMAGE_NAME/GCP_IMAGE_ID ?

src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
```diff
@@ -17,14 +17,15 @@ configMapGenerator:
...
   - CLOUD_PROVIDER="gcp"
-  #- PAUSE_IMAGE="" # Uncomment and set if you want to use a specific pause image
-  #- TUNNEL_TYPE="" # Uncomment and set if you want to use a specific tunnel type. Defaults to vxlan
-  #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
   - PODVM_IMAGE_NAME="" # set from step "Build Pod VM Image" in gcp/README.md
   - GCP_PROJECT_ID="" # set
   - GCP_ZONE="" # set e.g. "us-west1-a"
   - GCP_MACHINE_TYPE="e2-medium" # replace if needed. caa defaults to e2-medium
   - GCP_NETWORK="global/networks/default" # replace if needed.
+  #- GCP_DISK_TYPE="pd-standard" # Uncomment it if you want to change
```
bpradipt, line 25:
```Suggestion
  #- GCP_DISK_TYPE="" # Uncomment and set if you want to use a specific disk type.
Defaults to pd-standard
```

src/cloud-providers/gcp/provider.go
```diff
@@ -141,6 +142,39 @@ func (p *gcpProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 			},
 		},
 	}
+
+	if !p.serviceConfig.DisableCVM {
+		confidentialInstanceTypes := map[string]string{
+			"c3-":  "TDX",
+			"n2d-": "SEV_SNP",
+			"t2d-": "SEV_SNP",
```
bpradipt, line 150: Is it really need to check the instance type here? I doubt there will be a guarantee that conf instance types will only have the above specific instance prefix going forward and we'll need to keep updating it to account for new and retired instance types.
May be you can offload the detection to the gcp creation API . It should fail if the instance type is incorrect when EnableConfidentialCompute is enabled.

src/cloud-api-adaptor/entrypoint.sh
beraldoleal: I tried to avoid changing the env variable, because is public facing (Kustomization). But if we are ok with that, given we might not have that much users at this point, we can change it too.

src/cloud-api-adaptor/entrypoint.sh
beraldoleal: I mostly agree with you, but I have mixed feelings. I like the idea of using the provider’s naming internally, but for the user interface, I think it makes sense to go with more generic names when possible (which we don’t really do today). That way, if users switch providers, it’s less disruptive for them.

I haven’t thought too much about it, so maybe it wouldn’t work for every variable because of specific nuances. But in general, it’d help keep the interface more consistent across providers.

For now, I’m good with keeping machine-type.

src/cloud-providers/gcp/provider.go
beraldoleal: Yeah, I didn’t like this either. Totally agree.

If c3-* instances ever support both TDX and SEV_SNP, GCP can´t figure out based on the instancetype. It *could* figure it out based on the GuestOsFeatures. Assuming they actually could check this, I have no idea what happens if an image has both. But I'm not even sure if it makes sense to enable both in one image, but I couldn’t find anything about it in today’s docs.

So maybe we do need to expose the type for users to pick? (Which I was hoping to avoid). WTDY?

src/cloud-api-adaptor/entrypoint.sh
bpradipt: It's good to change now imho, since the GCP support is still new.

src/cloud-providers/gcp/provider.go
bpradipt: From the OS side, usually the same image should work for TDX and SNP (eg. Azure). In some unique cases, there may be a need for custom image (due to kernel requirements or other dep). For such cases, user can always override the default OS image via pod annotation. Likewise if there is a need for a specific instance (machine) type, a user can use the pod annotation to override it.  So options to pick up a specific OS image or an instance type already exists for the user.

Adding the check for supported machine type based on prefix in the code will become a pain point going forward. But if there is a provider specific API to check if the machine type supports confidentiality, then may be you can use it to give early feedback to the user on any wrong selection of machine type. Otherwise we just rely on the Create API from the provider to throw error in case of any incompatibility.
Having said that if there is a strong reason to support only specific machine types in CAA for now, then the current check in the code makes sense.

src/cloud-api-adaptor/entrypoint.sh
beraldoleal: Ok, I was about to change this, but I guess I will keep network here for now...

Currently, we are using the "network" logic, which let GCP choose the subnet. So here, is actually network, not subnet. And this was working fine with our tests. So I would be more comfortable to change to sub network on a different PR.

src/cloud-providers/gcp/provider.go
beraldoleal: >  Having said that if there is a strong reason to support only specific machine types in CAA for now, then the current check in the code makes sense.

Those are the only instance types GCP supports for now. Also, avoiding requests we know will fail gives users early feedback, which makes things easier and I believe is a good argument.

I don’t love having these hardcoded either, but with the current logic, I haven’t found a better way.

Also, I tried to delegate to the GCP API before, and got the warning below:

```bash
$ gcloud compute instances create test-confidential-vm \
  --zone=us-central1-a \
  --machine-type=c3-standard-4 \
  --image=fedora-mkosi-tee-intel-1-11-0 \
  --confidential-compute
WARNING: The --confidential-compute flag will soon be deprecated. Please use `--confidential-compute-type=SEV` instead
```

I only saw this deprecation here, could not find any API docs, but it seems like GCP is moving away from auto-detection.
 
(So, I’m not sure about all the details of DISABLECVM, but just an wild suggestion: what if we eventually replace it with TEE_PLATFORM="none|intel|amd"? That way, users wouldn’t need to set two variables, and we wouldn’t have to guess. Just an idea for the future.)

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -92,12 +92,12 @@ azure() {
 }
 
 gcp() {
-    test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
+    test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE GCP_IMAGE_NAME
 
-    [[ "${PODVM_IMAGE_NAME}" ]] && optionals+="-gcp-image-name ${PODVM_IMAGE_NAME} "
+    [[ "${GCP_IMAGE_NAME}" ]] && optionals+="-imageid ${GCP_IMAGE_NAME} "
```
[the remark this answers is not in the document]
wainersm: ah, you mentioned already the isn't standard across the providers, but my preference is for using the `PODVM_` prefix. Just saying, I'm ok with either way.

src/cloud-providers/gcp/manager.go
```diff
@@ -18,13 +18,12 @@ func init() {
...
-	flags.StringVar(&gcpcfg.ProjectId, "gcp-project-id", "", "GCP Project ID")
-	flags.StringVar(&gcpcfg.Zone, "gcp-zone", "", "Zone")
-	flags.StringVar(&gcpcfg.ImageName, "gcp-image-name", "", "Pod VM image name")
-	flags.StringVar(&gcpcfg.MachineType, "gcp-machine-type", "e2-medium", "Pod VM instance type")
-	flags.StringVar(&gcpcfg.Network, "gcp-network", "", "Network ID to be used for the Pod VMs")
+	flags.StringVar(&gcpcfg.GcpProjectId, "gcp-project-id", "", "GCP Project ID")
+	flags.StringVar(&gcpcfg.GcpZone, "gcp-zone", "", "GCP Zone")
+	flags.StringVar(&gcpcfg.ImageId, "imageid", "", "Pod VM image id that is available at GCP Images. Usually a name like 'podvm-image'")
+	flags.StringVar(&gcpcfg.MachineType, "machine-type", "e2-medium", "Pod VM Machine type")
+	flags.StringVar(&gcpcfg.GcpNetworkId, "gcp-network", "default", "GCP Network ID for the VMs")
```
wainersm, line 26: maybe `gcp-network-id` to be aligned with other ID like parameters (e.g. `gcp-project-id`). Although `imageid` breaks the rule too.

src/cloud-providers/gcp/provider.go
wainersm: I don't like the static mapping table on the code too.

Also, the table is ambiguous because according to https://cloud.google.com/confidential-computing/confidential-vm/docs/create-a-confidential-vm-instance#create-instance (and IIUC) N2D machines types can be either SEV or SEV-SNP.

What if we set `DISABLECVM=true` by default, and having the users to switch to `DISABLECVM=false` and fill the fields properly if they want confidential VMs?

src/csi-wrapper/pkg/wrapper/controllerservice.go
```diff
@@ -67,7 +67,7 @@ func NewControllerService(targetEndpoint, namespace string, peerpodvolumeClientS
 }
 
 func (s *ControllerService) redirect(ctx context.Context, req interface{}, fn func(context.Context, csi.ControllerClient)) error {
-	conn, err := grpc.Dial(s.TargetEndpoint, grpc.WithBlock(), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
wainersm, line 70: hmm  afaik we don't have automated tests for the csi-wrapper component. How to test it doesn't introduce a regression?

wainersm:
Hi @beraldoleal , I reviewed this PR fully and I have no further comments. thanks!

src/csi-wrapper/pkg/wrapper/controllerservice.go
beraldoleal: I'm not familiar with the csi-wrapper , so not sure. I just followed the documentation with the new way to do. I tried the code in a standalone method as part of a hello world example, not as part of the csi-wrapper.

src/cloud-providers/gcp/provider.go
beraldoleal: > What if we set DISABLECVM=true by default, and having the users to switch to DISABLECVM=false and fill the fields properly if they want confidential VMs?

The question is how we are going to derive the confidential_type? or you mean letting users to choose the type? With another variable?

src/cloud-providers/gcp/manager.go
beraldoleal: TBH, imo all should be without gcp as prefix. because all are parameters of GCP. Its kind of redundant. So I will probably fix this better in a following PR, if you don't mind.

src/cloud-providers/gcp/provider.go
wainersm: Yes, I thought in have another variable for confidential_type. So for enabling CVM users would need to set the following for SNP for instance:
```
DISABLECVM="false"
GCP_MACHINE_TYPE="n2d-standard-2"
GCP_CONFIDENTIAL_TYPE="SEV_SNP"    # the new variable
```

beraldoleal: I was trying to avoid having another variable, but yes, I'm ok with that. I will send a v2.

beraldoleal:
Thanks @bpradipt and @wainersm . Because of the nature of this PR and the critical changes on csi-wrapper, I will close this one and split the work into smaller PRs. Here is the first one: #2281 

Please review there, when you have some spare cycles.

PR #2268, opened 2025-01-29
podvm: Add --pull option to import.sh and download-image.sh

frankbu:
Currently, download-image.sh uses pull=always when it creates the container for the specified docker image, which fails if the image version is not available in the remote registry. Adding a pull option will allow users to import a docker image that is only available locally by specifying --pull=never.

### DIFF
--- a/src/cloud-api-adaptor/ibmcloud/image/import.sh
+++ b/src/cloud-api-adaptor/ibmcloud/image/import.sh
@@ -12,7 +12,7 @@ error(){
 script_dir=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)
 
 function usage() {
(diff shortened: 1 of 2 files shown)


mkulke (approved):
looks reasonable, although un/wrapping an qcow2 image in an oci image is probably not requried if the qcow2 is available locally to begin with

stevenhorsman:
> looks reasonable, although un/wrapping an qcow2 image in an oci image is probably not requried if the qcow2 is available locally to begin with

Interesting point - import.sh supports a local qcow2 file rather than an oci image, but I guess out packer build approach just copies the qcow2 file from the builder image into the podvm-image rather than to the local file system. We could try and iterate on this, but I guess if mkosi & oras is more strategic it probable isn't worth it.

stevenhorsman (approved):
The code LGTM. Thanks

PR #2269, opened 2025-01-30
ci: add delay on podvm smoke-test

mkulke:
1s delay between claiming an IP address and launching kata-agent is not enough in some cases, resulting in a flaky test.

Changing the delay before attempting to connect to the agent to 5s seems to resolve the issue. Running the test in a loop with 25+ launches didn't produce a failure, while doing the same with the original 1s delay, will result in "connection refused" failure after a couple of runs.


stevenhorsman (approved):
The code looks fine, DCO sign off is expecting `mkulke@gmail.com`, so failing though

beraldoleal (approved):
LGTM, thanks.

PR #2270, opened 2025-01-30
V0.12.0 post release tasks

stevenhorsman:
Update the citation and revert the CAA image pin.
> [!NOTE]
> I've not reverted the operator pin yet, pending slack discussion on the topic


wainersm (approved):
And all e2e (including docker) passing \o/ Deserves a printscreen!

Thanks @stevenhorsman !

PR #2271, opened 2025-01-30
workflow: podvm_mkosi: Update registry

stevenhorsman:
Update the registry for the docker provider image, so that it uses the input variable, rather than being hard-coded.


mkulke (approved):

wainersm (approved):
Good catch!

PR #2272, opened 2025-02-03
Test ubuntu 24.04 base image

stevenhorsman:
We tested 22.04 and it didn't work, but I've heard rumours that 24.04 does, so created this PR to test that out.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -5,6 +5,6 @@ authors:
 title: Cloud API Adaptor
 abstract: Ability to create Kata pods using cloud provider APIs aka the peer-pods approach
 license: Apache-2.0
-version: 0.11.0
-date-released: 2024-11-29
-commit: 7a5ec0e278a20ba0e7d349e6f08c8dc179568333
+version: 0.12.0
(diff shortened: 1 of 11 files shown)


stevenhorsman:
Okay, this doesn't look like it is working: 24.04 based image is still failing with s390x & packer: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/13117485000/job/36595501989

squarti:
@stevenhorsman FYI, I had better in the packer build after changing the builder image to ubuntu 24.04 ([here](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder#L9)).

stevenhorsman:
> @stevenhorsman FYI, I had better luck in the packer build after changing the builder image to ubuntu 24.04 ([here](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder#L9)).

Thanks for the info. I'll give that a try.

PR #2273, opened 2025-02-03
Support external network connectivity for the pod via pod VM network

bpradipt:


src/cloud-api-adaptor/pkg/podnetwork/workernode.go
```diff
@@ -186,6 +189,26 @@ func (n *workerNode) Inspect(nsPath string) (*tunneler.Config, error) {
...
+				continue
+			}
+			route := &tunneler.Route{
+				Dst: prefix,
+				// The gateway address is 0.0.0.0
+				GW:       netip.Addr{},
+				Dev:      podInterface,
+				Scope:    netops.RouteScope(netlink.SCOPE_LINK),
+				Protocol: netops.RouteProtocol(unix.RTPROT_KERNEL),
+			}
```
yoheiueda, line 207: GW should be the actual gateway address that is specified in the default route.

The `findPrimaryInterface` function that is called at Line 129 internally identifies the default route. I think we can modify the function to return the gateway address of the default route.
```
	podInterface, gatewayAddr, err := findPrimaryInterface(podNS)
```
```
			route := &tunneler.Route{
				Dst: prefix,
				GW: gatewayAddr,
				Dev: podInterface,
			}
```

src/cloud-api-adaptor/docs/external-network.md
```diff
@@ -0,0 +1,35 @@
...
+
+From a networking standpoint, this means that there should be bidirectional connectivity between the cloud and your development setup.
+
+One solution is to use a VPN, which complicates the setup. A relatively easier alternative is to use a public IP associated with the cloud instance for control-plane traffic and use the cloud instance network for external connectivity requirements (like accessing cloud object storage for data) instead of routing all traffic via your development setup.
+
+The external network connectivity for the pod via pod VM network is enabled via the
+option `ext-network-via-podvm` in cloud-api-adaptor. The equivalent option in the `peer-pods-cm` is `EXTERNAL_NETWORK_VIA_PODVM`
+
+The prerequisite is for the pod VM to have a secondary interface with an IP. This interface will be moved to the pod network namespace and default routes adjusted so that pod network traverses via worker node, and any other traffic uses the secondary interface.
+
```
yoheiueda, line 17: It might be helpful to mention that secondary NIC is only supported in AWS for now.

yoheiueda (commented):
Sorry for the delay. Overall LGTM. I only have minor comments.

bpradipt:
> Sorry for the delay. Overall LGTM. I only have minor comments.

Thanks @yoheiueda . I updated the code based on your feedback.. Kindly take a re-look

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -4,42 +4,270 @@
...
+			fmt.Printf("Secondary interface %q found with IPv4 address %q and route %v\n", link.Name(), addr, route)
+			return link.Name(), addr, route, nil
+		}
+
+		fmt.Printf("Route is nil, checking if %q is in the same subnet as %q\n", addr, primaryInterface)
+		priAddrCIDR, defRoute, err := getInterfaceDetails(ns, primaryInterface)
+		if err != nil {
+			return "", netip.Prefix{}, nil, err
+		}
+		if priAddrCIDR.Bits() == addr.Bits() {
```
yoheiueda, line 186: Does this condition checks whether the primary and secondary interface are in the same subnet?
```
priAddrCIDR.Bits() == addr.Bits()
``` 
I think `Bits()` returns the lenght of CIDR in bits. For example, this function returns 24 for both`192.168.1.0/24` and `192.168.2.0/24`. But these two subnet are not identical.

I think we can compare two Prefix objects like this.

```
		if priAddrCIDR.Prefix() == addr.Prefix() {
```

yoheiueda (approved):
LGTM. Thank you very much!

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -122,6 +122,9 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		flags.StringVar(&cfg.networkConfig.HostInterface, "host-interface", "", "Host Interface")
 		flags.IntVar(&cfg.networkConfig.VXLAN.Port, "vxlan-port", vxlan.DefaultVXLANPort, "VXLAN UDP port number (VXLAN tunnel mode only")
 		flags.IntVar(&cfg.networkConfig.VXLAN.MinID, "vxlan-min-id", vxlan.DefaultVXLANMinID, "Minimum VXLAN ID (VXLAN tunnel mode only")
+		flags.BoolVar(&cfg.networkConfig.ExternalNetViaPodVM, "ext-network-via-podvm", false, "Enable external networking via pod VM")
+		// Local pod subnets. This will be used by APF to create routes for local pod subnets when using external networking via pod VM
+		flags.Var(&cfg.networkConfig.PodSubnetCIDRs, "pod-subnet-cidrs", "Local pod subnets")
```
beraldoleal, line 127: Maybe this as description: "Comma separated CIDRs for local pod subnets"?

src/cloud-api-adaptor/docs/external-network.md
```diff
@@ -0,0 +1,37 @@
...
+  
+```sh
+SVCRANGE=$(echo '{"apiVersion":"v1","kind":"Service","metadata":{"name":"tst"},"spec":{"clusterIP":"1.1.1.1","ports":[{"port":443}]}}' | kubectl apply -f - 2>&1 | sed 's/.*valid IPs is //')
+echo $SVCRANGE
+```
+
+* The default pod CIDR for Flannel CNI: `10.244.0.0/16`
+* AWS VPC CNI uses VPC CIDR as the pod CIDR
+* Azure CNI uses the subnet CIDR as the pod CIDR
+* The default pod CIDR for OVN-Kubernetes: `10.128.0.0/14`
```
beraldoleal, line 37: Even this is only working for aws, maybe adding this for gcp?

 * GCP GKE with VPC-native: Pods get IPs from a secondary range on the subnet, so the pod CIDR is defined by that secondary range.

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -33,6 +33,8 @@ optionals+=""
 [[ "${SECURE_COMMS_PP_OUTBOUNDS}" ]] && optionals+="-secure-comms-pp-outbounds ${SECURE_COMMS_PP_OUTBOUNDS} "
 [[ "${SECURE_COMMS_KBS_ADDR}" ]] && optionals+="-secure-comms-kbs ${SECURE_COMMS_KBS_ADDR} "
 [[ "${PEERPODS_LIMIT_PER_NODE}" ]] && optionals+="-peerpods-limit-per-node ${PEERPODS_LIMIT_PER_NODE} "
+[[ "${EXTERNAL_NETWORK_VIA_PODVM}" ]] && optionals+="-ext-network-via-podvm  "
+[[ "${POD_SUBNET_CIDRS}" ]] && optionals+="-pod-subnet-cidrs ${POD_SUBNET_CIDRS} "
```
beraldoleal, line 37: If this is only valid for aws, maybe move to aws() method? or maybe add a note to the options descriptions is valid for aws, for now?

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -4,42 +4,270 @@
...
+			continue
+		}
+
+		if route != nil {
+			fmt.Printf("Secondary interface %q found with IPv4 address %q and route %v\n", link.Name(), addr, route)
+			return link.Name(), addr, route, nil
+		}
+
+		fmt.Printf("Route is nil, checking if %q is in the same subnet as %q\n", addr, primaryInterface)
+		priAddrCIDR, defRoute, err := getInterfaceDetails(ns, primaryInterface)
```
beraldoleal, line 182: Unless I'm missing something, looks like this `getInterfaceDetails(ns, primaryInterface)` could be called outside the loop?

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -4,42 +4,270 @@
...
+		if err != nil {
+			return "", netip.Prefix{}, nil, err
+		}
+		if priAddrCIDR.Bits() == addr.Bits() {
+			secIface = link.Name()
+			secRoute = &netops.Route{
+				Destination: defRoute.Destination,
+				Gateway:     defRoute.Gateway,
```
beraldoleal, line 191: Here you're assuming defRoute is not nil. IIUC, the loop on line 137 may not find a default route, and in that case the method returns nil without an error.

Should we check defRoute != nil before using it here to avoid a possible nil dereference?

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -4,42 +4,270 @@
...
+	// Bring up the network interface in the new namespace
+	err = link.SetUp()
+	if err != nil {
+		return fmt.Errorf("failed to bring up link %q: %w", iface, err)
+	}
+
+	// Get existing default route in the new namespace
+	defRoutes, err := dstNs.GetDefaultRoutes()
+	if err != nil {
+		return fmt.Errorf("failed to get default route in namespace %q: %w", dstNs.Path(), err)
```
beraldoleal, line 256: Same here? in case the loop could finish without finding the default routes? 

Some explicit checks like this?

```
if len(defRoutes) == 0 {
	return nil, fmt.Errorf("no default route found")
}
```

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -4,42 +4,270 @@
...
+		}
+	}
+
+	routes, err := ns.RouteList(&netops.Route{Destination: netops.DefaultPrefix})
+	if err != nil {
+		return netip.Prefix{}, nil, fmt.Errorf("failed to get routes on namespace %q: %w", ns.Path(), err)
+	}
+
+	for _, r := range routes {
+		if r.Device == iface && r.Destination.Bits() == 0 && r.Priority < math.MaxInt {
```
beraldoleal, line 138: I find it a bit odd to have the default route logic duplicated here. What do you think about modifying `GetDefaultRoutes` to optionally accept an interface name and apply the filtering inside? That way, this code could reuse `GetDefaultRoutes` ?

src/cloud-api-adaptor/go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.15.11
```
beraldoleal, line 14: Now I found it :)

I believe you could split this PR and send this bump in its own PR to unlock our CI. This is a big PR, l and I'm already missing this bump itself.

src/cloud-api-adaptor/pkg/podnetwork/common.go
```diff
@@ -183,7 +183,8 @@ func getSecondaryInterfaceDetails(ns netops.Namespace, primaryInterface string)
 		if err != nil {
 			return "", netip.Prefix{}, nil, err
 		}
-		if priAddrCIDR.Bits() == addr.Bits() {
+
+		if priAddrCIDR.Masked() == addr.Masked() {
```
beraldoleal, line 187: Please, squash this commit with the one that introduced the Bits().

bpradipt:
@beraldoleal I have created a separate PR with the CI related fixes - https://github.com/confidential-containers/cloud-api-adaptor/pull/2351
I'll address your other comments tomorrow

huoqifeng (approved):
LGTM

src/cloud-api-adaptor/docs/external-network.md
```diff
@@ -0,0 +1,37 @@
+# Introduction
+
+Currently, all pod traffic is routed via the worker node (WN).
+This is by design and is  fine as all existing network policies that were applicable to pods running on the WN also apply to peer pods.
```
huoqifeng, line 4: nit: 2 whitespace between `is   fine`

src/cloud-api-adaptor/pkg/podnetwork/common.go
bpradipt: Yes it can be outside the loop. I kept it inside the loop to call it only if there is a valid secondary interface.

src/cloud-api-adaptor/entrypoint.sh
bpradipt: Yeah good point. I'll adjust it accordingly.

src/cloud-api-adaptor/pkg/podnetwork/common.go
bpradipt: defRoutes is a slice and will return empty slice if no defRoutes exist. So no need to recheck it again.

src/cloud-api-adaptor/pkg/podnetwork/common.go
bpradipt: For now, I'll keep it as is and avoid changing the interfaces since this code is experimental. It's also possible to provide another method as well.  When we work on adding support for other providers, I'll look at refactoring/redesigning it.

src/cloud-api-adaptor/docs/external-network.md
bpradipt: Fixed.

src/cloud-api-adaptor/docs/external-network.md
bpradipt: Yeah.. next phase will be to add it for Azure and GCP

src/cloud-api-adaptor/pkg/podnetwork/common.go
beraldoleal: Not sure I follow your point here. Since ns and primaryInterface don’t change inside the loop, and the earlier conditions only skip some iterations, getInterfaceDetails(ns, primaryInterface) still ends up being called unnecessarily on each pass. Maybe I’m missing something, but as far as I can tell, there’s no condition that prevents it from being called repeatedly.

beraldoleal (commented):
Hi @bpradipt thanks for addressing the comments. I just reopened one thread, before I give my ack here.

beraldoleal (approved):
LGTM, thanks.

bpradipt:
@beraldoleal I made the changes as suggested by you, but didn't test the complete flow.

PR #2278, opened 2025-02-05
Update kbs-client package name for oras

portersrc:
The release process was skipping kbs-client, so I made the fix [in trustee here.](https://github.com/confidential-containers/trustee/pull/691). While fixing it, I noticed the package naming was slightly inconsistent with other packages, so I updated it. I found one place in our code base -- right here in cloud-api-adaptor -- that might be affected. Maybe @stevenhorsman PTAL? I don't know what triggers this script and `oras pull` in cloud-api-adaptor.


stevenhorsman:
Hi @portersrc - thanks for the head's up on this, it would have been a tricky one to debug. We need to put a hold on this as we can't take the change until we update the trustee version to be at/after the commit that your change when it otherwise it will break us.

portersrc:
got it; yeah it seemed to be a minor change but an obnoxious sync'ing problem; sorry for this

stevenhorsman:
Included in #2340

PR #2280, opened 2025-02-05
ibmcloud: fix import.sh bucket selection

frankbu:
Choosing the first bucket in an instance wil not work if the first bucket is not in the target region. This PR fixes the selection logic to choose the first bucket that is in the target region.


stevenhorsman (approved):
The code LGTM. Thanks!

mkulke (approved):

PR #2281, opened 2025-02-05
gcp: allow users to set podvm disk-type and align CLI options

beraldoleal:
This PR allows users to set the PodVM disk type for GCP instead of using the hardcoded pd-standard.

Also cleaned up some CLI options to be more consistent across providers, by dropping extra gcp- prefix.


bpradipt (approved):
/lgtm

snir911 (approved):
LGTM thanks

PR #2289, opened 2025-02-07
ibmcloud: add option to enable podvm confidential computing (tdx)

frankbu:
Added property CONFIDENTIAL_COMPUTING to configure ibmcloud provider
to set confidential_compute_mode=tdx and enable_secure_boot=true
when creating a podvm.

This also required updating the version of vpc_go_sdk from 0.35.0 -> 0.64.0.


src/cloud-providers/ibmcloud/provider.go
```diff
@@ -177,6 +177,13 @@ func (p *ibmcloudVPCProvider) getInstancePrototype(instanceName, userData, insta
 			Enabled:  core.BoolPtr(true),
 			Protocol: core.StringPtr(vpcv1.InstanceMetadataServicePatchProtocolHTTPConst),
 		},
+		ConfidentialComputeMode: core.StringPtr(vpcv1.InstanceConfidentialComputeModeDisabledConst),
+		EnableSecureBoot:        core.BoolPtr(false),
+	}
+
+	if p.serviceConfig.ConfidentialComputing {
```
bpradipt, line 184: Why not use the existing option of disablecvm used with existing providers ?

src/cloud-api-adaptor/go.mod
```diff
@@ -166,7 +166,6 @@ require (
 	github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/oklog/ulid v1.3.1 // indirect
-	github.com/onsi/ginkgo/v2 v2.15.0 // indirect
```
stevenhorsman: Hi Frank, do you know why this was removed here? Should it have been part of the previous commit?

src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_initializer.go
```diff
@@ -56,8 +56,9 @@ type IBMCloudProperties struct {
 	WorkerCount   int
 	IsSelfManaged bool
 
-	VPC        *vpcv1.VpcV1
-	ClusterAPI containerv2.Clusters
+	VPC                   *vpcv1.VpcV1
+	ClusterAPI            containerv2.Clusters
+	ConfidentialComputing bool
```
stevenhorsman, line 61: I can't see in the test code where this is being used? What was the reason for adding this logic now?

src/cloud-api-adaptor/go.mod
frankbu: Yes. I'm not sure what happened the first time. I ran `go mod tidy` before the first commit, but when I created the PR, the lint failed because of this. So I ran `go mod tidy` again and then squashed the commit into the previous one (but I guess that was not the one it should actually have been squashed with). Not sure what to do now. How best to fix this?

stevenhorsman: I think the best way is to edit the first commit and move the change into there and then it should go from this one. You can do that with an `edit` rather than `squash/fixup` in `git rebase -i`, but let me know if you have issues and we can try and work through it together.

src/cloud-providers/ibmcloud/provider.go
frankbu: Good point! I missed that so just made up a property myself. I will change it.

src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_initializer.go
frankbu: I wanted to have an easy way to turn on/off the cc while we're working on the code. Is there another better approach I should be using?

stevenhorsman: So this is in the e2e test code space, so if you are planning to add some e2e tests it's needed, but I don't think you are using this anywhere in any tests?

stevenhorsman (approved):
LGTM.

src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_initializer.go
stevenhorsman: The code's gone now so I can't double check, but my point was that this new `ConfidentialComputing` field you are reading from the properties file didn't seem to be used and wasn't setting the `IBM_PEER_POD_CONFIDENTIAL_VM` or whatever the env you created was as far as I could see, so I'm not sure how it would impact anything. Maybe I'm mis-remembering though

frankbu: I had a line in provision_common.go at line 968 that uses it. It might have been hard to see in the diff because the whole section was updated for formatting.

image

src/cloud-providers/ibmcloud/provider.go
frankbu: Removed this part and will do it properly in a followup PR. This PR is now just updating the dependency.

mkulke (approved):
lgtm

PR #2293, opened 2025-02-11
versions: Attempt to use Ubuntu 24.04 in packer

stevenhorsman:
See if we can bump to a 24.04 base for the packer
podvm image as 20.04 is going EoL

### DIFF
--- a/src/cloud-api-adaptor/Makefile.defaults
+++ b/src/cloud-api-adaptor/Makefile.defaults
@@ -19,7 +19,7 @@ $(or $(shell $(YQ_COMMAND) '.$(1)' $(VERSIONS_SRC)), \
 	$(error $(1) is not set, and could not be automatically set from $(VERSIONS_SRC)))
 endef
 
-ubuntu_RELEASE ?= focal
+ubuntu_RELEASE ?= noble
 rhel_RELEASE ?= 9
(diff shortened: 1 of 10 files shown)


ldoktor (commented):
Looks quite good and seems to work well (only tested x86_64). Could you please split the ubuntu and style changes into 2 commits (in case of reverts)? Also do you know if the versions.yaml is actually used anywhere in the code or is it there just for a reference?

beraldoleal (commented):
Hi @stevenhorsman lgtm, but I found `focal` under `src/cloud-api-adaptor/podvm/qcow2/rhel|ubuntu/variables.pkr.hcl` as default values.

Should we update there too?

stevenhorsman:
> Hi @stevenhorsman lgtm, but I found some `focal` strings under `src/cloud-api-adaptor/podvm/qcow2/rhel|ubuntu/variables.pkr.hcl` as default values.
> 
> Should we update there too?

Yes - we should - thanks for pointing this out - I did a search for 20.04 and fixed the references I would there, but didn't think to search for focal and replace them, so I've done that now and hopefully the tests will still be good.

beraldoleal (approved):
LGTM, thanks

PR #2295, opened 2025-02-11
ibmcloud: add option to enable/disable confidential VM

frankbu:
Added the DISABLECVM property to the ibmcloud adaptor config and to the caa-provisioner-cli properties for ibmcloud.

### DIFF
--- a/src/cloud-api-adaptor/install/overlays/ibmcloud/kustomization.yaml
+++ b/src/cloud-api-adaptor/install/overlays/ibmcloud/kustomization.yaml
@@ -30,6 +30,7 @@ configMapGenerator:
   - IBMCLOUD_VPC_SG_ID="" #set
   - IBMCLOUD_VPC_ID="" #set
   - CRI_RUNTIME_ENDPOINT="/run/cri-runtime/containerd.sock"
+  - DISABLECVM="true" # Set to false to enable confidential VM
   #- PAUSE_IMAGE="" # Uncomment and set if you want to use a specific pause image
   #- TUNNEL_TYPE="" # Uncomment and set if you want to use a specific tunnel type. Defaults to vxlan
   #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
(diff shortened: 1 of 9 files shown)


src/cloud-providers/ibmcloud/provider.go
```diff
@@ -177,6 +177,13 @@ func (p *ibmcloudVPCProvider) getInstancePrototype(instanceName, userData, insta
 			Enabled:  core.BoolPtr(true),
 			Protocol: core.StringPtr(vpcv1.InstanceMetadataServicePatchProtocolHTTPConst),
 		},
+		ConfidentialComputeMode: core.StringPtr("tdx"), // TODO when available: vpcv1.InstanceConfidentialComputeModeTdxConst
```
bpradipt, line 180: Will the ConfidentialComputeMode be always tdx? No other TEEs (eg. snp) ?

frankbu: My understanding is at this point TDX is the only TEE we plan to support in IBM Cloud. In the future we could add another config parameter to support others, but unless I'm missing something, I don't think there would be a backward compatibility issue when we do that.

src/cloud-api-adaptor/install/overlays/ibmcloud/kustomization.yaml
```diff
@@ -30,6 +30,7 @@ configMapGenerator:
   - IBMCLOUD_VPC_SG_ID="" #set
   - IBMCLOUD_VPC_ID="" #set
   - CRI_RUNTIME_ENDPOINT="/run/cri-runtime/containerd.sock"
+  #- DISABLECVM="true" # Uncomment it if you want a generic VM
```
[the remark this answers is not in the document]
frankbu: Good point! TDX is currently only available with very specific profile in certain regions. I think we should uncomment it by default.

src/cloud-providers/ibmcloud/provider.go
frankbu: I confirmed that TDX is all that IBM Cloud plans to support for a while, at least through 2025.

src/cloud-api-adaptor/install/overlays/ibmcloud/kustomization.yaml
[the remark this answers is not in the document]
frankbu: done

bpradipt (approved):
/lgtm
Thanks @frankbu

PR #2301, opened 2025-02-20
caa: Enable running without Kubernetes

spotlesstofu:
Log the error instead of exiting when Kubernetes configuration is missing.

Similar to what already happens in `cloud.go`:
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/pkg/adaptor/cloud/cloud.go#L137

### DIFF
--- a/src/cloud-api-adaptor/pkg/adaptor/server.go
+++ b/src/cloud-api-adaptor/pkg/adaptor/server.go
@@ -78,9 +78,11 @@ func (s *server) Start(ctx context.Context) (err error) {
 		}
 	}
 	// Advertise node resources
-	err = k8sops.AdvertiseExtendedResources(s.PeerPodsLimitPerNode)
-	if err != nil {
-		return err
+	if k8sops.IsKubernetesEnvironment() {
+		err = k8sops.AdvertiseExtendedResources(s.PeerPodsLimitPerNode)
+		if err != nil {
+			return err
+		}
 	}
 
 	ttRpc, err := ttrpc.NewServer()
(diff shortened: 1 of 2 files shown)


src/cloud-api-adaptor/pkg/adaptor/server.go
```diff
@@ -80,7 +80,7 @@ func (s *server) Start(ctx context.Context) (err error) {
 	// Advertise node resources
 	err = k8sops.AdvertiseExtendedResources(s.PeerPodsLimitPerNode)
 	if err != nil {
-		return err
+		logger.Printf("failed AdvertiseExtendedResources: %s", err)
```
wainersm, line 83: Hi @spotlesstofu !

I'm afraid that logging a message instead of return an error it will be masking a fatal failure, when running with Kubernetes.

I looked at `AdvertiseExtendedResources()` implementation, in the line https://github.com/confidential-containers/cloud-api-adaptor/blob/5708832ad53c8450baa314bc71c059de4ba7d433/src/cloud-api-adaptor/pkg/adaptor/k8sops/node.go#L28, it gets the `NODE_NAME` which should be non-empty when running from k8s. What if you put a guard there, so that if `NODE_NAME` is empty then it returns `nil` (logging a message too)?

spotlesstofu: Thanks @wainersm.

Building on your feedback, I introduced a function to gate the Kubernetes-specific feature.

What do you think?

wainersm (approved):
Hi @spotlesstofu ,

I don't know of any alternative way of checking whether is k8s env or not. I think that checking NODE_NAME should be enough for caa.

bpradipt (approved):
/lgtm

bpradipt:
@spotlesstofu can you rebase and push. It'll fix the govulncheck action

PR #2306, opened 2025-02-25
ibmcloud: Use beta go vpc sdk to enable TDX configuration

frankbu:
The beta version of the go vpc sdk needs to be used to configure TDX confidential compute mode because it is currently a beta feature.

The TDX feature GA is March 27th, and the SDK update about a week or so after that. We will switch back to the SDK as soon as the TDX support is available.


stevenhorsman:
It might be helpful to split out the code changes into a separate commit from the go.mod/go.sum ones as then once we have the new version of the VPC SDK we can just revert the code changes commit?

frankbu:
> It might be helpful to split out the code changes into a separate commit from the go.mod/go.sum ones as then once we have the new version of the VPC SDK we can just revert the code changes commit?

I can split it out if you feel strongly that I should, but I'm happy to do it manually in April - the way it is, it will take me less than 5 mins to change back (I can find all the places with one simple search).

The downside of leaving both SDKs around is that it will be harder to make sure all the places are changed. I see I missed a couple of places that are failing lint test - I missed them because I assumed that once `make all` worked, I had compiled all the code, but looks like it didn't.

Anyway, let me know if you want two PRs - if not, I'll fix up the remaining lint errors in this one. Thanks

stevenhorsman:
I force pushed to remove the merge commit

stevenhorsman (approved):
Looks fine to me. Thanks @frankbu

mkulke (approved):

PR #2307, opened 2025-02-26
versions: Bump golang.org/x/crypto

stevenhorsman:
Bump golang.org/x/crypto to fix GO-2025-3487


stevenhorsman:
This has got more complicated.
Do to the major release bump I think we need to bump the builder image first this time, so I've created https://github.com/confidential-containers/cloud-api-adaptor/pull/2308 to try and do this.

wainersm:
Hi @stevenhorsman !

> This has got more complicated. Do to the major release bump I think we need to bump the builder image first this time, so I've created #2308 to try and do this.

Then update the builder version in https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Dockerfile#L2 ?

src/cloud-api-adaptor/test/e2e/assessment_helpers.go
```diff
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
wainersm, line 16: I think you did, but double-checking: compiled e2e tests for each provider? (this is something good to have on CI too).

stevenhorsman: I ran local linting, but haven't compiled them for each provider, so I can try that out

wainersm (approved):
Go version everywhere! I don't think you need to change any other files, at least I couldn't find any. Thanks sir!

beraldoleal (approved):
LGTM, thanks. I will rebase the other PR after this get merged.

PR #2308, opened 2025-02-26
version: Bump golang version in builder

stevenhorsman:
We need to bump golang to pick up the fix
to x/crypto which requires golang 1.23.

Due to the change in major release version our
builds for components can't run on 1.22.x, so we
need to lead with the builder bump and then follow with the rest.


stevenhorsman:
We are going to need to accept the failing govulncheck for this one as we need to bump the builder image in order to bump the builds of components to use go 1.23 in order to bump x/crypto to 0.35 and resolve the `govulncheck` failure. I'm not sure if this is a reason to stop that check being required, or just to use admin rights to merge this once approved, I think the latter is probably the way to go for now.

wainersm:
> We are going to need to accept the failing govulncheck for this one as we need to bump the builder image in order to bump the builds of components to use go 1.23 in order to bump x/crypto to 0.35 and resolve the `govulncheck` failure. I'm not sure if this is a reason to stop that check being required, or just to use admin rights to merge this once approved, I think the latter is probably the way to go for now.

IMO we should skip the checker, push as admin.

beraldoleal (approved):
LGTM, thanks.

PR #2309, opened 2025-02-27
workflow: e2e_libvirt running out of space

stevenhorsman:
We are seeing the e2e_libvirt test jobs running out of space on the runner. From a quick google, there are a few suggested directories that can be removed to help with this.

### DIFF
--- a/.github/workflows/e2e_libvirt.yaml
+++ b/.github/workflows/e2e_libvirt.yaml
@@ -70,6 +70,16 @@ jobs:
         run: |
           ./hack/ci-helper.sh rebase-atop-of-the-latest-target-branch
 
+      - name: Remove unnecessary directories to free up space
+        run: |
+          sudo rm -rf /usr/local/.ghcup
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          sudo rm -rf /usr/local/lib/android/sdk/ndk
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /usr/local/share/boost
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+
       - name: Read properties from versions.yaml
         run: |
           sudo snap install yq


stevenhorsman:
Tested https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/13564653973

mkulke (approved):

wainersm:
Hi @stevenhorsman !

Apart from that, we could create the cluster's workers with smaller disks. I've tested it on my work to introduce a job to AWS, apparently it works fine:
```
commit d615b325b67b0620ec1782d2ba07ed3636c6b5a8
Author: Wainer dos Santos Moschetta <wainersm@redhat.com>
Date:   Thu Feb 6 16:00:44 2025 -0300

    workflows/e2e_aws: create cluster using small disk
    
    Create the kcli k8s cluster with nodes using 15G of disk instead
    of the default 20G, this way the github runner doesn't run out of
    disk space easily.
    
diff --git a/.github/workflows/e2e_aws.yaml b/.github/workflows/e2e_aws.yaml
index dcb1fb50..a66e4693 100644
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -157,6 +157,9 @@ jobs:
           echo "::endgroup::"
           export CONTAINER_RUNTIME=${{ inputs.container_runtime }}
           echo "::group::Create cluster with $CONTAINER_RUNTIME"
+          # Use less space as possible so the Git runner doesn't run out of
+          # disk space.
+          CLUSTER_DISK_SIZE=15
           ./libvirt/kcli_cluster.sh create
           echo "KUBECONFIG=$HOME/.kcli/clusters/peer-pods/auth/kubeconfig" >> "$GITHUB_ENV"
           echo "::endgroup::"
```

I can have the same change for the e2e_libvirt if you think it's a good idea.

wainersm (approved):
Thanks @stevenhorsman !

stevenhorsman:
> I can have the same change for the e2e_libvirt if you think it's a good idea.

Yeah, I think that's also a good idea. Once we have support for variable resource libvirt podvm's I'm also planning to reduce the memory and cpu requests for hopefully better stability.

stevenhorsman:
The govuln is failing, so we are waiting on https://github.com/confidential-containers/cloud-api-adaptor/pull/2307 to get merged and then can rebase.

ldoktor (commented):
lgtm but I think it'd make sense to inspire from our repos. We have similar but slightly different fix in the https://github.com/confidential-containers/operator/blob/main/.github/workflows/ccruntime_e2e.yaml#L72

PR #2312, opened 2025-03-03
workflow:e2e_run_all: Switch s390x image build runner

stevenhorsman:
With a recent change in the s390x CoCo runners, we now can only build the mkosi image on larger s390x runner (that runs on a different physical system), so update the workflow to reflect this.

### DIFF
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -238,7 +238,7 @@ jobs:
       dev_tags: ${{ inputs.caa_image_tag }}-s390x-dev
       release_tags: ${{ inputs.caa_image_tag }}-s390x
       git_ref: ${{ inputs.git_ref }}
-      runner: 's390x'
+      runner: s390x-large
     secrets: inherit
 
   libvirt_e2e_arch_prep:


BbolroC (approved):
LGTM thanks @stevenhorsman !

ldoktor (approved):
seems fine, but perhaps it'd be nice to include the reason in the commit message so we can revert in case the build becomes possible on smaller runners again, what do you think? (I'm actually wondering myself why it requires larger machine now...)

stevenhorsman:
> seems fine, but perhaps it'd be nice to include the reason in the commit message so we can revert in case the build becomes possible on smaller runners again, what do you think? (I'm actually wondering myself why it requires larger machine now...)

That's a fair request. I might need to lean on @BbolroC for the root cause of the marist systems not being able to build the image now?

BbolroC:
> > seems fine, but perhaps it'd be nice to include the reason in the commit message so we can revert in case the build becomes possible on smaller runners again, what do you think? (I'm actually wondering myself why it requires larger machine now...)
> 
> That's a fair request. I might need to lean on @BbolroC for the root cause of the marist systems not being able to build the image now?

It is explained https://github.com/confidential-containers/cloud-api-adaptor/issues/2310. Thanks!

stevenhorsman:
@ldoktor - I've updated the commit message to include the information from Choi. Thanks for pushing for more clarity here and let me know if that helps.

ldoktor (approved):
Thanks, perhaps we could even add an issue to roll back when the #2310 is resolved.

PR #2315, opened 2025-03-03
adaptor: log error causing agent proxy connection failure

frankbu:
Print the error instead of the address in the log message when retrying a failed agent proxy connection attempt.


stevenhorsman (commented):
Hi Frank, this code impacts all the cloud providers, so if you could update the commit message header to have the subsystem of `adaptor` rather than `ibmcloud` that would be great.

frankbu:
> The commit message needs to be changed as mentioned by @stevenhorsman

Done. Sorry about that ... that was because of copy-paste my standard commit message template 😄

stevenhorsman (approved):
LGTM. Thanks!

PR #2317, opened 2025-03-04
podvm_mkosi: Use container to run mkosi on s390x

BbolroC:
This PR runs mkosi inside a Fedora-based container for building images. This would provide a more consistent environment and prevent issues caused by upgrading Ubuntu on GHA self-hosted runners.

The change has been verified on a Ubuntu 24.04 runner: https://github.com/BbolroC/cloud-api-adaptor/actions/runs/13654124912/job/38169137973

I confirmed that the change works on both Ubuntu 22.04 and Ubuntu 24.04.

Fixes: #2310

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Makefile
+++ b/src/cloud-api-adaptor/podvm-mkosi/Makefile
@@ -30,6 +30,14 @@ ifeq ($(ARCH),s390x)
 YQ_CHECKSUM = $(YQ_CHECKSUM_s390x)
 endif
 
+define run_mkosi_in_container
+	docker run --rm --privileged \
+		-v "$(shell pwd)":/workspace \
+		-w /workspace \
+		fedora:40 \
+		bash -c "dnf install -y mkosi && mkosi --tools-tree=default --tools-tree-release=40 $(1)"
+endef
+
 binaries:
 	@echo "Building binaries..."
 	rm -rf ./resources/binaries-tree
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):
It would be nice to find a way to reduce some of the duplication here, but that's an issue for another PR, so LGTM. Thanks!

BbolroC:
> It would be nice to find a way to reduce some of the duplication here, but that's an issue for another PR, so LGTM. Thanks!

yeah, I will make the duplicate code as a function. thanks for the feedback.

src/cloud-api-adaptor/podvm-mkosi/Makefile
```diff
@@ -71,10 +79,10 @@ image: insecure-builder
 	rm -rf ./build
 	@echo "Building image..."
 ifeq ($(SE_BOOT),true)
-	sudo mkosi --profile production --image system
+	$(call run_mkosi_in_container,--profile production --image system)
```
stevenhorsman, line 82: Sorry, I know this wasn't your change, but I think the SE_BOOT prod version should have `--tools-tree=default --tools-tree-release=40` too, so if you could fix that, it would be great.

BbolroC: yes, sir!

stevenhorsman: Thank you!

BbolroC:
The CI failure (e2e tests on amd64) has nothing to do with this change. Thanks!

liudalibj (approved):
LGTM

PR #2319, opened 2025-03-06
ibmcloud: add another security group in ROKS demo instructions

frankbu:
IBM Cloud doesn't allow traffic between a cluster and its VPC by default anymore, so we need to add another security group to allow it to communicate with the peer pod VSIs.


stevenhorsman (approved):
LGTM

bpradipt (approved):
/lgtm

PR #2320, opened 2025-03-10
Fixes for RHEL/s390x PODVM builds

ifireball:
- Bump protoc to 3.16.0 because 3.15.0 is not available on s390x
- Ensure jq is installed in the builder image
- Ensure yq is in PATH in the builder image
- Ensure subscription credentials are passed to packer when running in a container

### DIFF
--- a/src/cloud-api-adaptor/podvm/Dockerfile.podvm.rhel
+++ b/src/cloud-api-adaptor/podvm/Dockerfile.podvm.rhel
@@ -13,6 +13,8 @@ FROM ${BUILDER_IMG} AS podvm_builder
 ARG CLOUD_PROVIDER
 ARG PODVM_DISTRO=rhel
 ARG UEFI=false
+ARG ORG_ID
+ARG ACTIVATION_KEY
 
 # If not provided, uses system architecture
 ARG ARCH
@@ -57,7 +59,7 @@ COPY . /src
 
 WORKDIR /src/cloud-api-adaptor/podvm
 
-RUN LIBC=gnu make image
+RUN LIBC=gnu ORG_ID=${ORG_ID} ACTIVATION_KEY=${ACTIVATION_KEY} make image
 
 # The below instructions can be used if you prefer to rebuild all the binaries
 #RUN make binaries
--- a/src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder.rhel
+++ b/src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder.rhel
@@ -42,7 +42,7 @@ RUN subscription-manager repos --enable codeready-builder-for-rhel-9-${ARCH/amd6
     dnf groupinstall -y 'Development Tools'; \
     dnf install -y yum-utils gnupg git --allowerasing curl pkg-config clang perl libseccomp-devel gpgme-devel \
     device-mapper-devel qemu-kvm unzip wget libassuan-devel genisoimage cloud-utils-growpart cloud-init \
-    perl-FindBin openssl-devel tpm2-tss-devel
+    perl-FindBin openssl-devel tpm2-tss-devel jq
 
 RUN curl -L -o /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_${YQ_ARCH} \
     && echo "${YQ_CHECKSUM#sha256:}  /usr/local/bin/yq" | sha256sum -c -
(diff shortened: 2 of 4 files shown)


stevenhorsman:
@ifireball - your commit doesn't meet the community's [required format](https://github.com/confidential-containers/confidential-containers/blob/main/CONTRIBUTING.md#commit-format), so I would suggest something like:
```
podvm_builder: Fixes for RHEL/s390x
- Bump protoc to 3.16.0 because 3.15.0 is not available on s390x
- Ensure jq is installed in the builder image
- Ensure yq is in PATH in the builder image

<Your DCO sign-off> 
```

ldoktor:
> @ifireball - your commit doesn't meet the community's [required format](https://github.com/confidential-containers/confidential-containers/blob/main/CONTRIBUTING.md#commit-format), so I would suggest something like:
> 
> ```
> podvm_builder: Fixes for RHEL/s390x
> - Bump protoc to 3.16.0 because 3.15.0 is not available on s390x
> - Ensure jq is installed in the builder image
> - Ensure yq is in PATH in the builder image
> 
> <Your DCO sign-off> 
> ```

Yep, let me just add that whenever one needs to use itemization or the word "AND" it usually means the patch should be split into individual changes. So I'd actually suggest 3 commits (which might sound like an overkill, but in the long-run such commits are easier reviewed, better revertable and usually well received).

ifireball:
@ldoktor I was debating with myself whether to go for separate commits given that the changes are quite small and simple. Would you say I need to go for multiple PRs as well or will separate commits within the same PR be sufficient?

src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder.rhel
```diff
@@ -50,7 +50,7 @@ RUN chmod a+x /usr/local/bin/yq && \
     curl https://dl.google.com/go/go${GO_VERSION}.linux-${YQ_ARCH}.tar.gz -o go${GO_VERSION}.linux-${YQ_ARCH}.tar.gz && \
     rm -rf /usr/local/go && tar -C /usr/local -xzf go${GO_VERSION}.linux-${YQ_ARCH}.tar.gz && \
     rm -f go${GO_VERSION}.linux-${YQ_ARCH}.tar.gz
-ENV PATH="/usr/local/go/bin:${PATH}"
+ENV PATH="/usr/local/go/bin:/usr/local/bin${PATH}"
```
ldoktor, line 53: You're missing the `:` separator between `/usr/local/bin` and `${PATH}`.

src/cloud-api-adaptor/versions.yaml
```diff
@@ -27,7 +27,7 @@ tools:
   golang: 1.23.6
   kcli: 99.0.202408152044
   mkosi: v22
-  protoc: 3.15.0
+  protoc: 3.16.0
```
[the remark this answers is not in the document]
ifireball: I really don't know what such a big bump would imply/cause, we seem to be using a fairly old version, I just assumed there was a good reason for that.

src/cloud-api-adaptor/podvm/Dockerfile.podvm.rhel
```diff
@@ -57,7 +59,7 @@ COPY . /src
 
 WORKDIR /src/cloud-api-adaptor/podvm
 
-RUN LIBC=gnu make image
+RUN LIBC=gnu ORG_ID=${ORG_ID} ACTIVATION_KEY=${ACTIVATION_KEY} make image
```
ldoktor, line 62: I see, we do have it in `Dockerfile.podvm_builder.rhel` but it was missing in here. Nice...

ldoktor (commented):
Thanks, Barak, looks nicer now. I only found one missing `:` but apart from that it looks good.

src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder.rhel
ifireball: will fix, saw typos in the commit messages so a good chance to fix those too

src/cloud-api-adaptor/podvm/Dockerfile.podvm.rhel
ifireball: yep...

src/cloud-api-adaptor/podvm/Dockerfile.podvm_builder.rhel
ifireball: fixed.

src/cloud-api-adaptor/versions.yaml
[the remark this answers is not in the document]
ldoktor: Sure and I agree with your approach. This note was more towards maintainers as I noticed the version is fairly old.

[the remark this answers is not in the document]
stevenhorsman: Yeah, I think this makes sense - lets get this PR in and then we can see about bumping the version after.

src/cloud-api-adaptor/versions.yaml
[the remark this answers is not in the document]
ifireball: fixed.

PR #2322, opened 2025-03-10
GHA: Use 390x runners again for building libvirt test artifacts

BbolroC:
As mentioned in #2310, this commit reverts #2312 and #2313 to use all `s390x` runners again for building libvirt test artifacts.

FYI: the libvirt e2e test should run on the `s390x-large` runner due to resource constraints. 

CC: @stevenhorsman @ldoktor

### DIFF
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -238,7 +238,7 @@ jobs:
       dev_tags: ${{ inputs.caa_image_tag }}-s390x-dev
       release_tags: ${{ inputs.caa_image_tag }}-s390x
       git_ref: ${{ inputs.git_ref }}
-      runner: s390x-large
+      runner: 's390x'
     secrets: inherit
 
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@BbolroC - do you think this still has a dependency on #2321, or can it go in beforehand?

stevenhorsman:
> > @BbolroC - do you think this still has a dependency on #2321, or can it go in beforehand? Maybe we should just add the `git revert` of those commits into that PR, if it's tied?
> 
> Yeah, we can revert them independently of #2321 because the PR for kcli is all about the e2e test this PR does not cover. When it comes to Ubuntu update, the manual patch for kcli on the s390x runner looks working well.
> 
> I've re-pushed two reverted commits. Thanks!

I think you need `git revert -S` (or to manually edit the commit message) so that it is still signed-off, which I'll admit may not make that much sense!

stevenhorsman (approved):
LGTM. Thanks!

PR #2329, opened 2025-03-13
tests/e2e: Add option to override CAA namespace

stevenhorsman:
I've realised that some downstream projects might run the cloud-api-adaptor in a custom namespace, so allow an option to override the default `confidential-containers-system` via a new `TEST_CAA_NAMESPACE` env


stevenhorsman:
@BbolroC - what version of yq is install on the s390x runner as we are seeing:
```
unsupported yq version
```
here

BbolroC:
> @BbolroC - what version of yq is install on the s390x runner as we are seeing:
> 
> ```
> unsupported yq version
> ```
> 
> here

v4.44.5 by snap. Do I have to delete this and make the system depend on the CI job for installation?

stevenhorsman:
> v4.44.5 by snap. Do I have to delete this and make the system depend on the CI job for installation?

No that should be fine as we check:
```
if yq --version | grep '^.* version v4.*$' > /dev/null; then
```
which should have matched?

BbolroC:
I think I know a reason for this. Let me fix it first. Please do not re-trigger the test. 😉

stevenhorsman:
@bpradipt @wainersm - this feature might be useful to you?

bpradipt:
@stevenhorsman is this relevant only when `TEST_PROVISION=yes` ?

stevenhorsman:
> @stevenhorsman is this relevant only when `TEST_PROVISION=yes` ?

No? Sorry, I don't understand the question

bpradipt:
> > @stevenhorsman is this relevant only when `TEST_PROVISION=yes` ?
> 
> No? Sorry, I don't understand the question

So I used the existing test framework to run the tests against an existing OpenShift cluster and didn't face any issues. So I was curious if this PR is specific for the scenario where CAA is deployed by the test framework. My question was without looking at the PR. I looked at it now and think I understood why I could run the tests against my existing cluster.

bpradipt (approved):
/lgtm
nit: probably adding TEST_CAA_NAMESPACE to README will be useful

stevenhorsman:
> So I used the existing test framework to run the tests against an existing OpenShift cluster and didn't face any issues. So I was curious if this PR is specific for the scenario where CAA is deployed by the test framework. My question was without looking at the PR. I looked at it now and think I understood why I could run the tests against my existing cluster.

Ok, I'm guessing that you didn't run any tests that required checking the CAA pod logs for success & failure messages as that was hard-coded and getting debug info for the CAA in case of failures also wouldn't work?

> nit: probably adding TEST_CAA_NAMESPACE to README will be useful

I deliberately avoided this as I didn't necessarily want to complicate upstream docs with downstream details, but I can change this if you disagree?

Thanks!

bpradipt:
> > So I used the existing test framework to run the tests against an existing OpenShift cluster and didn't face any issues. So I was curious if this PR is specific for the scenario where CAA is deployed by the test framework. My question was without looking at the PR. I looked at it now and think I understood why I could run the tests against my existing cluster.
> 
> TBH I'm not expecting downstream to install via the e2e test code as their deployment is not likely to be using the operator, but I wanted to modify that code for completeness rather than having hard-coded strings in multiple places. I'm guessing that you didn't run any tests that required checking the CAA pod logs for success & failure messages as that was hard-coded and getting debug info for the CAA in case of failures also wouldn't work?
>

 Right

> 
> > nit: probably adding TEST_CAA_NAMESPACE to README will be useful
> 
> I deliberately avoided this as I didn't necessarily want to complicate upstream docs with downstream details, but I can change this if you disagree?
> 

My thinking behind the comment:  If this is documented then people like me who occasionally use upstream CI tests against existing setups (both downstream and upstream) will know what options exists :-) 
But this is upto you. I don't have a hard opinion on this. 

> Thanks!

stevenhorsman:
> My thinking behind the comment: If this is documented then people like me who occasionally use upstream CI tests against existing setups (both downstream and upstream) will know what options exists :-) But this is upto you. I don't have a hard opinion on this.

Updated - thanks for the suggestion and explanation

bpradipt (approved):

mkulke (approved):

PR #2303, opened 2025-02-20
mkosi: fix Fedora GPG key issue in mkosi build

beraldoleal:
Attempt to bump to F40. Since F39 is EOL since 2024-11-26. Which will conflict with 165e989

PR #2330, opened 2025-03-13
docker: bump fedora image to 41

spotlesstofu:
- Bump image tag.
- Update dnf commands to the newer syntax.
- Add necessary dependencies that are not included in the new image.

### DIFF
--- a/hack/Dockerfile.golang
+++ b/hack/Dockerfile.golang
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.5-labs
 
-ARG BASE_IMAGE=registry.fedoraproject.org/fedora:40
+ARG BASE_IMAGE=registry.fedoraproject.org/fedora:41
 FROM --platform=$TARGETPLATFORM ${BASE_IMAGE} AS base
 
 # DO NOT UPDATE THIS BY HAND !!
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -1,6 +1,6 @@
 ARG BUILD_TYPE=dev
 ARG BUILDER_BASE=quay.io/confidential-containers/golang-fedora:1.23.8-40
-ARG BASE=registry.fedoraproject.org/fedora:40
+ARG BASE=registry.fedoraproject.org/fedora:41
 
 # This dockerfile uses Go cross-compilation to build the binary,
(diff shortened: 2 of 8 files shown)


stevenhorsman:
Hey @spotlesstofu - just to confirm that whilst this PR bumps a bunch of base images across different build environments, it's not bumping the OS that mkosi is built on in case you intended that? Also we will need a follow up bump of the builder base to `quay.io/confidential-containers/golang-fedora:1.23.6-41` once it's built in a separate PR.

stevenhorsman:
> Hi @stevenhorsman, does my latest commit address your comment?
> 
> About the follow-up pull, I can commit to that.

So I'd still like to check if you intend to bump the version of fedora that is the mkosi base, or just that we use in builds?

spotlesstofu:
Yes, my main goal is to bump the version that ends up in the PodVM image. That's the "mkosi base" I guess.

stevenhorsman:
> Yes, my main goal is to bump the version that ends up in the PodVM image. That's the "mkosi base" I guess.

I think you are missing some places then. https://github.com/confidential-containers/cloud-api-adaptor/pull/2303/commits/ca860109057b2a2e1a1e4d217df28b0f4a4fc0f6 might be a useful reference?

spotlesstofu:
Indeed, yes. Thank you. I updated the commit.

mkulke:
> Yes, my main goal is to bump the version that ends up in the PodVM image. That's the "mkosi base" I guess.

in this case you also need to adjust the fedora versions in the mkosi configuration: `git grep 40 src/cloud-api-adaptor/podvm-mkosi/`, the "base images" are just the docker images in which the build tool is invoked.

note that some packages might not be available anymore or renamed in fedora 41. The last time I looked into this e.g. the neofetch package was dropped. This is arguably not critical, but we lose the sparkling-pretty coco login banner on debug builds 😿 

maybe someone feels like drawing a new one for [fastfetch](https://github.com/fastfetch-cli/fastfetch)?

stevenhorsman:
As @spotlesstofu said they don't have capacity to work on this and part of this has been completed in #2419, I've take over this and rebased it on main, so it should just be focused on bumping the builder now that the base image bump has already happened.

mkulke (approved):

bpradipt (approved):

PR #2332, opened 2025-03-14
e2e: fix initdata templating

mkulke:
The cert values weren't set in all places where initdata is used, which broke e2e tests on azure. The templating was put into its own fn to make this more obvious.


stevenhorsman (approved):
LGTM. Thanks for the fix @mkulke!

bpradipt (approved):
/lgtm

PR #2337, opened 2025-03-18
tests/e2e : Add initdata related test cases, invalied initdata 

rafsalr:
Adding an end-to-end (E2E) test case to validate system behavior when invalied initdata body is provided.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/common.go
+++ b/src/cloud-api-adaptor/test/e2e/common.go
@@ -86,14 +86,15 @@ default WaitProcessRequest := true
 default WriteStreamRequest := true
 '''
 `
+var testEmptyInitdata string = " "
 
-func buildInitdataBody(kbsEndpoint string) (string, error) {
+func buildInitdataBody(kbsEndpoint string, testInitdataVal string) (string, error) {
 	content, err := os.ReadFile("../trustee/kbs/config/kubernetes/base/https-cert.pem")
 	if err != nil {
 		return "", err
 	}
 	certContent := string(content)
-	body := fmt.Sprintf(testInitdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
+	body := fmt.Sprintf(testInitdataVal, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
 	return body, nil
 }
 
@@ -254,7 +255,7 @@ func WithInitdata(kbsEndpoint string) PodOption {
 			p.ObjectMeta.Annotations = make(map[string]string)
 		}
 		key := "io.katacontainers.config.runtime.cc_init_data"
-		initdata, err := buildInitdataBody(kbsEndpoint)
+		initdata, err := buildInitdataBody(kbsEndpoint, testInitdata)
 		if err != nil {
 			log.Fatalf("failed to build initdata %s", err)
 		}
@@ -379,8 +380,8 @@ func NewBusyboxPodWithName(namespace, podName string) PodOrError {
 	return fromPod(NewPod(namespace, podName, "busybox", busyboxImage, WithCommand([]string{"/bin/sh", "-c", "sleep 3600"})))
 }
 
-func NewBusyboxPodWithNameWithInitdata(namespace, podName string, kbsEndpoint string) PodOrError {
-	initdata, err := buildInitdataBody(kbsEndpoint)
+func NewBusyboxPodWithNameWithInitdata(namespace, podName string, kbsEndpoint string, testInitdataVal string) PodOrError {
+	initdata, err := buildInitdataBody(kbsEndpoint, testInitdataVal)
 	if err != nil {
 		return fromError(err)
 	}
--- a/src/cloud-api-adaptor/test/e2e/common_suite.go
+++ b/src/cloud-api-adaptor/test/e2e/common_suite.go
@@ -595,7 +595,7 @@ func DoTestSealedSecret(t *testing.T, e env.Environment, assert CloudAssert, kbs
 // as test cases might be run in parallel
 func DoTestKbsKeyRelease(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint, resourcePath, expectedSecret string) {
 	t.Log("Do test https kbs key release")
-	pod := NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-key-release", kbsEndpoint).GetPodOrFatal(t)
+	pod := NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-key-release", kbsEndpoint, testInitdata).GetPodOrFatal(t)
 	testCommands := []TestCommand{
 		{
 			Command:       []string{"wget", "-q", "-O-", "http://127.0.0.1:8006/cdh/resource/" + resourcePath},
@@ -619,7 +619,7 @@ func DoTestKbsKeyRelease(t *testing.T, e env.Environment, assert CloudAssert, kb
 // as test cases might be run in parallel
 func DoTestKbsKeyReleaseForFailure(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint, resourcePath, expectedSecret string) {
 	t.Log("Do test kbs key release failure case")
-	pod := NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-failure", kbsEndpoint).GetPodOrFatal(t)
+	pod := NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-failure", kbsEndpoint, testInitdata).GetPodOrFatal(t)
 	testCommands := []TestCommand{
 		{
 			Command:       []string{"wget", "-q", "-O-", "http://127.0.0.1:8006/cdh/resource/" + resourcePath},
@@ -792,4 +792,11 @@ func DoTestPodWithCpuMemLimitsAndRequests(t *testing.T, e env.Environment, asser
 	// Custom resource added as req/limit - "kata.peerpods.io/vm
 
 	NewTestCase(t, e, "PodWithCpuMemLimitsAndRequests", assert, "Pod with cpu and memory limits and requests").WithPod(pod).Run()
+}
+
+// Test to check scenarios where initdata body is empty
+func DoTestPodWithoutInitdataAnnotations(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint string) {
+	pod := NewBusyboxPodWithNameWithInitdata(E2eNamespace, "podwithout-initdata", kbsEndpoint, testEmptyInitdata).GetPodOrFatal(t)
+	expectedErrorString := "unmarshalling initdata: toml: invalid character at start of key"
+	NewTestCase(t, e, "PodwithoutInitdataAnnotations", assert, "PodVM without Initdata Annotation is created").WithPod(pod).WithExpectedPodEventError(expectedErrorString).WithCustomPodState(v1.PodPending).Run()
 }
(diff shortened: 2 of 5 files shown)


src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -768,3 +768,21 @@ func DoTestPodWithCpuMemLimitsAndRequests(t *testing.T, e env.Environment, asser
...
+		"io.katacontainers.config.runtime.cc_init_data": initdataString,
+	}
+
+	pod := NewPod(E2eNamespace, podName, containerName, imageName, WithCommand([]string{"/bin/sh", "-c", "sleep 3600"}), WithAnnotations(annotationData))
+
+	t.Logf("Details of pod edited : RR")
+	t.Logf("Pod details: %+v", pod)
+	t.Logf("Pod details 2: %+v", pod.GetAnnotations())
+	t.Logf("Details of pod edited : RR")
+	t.Logf("Pod details 3: %+v", pod.ObjectMeta.Annotations)
```
stevenhorsman, line 793: Hi Rafsal, this looks like your debug you've left in? Is this PR completed? If not please mark it as draft.

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -273,6 +273,18 @@ func WithInitContainers(initContainers []corev1.Container) PodOption {
 	}
 }
 
+func WithEmptyInitdata(kbsEndpoint string) PodOption {
```
stevenhorsman, line 291: I guess I'm missing something but what about this is empty initdata as you are creating it here?

src/cloud-api-adaptor/test/e2e/assessment_runner.go
```diff
@@ -169,6 +170,11 @@ func (tc *TestCase) WithNydusSnapshotter() *TestCase {
 	return tc
 }
 
+func (tc *TestCase) WithExpectedInitdataAnnotation(initdataString string) *TestCase {
```
stevenhorsman, line 179: What is the purpose of this function?

stevenhorsman (commented):
Hi Rafsal, we need some more explanation of what you are trying to achieve with this PR, can you spell out the tests cases in the commit message please. It also looks partially developed, so please mark it as draft if this is the case?

src/cloud-api-adaptor/test/e2e/common_suite.go
rafsalr: Hi @stevenhorsman, sorry i missed removing the debugging steps i have changed the status to draft. will remove the debugging steps.

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -87,13 +87,21 @@ default WriteStreamRequest := true
 '''
 `
 
-func buildInitdataBody(kbsEndpoint string) (string, error) {
+func buildInitdataBody(kbsEndpoint string, initdataannotation string) (string, error) {
```
stevenhorsman, line 90: I don't like this way of passing a flag. Could we instead pass in the initdata value to this function?

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -793,3 +793,12 @@ func DoTestPodWithCpuMemLimitsAndRequests(t *testing.T, e env.Environment, asser
 
 	NewTestCase(t, e, "PodWithCpuMemLimitsAndRequests", assert, "Pod with cpu and memory limits and requests").WithPod(pod).Run()
 }
+
+func DoTestPodwithoutInitdataAnnotations(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint string) {
```
stevenhorsman, line 797: nit: formatting
```suggestion
func DoTestPodWithoutInitdataAnnotations(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint string) {
```

src/cloud-api-adaptor/test/e2e/libvirt_test.go
```diff
@@ -254,3 +254,13 @@ func TestLibvirtCreateWithCpuAndMemRequestLimit(t *testing.T) {
 	assert := LibvirtAssert{}
 	DoTestPodWithCpuMemLimitsAndRequests(t, testEnv, assert, "100m", "100Mi", "200m", "200Mi")
 }
+
+// Test case for empty initdata
+func TestLibvirtCreatePodwithoutInitdataAnnotations(t *testing.T) {
```
stevenhorsman, line 259:
```suggestion
func TestLibvirtCreatePodWithoutInitdataAnnotations(t *testing.T) {
```

src/cloud-api-adaptor/test/e2e/common.go
stevenhorsman: If we want/need to keep the current behavior with the default `testInitdata` we could always do something like:
```
func buildInitdataBody(kbsEndpoint string) (string, error) {
	body, err := buildInitdataBodyFromInitdata(kbsEndpoint, testInitdata)
	if err != nil {
		return "", err
	}
	return body, nil
}

func buildInitdataBodyFromInitdata(kbsEndpoint string, initdata string) (string, error) {
	content, err := os.ReadFile("../trustee/kbs/config/kubernetes/base/https-cert.pem")
	if err != nil {
		return "", err
	}
	certContent := string(content)
	body := fmt.Sprintf(initdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
	return body, nil
}
```
?

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -87,13 +87,21 @@ default WriteStreamRequest := true
...
 	}
 	certContent := string(content)
-	body := fmt.Sprintf(testInitdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
+	var body string
+	if initdataannotation == "with-initdata-annotation" {
+		body = fmt.Sprintf(testInitdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
+	} else {
+		testemptyInitdata := ""
+		body = fmt.Sprintf(testemptyInitdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
+		log.Printf("body in initdata RR: %v", body)
```
stevenhorsman, line 103: Remove the debug please

stevenhorsman (commented):
A few suggestions, we also need to squash all the commits down and provide a better description. Thanks

src/cloud-api-adaptor/test/e2e/common.go
rafsalr: Hi @stevenhorsman, even if we use this method still we need to pass a flag to decide which initdata to be picked up right.

stevenhorsman: So in the existing methods you call the same code as before, in your new methods that need blank initdata you call `buildInitdataBodyFromInitdata(kbsEndpoint, "")` ?

rafsalr: But the issue here is `buildInitdataBody` is called from function `NewBusyboxPodWithNameWithInitdata`. If we need to eliminate flag we need to recreate 2 functions. NewBusyboxPodWithNameWithInitdata and buildInitdataBody. will that be ok ?

stevenhorsman: I see, could we move:
```
	initdata, err := buildInitdataBody(kbsEndpoint)
	if err != nil {
		return fromError(err)
	}
```
outside of `NewBusyboxPodWithNameWithInitdata` and then pass in initdata as an argument to that function. Then the couple of existing uses would just call:
```
  initdata, err := buildInitdataBody(kbsEndpoint)
  if err != nil {
    // Handle error
  }
   NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-key-release", kbsEndpoint, initdata)
```
and your new blank one will call something like
```
initdata, err := buildInitdataBody(kbsEndpoint, "")
  if err != nil {
    // Handle error
  }
NewBusyboxPodWithNameWithInitdata(E2eNamespace, "kbs-key-release", kbsEndpoint, initdata)`
```

I'm trying to work out if we can do a refactor and a new `withInitData() PodOption` method following the builder pattern more, but I think that's probably best as a new PR as I'm not sure how achievable it is right now

src/cloud-api-adaptor/test/e2e/common.go
rafsalr: Its removed. thanks

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -89,13 +89,25 @@ default WriteStreamRequest := true
...
 	content, err := os.ReadFile("../trustee/kbs/config/kubernetes/base/https-cert.pem")
+
 	if err != nil {
 		return "", err
 	}
 	certContent := string(content)
 	body := fmt.Sprintf(testInitdata, kbsEndpoint, kbsEndpoint, certContent, kbsEndpoint, certContent)
 	return body, nil
 }
+func buildInitdataBodyFromEmptyInitdata(kbsEndpoint string) (string, error) {
```
stevenhorsman, line 100: This isn't what I suggested - you have too much duplicated code here and have hardcoded the initdata string?

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -254,7 +256,7 @@ func WithInitdata(kbsEndpoint string) PodOption {
 			p.ObjectMeta.Annotations = make(map[string]string)
 		}
 		key := "io.katacontainers.config.runtime.cc_init_data"
-		initdata, err := buildInitdataBody(kbsEndpoint)
+		initdata, err := buildInitdataBody(kbsEndpoint, testInitdata)
```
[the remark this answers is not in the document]
rafsalr: Hi Steve, removing `NewBusyboxPodWithNameWithInitdata` all together is something where we need more investigation on. As you were suggesting it will be better if we can take care of that in a separate PR. Thanks.

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -593,9 +593,9 @@ func DoTestSealedSecret(t *testing.T, e env.Environment, assert CloudAssert, kbs
 
 // DoTestKbsKeyRelease and DoTestKbsKeyReleaseForFailure should be run in a single test case if you're chaining opa in kbs
 // as test cases might be run in parallel
-func DoTestKbsKeyRelease(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint, resourcePath, expectedSecret string) {
+func DoTestKbsKeyRelease(t *testing.T, e env.Environment, assert CloudAssert, kbsEndpoint, resourcePath, expectedSecret string) PodOrError {
```
stevenhorsman, line 596: Can we remove the PodOrError return here too please if it's not used?

src/cloud-api-adaptor/test/e2e/common_suite.go
```diff
@@ -645,6 +646,7 @@ func DoTestKbsKeyReleaseForFailure(t *testing.T, e env.Environment, assert Cloud
 	}
 
 	NewTestCase(t, e, "DoTestKbsKeyReleaseForFailure", assert, "Kbs key release is failed").WithPod(pod).WithTestCommands(testCommands).Run()
+	return fromPod(pod)
```
stevenhorsman, line 649: Same here

stevenhorsman (commented):
A couple of other points that new returns were added that aren't being used. Can you also squash the commits together please. Thanks

stevenhorsman (commented):
Your commit message doesn't meet the requirement - you need a commit body that ideally summarises what you did and why.

mkulke (approved):

PR #2345, opened 2025-03-21
gcp: fetch disk image size dynamically and avoid using node's metadata service creds

beraldoleal:
Vide commit logs.


src/cloud-providers/gcp/provider.go
```diff
@@ -76,6 +76,29 @@ func getIPs(instance *computepb.Instance) ([]netip.Addr, error) {
 	return podNodeIPs, nil
 }
 
+func (p *gcpProvider) getImageSizeGB(ctx context.Context, image string) (int64, error) {
```
bpradipt, line 79: May be you can use rootVolSize to align with other code - https://github.com/confidential-containers/cloud-api-adaptor/blob/68df7a40ea5906884b3f9b34252a26fb00af2056/src/cloud-providers/aws/provider.go#L100
Also can the size be changed based on input value during VM creation? If yes, then you can follow a similar approach like existing AWS code.

beraldoleal: @bpradipt completely missed that logic. I just force-pushed with a small change, that I believe is doing the same. I tested and it worked.

bpradipt (approved):
/lgtm
Thanks @beraldoleal

src/cloud-providers/gcp/provider.go
```diff
@@ -76,6 +76,29 @@ func getIPs(instance *computepb.Instance) ([]netip.Addr, error) {
...
+	imageName := parts[len(parts)-1]
+
+	req := &computepb.GetImageRequest{
+		Project: p.serviceConfig.ProjectId,
+		Image:   imageName,
+	}
+
+	img, err := client.Get(ctx, req)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to get image size for %s: %w", image, err)
```
littlejawa, line 96: nit:
```suggestion
		return 0, fmt.Errorf("Failed to get image for %s: %w", image, err)
```

littlejawa (approved):
Very small nit, otherwise LGTM
Thanks @beraldoleal

PR #2364, opened 2025-04-07
ibmcloud: add INITDATA to CLI peerpod properties

frankbu:
Add new CLI property to set the INITDATA field in the peer-pods-cm configmap.


stevenhorsman (approved):
LGTM.

PR #2370, opened 2025-04-09
version: Bump golang version in builder

stevenhorsman:
We need to bump golang to pick up the fix for CVE `GO-2025-3563`

We'll start with the builder image and version to try and get the govuln to pass,
then follow up with a separate PR to do the rest of the bump when our base image is available.


bpradipt (approved):
/lgtm

stevenhorsman:
Closing and re-opening to try and wake up the DCO check

stevenhorsman:
It didn't help, so I think I will just have to force push this PR. Our commit message check passes which I think confirms that the Sign off is correct:
```
```

Issue #2219, opened 2024-12-19
Multiple installations of cloud-api-adaptor on the same cluster


bpradipt:
In some scenarios there is a need to support multi-cloud pod VM creation from the same K8s cluster. For example, the K8s cluster could be on-premise, but it's creating peer-pod on AWS, Azure and IBM cloud. 

Creating this tracker issue to discuss on how such a deployment model be supported?

cc @stevenhorsman @mkulke

mkulke:
It probably makes sense to think first about what a good ux would be and make also reference this with the "profile" idea.

I imagine the best UX would be to offer multiple runtime classes, e.g. `kata-remote-azure` or `kata-remote-gcp-large` `kata-remote-aws-gpumedium` and a single CAA installation is owning those RCs and dispatching it to a provisioner and profile

bpradipt:
Yeah different cloud-providers can be supported via different runtimeClasses 

At a high level it may look like this

Kata configs

```
configuration-remote-azure.toml
...
[hypervisor.remote]
remote_hypervisor_socket = "/run/peerpod/hypervisor-azure.sock"
...

configuration-remote-aws.toml
...
[hypervisor.remote]
remote_hypervisor_socket = "/run/peerpod/hypervisor-aws.sock"
...

configuration-remote-ibmcloud.toml
...
[hypervisor.remote]
remote_hypervisor_socket = "/run/peerpod/hypervisor-ibmcloud.sock"
...
```

Then comes the CAA daemonset and it's associated configs (peer-pods-secret, peer-pods-cm and in future profiles).

May be different namespaces for each runtimeClass ?

snir911:
I agree runtimeclasses would be the best UX, however, there's the use case of same cloud-provider but different accounts, IDK how commonly it's needed but i heard some interest in that.

Would it make sense to use current workflow and on top of it allow some kind of profiles? by default caa will use the global configuration unless profile is explicitly set to a pod (or something like that, maybe it's too flexible approach )

mkulke:
hmm, should there be different socket for different cloud providers?

theoretically all the -remote runtimeclasses could carry only informational value. different  `kata-remote-*` RC's would being handled by the same CAA instance, which uses it to dispatch to different profiles. In this sense a cloud-provider could be a field in a CAA profile: kata-remote-aws => `{ cloud-provider: aws, provider-config: { region: east-us, vpc: abc }}`

mkulke:
[...]

PR #2381, opened 2025-04-22
peerpod-ctrl: Support multiple adaptors in peerpod-ctrl

squarti:
This PR processes peerpod resources from multiple adaptors.

Fixes: TBD


src/peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -37,8 +37,8 @@ import (
 // PeerPodReconciler reconciles a PeerPod object
 type PeerPodReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Provider provider.Provider
+	Scheme    *runtime.Scheme
+	Providers map[string]provider.Provider
```
bpradipt, line 41: Any thoughts on how the different provider configs will be stored in the same cm/secret?

squarti: I have been focusing only on the `ibmcloud` and `ibmcloud-powervs` providers which do not have any conflicting env vars. This let's me add all env vars to the same cm/secret.

Also, the set of env vars used by peerpod-ctrl is small since it only makes a call to delete the PodVM at the moment. All it requires is `IBMCLOUD_API_KEY` secret.

If we continue with this approach for the wider [issue](https://github.com/confidential-containers/cloud-api-adaptor/issues/2219), we may need to add prefix to conflicting env vars. I believe that is already the case for many env vars.

bpradipt (approved):
/lgtm
Thanks @squarti

stevenhorsman:
@squarti - sorry for the delay on this - are you able to rebase to resolve the govuln issue please. Thanks

stevenhorsman (approved):
Looks okay to me. Thanks

PR #2401, opened 2025-05-08
podvm: fix GO_ARCH env

Xynnn007:
The arch of golang to be downloaded is determined by env `GO_ARCH` rather than `YQ_ARCH`. I think this is a typo.


mkulke (approved):
yeah, probbably. rhel builds aren't covered by the CI, so maybe someone from RH can take a look.

stevenhorsman (approved):
I think this is a fairly clear cut typo. It's worth noting from:
```
ARG GO_ARCH="amd64"
ARG YQ_ARCH="amd64"
```
that they both are the same format, so this change shouldn't change any behaviour at all, but is good to have cleared up. Thanks @Xynnn007!

xutao323:
Good catch. `GO_ARCH` was added here recently when sorting out ARCH, and `YQ_ARCH` has been used interchangeably for a long time. Not changing behaviour, but code is cleaner.

PR #2406, opened 2025-05-12
gcp: allow users to set podvm labels

beraldoleal:
Adds the "Labels" field to allow users specifying custom labels on PodVMs. In GCP, "labels" are used for resource organization, while "tags" are reserved for network purposes. Because of this distinction, I chose to use "labels" instead of "tags" as in other providers.

### DIFF
--- a/src/cloud-api-adaptor/entrypoint.sh
+++ b/src/cloud-api-adaptor/entrypoint.sh
@@ -108,6 +108,7 @@ gcp() {
     [[ "${GCP_DISK_TYPE}" ]] && optionals+="-disk-type ${GCP_DISK_TYPE} "                          # defaults to 'pd-standard'
     [[ "${GCP_CONFIDENTIAL_TYPE}" ]] && optionals+="-confidential-type ${GCP_CONFIDENTIAL_TYPE} "  # if not set raise exception only when disablecvm = false
     [[ "${ROOT_VOLUME_SIZE}" ]] && optionals+="-root-volume-size ${ROOT_VOLUME_SIZE} "             # Specify root volume size for pod vm
+    [[ "${LABELS}" ]] && optionals+="-labels ${LABELS} "                                           # Custom labels applied to pod vm
 
     set -x
 
--- a/src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
+++ b/src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
@@ -31,6 +31,7 @@ configMapGenerator:
   #- ROOT_VOLUME_SIZE="10" # Uncomment and set if you want to use a specific root volume size. Defaults to 10
   #- TUNNEL_TYPE="" # Uncomment and set if you want to use a specific tunnel type. Defaults to vxlan
   #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
+  #- LABELS="" # Uncomment and add key1=value1,key2=value2 etc if you want to use specific labels for podvm
 ##TLS_SETTINGS
   #- CACERT_FILE="/etc/certificates/ca.crt" # for TLS
   #- CERT_FILE="/etc/certificates/client.crt" # for TLS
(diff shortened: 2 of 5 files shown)


bpradipt:
> while "tags" are reserved for network purposes

Any references stating this ?

vvoronko:
@beraldoleal does that method answer the base requirement this story was intended to implement? I would like to have design review especially if something is not unified cross platform to evaluate possible risks of such implementation for future

beraldoleal:
@bpradipt @vvoronko let me see if I can explain better:

In GCP, the term can be confusing. We have a few possible alternatives:

1. Tags
2. Resource Manager Tags (I'm calling them that to avoid confusion with 1, see below)
3. Labels
4. Metadata

### 1. Tags:

My first attempt. But then I discovered what those actually mean. When using the API during instance creation:

We do this:

```
Tags: &computepb.Tags{
  Items: []string{"test-tag-1", "test-tag-2"},
}
```

And with the CLI:

```
gcloud compute instances create my-vm --tags=peerpods
```

These are just a list of strings (not key-value) and show up under "Compute network details." Their purpose is specific: firewall rules and routes. e.g., "allow port 1550 to all machines with tag peerpods."

Reference:

https://cloud.google.com/vpc/docs/add-remove-network-tags

Any user familiar with GCP using `--tags` would expect things to work this way, same as with the `gcloud` CLI. Using the word "tags" for something else feels wrong to me and could confuse GCP users.

### 2. Resource Manager Tags:

"Resource Manager Tags" could work.... it's key-value but a bit tricky: they’re defined at the org or project level and need extra permissions, so it felt too heavy for this. It’s a resource itself mostly used for policies.

###  3. Labels:

Using labels is the "keep it simple" approach. You can filter with them, like:

```
gcloud compute instances list --filter="labels.env=prod"
```
(and via API too).

Labels are key-value pairs to organize and filter resources. Mostly used to track costs, monitor usage, and manage things like VMs, disks, and buckets.

Reference:

https://cloud.google.com/compute/docs/labeling-resources

### 4. Metadata:

Metadata could also work, but I didn’t go that route. GCP’s tag naming is already messy, and adding another "tag" into metadata just adds to confusion, IMO. Also, metadata is more for startup scripts, env vars, SSH keys, etc (but yes, we could used it).

I'm open to using something else, but I’d strongly suggest using the proper name to avoid confusion with users that are already managing those cloud resources. If our goal is to group resources, labels should work fine, unless I’m missing something. 

In the future, if we want to add policies based on tags, then maybe evolve to 2. But then we should evaluate this for the other providers too.

Sorry for the long answer.... not sure if that makes sense to you guys, please let me know wdyt.

beraldoleal:
Thanks @bpradipt ...

Just to confirm: are those really resource tags in AWS and Azure?

```
# AWS
curl -s http://169.254.169.254/latest/meta-data/tags/instance/key1

# Azure
curl -s -H "Metadata:true" "http://169.254.169.254/metadata/instance/compute/tags?api-version=2023-07-01&format=text"
```

It seems we’re using those already, which gave the impression that AWS/Azure treat them as lightweight metadata.

That said, iiuc, GCP doesn’t expose resource tags via metadata, only through the Resource Manager API, which means a different access path as well.

But my understanding from your msg, is that they are resource tags on azure and aws, is that right? If so, makes sense, we just need to be clear that access will still differ across providers.

bpradipt:
> Thanks @bpradipt ...
> 
> Just to confirm: are those really resource tags in AWS and Azure?
> 
> ```
> # AWS
> curl -s http://169.254.169.254/latest/meta-data/tags/instance/key1
> 
> # Azure
> curl -s -H "Metadata:true" "http://169.254.169.254/metadata/instance/compute/tags?api-version=2023-07-01&format=text"
> ```
> 
> It seems we’re using those already, which gave the impression that AWS/Azure treat them as lightweight metadata.
> 
> That said, iiuc, GCP doesn’t expose resource tags via metadata, only through the Resource Manager API, which means a different access path as well.
> 
> But my understanding from your msg, is that they are resource tags on azure and aws, is that right? If so, makes sense, we just need to be clear that access will still differ across providers.

The primary use case for the tags are not to retrieve it via metadata service. It's to be used in billing etc. Iow, how the tags are queried (apis, console, cli etc) is not of primary concern. 
But "what exactly are tags" is important in the context of CAA to have consistency and less cognitive load on the admin managing multiple cloud providers. 
Afaik, there is only one kind of tags in AWS and Azure and they don't have a concept of labels like GCP. Whether it's light weight or not it's an implementation aspect. Further, all tags are applied to resources. You can also have organisational policies for the tags. This is a good material to understand tag usage in AWS - https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/tagging-best-practices.html

Point is, you can use labels for GCP, but I don't think we can call it `tags` and it results in a deviation from the other existing providers. I don't see a strong reason to make this deviation.

beraldoleal:
> Point is, you can use labels for GCP, but I don't think we can call it `tags` and it results in a deviation from the other existing providers. I don't see a strong reason to make this deviation.

That is the reason I'm calling here labels, and not tags. :)

Just to be clear, I’m not against using resource tags for GCP. The only reason I went with labels it was to avoid forcing GCP users to set up tags ahead of time. But if we’re all aligned on going this route, I’m good with it.

Closing this one and will send a version using resource tags.

Thanks again, @bpradipt

PR #2407, opened 2025-05-13
ci: Skip checklinks job

stevenhorsman:
The lychee action isn't working properly, so
skip the job until we have a CoCo wide strategy
see https://github.com/confidential-containers/confidential-containers/issues/301


ldoktor (approved):
yep, this should do...

mkulke (approved):

PR #2411, opened 2025-05-15
Revert "ci: add yq shim for v3/v4 support"

stevenhorsman:
We had to create yq shim to handle both yq v3 (that was installed as part of the kata-agent build) and yq v4 that we installed. Given that we don't build the kata-agent anymore, we shouldn't need yq v3 toleration, so revert this change.


stevenhorsman:
@mkulke @BbolroC - I can't guarantee this will resolve all the s390x yq issues, but I guess it's worth a try and reduces the codebase a bit?

mkulke:
> We had to create yq shim to handle both yq v3 (that was installed as part of the kata-agent build) and yq v4 that we installed. Given that we don't build the kata-agent anymore, we shouldn't need yq v3 toleration, so revert this change.

well, there is a subtle amiguity: we might not build kata-agent in this repo any more, but is this also true for the private ibm runners? if the kata build will install yq v3 on those machines, then we'd still have problems

stevenhorsman:
> > We had to create yq shim to handle both yq v3 (that was installed as part of the kata-agent build) and yq v4 that we installed. Given that we don't build the kata-agent anymore, we shouldn't need yq v3 toleration, so revert this change.
> 
> well, there is a subtle amiguity: we might not build kata-agent in this repo any more, but is this also true for the private ibm runners? if the kata build will install yq v3 on those machines, then we'd still have problems

You can't share a runner across orgs, so the s390x runners for the coco org are different from the kata ones

mkulke (approved):

BbolroC (approved):
LGTM, thanks @stevenhorsman !

PR #2412, opened 2025-05-15
initdata: skip double encoding if initdata is already encoded

pawelpros:
Peerpods can use annotation io.katacontainers.config.runtime.cc_init_data which should be gzip + base64. This fix skip double encoding if annotation already contains base64 string.


mkulke (commented):
I would strongly recommend to not treat input data ambiguously like this. If we end up with unexpected data/encodings that should produce a hard fail IMO.

mkulke:
with resolution of #2416, can we close this, too?

PR #2413, opened 2025-05-15
docs: added gzip + base64 encoding info

pawelpros:
Update documentation to reflect that initdata needs to be gzipped, then base64 encoded


mkulke (commented):
Oh, I wasn't aware of the protocol bumps. There are more sections where we produce Initdata I suppose. We also should make sure that we assert version == 0.2.* in the code that parses initdata, so we can give a clear error if someone sets the wrong body

fitzthum:
I'm not aware of us bumping the init-data version. For bare metal we are [using v0.1.0](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-initdata.bats#L96) We did recentlyish update the KBS protocol, but that's unrelated.

Xynnn007:
Yup, 0.1.0 is what we are using. Probably it is adding of the compression that looks like a new version, but not because we do not change the body format.

pawelpros:
Thanks guys for pointing that out -> I'll drop changes related to version (I thought it was protocol version)

stevenhorsman (commented):
Your commit message doesn't match our required format. Could you switch it to something like:
```


docs: added gzip + base64 encoding info

Update documentation to reflect that initdata needs to be gzipped, then base64 encoded

Signed-off-by: Pawel Proskurnicki <pawel.proskurnicki@intel.com>
```
?

stevenhorsman (approved):
Looks okay to me. Thanks

mkulke (approved):

PR #2422, opened 2025-05-22
gcp: adding ability to bind tags to instances

beraldoleal:
This will allow users to set Resource Tags to the podvms. However, since GCP works differently from Azure,AWS, we cannot create tags during the instance creation, we need to bind. Because of permission required, we are assuming those tags will exist at the project level.


src/cloud-api-adaptor/install/overlays/gcp/kustomization.yaml
```diff
@@ -31,6 +31,7 @@ configMapGenerator:
   #- ROOT_VOLUME_SIZE="10" # Uncomment and set if you want to use a specific root volume size. Defaults to 10
   #- TUNNEL_TYPE="" # Uncomment and set if you want to use a specific tunnel type. Defaults to vxlan
   #- VXLAN_PORT="" # Uncomment and set if you want to use a specific vxlan port. Defaults to 4789
+  #- TAGS="" # Uncomment and add key1=value1,key2=value2 etc if you want to use specific tags for podvm
```
wainersm, line 34: Hi @beraldoleal !

I suggest a comment similar to "Tags must already exist in the GCP project" here too.

wainersm (approved):
Hi @beraldoleal ,

I'm glad we should not deal with gcloud binary in the caa image \o/
LGTM. Thanks!

PR #2438, opened 2025-05-31
aws: Update sdk to fix imds issue on EKS

bpradipt:
Ref: https://github.com/aws/aws-sdk-go-v2/issues/2449


bpradipt:
The error from CAA logs was this
```
[adaptor/cloud/aws] NewMetadataRetriever is initialized without mac (not found, Signing)
```
After updating the SDK as mentioned in the references github issue, CAA is able to retrieve the details using IMDS.

bpradipt:
> NewMetadataRetriever is initialized without mac (operation error ec2imds: GetMetadata, request canceled, context deadline exceeded)

The error is different in your case. 
You can check if the cluster worker node has IMDS access by running a sample runc pod

```
apiVersion: v1
kind: Pod
metadata:
  name: aws-imds-test 
spec:
  hostNetwork: true
  containers:
    - name: aws-cli
      image: amazon/aws-cli:latest
      command:
        - sleep
        - infinity
```

```
kubectl exec -it aws-imds-test -- curl -w "\n" http://169.254.169.254/latest/meta-data/mac
```

bpradipt:
@ldoktor also are you running the cluster on AWS or are you running the cluster on your own server? 
I'm asking since you mentioned public ip=true. 
IMDS won't work if the cluster is not on AWS.

PR #2446, opened 2025-06-04
workflows: Replace secrets: inherit

stevenhorsman:
Having secrets unconditionally being inherited is
bad practice, so update the workflows to only pass through the minimal secrets that are needed


stevenhorsman:
Superceded by #2457 - which is on a topic branch, so some workflows can be tested

Issue #2455, opened 2025-06-05
caa: fail to pull images with some docker pull secrets


squarti:
### Describe the bug

Kubernetes (ie. containerd) allows docker pull secrets that only specify username/password properties and have the `auth` property missing.  In this case, the username/password should be used instead.

### How to reproduce

Create a docker pull secret that only specifies username/password and run a pod that requires that pull secret to download the image.

### CoCo version information

0.14.0

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell

```

PR #2456, opened 2025-06-05
caa: allow docker image pull secrets with username/password only

squarti:
This PR reads username/password in docker pull secrets and initializes auth param if missing.

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/2455


bpradipt (approved):
/lgtm

PR #2457, opened 2025-06-06
workflows: Remove inherit secrets

stevenhorsman:
Having secrets unconditionally being inherited is
bad practice, so update the workflows to only pass through the minimal secrets that are needed.

As part of this also shuffle some things that were secrets to variables to clean up the workflows

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -19,6 +19,13 @@ on:
       podvm-image-id:
         type: string
         description: prebuilt podvm image
+    secrets:
+      AZURE_CLIENT_ID:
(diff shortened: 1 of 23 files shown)


stevenhorsman:
This is in draft as there is a pre-req to had the Azure vars:
AZURE_MANAGED_IDENTITY_NAME
AZURE_PODVM_GALLERY_NAME
AZURE_REGION
AZURE_RESOURCE_GROUP

set in github in order for it to run and I don't know the values

stevenhorsman:
Topic branch tests passed: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/15487324390

PR #2477, opened 2025-06-17
e2e-test: add id-token: write when invoking wf's

mkulke:
The updated workflow permission in b02ef1ea need further adjustments, since azure-e2e-test is invoked as a nested workflow and requires `id-token: write` permissions.


mkulke:
> Sorry for the inconvenience in not getting this right the first time - I didn't find a proper way to validate it.

np, actionlint doesn't detect it either.

PR #2479, opened 2025-06-19
versions: Bump to kata 3.18.0

stevenhorsman:
Bump the kata-agent and runtime to 3.18.0
along with the matching guest-components
and trustee, to pick up the new image pull
in cdh approach

### DIFF
--- a/.github/workflows/podvm_smoketest.yaml
+++ b/.github/workflows/podvm_smoketest.yaml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Install kata-agent-ctl
         run: |
-          oras pull "${KATA_REG}/agent-ctl:${KATA_REF}-x86_64"
+          oras pull "${KATA_REG}/agent-ctl:latest-main-x86_64"
           tar xf kata-static-agent-ctl.tar.xz
           cp opt/kata/bin/kata-agent-ctl /usr/local/bin
 
--- a/src/cloud-api-adaptor/go.mod
+++ b/src/cloud-api-adaptor/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/google/uuid v1.6.0
-	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20250521110624-a897bce29f3c
+	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20250627024506-a43e06e0ebd5
 	github.com/opencontainers/runtime-spec v1.2.0
 	github.com/stretchr/testify v1.10.0
 	github.com/vishvananda/netlink v1.2.1-beta.2
--- a/src/cloud-api-adaptor/go.sum
+++ b/src/cloud-api-adaptor/go.sum
@@ -415,8 +415,8 @@ github.com/json-iterator/go v1.1.10/go.mod h1:KdQUCv79m/52Kvf8AW2vK1V8akMuk1QjK/
 github.com/json-iterator/go v1.1.12 h1:PV8peI4a0ysnczrg+LtxykD8LfKY9ML6u2jnxaEnrnM=
 github.com/json-iterator/go v1.1.12/go.mod h1:e30LSqwooZae/UwlEbR2852Gd8hjQvJoHmT4TnhNGBo=
 github.com/jtolds/gls v4.20.0+incompatible/go.mod h1:QJZ7F/aHp+rZTRtaJ1ow/lLfFfVYBRgL+9YlvaHOwJU=
-github.com/kata-containers/kata-containers/src/runtime v0.0.0-20250521110624-a897bce29f3c h1:rS5zT2DEWTSZhzjPtXp5NwsM4b/29dMkHZpqKx/BLlE=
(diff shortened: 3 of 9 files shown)


stevenhorsman:
@Xynnn007 - Among other tests we are seeing a the authenticated registry tests fail in peer pods as we write the auth file location in agent-config.toml as we can't use the kernel_params to update `/proc/cmdline`: https://github.com/confidential-containers/cloud-api-adaptor/blob/1a16fc3b9965797e8fb1a57f065ead72080a0c9f/src/cloud-api-adaptor/podvm/files/etc/agent-config.toml#L3. As the agent doesn't process this now, does that mean for this scenario initdata is mandatory and we need the:
```
[image]
authenticated_registry_credentials_uri  "file:///run/peerpod/auth.json"
```
setting as part of it? I'm just trying to work out the best way to do this as at the moment we transparently (from a user perspective) lift the credentials from image-pull secrets, so I'm wondering if we just inject the above setting into cdh.toml in our process user data? It sounds a bit messy to me, so there might be a more reliable approach?

mkulke:
> @Xynnn007 - Among other tests we are seeing a the authenticated registry tests fail in peer pods as we write the auth file location in agent-config.toml as we can't use the kernel_params to update `/proc/cmdline`:
> 
> https://github.com/confidential-containers/cloud-api-adaptor/blob/1a16fc3b9965797e8fb1a57f065ead72080a0c9f/src/cloud-api-adaptor/podvm/files/etc/agent-config.toml#L3
> 
> . As the agent doesn't process this now, does that mean for this scenario initdata is mandatory and we need the:
> ```
> [image]
> authenticated_registry_credentials_uri  "file:///run/peerpod/auth.json"
> ```
> 
> setting as part of it? I'm just trying to work out the best way to do this as at the moment we transparently (from a user perspective) lift the credentials from image-pull secrets, so I'm wondering if we just inject the above setting into cdh.toml in our process user data? It sounds a bit messy to me, so there might be a more reliable approach?

does it matter who is reading the file. afair we provide an empty dummy file on all podvms and process-user-data is overwriting it from user-data?

mkulke:
> > does it matter who is reading the file. afair we provide an empty dummy file on all podvms and process-user-data is overwriting it from user-data?
> 
> That's true - I'm not proposing changing the file at all - it's the question of configuring how the file is read. We already have agent-config.toml hardcoded in the podvm, so that was easy, but we don't have cdh.toml and it's generated from initdata, but I don't think we want to mandate that users set
>

I see, so the problem is that an init-data provided cdh.toml would overwrite the cdh.toml that would be present on the machine?

that's tricky, we could consider merging this property into the toml config, but I agree that's not great. My first best guess would be a `CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS` (*gasp*) env that is considered by CDH if `image.authenticated_registry_credentials_uri` is not set.

and that env is set by all peerpod podvms and pointing to the old location.

stevenhorsman:
So I guess we need the guest-component change to read `CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS`, if the cdh.toml isn't set and then we can update kata to pull in the latest gues-components, and bump this to pull in that kata-agent and set the `CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS` env in the cdh service?

stevenhorsman:
> ok, so the smoke-tests fails b/c for `agent-ctl` kata's CI doesn't seem to tag oci builds with their commit sha. unless we perform some gymnastics, we cannot easily go from a commit sha to an oci tag. pushing the tag on kata's side is probably uncontroversial, but the build machinery is not trivial to grok. I would suggest we hard-code this to `latest-main-x86_64` (instead of using versions.yaml) and hope the agent rpc protocol remains somewhat stable.

Yeah the tagging of builds is the exception, rather than the rule as I only tagged the ones that we needed to consume. I can add code that tags agent-ctl, but in the meantime I'll try the suggestion to pick up the latest to unblock us here.

stevenhorsman:
> about the FOSSA alert, it looks like it detects kata-containers as GPL 3.0 project (direct dependency via the pkg: `github.com/kata-containers/kata-containers/src/runtime`). It's probably nonsense, but we might want to understand why it occurs now.
> 
> or just ignore it w/ a remark "false positive"

I can't see any reference to GPL-3.0 at all in the kata project, so I've raised a fossa issue to report the mix up/ask for more information.

stevenhorsman:
With the hard-coding of the agent-ctl (pending https://github.com/kata-containers/kata-containers/pull/11493 merging to make this nicer) and a slightly increase in the memory limit in the libvirt test, which I guess reflects the increased memory needed, the e2e tests are back working now.

mkulke:
> With the hard-coding of the agent-ctl (pending [kata-containers/kata-containers#11493](https://github.com/kata-containers/kata-containers/pull/11493) merging to make this nicer) and a slightly increase in the memory limit in the libvirt test, which I guess reflects the increased memory needed, the e2e tests are back working now.
> 
> I didn't get a response from the FOSSA team about the kata-containers GPL-3.0 classification, so I guess we mark as a false positive?

Yeah, I created an ignore rule with a note. you probably need to push again, I don't think we can retrigger the fossa test otherwise

stevenhorsman:
> Yeah, I created an ignore rule with a note. you probably need to push again, I don't think we can retrigger the fossa test otherwise

I managed to re-scan using the FOSSA website and it's green now. thanks!

mkulke (approved):

PR #2490, opened 2025-06-30
ibmcloud-powervs: remove trailing zeros while printing instance details

Amulyam24:
Currently the output for printing the instance capacity  looks like
```
[adaptor/cloud/ibmcloud-powervs] Instance type selected by the cloud provider based on config: s922-0.500000x2.000000
```
For better readability, the trailing zeros can be removed by using `%g` instead which has been added via this PR.


stevenhorsman (approved):
LGTM

PR #2494, opened 2025-07-03
agent-protocol-forwarder: daemon.json => apf.json

mkulke:
daemon.json is not the most descriptive name and it can be confusing when depicting the project's architectures, since it's only one config file next to others and scoped to agent-protocol-forwarder.

There is one exception: daemon.json is now also used in the just-merged scratch-space script. But arguably daemon is not a fitting name for this script either. I'd address it in a follow up PR, we can probably just use a discrete file `scratch.json`.

This change renames it to `apf.json` to align with the convention for the rest of the configfiles.

### DIFF
--- a/src/cloud-api-adaptor/docs/initdata.md
+++ b/src/cloud-api-adaptor/docs/initdata.md
@@ -143,10 +143,10 @@ spec:
 ```
 
 ## Structure in `write_files`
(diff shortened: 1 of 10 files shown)


bpradipt (approved):
/lgtm

PR #2495, opened 2025-07-03
scratch-space: use .env file for configuration

mkulke:
The scratch configuration was located in the configuration for agent-protocol-forwarder where it doesn't really belong.

We can write a dedicated .env file that will be passed to the shell script, so we also do not need to extract the values with grep expressions. We need to perform some validation, since this is untrusted input and the host might inject envs that subvert encryption.

drive-by fix: refine some naming.

We are referring to scratch disks, which could be a bit misleading, since often just excess space on an existing disk is being used.


bpradipt (approved):
/lgtm

stevenhorsman (approved):
LGTM. Thanks @mkulke!

PR #2500, opened 2025-07-08
podvm_mkosi: Update s390x run_mkosi_in_container

stevenhorsman:
- Bump the container image being used to fedora:41 as 40 is EOL
- Install the pinned version of mkosi from git, as the dnf version isn't compatible

PR #2501, opened 2025-07-08
podvm_mkosi: Unify build containers

stevenhorsman:
Try and share the same build process for s390x and x86 where possible to reduce the duplication and differences in our builds


stevenhorsman:
> I think this approach is an improvement over #2500

I agree with this, but I've not been able to get it working and given that the strategic direction is for a ubuntu based image, I'm not sure it's worth spending many cycles on this, so I'm proposing failing back to the more basic fix

PR #2530, opened 2025-08-15
workflows: Add scorecard workflow

stevenhorsman:
Add a workflow to update our scorecard score on
each change that's merged


ldoktor (approved):
LGTM although I'm not sure this workflow's terms of service (but I assume for OSS project it should be fine?)

PR #2534, opened 2025-08-19
versions: Bump golang to 1.24.7

stevenhorsman:
Golang 1.23 is EoL, so move to 1.24

PR #2535, opened 2025-08-19
versions: Bump the builder image

stevenhorsman:
Update the builder base of the Docker files
to pick up both the 1.24.6 golang based image

Depends on #2534


wainersm (approved):
PR #2534 is already merged and quay.io/confidential-containers/golang-fedora:1.24.7-41 already built.

Thanks @stevenhorsman !

ldoktor (approved):
Seems to cover all places and the images are there

PR #2536, opened 2025-08-19
cloud: Prioritise instance type annotation for selection

bpradipt:
This helps in the following cases:
- When GPU count is not available. For example in Azure, the GPU count is not available via the API used to get the cpu/mem details
- When a specific instance type must be used irrespective of the resource spec in the pod manifest

Also add a documentation on the instance selection flow.


stevenhorsman (approved):
Looks good, one minor clarification might be helpful

PR #2540, opened 2025-08-21
test: Rename test names to follow a well defined pattern

bpradipt:
Tests are named using the pattern Test<Category><Provider><TestName> This helps to filter tests and run specific categories as needed.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/README.md
+++ b/src/cloud-api-adaptor/test/e2e/README.md
@@ -54,6 +54,39 @@ you should export an unanchored regular expression in the `RUN_TESTS` variable t
 $ RUN_TESTS=CreateSimplePod TEST_PROVISION=yes TEST_PODVM_IMAGE="path/to/podvm-base.qcow2" CLOUD_PROVIDER=libvirt make test-e2e
 ```
 
+## Running tests by category
+
+The E2E tests are organized into categories using the naming convention `Test<Category><Provider><TestCaseName>`. This allows you to run specific types of tests across all providers or for specific providers.
+
+### Test Categories
+
+- **Basic** (37 tests) - Core pod operations like creation, deletion, ConfigMaps, Secrets, Jobs, and Deployments
+- **Net** (14 tests) - Network functionality including external IP access, service communication, and mTLS
+- **Sec** (13 tests) - Security features such as authenticated images, device annotations, and access control
+- **Conf** (10 tests) - Confidential computing features including image decryption, attestation, and KBS integration
+- **Res** (32 tests) - Resource management including CPU/memory limits, annotations, and logging
+- **Img** (6 tests) - Image handling including large images and alternate images
+- **Store** (2 tests) - Storage functionality such as Persistent Volume Claims (PVC)
+
+### Category-based filtering examples
+
+```bash
+# Run all basic tests for libvirt provider
+$ RUN_TESTS='^TestBasic' CLOUD_PROVIDER=libvirt make test-e2e
+
+# Run all networking tests for azure provider
+$ RUN_TESTS='^TestNet' CLOUD_PROVIDER=azure make test-e2e
+
+# Run all confidential computing tests for libvirt provider
+$ RUN_TESTS='^TestConf' CLOUD_PROVIDER=libvirt make test-e2e
+
+# Run all security tests for aws provider
+$ RUN_TESTS='^TestSec' CLOUD_PROVIDER=aws make test-e2e
+
+# Run all resource management tests for azure provider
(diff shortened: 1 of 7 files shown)


bpradipt:
Folks, ptal a look and let me know if this approach sounds good. With the current code, it's really hard to run specific set of tests on need basis so I'm starting with a well-defined and opinionated patter for the names.
I haven't modified the gh actions yet.

bpradipt:
> > > > One approach I am thinking is to have separate labels for each categories and then triggering a suite (say nightly or bvt etc) via grouping the labels. Do you think this approach will work with GHA ?
> > > 
> > > 
> > > Sorry, I don't understand this proposal? What do you mean by labels?
> > 
> > 
> > I meant the labels that you use today to trigger the e2e tests. Like `test_e2e_docker`
> 
> Magnus' proposal is that we run a cut down version of the e2e tests on nightly to improve the reliability (what I'm calling the BVT), but run the full e2e tests on PRs.
> 
> It sounds like you are working on an orthogonal idea and you don't want to run the full e2e tests on PRs, but only selective tests driven by a label? What's the reason behind this as it feels like we run full tests on PRs to try and ensure that we don't break the `main` branch?

Sorry for the confusion. In this PR, my only intention is to make it easier to run group of tests via regular expression (via setting the RUN_TESTS variable) which is not possible today.
This name change doesn't impact the current way of running the tests. IOW, you continue to run the full set of tests for every PR.

However we do need a way to run some specific tests for some PR so as to not overload the runners. This is where I was thinking we can have GH labels which can be mapped to specific regular expressions to allow a subset of tests to run. 
As for nightlies, since those are run via scheduled jobs, we can use regular expressions for the providers as needed.

Does this help ?

stevenhorsman:
> Does this help ?

Not really - I don't understand when we would use the labels? However, my naive approach for the BVT was going to be a relabel that clashes with what you did e.g. `TestBVTCreateSimplePod`, but we could just switch that to the end and then not clash e.g. `TestBasicCreateSimplePodBVT`, so I don't think this will get in the way.

bpradipt:
Looks like the docker e2e is broken due to some recent changes :-(. I'm looking into it. Will send PR soon.

stevenhorsman:
> > > Does this help ?
> > 
> > 
> > Not really - I don't understand when we would use the labels? However, my naive approach for the BVT was going to be a relabel that clashes with what you did e.g. `TestBVTCreateSimplePod`, but we could just switch that to the end and then not clash e.g. `TestBasicCreateSimplePodBVT`, so I don't think this will get in the way.
> 
> Can you run the following as part of bvt.yaml GHA workflow for example ?
> 
> ```
> RUN_TESTS='^TestBasic' CLOUD_PROVIDER=libvirt make test-e2e
> ```

I don't think your definition of the labels matches with what we'd want in the BVT. I'd expect us to have some of `Basic` like `TestBasicAwsCreateSimplePod`, `TestBasicAwsCreateSimplePod` for example but not all things like the nginx deployment that is fussy, or variants with configmap and secret tests. However I'd then expect some tests from other categories e.g. key release, image decrypt and maybe the golden path in Sec - `TestSecDockerCreatePeerPodWithAuthenticatedImageWithImagePullSecretOnPod`, but not the negative tests. Does that help, or do you think I've got the breakdown wrong?

BTW - thanks for the starting this conversation - it's really important!

stevenhorsman:
> Let's ignore the labels for now and see how we can run specific set of tests. I think that's the real issue. Or am I missing something? Is there a way other than using regex in `go test...` ?
> 
> If not, then may be we can think about naming the tests in a way to include the category like BVT, Nightly etc. ?

Yeah, AFAIK the regex is in the only way to do it, so I figured we just have a list of the tests and rename the cut down version TestBVT<name> and then in the nightly gha with just set RUN_TESTS to be `TestBVT`?

bpradipt:
> > Let's ignore the labels for now and see how we can run specific set of tests. I think that's the real issue. Or am I missing something? Is there a way other than using regex in `go test...` ?
> > If not, then may be we can think about naming the tests in a way to include the category like BVT, Nightly etc. ?
> 
> Yeah, AFAIK the regex is in the only way to do it, so I figured we just have a list of the tests and rename the cut down version TestBVT and then in the nightly gha with just set RUN_TESTS to be `TestBVT`?

When you get time, can you list down the tests you would want to run for BVT? I'll have a sample pattern and then we can iterate to see if it makes sense.

stevenhorsman:
I created https://github.com/confidential-containers/cloud-api-adaptor/issues/2542 to cover this discuss and as a starting point. I hope this helps

PR #2549, opened 2025-08-28
Initial implementation for pre-created VM in Azure

bpradipt:
This is based on the work done by @Amulyam24 on SFTP support.

I'm creating this draft to kickstart the design discussions. This is a working code if someone wants to test it to get more ideas.

To properly support pre-created VMs, we need a way to reinitialise/restart the pod environment in the VM on pod deletion. Also for pre-created VM, ssh public key must already exist in the pod VM as we can't use userdata. 
The SSH public key can be embedded in the image or should be made available out-of-band. The corresponding private key must be configured in CAA.

The implementation support multiple ways to handle pre-created VM
- global default
- based on instance type (in the pod annotation)
- based on pod name

The reason I went ahead with multiple options is to make it flexible. For example we might want to use pre-created VM only for large GPU instances which takes a huge amount of time to boot.

Also the current implementation enables SFTP by default for the debug image. This can be made configurable by adding another mkosi profile specifically for SFTP.


bpradipt:
@mkulke @stevenhorsman @snir911 @Xynnn007 and others wdyt about a separate provider to support pre-created instances (`byom`) ?

Xynnn007:
@bpradipt If I understand correctly, the idea is to leverage an existing VM via ssh, ragardless of where it is and how it comes.
Am I right?

If so, probably we do not need the provider logic to do actually the creation work, thus a new provider would make sense. A naive question is how we can ensure the software stack inside the VM is as expected, thus we can change it at runtime using the defined way?

bpradipt:
For context - separate provider for pre-created machines - https://github.com/bpradipt/cloud-api-adaptor/tree/byom

PR #2555, opened 2025-09-05
test/e2e: adding pvc options and refactoring pvc/secret binding

ramachandrach:
Enhanced `WithSecretBinding` and `WithPVCBinding` to handle multi-container scenario.

Enhanced `NewPVC` to handle different type of volumes by using PVCOptions. 

Introduced a delay function while creating PVC, as it takes few seconds to bound to Persistent Volume. Without this delay we get a warning in PodEvents as the PVC will not be ready before the pod creation. Which causes the testcases to fail because of a transient warning.

### DIFF
--- a/src/cloud-api-adaptor/test/e2e/common.go
+++ b/src/cloud-api-adaptor/test/e2e/common.go
@@ -141,6 +141,16 @@ func encodePolicyFile(policyFilePath string) string {
 
 type PodOption func(*corev1.Pod)
 
+func GetContainerReference(pod *corev1.Pod, containerName string) *corev1.Container {
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == containerName {
+			return &pod.Spec.Containers[i]
+		}
+	}
+	log.Error(fmt.Sprintf("Failed getting container reference for container %s", containerName))
+	return nil
+}
+
 func WithRestartPolicy(restartPolicy corev1.RestartPolicy) PodOption {
 	return func(p *corev1.Pod) {
 		p.Spec.RestartPolicy = restartPolicy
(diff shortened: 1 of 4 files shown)


src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -236,18 +245,22 @@ func WithConfigMapBinding(mountPath string, configMapName string) PodOption {
...
 }
 
-func WithSecretBinding(mountPath string, secretName string) PodOption {
+func WithSecretBinding(mountPath string, secretName string, containerName string) PodOption {
 	return func(p *corev1.Pod) {
-		p.Spec.Containers[0].VolumeMounts = append(p.Spec.Containers[0].VolumeMounts, corev1.VolumeMount{Name: "secret-volume", MountPath: mountPath})
-		p.Spec.Volumes = append(p.Spec.Volumes, corev1.Volume{Name: "secret-volume", VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{SecretName: secretName}}})
+		if c := GetContainerIndex(p, containerName); c != nil {
+			c.VolumeMounts = append(c.VolumeMounts, corev1.VolumeMount{Name: "secret-volume-" + containerName, MountPath: mountPath})
+		}
```
stevenhorsman, line 252: Do we need an error/warning if we can't find the container? Does this function still make sense if this happens?

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -236,18 +245,22 @@ func WithConfigMapBinding(mountPath string, configMapName string) PodOption {
...
 
-func WithPVCBinding(mountPath string, pvcName string) PodOption {
+func WithPVCBinding(mountPath string, pvcName string, containerName string) PodOption {
 	propagationHostToContainer := corev1.MountPropagationHostToContainer
 	return func(p *corev1.Pod) {
-		p.Spec.Containers[2].VolumeMounts = append(p.Spec.Containers[2].VolumeMounts, corev1.VolumeMount{Name: "pvc-volume", MountPath: mountPath, MountPropagation: &propagationHostToContainer})
-		p.Spec.Volumes = append(p.Spec.Volumes, corev1.Volume{Name: "pvc-volume", VolumeSource: corev1.VolumeSource{PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{ClaimName: pvcName}}})
+		if c := GetContainerIndex(p, containerName); c != nil {
+			c.VolumeMounts = append(c.VolumeMounts, corev1.VolumeMount{Name: "pvc-volume-" + containerName, MountPath: mountPath, MountPropagation: &propagationHostToContainer})
+		}
```
stevenhorsman, line 262: Same question here

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -291,13 +304,31 @@ func WithInitContainers(initContainers []corev1.Container) PodOption {
...
 	runtimeClassName := "kata-remote"
 	annotationData := map[string]string{}
 
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{Name: podName,
-			Namespace:   namespace,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"app": podName,
+			},
```
stevenhorsman, line 331: What's the reason for this? We have the `WithLabel` PodOption that can be passed in to achieve the same effect when it's required?

stevenhorsman (commented):
Generally I think that the changes make sense. I've left a few questions and we also need the commit message to be updated to match our commit message standards - notably adding a "subsystem" prefix to the header (`tests/e2e:` in this case) and adding a message body to describe (the how/why) of the changes. Thanks!

src/cloud-api-adaptor/test/e2e/common.go
ramachandrach: I had assumed we pass the same variable which is being passed to `NewPod` while creation of podSpec. Will include `Fatalf` in `GetContainerIndex` to exit if the container is not found.

src/cloud-api-adaptor/test/e2e/ibmcloud_test.go
```diff
@@ -138,7 +138,7 @@ func TestCreatePeerPodWithPVC(t *testing.T) {
 		csiContainerName := "ibm-vpc-block-podvm-node-driver"
 		csiImageName := "gcr.io/k8s-staging-cloud-provider-ibm/ibm-vpc-block-csi-driver:v5.2.0"
 
-		myPVC := NewPVC(nameSpace, pvcName, storageClassName, storageSize, corev1.ReadWriteOnce)
+		myPVC := NewPVC(nameSpace, pvcName, storageSize, corev1.ReadWriteOnce, WithStorageClass(storageClassName))
 		myPodwithPVC := NewPodWithPVCFromIBMVPCBlockDriver(nameSpace, podName, containerName, imageName, csiContainerName, csiImageName, WithPVCBinding(mountPath, pvcName))
```
stevenhorsman, line 142: the `WithPVCBinding` needs the containerName passed through

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -236,18 +246,22 @@ func WithConfigMapBinding(mountPath string, configMapName string) PodOption {
 	}
 }
 
-func WithSecretBinding(mountPath string, secretName string) PodOption {
+func WithSecretBinding(mountPath string, secretName string, containerName string) PodOption {
```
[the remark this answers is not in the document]
ramachandrach: @wainersm containerName is already defined in the testcase when creating the pod With `NewPod`, so we can reuse it here to keep bindings container-specific and consistent.

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -141,6 +141,16 @@ func encodePolicyFile(policyFilePath string) string {
 
 type PodOption func(*corev1.Pod)
 
+func GetContainerIndex(pod *corev1.Pod, containerName string) *corev1.Container {
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == containerName {
+			return &pod.Spec.Containers[i]
+		}
+	}
+	log.Fatalf("container %q not found in pod spec", containerName)
```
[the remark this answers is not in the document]
ramachandrach: @stevenhorsman @wainersm If i update the function to receive the `*testing.T` and use the t.Fatalf to fail the testcase? 
This method will not impact the existing tests.

src/cloud-api-adaptor/test/e2e/common.go
```diff
@@ -291,6 +305,21 @@ func WithInitContainers(initContainers []corev1.Container) PodOption {
 	}
 }
 
+type PVCOption func(*corev1.PersistentVolumeClaim)
+
+// Option to add StorageClass
+func WithStorageClass(sc string) PVCOption {
+	return func(pvc *corev1.PersistentVolumeClaim) {
+		pvc.Spec.StorageClassName = &sc
```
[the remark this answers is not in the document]
wainersm: ah, ok, I saw `VolumeName` being passed as a value then I found it weird. ok.

ramachandrach:
@stevenhorsman Could you please review and merge the code. Also could you please re-trigger the failed check.
Thank you!

PR #2559, opened 2025-09-10
e2e: disable tdx e2e tests on azure temporarily

mkulke:
The public preview of TDX v5 instance sizes is ending this week, to be superseded with a public preview of TDX v6. Since the new instance types will be different in some regards (OpenHCL, NVMe, limited region availability initally) we need to perform the necessary adjustments. In the meantime, to avoid a broken CI, the TDX tests are disabled.

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -82,9 +82,9 @@ jobs:
     strategy:
       matrix:
         parameters:
-          - id: "tdx"
-            machine_type: "Standard_DC2es_v5"
-            jitter: 0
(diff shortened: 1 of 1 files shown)


ldoktor (approved):
I guess there won't be any transition period where we could still keep using the `_V5` before the `_v6` is "stabilized"? Anyway the change looks alright, just wondering if there are no ways around it...

mkulke:
> I guess there won't be any transition period where we could still keep using the `_V5` before the `_v6` is "stabilized"? Anyway the change looks alright, just wondering if there are no ways around it...

sadly no, this is rather abrupt.

PR #2563, opened 2025-09-11
Adding devcontainers generally and for mkosi 

nmwael:


.devcontainer/development/scripts/postCreateCommand.sh
```diff
@@ -0,0 +1,47 @@
...
+echo "DISTRO_ARCH=$DISTRO_ARCH"
+
+# Build URL
+YQ_DOWNLOAD_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_${DISTRO_OS}_${DISTRO_ARCH}"
+
+echo "Downloading yq version ${YQ_VERSION} for ${DISTRO_OS}/${DISTRO_ARCH} from:"
+echo "${YQ_DOWNLOAD_URL}"
+
+sudo wget -qO /usr/bin/yq "${YQ_DOWNLOAD_URL}"
+sudo chmod +x /usr/bin/yq
```
mkulke, line 45: I suppose is this happening in a sandbox and won't overwrite anything on a developer machine?

.devcontainer/development/scripts/postCreateCommand.sh
```diff
@@ -0,0 +1,47 @@
...
+
+go install github.com/edgelesssys/uplosi@latest
```
mkulke, line 47: should we pin the version?

mkulke (commented):
conceptually this looks good to me, but personally I'm not a vscode user, so someone who is using an IDE that has devcontainer integration should take a look/test that PR.

the commit message needs fixing (a proper descriptive commit message, and a subsystem/scope should be in the header, sth like: "ide: adding devcontainer files"

also, the branch probably needs a rebase

.devcontainer/development/scripts/postCreateCommand.sh
nmwael: Good question, I'm a little worried about maintenance.

PR #2579, opened 2025-09-22
test/e2e: Increase memory limit

stevenhorsman:
In our nightly runs, we quite often see failure in the `TestLibvirtCreateWithCpuAndMemRequestLimit`
test, where the containers doesn't seem to start.
Given that it has a much more restricted memory
than all the other pods, try increasing the limit to see if this helps with stability.


bpradipt (approved):

stevenhorsman:
Sorry, I should have left this in draft as I want to do a few rounds of tests to see if it is still failing, before I merge.

PR #2584, opened 2025-09-23
ci: add azure workflow + infra for podvm release

mkulke:
PodVM release images should be built on the CI using a workflow. This is preferable over manual steps, but also useful since we have provenance attestation for the measurements. The measurements can only be verified to be valid if we generate them on the CI.

The required terraform infra has been added, which allows github actions to write to the image gallery that holds release images.

A azure-podvm-release.yml workflow has been added that overrides a couple of defaults in the azure-podvm-build.yml workflow.

drive-by-fix: the podvm build workflow has been adjusted to expand input/secret/vars fields in a step's env block (to avoid injections problems).

(the workflow has been tested in a fork [here](https://github.com/mkulke/cloud-api-adaptor/actions/runs/17946385520))

### DIFF
--- a/.github/workflows/azure-podvm-image-build.yml
+++ b/.github/workflows/azure-podvm-image-build.yml
@@ -6,6 +6,20 @@ on:
       image-version:
         type: string
         required: true
+      image-variant:
+        type: string
+        default: production
+      image-gallery:
+        type: string
+      image-definition:
+        type: string
+      community-gallery-name:
+        type: string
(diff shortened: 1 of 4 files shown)


ldoktor (approved):
Changes makes sense, I'm not 100% sure about the tf configuration correctness but it seems to be working well and we definitely need to add something similar to AWS once @wainersm finishes his AWS workflows. Thanks

PR #2589, opened 2025-09-26
ci: Check for AGENT_TOOLSDIRECTORY being set

stevenhorsman:
AGENT_TOOLSDIRECTORY is a gh-hosted runner variable, but not necessarily set in self-hosted runners. As we now call `set -o nounset` before deleting it, we see problems on the s390x runs. Unfortunately we couldn't test this code on the self-hosted runners previously as it was a workflow change though!


BbolroC (approved):
LGTM, I've tested it locally and it worked. Thanks @stevenhorsman for a quick fix!

wainersm (approved):
Good catch! Thanks @stevenhorsman !

PR #2633, opened 2025-10-14
e2e: readd tdx to azure e2e tests

mkulke:
The TDX v6 instance sizes are available in the subscription now, hence we can re-enable the tests. We don't need further adjustments

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -82,10 +82,10 @@ jobs:
     strategy:
       matrix:
         parameters:
-          # - id: "tdx"
-          #   machine_type: "Standard_DC2es_v5"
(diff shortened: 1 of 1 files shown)


.github/workflows/azure-e2e-test.yml
```diff
@@ -84,9 +84,11 @@ jobs:
         parameters:
           # - id: "tdx"
           #   machine_type: "Standard_DC2es_v5"
+          #   location: "westeurope"
           #   jitter: 0
```
stevenhorsman: Can we just remove the v5 versions as I thought they were gone now?

stevenhorsman: I see - the next commit switched them for v6

Issue #2635, opened 2025-10-14
ci: nighlty libvirt jobs with secure_comms are unstable


wainersm:
As in $SUBJECT, mostly fails are with `crio+secure_commons` but sometimes `containerd+secure_commons` too.

Last 7 nighties runs had 4 fails:

`containerd+secure_commons` fails:
 * https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/18424378437/job/52503943545
 * https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/18396409954/job/52417334365
 
`crio+secure_commons` fails:
 * https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/18396409954/job/52417334376
 * https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/18365473093/job/52318215485
 * https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/18333741713/job/52214464948
 
Following are the tests failing. Usually it's one or another that fails in a run, haven't seen them all fail at same time:

* TestLibvirtCreateWithCpuAndMemRequestLimit
* TestLibvirtCreateSimplePod
* TestLibvirtCreateNginxDeployment
* TestLibvirtCreateSimplePodWithSecureCommsIsValid

Also other tests of the suite are passing, even tests that are supposed to be more "complex".

The `TestLibvirtCreateSimplePod` fails and you can see the following caa logs:

```
2025/10/10 05:01:48 [adaptor/cloud] created an instance podvm-simple-test-6f543dff for sandbox 6f543dff8597df1648d9b85002e2ddd8a7b6e2ca5c18355e7ea0b7932c798b1e
        2025/10/10 05:01:48 [secure-comms] Attestation phase: starting
        2025/10/10 05:01:48 [secure-comms] Attestation phase: unable to Dial 192.168.122.19:2222: dial tcp 192.168.122.19:2222: connect: connection refused
        2025/10/10 05:01:49 [secure-comms] Attestation phase: unable to Dial 192.168.122.19:2222: dial tcp 192.168.122.19:2222: connect: connection refused
        2025/10/10 05:01:49 [secure-comms] Attestation phase: unable to Dial 192.168.122.19:2222: dial tcp 192.168.122.19:2222: connect: connection refused
        2025/10/10 05:01:50 [secure-comms] Attestation phase: unable to Dial 192.168.122.19:2222: dial tcp 192.168.122.19:2222: connect: connection refused

[...]

PR #2636, opened 2025-10-14
CI: promote libvirt CRI-O jobs to stable but secure_comms unstable

wainersm:
The nightly libvirt e2e jobs for CRI-O can be considered stable, except when configured to run with secure_comms. As well as containerd configured with secure_comms are now unstable.

Let's keep running the jobs with secure_comms but disregards their finish status until we don't have a fix.

Refers to #2635


stevenhorsman:
> > Let's keep running the jobs with secure_comms but disregards their finish status until we don't have a fix.
> 
> Does this mean the tests will keep failing? In this case I would disagree, we shouldn't keep a mental map in our head which tests to disregard and which ones to accept, this creates a burden to inspect the logs all the time and is obscure for people who are not aware of this.

So for the nightlies this approach works okay as the overall status that we reflect on the badge and the job ignores these unstable jobs, but I agree that with the current approach of running the same in PRs as. nightlies then reviewers do need to remember these more or might think that there are blocking failures. I assume the github ignores them in the status in the PR list though, so maybe that's good enough?

PR #2643, opened 2025-10-20
workflows: do not persist credentials on checkout

wainersm:
zizmor flags that code checkout with persist credentials is a security risk. Default behavior of actions/checkout is to always persist so zizmor has complained about that in a lot of our workflows. We don't need to persist the credentials, hence, explicitly setting `persist-credentials: false` in all workflows that checkout code.

---

This is a follow up of https://github.com/confidential-containers/cloud-api-adaptor/pull/2607 (discussed in https://github.com/confidential-containers/cloud-api-adaptor/pull/2607#discussion_r2406284461) . I was supposed to set `persist-credentials: false` only on e2e_aws workflow but then zizmor would be complaining about the same issue in a bunch of workflows. So I'm applying that fix on all workflows that checkout code.

This is also a split of https://github.com/confidential-containers/cloud-api-adaptor/pull/2641

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -27,6 +27,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Install actionlint gh extension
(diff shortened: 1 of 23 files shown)


wainersm:
I don't see how the CI fails are related with this PR, so gonna merge it.

PR #2646, opened 2025-10-21
azure: use Standard SSD for CVM storage

snir911:
Change the storage account type for Azure Confidential VMs from Standard HDD (Standard_LRS) to Standard SSD (StandardSSD_LRS) for better performance.


esposem:
@snir911   you also need to update for peer-pods

bpradipt:
Adding the tag do-not-merge to avoid merging this unless the disk type change is added for both CVM and non-CVM case

bpradipt (approved):
/lgtm

PR #2656, opened 2025-10-27
RFC: install: charts for peerpods

beraldoleal:
This is a first draft of our deployment via charts. CoCo project is moving from the operator to a umbrela chart that will have as dependency other charts, including kata-containers.

This way we can list peerpods as dependency of that chart.

Ideally, our kustomize folder could be replaced by this chart, to avoid duplication. But I would like to start simple with the chart first and replace later.

### DIFF
--- a/src/cloud-api-adaptor/podvm/Makefile.inc
+++ b/src/cloud-api-adaptor/podvm/Makefile.inc
@@ -55,7 +55,7 @@ endif
 
 # probe version if the ref does not appear to be a git commit digest
 ifeq ($(shell echo $(KATA_REF) | grep -E "^[0-9a-f]{40}$$"),)
-KATA_COMMIT := $(shell curl \
+KATA_COMMIT := $(shell \
 	$(GH_CURL) \
 	$(GH_API_BASE)/repos/$(KATA_REPO)/commits/$(KATA_REF) \
 	| jq -r '.sha // empty')


src/cloud-api-adaptor/install/charts/peerpods/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v2
+name: peerpods
+description: Cloud API Adaptor (Peerpods) Helm Chart
+type: application
+version: 0.1.0
```
stevenhorsman: Is this chart version designed to be independent of the appVersion, or is there any value in sync-ing it like we do in Kata and the coco Chart.yaml?

stevenhorsman (commented):
Thanks for starting this @beraldoleal - I am very ignorant on helm, so really appreciated it. As Wainer has said - lets get some CI on this as soon as we can. I'm not sure about all the providers being together and whether that is confusing for folks that only deal with one of them, but we can canvas opinion wider in the community for this.

src/cloud-api-adaptor/install/charts/peerpods/Chart.yaml
beraldoleal: Chart version should be independent of appVersion. Chart version tracks changes to the chart itself (templates, structure), while appVersion tracks which CAA version gets deployed. This lets us fix chart bugs or support different CAA versions without forcing version locksetp.

For instance: independent versioning lets us fix chart bugs without waiting for a new CAA release.

src/cloud-providers/Makefile
stevenhorsman: Can these changes go into a prior commit in the tool update PR as although they are the mechanism to generate charts, I think separating the code out will help keep this commit clean (which I think is needed as I'm hoping to get some people who have more charts experience involved in the review and they don't need the distraction of the mechanism we use to create the templates). Does that make sense?

src/cloud-providers/cmd/config-extractor/README.md
```diff
@@ -17,14 +17,16 @@ make config-extractor
 ## Usage
 
 ```bash
-./bin/config-extractor [-o json|table] <provider-name>
+./bin/config-extractor [-o json|table] [-no-secrets|-secrets-only] <provider-name>
```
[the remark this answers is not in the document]
beraldoleal: yes, usually we don't commit values.yaml and values-secrets.yaml together, so users can keep their own local copy of those secrets (that should not be versioned). Only a values-secrets.yaml.template should be on git.

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
```diff
@@ -0,0 +1,69 @@
...
+    qemu-coco-dev-runtime-rs:
+      enabled: false
+    remote:
+      enabled: true
+      supportedArches:
+        - amd64
+        - s390x
+      allowedHypervisorAnnotations: []
+      containerd:
+        snapshotter: "nydus"
```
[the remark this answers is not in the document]
fidencio: https://github.com/kata-containers/kata-containers/pull/12243

Also, please, set
snapshotter:
  setup: ["nydus"]

beraldoleal:
> > This was fixed on my latest push, while we still need to some tweaks on the kata chart side.
> 
> Is there an issue on the kata side that I can follow?

I think you covered already with your last commit. I'm pushing t he changes you suggested here...

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
```diff
@@ -0,0 +1,69 @@
+# Default values for peerpods (CAA) chart
+
+# Namespace where peerpods resources will be deployed
+namespace: confidential-containers-system
+
+# CAA (Cloud API Adaptor) image
+image:
+  name: quay.io/confidential-containers/cloud-api-adaptor
+  # tag defaults to Chart.appVersion if not set
+  tag: ""
```
wainersm: Hi @beraldoleal ,

Note that we release two "variants" of the cloud-api-daptor:
* release: doesn't have libvirt-dev libraries
* dev: does have libvirt-dev libraries

The only provider that (I know of) uses the dev image is libvirt.

if the tag is based on `Chart.appVersion` then we will need to distribute two charts image (IIUC). Otherwise, we will need to use the `Values.provider` to figure out the proper tag.

wainersm:
> @wainersm @stevenhorsman this depends on #2712 so I can properly generate secrets, required and optional values.

IIUC we will use the extractor to generate the values.yaml for each provider, and have that yaml check in the repo. So could you please generate these files and commit on this PR so we can 1) see how it will look like 2) try it out

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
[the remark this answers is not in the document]
beraldoleal: @wainersm is already, `shims.remote.enabled` is set to true (line 63 in values.yaml).

beraldoleal:
@wainersm,

> Other than that, could we make a fully working version for docker or libvirt so that we could run it locally to experiment with?

I have generated the files for all providers, and updated the e2e libvirt ci job for us to test. Is that what you had in mind?

src/cloud-api-adaptor/install/charts/peerpods/providers/alibabacloud.yaml
```diff
@@ -0,0 +1,136 @@
+# Auto-generated by: make sync-chart-values
```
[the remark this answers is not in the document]
beraldoleal: @Xynnn007 ... this is now made automatically via the new variable registration. Its just needs to add the doc string here:

https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-providers/alibabacloud/manager.go#L24

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
stevenhorsman: Sorry, can you clarify the context of this question please? During development we want it to be latest as we need to test against that, then during the release we either tag with dev (for libvirt & docker), or the release sha e.g. https://github.com/confidential-containers/cloud-api-adaptor/commit/0275f8b095db93ad3296a17b838cbcd6f810e0ae Are you suggesting that this approach doesn't work with charts and we can only have a single value?

stevenhorsman: I will be honest that I've lost track of how `latest` works for all when the "release" doesn't, but maybe we need to crack open this and check that libvirt still requires the cgo bindings which I think is the reason for two images

.github/workflows/e2e_libvirt.yaml
```diff
@@ -180,12 +193,42 @@ jobs:
...
+          # Extract image name and tag from caa_image (format: registry/image:tag)
+          IMAGE_FULL="${CAA_IMAGE}"
+          IMAGE_TAG="${IMAGE_FULL##*:}"
+          IMAGE_NAME="${IMAGE_FULL%:*}"
+
+          echo "Deploying with Helm:"
+          echo "  Image: ${IMAGE_NAME}"
+          echo "  Tag: ${IMAGE_TAG}"
+
+          helm install peerpods ./install/charts/peerpods \
```
wainersm: Hi @beraldoleal , is this a temporary solution so we can test this PR before merge? or you want it as a long term solution? If the later then I prefer the installation being carried out inside the test tool.

Hmmm changes on the workflow won't take effect anyway because they run on pull_request_target: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/e2e_on_pull.yaml#L9

wainersm: ok. could you add a `TODO` so we don't loose track of that change?

src/cloud-api-adaptor/install/charts/peerpods/providers/alibabacloud.yaml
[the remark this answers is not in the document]
Xynnn007: Thanks, as Alibaba Cloud is still not involved into CI system, I can perform a manual deployment after the PR and update the needed documents.

wainersm:
Hi @beraldoleal ,

Pushed this version to my branch and triggered the manual test (for libvirt): https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/20728754017/job/59510995175

It failed to install the chart:

```
Deploying with Helm:
  Image: ghcr.io/wainersm/cloud-api-adaptor
  Tag: latest-dev
Error: INSTALLATION FAILED: An error occurred while checking for chart dependencies. You may need to run `helm dependency build` to fetch missing dependencies: found in Chart.yaml, but missing in charts/ directory: kata-deploy
Error: Process completed with exit code 1.
```

.github/workflows/e2e_libvirt.yaml
```diff
@@ -172,22 +187,152 @@ jobs:
...
+          CONTAINER_RUNTIME: ${{ inputs.container_runtime }}
+          TEST_PROVISION: "yes"
+          TEST_INSTALL_CAA: "no"
+          TEST_TEARDOWN: "no"
+          TEST_PROVISION_FILE: ${{ github.workspace }}/src/cloud-api-adaptor/libvirt.properties
+          TEST_PODVM_IMAGE: ${{ env.PODVM_QCOW2 }}
+          TEST_E2E_TIMEOUT: "30m"
+        run: |
+          # Provision cluster only, no CAA install, no KBS, no tests
+          make test-e2e RUN_TESTS="^$"
```
mkulke: that looks odd, can we hide this in the Makefile: "make provision-cluster" (it should probably be in the cli tool, but that's out of scope for this PR)

.github/workflows/e2e_libvirt.yaml
```diff
@@ -172,22 +187,152 @@ jobs:
...
+          echo "  Tag: ${IMAGE_TAG}"
+
+          # Build chart dependencies (downloads kata-deploy from OCI registry)
+          echo "Building Helm chart dependencies..."
+          helm dependency build ./install/charts/peerpods
+
+          # TODO: Temporary solution. Providers lack Helm Apply() in test framework.
+          #       Long term: implement HelmInstallOverlay.Apply() similar to other ProviderInstallOverlay implementations.
+          # Read libvirt config from properties file (created by config_libvirt.sh)
+          source libvirt.properties
```
mkulke: we should create an issue describing what needs to be done, so we don't forget about it

.github/workflows/e2e_libvirt.yaml
```diff
@@ -141,14 +148,21 @@ jobs:
...
+        if: ${{ inputs.install_method == 'kustomize' }}
         run: |
           command -v kustomize >/dev/null || \
           curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | \
             sudo bash -s /usr/local/bin
 
+      - name: Install Helm
+        if: ${{ inputs.install_method == 'helm' }}
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
```
mkulke: why not helm 4?

can we install helm from github releasess? (with a specific version and a sha256 check)

.github/workflows/e2e_libvirt.yaml
```diff
@@ -38,6 +38,11 @@ on:
         description: Name of the container runtime. Either containerd or crio.
         required: false
         type: string
+      install_method:
+        default: 'kustomize'
+        description: Installation method. Either kustomize or helm.
+        required: false
+        type: string
```
mkulke: should it be type: choice then w/ options kustomize,helm?

beraldoleal:
@mkulke thanks for the review, I pushed with your suggestions.

src/cloud-providers/Makefile
```diff
@@ -1,7 +1,70 @@
...
+				"    # \(.description)", \
+				"    # (default: \"\(.default)\")", \
+				"    # \(.env_var): \"\(.default)\"" \
+			end, \
+			"" \
+		)' > $(CHART_PROVIDERS_DIR)/$(1).yaml
+endef
+
+define gen-provider-secrets
+	./bin/config-extractor -o json -only-secrets $(1) | jq -r ' \
```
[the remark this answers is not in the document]
mkulke: Not 100% certain, but I would assume  that `SecretGenerator` in kustomize will store the values as kubernetes secrets? if the helm values are not stored as clear text somewhere in the cluster, I guess that's similar.

mkulke (approved):

PR #2657, opened 2025-10-27
[DO NOT MERGE] CI: bump k8s to v1.34

wainersm:
I'd like to test peer pods with K8S 1.34.1 + CRI-O but on my laptop kcli has been hit by "Too many requests" errors from github. Let me try it on CI....


stevenhorsman:
> @stevenhorsman @bpradipt I did this PR just to test one thing for Fabiano but, hey, it worked out. Maybe we should update k8s version used in our tests? At least the version in libvirt and aws. I can try to bump docker too.

Yeah, I'm all for this

wainersm:
> > Could you change the permission? or maybe safer, download the image manually with kcli download ?
> 
> @wainersm here is an ACL output:
> 
> ```shell
> $ getfacl /var/lib/libvirt/images
> getfacl: Removing leading '/' from absolute path names
> # file: var/lib/libvirt/images
> # owner: root
> # group: root
> user::rwx
> user:ansible:rwx                #effective:--x
> group::--x
> mask::--x
> other::--x
> ```
> 
> So the `mask::--x` was a cause. I ran the following:
> 
> ```
> sudo setfacl -m m:rwx /var/lib/libvirt/images
> ```
> 
> Now the CI job should pass, I believe. 😉

It worked out @BbolroC ! thank you so much!

wainersm:
Removed the tags that triggers for aws and libvirt because I'm going to push a commit to update docker (let's only tests on that provider now).

I had to build a new kindnode image with cri-o, as kindest/node is containerd. Used the guide in https://github.com/cri-o/cri-o/blob/main/tutorials/crio-in-kind.md#build-node-image . Pushed the image to my area in quay.io; if tests pass then I will push it to confidential-container's repository.

PR #1719, opened 2024-02-27
move cloud-providers to top go module

liudalibj:
Remove cyclic dependency among cloud-api-adaptor and peerpod-ctrl.
Move cloud-providers to a new go mod and re-structure all the source codes as:
```
cloud-api-adaptor
    \src\cloud-api-adaptor
         cloud-providers
         peerpod-ctl
         peerpodcfg-ctl
         webhook
         csi-wrapper
```
And updated all workflow yamls to reflect new directory structure.

part of https://github.com/confidential-containers/cloud-api-adaptor/issues/1122

Signed-off-by: Da Li Liu <liudali@cn.ibm.com>
Co-authored-by: James Tumber <james.tumber@ibm.com>

PR #2659, opened 2025-10-29
Use local go modules

gkurz:
Leverage the replace directives already present in the go.mod files and use the local CAA modules instead of pulling them from GH. For the sake of consistency.


gkurz:
> > @stevenhorsman do you remember why we don't use the local go modules? We had a long discussion on splitting vs single repo then the internal references....but I forgot it all!
> 
> I think it was to avoid circular-dependencies, no?

@mkulke @wainersm, 

IIUC circular dependencies were fixed by #1719 : move modules around and introduce the `replace` directives in the various `go.mod` files. The `replace` directives take precedence on whatever appears in the `required` section, i.e. the version is irrelevant and go will consider the local files instead of pulling from GH.

- `src/cloud-api-adaptor/` depends on `cloud-providers` and `peerpod-ctrl`
- `src/csi-wrapper/` depends on `cloud-api-adaptor`
- `src/peerpod-ctrl/` depends on `cloud-providers`

I don't see any cycle in there and this PR isn't doing any change in this area. What this PR is doing is to _simply_ replace the real version with a pseudo one to make it clear that the modules are local, and avoid confusion to (who|what)ever would want to bump the versions for these modules. I'll clarify that in the commit message and re-push.

stevenhorsman (approved):
I think this PR doesn't change how things work. I think we had a question/theory at some point about removing the replace directive for releases, but I'm not sure we ever got this working, so I think I've been blindly bumping this to no effect since. We probably need a follow up to update the release process and then a discussion about whether we should have this pinned for a release and how that mechanism works with our tagging approach (probably tag some components and update others then tag them)?

bpradipt (approved):
/lgtm
The changes looks good. I'm looking at the CI failures..

PR #2670, opened 2025-11-03
parser: fix env variable precedence

### LINKED ISSUE #2669: Environment variables ignored when flags have hardcoded defaults
### Describe the bug

Environment variables are not being respected for configuration values that have hardcoded defaults in flag definitions. This creates an inconsistent precedence model where some environment variables work while others don't, depending on whether the flag has a hardcoded default or not.

This affects multiple cloud providers (Azure, AWS, GCP, etc.) and prevents users from reliably configuring the application via environment variables.

### How to reproduce

 1. Set an environment variable and run without the corresponding CLI flag:
```
AZURE_INSTANCE_SIZE="Standard_D4s_v3" cloud-api-adaptor azure -clientid "test" 2>&1 | grep Size
```
  2. Observe the output:

`Size:Standard_DC2as_v5`

The environment variable `AZURE_INSTANCE_SIZE` is completely ignored, and the hardcoded default Standard_DC2as_v5 is used instead.

What I was expecting:

Consistent precedence across all configuration options:

  1. CLI flag (highest priority)
  2. Environment variable
  3. Hardcoded default (lowest priority)

The output should show: `Size:Standard_D4s_v3`

Actual Behavior

Inconsistent precedence that varies by field:

Some fields follow: CLI flag > Environment variable > Hardcoded default. Example:
```
 AZURE_REGION="eastus" cloud-api-adaptor azure -clientid "test" 2>&1 | grep Region
```
Output: Region:eastus (works!)

This inconsistency makes it confusing for users to understand which environment variables will actually work.

### CoCo version information

caa devel

### What TEE are you seeing the problem on

None

beraldoleal:
Environment variables were being ignored for configuration options that had hardcoded defaults in their flag definitions. This happened because flags were registered and parsed before LoadEnv() was called, so fields already had values by the time environment variable loading was attempted. Vide #2669 for details.

Previously, providers used two separate methods:

- ParseCmd() to register flags with hardcoded defaults
- LoadEnv() to load environment variables (which didn't work for fields with non-empty defaults)

This introduces a `FlagRegistrar` helper that checks environment variables during flag registration. Providers can now use methods like `StringWithEnv()`, `IntWithEnv()`, `BoolWithEnv()`, and `ListWithEnv()` to register flags that respect environment variables, with the expected precedence: CLI flag > Environment variable > Hardcoded default.

This will also improve the consistency across providers regarding the way arguments are parsed, since they would be using common parsers.

With this approach, LoadEnv() is no longer needed and can become a no-op, that could be deprecated soon.

Fixes #2669

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/docker/provision_common.go
+++ b/src/cloud-api-adaptor/test/provisioner/docker/provision_common.go
@@ -133,7 +133,7 @@ func (l *DockerProvisioner) DeleteVPC(ctx context.Context, cfg *envconf.Config) 
 }
 
 func (l *DockerProvisioner) GetProperties(ctx context.Context, cfg *envconf.Config) map[string]string {
-	return map[string]string{
+	allProps := map[string]string{
 		"DOCKER_HOST":         DockerProps.DockerHost,
 		"DOCKER_API_VERSION":  DockerProps.ApiVer,
 		"CLUSTER_NAME":        DockerProps.ClusterName,
@@ -145,6 +145,16 @@ func (l *DockerProvisioner) GetProperties(ctx context.Context, cfg *envconf.Conf
 		"TUNNEL_TYPE":         DockerProps.TunnelType,
 		"VXLAN_PORT":          DockerProps.VxlanPort,
 	}
+
+	// Filter out empty values to avoid overriding defaults
+	props := make(map[string]string)
+	for k, v := range allProps {
+		if v != "" {
+			props[k] = v
+		}
+	}
+
+	return props
 }
 
 func (l *DockerProvisioner) UploadPodvm(imagePath string, ctx context.Context, cfg *envconf.Config) error {


snir911 (commented):
would be good to actually switch the LoadEnv appearances along with this pr, to make sure implementation supports all use cases.
( i especially refer to peerpod-ctrl which IIRC externally uses LoadEnv to fetch the essential values for VM deletion)

beraldoleal:
> would be good to actually switch the LoadEnv appearances along with this pr, to make sure implementation supports all use cases. ( i especially refer to peerpod-ctrl which IIRC externally uses LoadEnv to fetch the essential values for VM deletion)

Thanks @snir911, for pointing out peerpod-ctrl! I was only focusing on the main binary.

I think we can still consolidate this with the FlagRegistrar approach. Since environment variable loading now happens automatically during flag registration in ParseCmd(), peerpod-ctrl can just call ParseCmd() with a dummy FlagSet to trigger the env loading logic. Check commit 0332576f6549b46a27102f57626a2dbb924ac98d.

All other LoadEnv() calls are from the cloud-api-adaptor main binary, which I would replace if you guys are ok with this approach. wdyt?

src/cloud-providers/azure/manager.go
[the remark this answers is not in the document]
beraldoleal: It's more about consistency, yes. In current production scenarios it's working because we're effectively parsing configuration twice:

  1. entrypoint.sh gets env vars and passes them as CLI arguments (e.g., `-instance-size=${AZURE_INSTANCE_SIZE}`), so this bypass the "issue" I reported.
  2. manager.go has LoadEnv() that tries to parse some env variables again

However, 2 is selective and inconsistent about what it parses. I initially thought it was only for security/sensitive variables, but I'm not sure that's the complete picture:

  Examples:
  - AWS: LoadEnv() handles credentials BUT ALSO PODVM_INSTANCE_TYPE
  - Azure: LoadEnv() handles credentials BUT ALSO region, resource group, instance size
  - vSphere: LoadEnv() only handles credentials
  - GCP: LoadEnv() only handles GCP_CREDENTIALS

I could not find a clear pattern, and its leading to inconsistencies between providers (for instance: instance type is loaded from env in AWS/Azure but not GCP; region in Azure but not AWS).

In production we use daemonset with entrypoint.sh converting env vars to flags. But entrypoint.sh conditionally passes flags, if an env var is unset, it expects LoadEnv() to pick up the value, which doesn't actually work when there's a hardcoded default. 

Also, I think the cloud-api-adaptor code could work correctly on its own, relying on entrypoint.sh to work around this, feels a bit fragile. The flag docs say "defaults to
  `AZURE_INSTANCE_SIZE`" but the env var gets silently ignored, which is confusing.

What I would like to propose here with this approach is to create a common pattern: everything is read the same way, with the same precedence. We're not selective about which variables.

The pattern:
  
  1. Default value is what we have hardcoded
  2. That might be overridden by ENV variables
  3. That might be overridden by cmdline arguments

This way, if some value must be passed via ENV and not via cmdline (for security or any other reason), the logic will support this by default, just don't document the CLI flag or handle it specially in entrypoint.sh.

Does this makes some sense?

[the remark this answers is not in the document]
stevenhorsman: Just to confirm that I understand - "overridden by cmdline arguments" doesn't necessarily make much sense to me from a zoomed out user perspective - are you referring to the flags that entrypoint sets by this?

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.StringWithEnv(&azurecfg.ResourceGroupName, "resourcegroup", "", "AZURE_RESOURCE_GROUP", "Resource Group")
+	reg.StringWithEnv(&azurecfg.Size, "instance-size", "Standard_DC2as_v5", "AZURE_INSTANCE_SIZE", "Instance size")
+
+	// Flags without environment variable support (pass empty string for envVarName)
+	reg.StringWithEnv(&azurecfg.Zone, "zone", "", "", "Zone")
+	reg.StringWithEnv(&azurecfg.SubnetId, "subnetid", "", "", "Network Subnet Id")
+	reg.StringWithEnv(&azurecfg.SecurityGroupId, "securitygroupid", "", "", "Security Group Id")
+	reg.StringWithEnv(&azurecfg.ImageId, "imageid", "", "", "Image Id")
+	reg.StringWithEnv(&azurecfg.SSHKeyPath, "ssh-key-path", "", "", "Path to SSH public key")
+	reg.StringWithEnv(&azurecfg.SSHUserName, "ssh-username", "peerpod", "", "SSH User Name")
```
stevenhorsman: Are these used by anything? I have a feeling they were made redundant recently?

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.BoolWithEnv(&azurecfg.DisableCVM, "disable-cvm", false, "", "Use non-CVMs for peer pods")
```
stevenhorsman: This is a shared option, so can we move the processing to `cloud-api-adaptor/main.go` (and it uses the `DISABLECVM` env) to reduce duplication please?

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.BoolWithEnv(&azurecfg.EnableSecureBoot, "enable-secure-boot", false, "", "Enable secure boot for the VMs")
```
stevenhorsman: This uses the `ENABLE_SECURE_BOOT` env

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.BoolWithEnv(&azurecfg.UsePublicIP, "use-public-ip", false, "", "Assign public IP to the PoD VM and use to connect to kata-agent")
```
stevenhorsman: This uses `USE_PUBLIC_IP` env, but is in 4 providers, so we could consider making it shared and moving to `main.go`

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.IntWithEnv(&azurecfg.RootVolumeSize, "root-volume-size", 0, "", "Root volume size in GB. Default is 0, which implies the default image disk size")
+
+	// Custom flag types (comma-separated lists)
+	reg.CustomTypeWithEnv(&azurecfg.InstanceSizes, "instance-sizes", "", "", "Instance sizes to be used for the Pod VMs, comma separated")
+	reg.CustomTypeWithEnv(&azurecfg.Tags, "tags", "", "", "Custom tags (key=value pairs) to be used for the Pod VMs, comma separated")
```
stevenhorsman: This uses `TAGS`

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
-	flags.BoolVar(&azurecfg.EnableSecureBoot, "enable-secure-boot", false, "Enable secure boot for the VMs")
-	flags.BoolVar(&azurecfg.UsePublicIP, "use-public-ip", false, "Assign public IP to the PoD VM and use to connect to kata-agent")
-	flags.IntVar(&azurecfg.RootVolumeSize, "root-volume-size", 0, "Root volume size in GB. Default is 0, which implies the default image disk size")
+	reg := provider.NewFlagRegistrar(flags)
+
+	// Flags with environment variable support
+	reg.StringWithEnv(&azurecfg.ClientId, "clientid", "", "AZURE_CLIENT_ID", "Client Id")
+	reg.StringWithEnv(&azurecfg.ClientSecret, "secret", "", "AZURE_CLIENT_SECRET", "Client Secret")
+	reg.StringWithEnv(&azurecfg.TenantId, "tenantid", "", "AZURE_TENANT_ID", "Tenant Id")
+	reg.StringWithEnv(&azurecfg.SubscriptionId, "subscriptionid", "", "AZURE_SUBSCRIPTION_ID", "Subscription ID")
```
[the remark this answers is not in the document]
beraldoleal: The way I'm seeing this is that we don't differentiate here those cases. Everything is an argument, that might have ENV variable or not. The caller (in our case, entrypoint.sh) will be the one deciding to hide or not. If we remove that here, we would need to parse the envs in a different way for each case, and that is thing I would like to remove with this PR. Have a single place to register arguments (followed or not by the envs). With this we minimize inconsistencies among the providers.

src/cloud-providers/azure/manager.go
beraldoleal: Yes I agree, I just would like to split the efforts if you don't mind. Right now its on those providers code, and I'm just doing "minimal" changes to migrate to the new registrar. wdyt?

src/cloud-providers/azure/manager.go
beraldoleal: same as before.

src/cloud-providers/azure/manager.go
```diff
@@ -18,37 +18,36 @@ func init() {
...
+	reg.StringWithEnv(&azurecfg.Region, "region", "", "AZURE_REGION", "Region")
+	reg.StringWithEnv(&azurecfg.ResourceGroupName, "resourcegroup", "", "AZURE_RESOURCE_GROUP", "Resource Group")
+	reg.StringWithEnv(&azurecfg.Size, "instance-size", "Standard_DC2as_v5", "AZURE_INSTANCE_SIZE", "Instance size")
+
+	// Flags without environment variable support (pass empty string for envVarName)
+	reg.StringWithEnv(&azurecfg.Zone, "zone", "", "", "Zone")
+	reg.StringWithEnv(&azurecfg.SubnetId, "subnetid", "", "", "Network Subnet Id")
```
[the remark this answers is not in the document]
beraldoleal: I guess I can, sure! I was just trying to do in baby steps! :)

src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -58,17 +68,7 @@ one_of() {
...
-    [[ "${PODVM_AMI_ID}" ]] && optionals+="-imageid ${PODVM_AMI_ID} "                                       # MUST if template is not used
-    [[ "${PODVM_INSTANCE_TYPE}" ]] && optionals+="-instance-type ${PODVM_INSTANCE_TYPE} "                   # default m6a.large
-    [[ "${PODVM_INSTANCE_TYPES}" ]] && optionals+="-instance-types ${PODVM_INSTANCE_TYPES} "
-    [[ "${SSH_KP_NAME}" ]] && optionals+="-keyname ${SSH_KP_NAME} "                    # if not retrieved from IMDS
-    [[ "${AWS_SUBNET_ID}" ]] && optionals+="-subnetid ${AWS_SUBNET_ID} "               # if not set retrieved from IMDS
-    [[ "${AWS_REGION}" ]] && optionals+="-aws-region ${AWS_REGION} "                   # if not set retrieved from IMDS
-    [[ "${TAGS}" ]] && optionals+="-tags $(cleanup_spaces "${TAGS}") "                 # Custom tags applied to pod vm
-    [[ "${USE_PUBLIC_IP}" == "true" ]] && optionals+="-use-public-ip "                 # Use public IP for pod vm
-    [[ "${ROOT_VOLUME_SIZE}" ]] && optionals+="-root-volume-size ${ROOT_VOLUME_SIZE} " # Specify root volume size for pod vm
+    # Global flags without env var support - still need conversion
```
stevenhorsman: Sorry, I don't quite understand this - the values are read here from envs aren't they, so what is the missing env var support?

src/cloud-api-adaptor/entrypoint.sh
stevenhorsman: Do you think there is an argument that the common env -> flag conversions should also be updated (in main.go I guess) to use your env processing code to avoid this switch from env -> flag elsewhere of your code like you've done with the provider specific ones?

src/cloud-api-adaptor/entrypoint.sh
stevenhorsman: Okay, is that a future PR?

beraldoleal: I just pushed.

stevenhorsman (commented):
A couple of comments, but otherwise this looks good. I think Magnus' suggestion to use an existing library is probably sensible, but as a lot of work has clearly gone in here to help tidy these, I think it's worth using this as a platform for that refactoring.

src/cloud-providers/azure/manager.go
stevenhorsman: I agreed not to move - but you missed adding the `DISABLECVM` env in azure which was added in the other repos to parse from the envs

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -102,35 +102,38 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
-		flags.StringVar(&cfg.serverConfig.ForwarderPort, "forwarder-port", daemon.DefaultListenPort, "port number of agent protocol forwarder")
-		flags.StringVar(&tlsConfig.CAFile, "ca-cert-file", "", "CA cert file")
-		flags.StringVar(&tlsConfig.CertFile, "cert-file", "", "cert file")
-		flags.StringVar(&tlsConfig.KeyFile, "cert-key", "", "cert key")
-		flags.BoolVar(&tlsConfig.SkipVerify, "tls-skip-verify", false, "Skip TLS certificate verification - use it only for testing")
+		reg := provider.NewFlagRegistrar(flags)
+
+		// Common flags with environment variable support
+		reg.StringWithEnv(&cfg.serverConfig.SocketPath, "socket", adaptor.DefaultSocketPath, "REMOTE_HYPERVISOR_ENDPOINT", "Unix domain socket path of remote hypervisor service")
+		reg.StringWithEnv(&cfg.serverConfig.PodsDir, "pods-dir", adaptor.DefaultPodsDir, "PODS_DIR", "base directory for pod directories")
```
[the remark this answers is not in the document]
beraldoleal: Yes, this inconsistency existed before this PR. The kustomization files have incorrectly documented PEER_PODS_DIR since the feature was added in d9b1b9a6, while entrypoint.sh has always actually read PODS_DIR.

beraldoleal:
@stevenhorsman please let me know if you still see some pending issues. I believe we fixed all the open issues.

beraldoleal:
@stevenhorsman are those know issues? I tried to see if it was caused by this change, but it doesn't look like.

src/cloud-providers/azure/manager.go
beraldoleal: fixed. sorry, I missed this msg.

beraldoleal:
@stevenhorsman docker and aws errors fixed... I'm trying to understand those libvirt issues...

beraldoleal:
@stevenhorsman @wainersm those libvirt issues seems to be related to https://github.com/confidential-containers/cloud-api-adaptor/issues/2635. I can see the same securecomms issue on the logs.

stevenhorsman:
> @stevenhorsman @wainersm those libvirt issues seems to be related to #2635. I can see the same securecomms issue on the logs.

Yeah, I wouldn't worry about the secure comms failures, that's why I just mentioned docker and aws earlier. Thanks!

stevenhorsman:
@beraldoleal - can you rebase this and hopefully we'll be able to get the CI passing enough to get this in?

stevenhorsman:
Enough of the tests are passing, so let's get this in now. Thanks @beraldoleal

mkulke:
the azure e2e tests are failing since we merged these changes I think, could this be related? It looks like the daemonset pod doesn't start: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/19724201799/job/56520519408

PR #2672, opened 2025-11-04
Add GCP subnetwork support with flexible format handling

mateo-moon:
This PR adds support for configuring GCP subnetworks when creating Pod VMs. The subnetwork parameter supports multiple formats for flexibility:

- **Full path**: `projects/<project>/regions/<region>/subnetworks/<subnetwork>`
- **Partial path**: `regions/<region>/subnetworks/<subnetwork>`
- **Short name**: `<subnetwork>` (automatically formatted to full path based on zone)

The subnetwork is required for custom subnet mode networks and is automatically formatted based on the zone configuration.

## Changes

- Add `Subnetwork` field to GCP Config type
- Add `-subnetwork` CLI flag in manager
- Add `GCP_SUBNETWORK` environment variable support in entrypoint.sh
- Implement subnetwork formatting logic in provider.go to handle different input formats

## Testing

- Tested with full path format
- Tested with short name format (auto-formatted)
- Verified backward compatibility when subnetwork is not specified


beraldoleal (commented):
Thanks for this @mateo-moon . I would just remove the first commit from this PR.

beraldoleal (approved):
LGTM, thanks

PR #2673, opened 2025-11-05
workflows: Switch CAA arch build to use ibm cloud runners

stevenhorsman:
Now we have onboarded the ibm cloud runners for s390x and ppc64le
try and use these for the relevant stages of CAA multi-arch, so that we can avoid emulation and speed up the builds significantly

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -7,3 +7,5 @@ self-hosted-runner:
   # Labels of self-hosted runner that linter should ignore
   labels:
     - S390X
+    - ubuntu-24.04-ppc64le
+    - ubuntu-24.04-s390x
--- a/.github/workflows/caa_build_and_push_per_arch.yaml
+++ b/.github/workflows/caa_build_and_push_per_arch.yaml
@@ -64,23 +64,18 @@ jobs:
   build_push_job:
     name: build and push
     needs: [upload_tags]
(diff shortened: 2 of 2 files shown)


BbolroC:
FYI: I've finished a proper review. It does look good to me. Thanks!

stevenhorsman:
I've created https://github.com/confidential-containers/cloud-api-adaptor/pull/2677 to try and address the issue with setup-go

wainersm:
> I think it's because setup-go is installing the non-le version of ppc:
> 
> ```
> Setup go version spec 1.24.9
> Attempting to download 1.24.9...
> matching 1.24.9...
> Not found in manifest.  Falling back to download directly from Go
> Install from dist
> Acquiring go1.24.9 from https://storage.googleapis.com/golang/go1.24.9.linux-ppc64.tar.gz
> ```
> 
> so maybe set-up go doesn't work properly with power?

Hi @stevenhorsman !

Any chance of the runner getting a ppc64 (big endian) by mistake?

I glanced at `setup-go` source it seems to be using TypeScript's `os.arch()` to get the arch in https://github.com/actions/setup-go/blob/main/src/installer.ts#L45 . In https://github.com/actions/setup-go/blob/main/src/system.ts#L19 there is a mapping to convert from system's arch to go's arch but no mapping for ppc64le. So it's really using the output of `os.arch()`.

PR #2676, opened 2025-11-06
config-extractor: an utility to get the provider variables

beraldoleal:
This would help us to get the default values, types and documentaion of each variable. We can use this to populate tables and our charts/kustomize templates.

This depends on #2670.

### DIFF
--- a/.github/workflows/validate-config-extractor.yaml
+++ b/.github/workflows/validate-config-extractor.yaml
@@ -1,0 +1,82 @@
+# (C) Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Validate config-extractor tool
+---
+name: validate-config-extractor
+
+on:
+  pull_request:
+    paths:
+      - 'src/cloud-providers/**/manager.go'
+      - 'src/cloud-providers/util.go'
+      - 'src/cloud-providers/cmd/config-extractor/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  test-extractor:
+    name: Test config-extractor tool
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: src/cloud-providers
+    steps:
+      - name: Checkout the pull request code
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
+          persist-credentials: false
+
+      - name: Read properties from versions.yaml
+        run: |
+          go_version="$(yq '.tools.golang' ../cloud-api-adaptor/versions.yaml)"
+          [ -n "$go_version" ]
+          echo "GO_VERSION=${go_version}" >> "$GITHUB_ENV"
+
+      - name: Setup Golang version ${{ env.GO_VERSION }}
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache-dependency-path: "**/go.sum"
+
+      - name: Build config-extractor
+        run: make config-extractor
+
+      - name: Test all providers parse successfully
+        run: |
+          # Find all directories with manager.go (excluding cmd and util)
+          for provider_dir in */; do
+            provider=$(basename "$provider_dir")
+
+            # Skip non-provider directories
+            if [[ "$provider" == "cmd" || "$provider" == "util" || "$provider" == "bin" ]]; then
+              continue
+            fi
+
+            # Check if manager.go exists
+            if [[ ! -f "${provider}/manager.go" ]]; then
+              echo "Error: ${provider}/manager.go not found"
+              echo "This could indicate a refactoring issue or incomplete provider implementation"
+              exit 1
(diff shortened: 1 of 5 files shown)


beraldoleal:
> Hi @beraldoleal !
> 
> > @stevenhorsman.... given the amount of config options we have across providers, this could help us to generate better templates (either helm or kustomize), in a way that we would avoid having the values hardcoded there. This would be my next PR, focusing on the helms.
> > Let me know what you think, so I can proceed.
> 
> I think I know what you want to do. But could you give us a concrete example?

Sure, for instance, for helm or kustomize templates, we could have something like this:

```go
// Example: ./config-extractor -o json gcp | go run build-values.go

  var config ProviderConfig
  json.NewDecoder(os.Stdin).Decode(&config)

  fmt.Printf("%s:\n", config.Provider)
  for _, flag := range config.Flags {
      fmt.Printf("  # %s\n", flag.Description)
      fmt.Printf("  # Default: \"%s\"\n", flag.Default)
      fmt.Printf("  # %s: \"\"\n\n", flag.EnvVar)
  }
```

And we could have an automatic generated yaml section like this:

  Output:
```
  gcp:
    # Google Application Credentials
    # Default: ""
    # GCP_CREDENTIALS: ""

    # GCP Project ID description here
    # Default: ""
    # GCP_PROJECT_ID: ""

    # Zone description here
    # Default: ""
    # GCP_ZONE: "" 
```

This way, we don ´t have to maintain templates or documentation about those values.

.github/workflows/validate-config-extractor.yaml
```diff
@@ -0,0 +1,79 @@
...
+            provider=$(basename "$provider_dir")
+
+            # Skip non-provider directories
+            if [[ "$provider" == "cmd" || "$provider" == "util" ]]; then
+              continue
+            fi
+
+            # Check if manager.go exists
+            if [[ ! -f "${provider}/manager.go" ]]; then
+              continue
```
stevenhorsman: Would it be worth error-ing here? It feels like if the manager.go file is missing then either we've refactored and need to add it to the non-provider directory, or a half-baked cloud-provider has gone it which is error worthy? Or am I missing something?

src/cloud-providers/cmd/config-extractor/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```bash
+./config-extractor [-o json|table] <provider-name>
+```
+
+### Arguments
+
+- `-o`: Output format (default: `json`)
+  - `json`: Outputs flags in JSON format
+  - `table`: Outputs flags in a formatted table
```
stevenhorsman: What is the use-case for the table option? Do we need it as we are not testing it?

.github/workflows/validate-config-extractor.yaml
beraldoleal: I think its ok to bail here. Doing ....

src/cloud-providers/cmd/config-extractor/README.md
```diff
@@ -0,0 +1,79 @@
...
+From the `cloud-providers` directory:
+
+```bash
+make config-extractor
+```
+
+## Usage
+
+```bash
+./config-extractor [-o json|table] <provider-name>
```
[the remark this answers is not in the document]
beraldoleal: yes, it was. But I moved now to the bin/ folder. And updated the readme.

src/cloud-providers/cmd/config-extractor/README.md
beraldoleal: The table option is for human readability... While AI is almost replacing me, I still prefer tables over JSON. :)

I'm not testing it because it's just pretty-printing the same data the JSON test already validates. If the parser works, the table works, testing behavior, not presentation.

.github/workflows/validate-config-extractor.yaml
```diff
@@ -0,0 +1,79 @@
...
+            echo "::group::Testing provider: $provider"
+            ./cmd/config-extractor/config-extractor -o json "$provider" > /tmp/"$provider".json
+
+            # Verify output is valid JSON
+            if ! jq empty /tmp/"$provider".json 2>/dev/null; then
+              echo "❌ Invalid JSON output for $provider"
+              exit 1
+            fi
+
+            # Verify at least some flags were extracted
```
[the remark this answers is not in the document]
beraldoleal: Yep, I get you.

.github/workflows/validate-config-extractor.yaml
```diff
@@ -0,0 +1,82 @@
...
+            provider=$(basename "$provider_dir")
+
+            # Skip non-provider directories
+            if [[ "$provider" == "cmd" || "$provider" == "util" || "$provider" == "bin" ]]; then
+              continue
+            fi
+
+            # Check if manager.go exists
+            if [[ ! -f "${provider}/manager.go" ]]; then
+              echo "❌ Error: ${provider}/manager.go not found"
```
wainersm: Hi @beraldoleal !

LLMs generate these nasty (IMHO) Unicode on messages. Could we avoid it in our code? Let's keep the minimum of dignity here :)

.github/workflows/validate-config-extractor.yaml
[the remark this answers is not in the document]
wainersm: How I understood this: eventually that action will generate some helm files and diff with the ones already checked in the repository. If they diverge then flags are out of sync, raise an error. So the verification done here is just bogus at this point (I'm okay to leave it as is)

PR #2677, opened 2025-11-06
workflows: Set the ppc64le arch explicitly in setup-go

stevenhorsman:
The setup-go action incorrectly installs the big-endian version of go, so try and hard-coded the architecture to work around this.


wainersm (approved):

ldoktor (approved):
lgtm and the bug is still opened so let's add this for now

PR #2682, opened 2025-11-12
versions: Bump kata runtime to 3.23.0 ~pre-~release

stevenhorsman:
The guest-components and trustee have been bumped on kata, so let's check ahead of time that they work for us too.


src/cloud-api-adaptor/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor
 
-go 1.24.6
+go 1.24.9
 
-toolchain go1.24.7
+toolchain go1.24.10
```
bpradipt: A question - can the toolchain and go version be in sync ?

PR #2687, opened 2025-11-18
Install UKI under BOOTX64 for GCE

### LINKED ISSUE #2686: PodVM fails to boot on GCE SEV nodes
## Summary
Building the podvm image with `make image` and uploading it to Google Cloud's SEV-capable `n2d-*` instances results in a firmware boot loop. The Unified Kernel Image is only placed under `\EFI\Linux\`, but the GCE firmware only probes the generic fallback path `\EFI\BOOT\BOOTX64.EFI`. Because of that the VM never loads the kernel.

## Steps to Reproduce
1. `cd src/cloud-api-adaptor/podvm-mkosi && DOCKER_DEFAULT_PLATFORM=linux/amd64 make image`
2. Upload `build/system.raw` to GCE (e.g. via `uplosi upload -c . build/system.raw`).
3. Create a pod that launches a PeerPod VM on a `n2d-standard-*` node and watch the serial console.

## Actual Result
The console repeatedly prints messages similar to the following and the VM never boots:
```
UEFI: Failed to load image.
Description: UEFI nvme_card-pd
FilePath: PciRoot(0x0)/Pci(0x4,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)
Status: Not Found.
```

## Expected Result
The VM should find the Unified Kernel Image and continue booting the pod VM OS.

## Additional Context
Copying the generated UKI (`/efi/EFI/Linux/*.efi`) to `/efi/EFI/BOOT/BOOTX64.EFI` inside the image fixes the issue because it satisifies the firmware's fallback lookup. The fix is safe for other platforms because it only runs when the UKI exists.

mateo-moon:
## Summary
Google Cloud's SEV-capable `n2d-*` instances only probe `\EFI\BOOT\BOOTX64.EFI`.  The podvm image previously left the Unified Kernel Image solely under `\EFI\Linux\`, so the firmware could not find it and the VM looped with:
```
UEFI: Failed to load image.
Description: UEFI nvme_card-pd
Status: Not Found.
```

This change copies the generated UKI to `\EFI\BOOT\BOOTX64.EFI` during the mkosi finalize step.  The copy only happens if the UKI exists, so it is safe for other environments.  The README now documents why we do this for GCE.

Fixes #2686.

## Testing
- `DOCKER_DEFAULT_PLATFORM=linux/amd64 ARCH=x86_64 make image`
- Uploaded the raw image with `uplosi upload -c . build/system.raw`
- Launched a PeerPod on `n2d-standard-4` and confirmed the VM boots instead of looping in firmware.

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/README.md
+++ b/src/cloud-api-adaptor/podvm-mkosi/README.md
@@ -51,6 +51,14 @@ If you want to use the image with libvirt, run the following to convert to qcow2
 qemu-img convert -f raw -O qcow2 build/system.raw build/system.qcow2
 ```
 
+#### Firmware expectations on GCE
+
+When the pod VM runs on Google Cloud's SEV-capable `n2d-*` instances the
+virtual firmware only probes `\EFI\BOOT\BOOTX64.EFI`.  The mkosi finalize
+step copies the generated Unified Kernel Image from `\EFI\Linux\*.efi` to
+that fallback path automatically, ensuring the VM boots without any manual
+intervention.
+
 ## Debug image
 
 There is a debug variant of the image that provides a specific configuration to debug things within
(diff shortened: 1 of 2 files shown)


mateo-moon:
> /lgtm It would be good to add a reference link if available.

Ref link? What is it?

PR #2688, opened 2025-11-18
GCP: support for using internal instance ips

EmmEff:
This PR implements support for using internal IPs of instances or external (NAT) IPs on GCP.

CAA setting `USE_PUBLIC_IP` toggles this behaviour and matches the functionality present in other providers.


bpradipt (approved):
/lgtm
@EmmEff please rebase

PR #2693, opened 2025-11-21
Release/0.17.0

stevenhorsman:
- Pin the CAA image to the latest release commit
- Update the release process doc to reflect the latest situation


docs/Release-Process.md
```diff
@@ -31,27 +31,22 @@ the kata-containers release version.
         ```
         in the [cloud-api-adaptor](../src/cloud-api-adaptor/) directory and [csi-wrapper](../src/csi-wrapper/) directory.
 1. The CoCo operator updates to use references to the other component releases and then releases itself
+    - In order to have the correct version of the kata-containers payload in our peer pods releases, we need to
+    wait for this CoCo operator release before we can start the peer pods release process. After this operator payload
```
mkulke: I understand this will be the last release of the the CoCo operator?

PR #2699, opened 2025-11-28
parser: ignore empty environment variables

beraldoleal:
Treat empty environment variables as "not set" to prevent template placeholders (e.g., AZURE_SUBSCRIPTION_ID="") from overwriting defaults.

This approach is debatable. If the community prefers empty strings to override defaults, we can revert this and comment out empty variables in kustomization templates instead.

This might be the root cause of pods failing with timeout because empty env vars cause cloud provider initialization to fail silently, preventing readiness probes from succeeding.

### DIFF
--- a/src/cloud-providers/util.go
+++ b/src/cloud-providers/util.go
@@ -178,7 +178,7 @@ func (r *FlagRegistrar) StringWithEnv(field *string, flagName, hardcodedDefault,
 	*field = hardcodedDefault
 
 	if envVarName != "" {
-		if envValue, exists := os.LookupEnv(envVarName); exists {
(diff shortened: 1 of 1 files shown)


beraldoleal:
@stevenhorsman sure, but I'm afraid I can't create a new branch here. It seems that I lack permissions.

mkulke:
https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/19768714726

stevenhorsman:
I passed in `/CommunityGalleries/cocopodvm-d0e4f35f-5530-4b9c-8596-112487cdea85/Images/podvm_image0/Versions/2025.11.28` which I got from the nightly, but I'm guessing Magnus would have given one too?

PR #2700, opened 2025-11-28
fix(gcp): fix PodVM networking interface, internal IP usage and cross-build

mateo-moon:
This PR fixes networking and build issues when running Confidential Containers on GCP with PodVM images built on non-x86 hosts.

**Fixes:**
1.  **Network Interface Naming:** Updated `mkosi.conf` to disable consistent network device naming (`net.ifnames=0 biosdevname=0`). This ensures the primary interface is always `eth0`, matching `systemd-networkd` configuration and `agent-protocol-forwarder` expectations. Previously, interfaces were named `ens3` or similar, causing DHCP and service failures.
2.  **PodVM Firewall:** Disabled `nftables` and `iptables` services in the PodVM image to prevent default rules from blocking the agent port (15150).
3.  **GCP Internal IP:** Updated the GCP cloud provider to prioritize the internal network IP of the PodVM instance. Previously, it was only looking for external NAT IPs, which caused connectivity issues if public IP access was restricted or not preferred.
4.  **Cross-Compilation:** Patched `hack/cross-build-extras.sh` to support cross-compilation on Fedora (e.g., building x86_64 image on Apple Silicon). It now creates a symlink for `gcc` to the cross-compiler when necessary, fixing build failures for components like `skopeo`.
5.  **GCP Subnetwork:** Added support for `Subnetwork` configuration in GCP provider to allow deploying PodVMs in specific subnets.

**Verification:**
- Built `podvm-mkosi-fedora` image on arm64 host for x86_64 target.
- Deployed on GKE cluster.
- Verified successful pod startup, image pulling, and agent connectivity via internal IP.

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.finalize.chroot
+++ b/src/cloud-api-adaptor/podvm-mkosi/mkosi.presets/system/mkosi.finalize.chroot
@@ -42,10 +42,9 @@ systemctl disable iptables.service || true
 systemctl mask nftables.service || true
 systemctl mask iptables.service || true
 
-# Disable setup-nat-for-imds.service for GCP (only needed for Azure/AWS)
-# GCP uses metadata.google.internal which doesn't require NAT setup
-echo "Disabling setup-nat-for-imds.service (not needed for GCP)"
-systemctl disable setup-nat-for-imds.service || true
+# Note: setup-nat-for-imds.service is left enabled by default (via 30-coco.preset)
+# It is needed for Azure/AWS IMDS access. On GCP, the service will fail gracefully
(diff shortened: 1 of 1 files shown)


src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -76,6 +76,19 @@ alibabacloud() {
 gcp() {
     test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
 
+    # TODO: The following manual conversions are kept for compatibility but should
```
beraldoleal: Could you split this commit into individual ones, please? With proper git log description, explaining why each change is needed?

PR #2708, opened 2025-12-04
ibmcloud: Align environment variable naming

Pacho20:
This pull request addresses inconsistencies in IBM Cloud environment variable naming:

- Renamed `IBMCLOUD_TAGS` back to `TAGS`
   The `TAGS` environment variable was accidentally renamed to `IBMCLOUD_TAGS` during the introduction of the ParseCmd method. Restoring the original name ensures alignment with other cloud providers and compatibility with IBMCloudProvisioner.

- Fixed cluster_id naming mismatch
  The IBM Cloud manager expects `IBMCLOUD_CLUSTER_ID`, while the provisioner returned `CLUSTER_ID`. Updated references to maintain consistency across components.

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_ibmcloud.properties
+++ b/src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_ibmcloud.properties
@@ -3,6 +3,8 @@ IBMCLOUD_PROVIDER="ibmcloud"
 APIKEY="${MY_VPC_APIKEY}"
 IAM_PROFILE_ID="${MY_IAM_PROFILE_ID}"
 CLUSTER_NAME="e2e-test1"
+# optional, normally CAA can auto-detect this
+CLUSTER_ID=""
 # Resource list -> storage -> a cos service ->
 COS_BUCKET="peerpod-cos-bucket"
 # Resource list -> storage -> a cos service -> instances -> an cos service instance -> service credentials -> resource_instance_id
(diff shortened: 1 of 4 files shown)


src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_ibmcloud.properties
```diff
@@ -3,6 +3,8 @@ IBMCLOUD_PROVIDER="ibmcloud"
 APIKEY="${MY_VPC_APIKEY}"
 IAM_PROFILE_ID="${MY_IAM_PROFILE_ID}"
 CLUSTER_NAME="e2e-test1"
+# optional, normally CAA can auto-detect this
+CLUSTER_ID=""
```
wainersm: Hi @Pacho20 ,

`IBMCLOUD_CLUSTER_ID` here too?

Pacho20: Not necessarily. If you check the [InitIBMCloudProperties](https://github.com/confidential-containers/cloud-api-adaptor/blob/2bdfc074d6570940d1e20735b4b0dd9117b3ccd8/src/cloud-api-adaptor/test/provisioner/ibmcloud/provision_initializer.go#L69) function, it uses a different name, just `CLUSTER_ID`. That’s why it’s different in the properties file. But I can change those too, so they’re consistent everywhere. What do you suggest?

PR #2710, opened 2025-12-05
podvm: Upgrade RHEL base images to ubi10.1

Rohit-Singh43-1:
Updated UBI base image to 10.1 in Dockerfile.podvm_builder.rhel.
Updated codeready-builder-for-rhel repository from 9 to 10.
Changed tool name from genisoimage to xorriso dependent to rhel10 
Added missing SE_VERIFY flag in Makefile during Docker build for PodVM image.


stevenhorsman (approved):
We need to be aware that packer is deprecated upstream, but it seems reasonable to update this to get CVE fixes etc before we remove it. Thanks @Rohit-Singh43-1

PR #2717, opened 2025-12-10
CI: testing AWS confidential VMs

wainersm:
This is the first step towards to run the attestation-aware tests on AWS. It was introduced a new job to run the non-CoCo tests on EKS clusters, using AMD SEV-SNP podvm instances.

In e2e framework there were some code to instantiate EKS but it was broken and CoCo wasn't getting installed on Amazon Linux. I switched most of the code from AWS Go SDK to  calling `eksctl` tool to provision/deprovision clusters as well as migrated to use Ubuntu 24.04 workers (where CoCo installs just fine). Along the way I had to do some fixes and adjustments here and there.

Although we have a clean up mechanism for dangling resources running in our AWS account, I've adapted (and fixed) the script that run after the job fail.

The new job will be running in `continue-on-error` until it's unstable. Yes, it's still a bit unstable because often the EKS cluster provision fails (need to investigate and fix).

Two executions that I used to test:

* Everything went well: https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/20031775323/job/57478524742
* Provision failed and clean up script ran: https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/20076771410/job/57594806839

### DIFF
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -68,6 +68,8 @@ jobs:
     needs: aws-credentials
     if: needs.aws-credentials.outputs.has_secrets == 'true'
     runs-on: ubuntu-22.04
+    # Remove this when the job becomes stable
+    continue-on-error: ${{ inputs.cluster_type == 'eks' }}
     defaults:
       run:
         working-directory: src/cloud-api-adaptor


src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
```diff
@@ -1249,6 +1249,13 @@ func NewEKSCluster(cfg aws.Config, vpc *Vpc, SshKpName string) *EKSCluster {
 func (e *EKSCluster) CreateCluster() error {
 	var err error
 
+	if e.Name != "" {
+		log.Infof("Using existing EKS cluster: %s", e.Name)
+		return nil
+	}
+
+	e.Name = e.Vpc.BaseName + "-k8s"
```
stevenhorsman: Is this always unique, or could the vpc basename be re-used?

.github/workflows/e2e_run_all.yaml
```diff
@@ -225,19 +225,28 @@ jobs:
...
         provider:
           - generic
         arch:
           - amd64
+        include:
+          - container_runtime: containerd
+            cluster_type: eks
+            os: ubuntu
+            provider: generic
+            arch: amd64
```
stevenhorsman: Could we refactor this into a table for more readability?

src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
wainersm: Hi @stevenhorsman , it can be re-used. You just need to pass the `resources_basename` property.

.github/workflows/e2e_run_all.yaml
wainersm: Yes, we can!

wainersm: done. that's what you wanted @stevenhorsman ?

PR #2726, opened 2025-12-15
ibmcloud: remove VpcServiceURL default value

Pacho20:
Removed the VpcServiceURL default value because it prevents CAA from auto-detecting the value when not provided, which results in using the wrong URL. Before FlagRegistrar, the empty environment variable overrode the default flag value.


ldoktor (approved):
thanks, it seems to be set in "provision_initializer.go" based on region when it's "" so it should be indeed better.

stevenhorsman:
> @stevenhorsman not sure about ibmcloud, do we need any labels to test these changes or is review-only sufficient for merge?

We don't have any ibmcloud e2e tests, so review (and unit tests I guess) is fine. Thanks for checking!

PR #2728, opened 2025-12-16
Reapply "versions: Pin operator version"

stevenhorsman:
This reverts commit b3cc5867b3100e58a0d515a231c4f1b2a8712ed4.

No that the operator is not supported and working, we need to pin to a stable version, so pick the last release


BbolroC (approved):
LGTM, thanks @stevenhorsman !

mkulke (approved):

Issue #2755, opened 2026-01-05
vsphere: Drop the vsphere cloud provider


stevenhorsman:
As discussed in the CoCo 0.17 release notes - the vsphere provider will be dropped due to it being unmaintained for a long time and non-strategic

PR #2756, opened 2026-01-05
Remove vsphere cloud provider

stevenhorsman:
Remove vsphere provider in line with the deprecation notice in the 0.17.0 release

Fixes: #2755


mkulke:
the failed tests are securecomms tests, I think we can ignore them

Issue #2773, opened 2026-01-11
alibabacloud: DeleteInstance always fails due to incorrect return value


tak-ka3:
## Summary

In `src/cloud-providers/alibabacloud/provider.go`, the `DeleteInstance` function always fails even when the instance is successfully deleted.

## Problem

At line 387, when `DeleteInstance` API call succeeds (`err == nil`), the code incorrectly returns:

```go
return false, fmt.Errorf("failed to describe instance %s: %+v", instanceID, err)
```

This causes `waitUntilTimeout` to:
1. Receive `err != nil` (since `fmt.Errorf` always returns non-nil error)
2. Return the error immediately
3. `DeleteInstance` reports failure even though the instance was successfully deleted

## Expected Behavior

When deletion succeeds, it should return `true, nil` to signal success to `waitUntilTimeout`.

## Fix

```go
// Before (line 387)
return false, fmt.Errorf("failed to describe instance %s: %+v", instanceID, err)

// After
return true, nil
```

## Additional Note

The error message also incorrectly says "failed to describe instance" when it should relate to `DeleteInstance`.

PR #2774, opened 2026-01-11
alibabacloud: Fix DeleteInstance always failing on success

tak-ka3:
## Summary

When `DeleteInstance` API call succeeds (`err == nil`), the code incorrectly returned `false, fmt.Errorf(...)` which caused `waitUntilTimeout` to immediately return an error instead of recognizing success.

This meant `DeleteInstance` would always report failure even when the instance was successfully deleted.

Fix by returning `true, nil` on successful deletion.

Fixes: #2773


Xynnn007:
Thanks. I missed this fix but it LGTM!

PR #2776, opened 2026-01-12
workflows: Auto go mod tidy dependabot PRs

stevenhorsman:
Do to our repo structure of go modules, we frequently hit issues that dependabout bumps versions, but other modules that reference the changes module don't get updated, so we have error with go mod and sum not being tidied. Attempt to create a workflow that will automatically do this change, so that we don't need to do it manually anymore.

### DIFF
--- a/.github/workflows/dependabot_go_mod_tidy.yaml
+++ b/.github/workflows/dependabot_go_mod_tidy.yaml
@@ -62,19 +62,37 @@ jobs:
 
       - name: Commit and push changes
         if: steps.go-mod-tidy-changed.outputs.changed == 'true'
-        uses: devops-infra/action-commit-push@8a2d9d73c3f506468129be2e4409e60dbed70357 # v1.0.3
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          commit_message: "chore: Auto-run go mod tidy after Dependabot update"
+        # uses: devops-infra/action-commit-push@8a2d9d73c3f506468129be2e4409e60dbed70357 # v1.0.3
+        # with:
+        #   github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


mkulke:
@stevenhorsman I was trying something similar, it's immensely frustrating: the culprit here seems to be that after a workflow-triggered change to the PR new test runs aren't triggered, because actions triggered by workflows are prohibited from triggering another automation (presumably to break loops).

I had a comment-based trigger in mind ("tidy-plz"), but I failed due make it work pleasantly due to the above restrictions.

mkulke:
I thought abou the same, but how will you do that? afaict you have to push an empty commit _manually_ , b/c you cannot run test for the workflow commit. if you have to do that, then you can just run mod tidy locally, too. It‘s mad…

mkulke:
> Ok, I've tried to update this to include the new proposal - do the go mod tidy, create a new PR (which hopeful will run the CI) and then close the original PR. Again - I'm not sure that it will work, or how to test it before merging, but maybe it's worth a try?

I think I tested it by pushing a (broken) dependabot PR's branch to my fork and then opening a PR (on my fork) myself

ldoktor:
I think we'll have to test it. I replaced some PRs by manual ones before, important is not to include "ignore" or such (which extends the dependabot ignore list). But brief check with AI wasn't much optimistic and it suggested using:

```
on:
  pull_request_target:
    types: [opened, synchronize, reopened]
```

instead. Anyway I'd prefer the closing if it works as I'd expect:

```
upstream  - 0.1
dependabot detects 0.2 and creates PR
workflow creates 0.2 + go mod tidy and closes dependabot 0.2 PR without saying to ignore this further
dependabot detects 0.3 and creates PR
workflow creates 0.3 + go mod tidy and closes dependabot 0.3
```

(when we merge the 0.2 or not). If it doesn't work this way we'd have to find another solution.

PR #2777, opened 2026-01-12
Simplify caa multi arch build 

stevenhorsman:
The CAA multi-arch build is confusing (at least to me) with uploaded files pointing to sha that are later downloaded to create a dynamic manifest based on them. I think by hardcoded the arches we include in the manifest (which doesn't change often), we can re-use the "standard" caa build and then create a manifest in a more standard way.

Along the way fix some inconsistency with tags and zizmor issues.


ldoktor (commented):
@stevenhorsman do you intend to convert this from draft? To me it looks better than the current approach. I haven't reviewed thoroughly the new image tags (which is why I'm not giving the ack) but the publishing seems very good and the new tags schema is cleaner.

stevenhorsman:
Ok, this seems to be tested and working okay in my fork, so maybe it's not risky like I though: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/21719552373

src/cloud-api-adaptor/hack/build.sh
```diff
@@ -40,8 +39,8 @@ function get_tag_string() {
 
 	# if we are building fora single arch, then also append this to the tag
 	if [[ -n "${arch:-}" ]] && [[ "${arch:-}" != *,* ]];then
-		# strip the `linux/` from the docker buildx platform format
-		arch_suffix="-${arch#"linux/"}"
+		# strip the `linux/` from the docker buildx platform format in the tag
+		arch_suffix="-${arch#${arch_prefix}}"
```
ldoktor: This confused me a bit, I was looking for the change but it's only an unrelated cleanup :D (very minor, mentioning it just to save other reviewers time)

ldoktor: Btw two lines above is also a typoe "if we ar ebuilding *for a* single arch" (again, not critical but if you happen to be updating it you might consider including that fix as well)

ldoktor (approved):
I did my best, your branch seems to be passing and I do like the naming unification. Let's get that in and hammer-out the details if needed. Thanks @stevenhorsman

src/cloud-api-adaptor/hack/build.sh
```diff
@@ -17,23 +17,35 @@ if [[ "$commit" = unknown ]]; then
 	[[ -n "$(git status --porcelain --untracked-files=no)" ]] && commit+='-dirty'
 fi
 
-dev_tags=${DEV_TAGS:-"latest,dev-${commit}"}
-release_tags=${RELEASE_TAGS:-"${commit}"}
+tags=""
+if [[ "${release_build}" == "true" ]]; then
+	tags=${RELEASE_TAGS:-"${commit}"}
+else
+	tags=${DEV_TAGS:?}
```
wainersm: Hi @stevenhorsman ,

is there a reason to not provide default tags like before and like RELEASE_TAGS? i.e. default would be `latest,${commit}-dev`

stevenhorsman: I'm not 100% sure as I wrote this weeks and weeks ago, but I think my view was that it's just somewhere else to duplicate logic that we have elsewhere and need to keep in sync

.github/workflows/caa_build_and_push_all_arches.yaml
```diff
@@ -201,8 +88,20 @@ jobs:
...
+        run: |
+          ./hack/publish.sh publish-multiarch-manifest
         env:
-          registry: ${{ inputs.registry }}
-        run:  |
-          hack/image-manifest.sh
+          IMAGE_REGISTRY: ${{ inputs.registry }}
+          IMAGE_NAME: "cloud-api-adaptor"
+          IMAGE_TAGS: ${{ inputs.dev_tags}}
+          ARCHES: "amd64,arm64,ppc64le,s390x"
```
wainersm, line 98: `publish.sh` provides default for `ARCHES`. To avoid keeping it in sync maybe we should just drop that line, or remove the defaults.

src/cloud-api-adaptor/hack/build.sh
wainersm: Without a default I suspect `azure-e2e-test.yaml` will fail because it does `cd src/cloud-api-adator & make image` under the hoods which turns out to run `hack/build.sh -i` and `DEV_TAGS` seems to not be exported in that code path. Then we should export `DEV_TAGS` from the workflow.

.github/workflows/caa_build_and_push_all_arches.yaml
stevenhorsman: I didn't have that originally, but I think Lukas asked me to add it above! Let me leave the option, but remove the defaults

.github/workflows/publish_images_on_push.yaml
```diff
@@ -19,9 +19,11 @@ permissions: {}
 
 jobs:
   caa:
-    uses: ./.github/workflows/caa_build_and_push_per_arch.yaml
+    uses: ./.github/workflows/caa_build_and_push_all_arches.yaml
     with:
       git_ref: ${{ github.sha }}
+      dev_tags: ${{ github.sha }}-dev
```
wainersm: Could we also keep publishing the `latest` dev image?

src/cloud-api-adaptor/hack/build.sh
stevenhorsman: Azure doesn't have dev tags, lets add the default back into the script

stevenhorsman: Thanks for the important catch!

src/cloud-api-adaptor/hack/publish.sh
```diff
@@ -0,0 +1,50 @@
...
+		echo "Validating architecture-specific images for tag: ${tag}"
+		for image in "${images[@]}"; do
+			if ! docker manifest inspect "${image}" > /dev/null 2>&1; then
+				echo "Error: Required image does not exist: ${image}" >&2
+				echo "Please ensure all architecture-specific images are built and pushed before creating the manifest." >&2
+				exit 1
+			fi
+		done
+
+		docker manifest create "${IMAGE_REGISTRY}/${IMAGE_NAME}:${tag}" "${images[@]}"
```
wainersm, line 35: There is a feedback from Claude that I could not check myself:
```
The old image-manifest.sh used docker buildx imagetools create which works server-side without pulling images. The new publish.sh uses docker   
  manifest create/push which works but requires images to be pullable (the docker manifest inspect validation helps here). This is functional, but
   the buildx imagetools approach is generally preferred for CI as it's faster and doesn't require local pulls.
```

I don't think it will increase dramatically the build and also the caa image is not exactly big. But it might worth trying to save pulls to avoid network issues. I am okay to have it address in a follow up PR if this makes sense to you.

wainersm (approved):
Thanks @stevenhorsman !

src/cloud-api-adaptor/hack/publish.sh
stevenhorsman: I asked my AI and it agreed with you:
> “The suggestion is reasonable, but since [src/cloud-api-adaptor/hack/publish.sh](vscode-webview://153nfalmvv5hnka1hn8aglvaotfm34dreuh590132m1i60astaev/src/cloud-api-adaptor/hack/publish.sh) is already functionally correct, I’d prefer to keep this PR scoped and treat a switch to Buildx imagetools as a separate cleanup unless we have evidence that the current approach causes problems.”

to the point that I asked it to implement the change and it keeps refusing and saying it's not neccessary 🫤

wainersm: It's hard to understand these kids :D

wainersm:
hey @stevenhorsman , let's merge it? or are you still polishing it?

PR #2779, opened 2026-01-14
helm: introducing peerpods-ctrl and webhook charts

beraldoleal:
This introduces Helm charts for peerpod-ctrl and webhook components as proper dependencies of the main peerpods chart, replacing the previous kustomize-based installer Job.

### DIFF
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -112,8 +112,12 @@ jobs:
           sudo snap install yq
           go_version="$(yq '.tools.golang' versions.yaml)"
           [ -n "$go_version" ]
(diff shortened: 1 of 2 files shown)


stevenhorsman (commented):
To my charts ignorance this looks good. I think we might need to check the e2e tests and ensure that we don't end up with a clash of both ways installing.

src/webhook/chart/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+This file is automatically regenerated when running:
+
+```bash
+make manifests
+```
+
+> **Note**: The generated file is committed to git so users can install the
+> chart without running `make manifests`. Developers should run this command
+> after modifying the webhook markers in the Go code.
```
mkulke: see comment above

PR #2790, opened 2026-01-21
helm: configure DS updateStrategy for --wait

beraldoleal:
Change DaemonSet updateStrategy from hardcoded maxUnavailable: 1 to
configurable values with defaults maxUnavailable: 0, maxSurge: 1.

With maxUnavailable: 1 on single node cluster, helm --wait can return successfully even when pods are not ready. Example: 1-node cluster with maxUnavailable: 1 means minReady = 0, so helm --wait doesn't actually wait for the pod. This forces tests to implement custom wait logic after helm install.

With maxUnavailable: 0, maxSurge: 1, at least one pod must be ready before helm --wait returns. Works correctly on both small and large clusters, provides zero-downtime rolling updates, and tests can now rely on helm --wait instead of custom waits.

Users can override these values for specific deployment needs.

### DIFF
--- a/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
@@ -12,7 +12,8 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 1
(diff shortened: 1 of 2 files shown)


wainersm (approved):
@beraldoleal code wise, looks good. I was testing it before approve. It is working indeed, I even have a problem with resourcesctrl not being ready and installation failing.

PR #2794, opened 2026-01-22
Coco 0.18.0 release versions

stevenhorsman:
Let's pin the components we consume to sync with the CoCo 0.18 release versions.


wainersm:
ok, tomorrow I can send the PR for enabling CI coverage to all providers with peerpods's helm. If things goes well then maybe in 1+ week max we may be able to put caa 0.18.

PR #2795, opened 2026-01-22
workflows: Disable go cache

stevenhorsman:
Disable the go cache for security reasons


wainersm (approved):
Thanks @stevenhorsman !

ps: next opportunity I can show all my power with Cursor at hand :)

PR #2808, opened 2026-01-26
helm: fix automatic namespace and subcharts patches 

beraldoleal:
Description on each commit.


stevenhorsman (approved):
With the disclaimer that I don't 100% understand all of the helm pieces, this seems fine to be and the CI passing is great.

PR #2809, opened 2026-01-28
Libvirt helm e2e tests

stevenhorsman:
Move the provisioning of helm in libvirt out of the workflow into the Helm part of the provisioner framework

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/libvirt/provision.go
+++ b/src/cloud-api-adaptor/test/provisioner/libvirt/provision.go
@@ -12,4 +12,5 @@ import (
 func init() {
 	pv.NewProvisionerFunctions["libvirt"] = NewLibvirtProvisioner
 	pv.NewInstallOverlayFunctions["libvirt"] = NewLibvirtInstallOverlay
+	pv.NewInstallChartFunctions["libvirt"] = NewLibvirtInstallChart
 }
--- a/src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
+++ b/src/cloud-api-adaptor/test/provisioner/libvirt/provision_common.go
@@ -6,6 +6,7 @@ package libvirt
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"os"
 	"os/exec"
@@ -25,6 +26,7 @@ const DefaultCPU = 2
 
 // LibvirtProvisioner implements the CloudProvisioner interface for Libvirt.
 type LibvirtProvisioner struct {
+	caa_image        string           // The CAA image to install
 	conn             *libvirt.Connect // Libvirt connection
 	containerRuntime string           // Name of the container runtime
 	network          string           // Network name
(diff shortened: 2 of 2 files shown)


wainersm:
> > Hi @stevenhorsman ,
> > Libvirt's CI uses default values set by kustomize: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/overlays/libvirt/kustomization.yaml#L19
> > In helm, it's not setting by default any value (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/charts/peerpods/providers/libvirt.yaml#L13), so it is leaving the defaults of manager.go (https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-providers/libvirt/manager.go#L19). Notice that default values of helm are different from kustomize, for example, DISABLECVM=false (vs true in kustomize) and the libvirt URI.... if we ensure the default of helm are same as kustomize then probably CI will pass.
> > ps: in #2811 @beraldoleal has fixed the parsing of default values in manager.go, but he is not changing the defaults of libvirt.
> 
> Before we dig into design choices... let me show here the current drifts we have: (Generated by claude and using the parser fix):
> #### libvirt
> Key 	Kustomize 	Helm
> DISABLECVM 	true 	false
> LIBVIRT_EFI_FIRMWARE 	/usr/share/OVMF/OVMF_CODE_4M.fd 	(empty)
> LIBVIRT_URI 	qemu+ssh://root@192.168.122.1/system?no_verify=1 	qemu:///system
> #### azure
> Key 	Kustomize 	Helm
> SSH_USERNAME 	(empty) 	peerpod
> #### ibmcloud
> Key 	Kustomize 	Helm
> DISABLECVM 	true 	false
> IBMCLOUD_IAM_ENDPOINT 	(empty) 	https://iam.cloud.ibm.com/identity/token
> 
> Apart from DISABLECVM that is a global one, do we want to move those defaults to the go code? or, is there any reason those are not the default in code?

I suggest to change the defaults of manager.go to be equal of kustomize so that we don't break the users (e.g. CI) that rely on default values.

As for `DISABLECVM` I don't know what to do for sure. My feeling is that if we make it configurable per provider, and defaulting to same value as in kustomize then we won't have further troubles.

beraldoleal:
> I suggest to change the defaults of manager.go to be equal of kustomize so that we don't break the users (e.g. CI) that rely on default values.

> As for DISABLECVM I don't know what to do for sure. My feeling is that if we make it configurable per provider, and defaulting to same value as in kustomize then we won't have further troubles.

This makes sense to me. 

We would just need to decide the following:

1 . LIBVIRT_URI

TBH, I'm not sure here. But it feels goes on the same as the previous ones.

2. SSH_USERNAME

It feels strange to have each provider with a different user, or why the kustomize was using a different value. It seems to me that we want the peerpods default that is already in place.

3. IBMCLOUD_IAM_ENDPOINT

Here, it was added as default in go code, so not sure why the kustomize was still using the "". Some good reason? Or just a drift that we forgot to update there too?

beraldoleal:
@stevenhorsman https://github.com/beraldoleal/cloud-api-adaptor/actions/runs/21646500732/job/62400894623

Libvirt e2e + helm is passing on my local branch... basically this PR + some minor fixes:

https://github.com/confidential-containers/cloud-api-adaptor/compare/main...beraldoleal:cloud-api-adaptor:main

stevenhorsman:
Ok - I think (thanks @beraldoleal and @wainersm) that this is working and cleaned up now. Tested on my fork: https://github.com/stevenhorsman/cloud-api-adaptor/actions/runs/21673526744/job/62488447963

PR #2810, opened 2026-01-28
RFC (post 0.18): libvirt and azure helm provisioning support with cleaner API (conflicts with #2809)

beraldoleal:
Hi folks, this seems big, but I hope its easy to review for those that are involved in the helm code at this point.

Before we implement the remaining providers, I'd like you to consider this approach if possible. This is groundwork for removing several map conversions in the code that can now be avoided with helm. Doing it now would avoid more rework later.

  1. This flow is helm-only; kustomize remains unchanged
  2. Eliminates boilerplate code and removes the need for each provider to implement certain APIs                                                                     
  3. Forces us to write helm values, which helps validate the chart
  4. Implemented for AWS, Docker, libvirt, and Azure, but testing so far has been basic (build and verifying templates generate correctly)

Risk looks low to me (though I might be colorblind to danger)... happy to postpone if needed.

If that makes sense, I can continue testing on that.

Note: I chose to replace helm.go with new code to make reviewing a bit easier, but most of the work there came from Wainer's interface.

### DIFF
--- a/.github/workflows/azure-e2e-test.yml
+++ b/.github/workflows/azure-e2e-test.yml
@@ -19,6 +19,11 @@ on:
       podvm-image-id:
         type: string
         description: prebuilt podvm image
+      install_method:
+        default: 'kustomize'
+        description: Installation method. Either kustomize or helm.
+        required: false
+        type: string
     secrets:
       AZURE_CLIENT_ID:
         required: true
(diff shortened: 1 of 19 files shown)


wainersm:
Hi @beraldoleal ,

I only glanced at the changes and honestly I disliked it because it changes the "philosophy" of the framework. I don't want to go into further details because I think it's not time to make that change, IMO we should focus on what is strictly needed for the 0.18 release. I understand it has boilerplates but it works; IMO we should focus on just covering the other providers.

PR #2824, opened 2026-02-02
helm: conditional kata-deploy dependency

beraldoleal:
Vide commits.

### DIFF
--- a/.github/workflows/publish-peerpods-chart.yaml
+++ b/.github/workflows/publish-peerpods-chart.yaml
@@ -25,9 +25,10 @@ jobs:
     name: Release PeerPods Helm Chart
     runs-on: ubuntu-22.04
     permissions:
-      contents: read   # Required for checkout
-      packages: write  # Required for pushing to GitHub Container Registry
-      id-token: write  # Required for attestation signing
+      contents: read       # Required for checkout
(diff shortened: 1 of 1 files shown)


src/cloud-api-adaptor/install/charts/peerpods/Chart.yaml
```diff
@@ -21,7 +21,7 @@ maintainers:
 
 dependencies:
   - name: kata-deploy
-    version: "3.25.0"
+    version: "3.26.0"
```
wainersm: Hi @beraldoleal , for upcoming release we want to match the version released with CoCo 0.18, which is kata 3.25

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
```diff
@@ -84,6 +84,10 @@ webhook:
 
 # kata-deploy subchart configuration
 kata-deploy:
+  # Installs kata and the kata-remote RuntimeClass required by peer pods.
```
wainersm: Can't helm deal with the fact that kata-deploy might be already deployed and then just skip? Or maybe have helm to threat dependencies install as idempotent?

wainersm (approved):
Thanks @beraldoleal !

PR #2826, opened 2026-02-04
dependabot: Add cooldown

stevenhorsman:
It's good practice to get dependabot to wait after a release before bumping to avoid it bumping to a release done seconds before, which could have supply-chain security implications, so add a 7 day cooldown to help with this.

See https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,8 @@ updates:
       # Check for updates to GitHub Actions every week on Saturday
       interval: "weekly"
       day: "saturday"
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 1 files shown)


wainersm (approved):
Hi @stevenhorsman ,

I was reading the blog post, very interesting indeed. My concern was about missing security fixes, but the text makes clear that cooldown usually does not apply for CVEs, and I checked in https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown- that's the case for dependabot.

Cool. Thanks!

PR #2845, opened 2026-02-11
charts: adding manual tls support

beraldoleal:
This is just adding manual tls support as we have today in kustomize. But I find redundant a bit, and could be improved later.


wainersm (approved):
Thanks @beraldoleal !

mkulke (approved):

PR #2847, opened 2026-02-11
install/helm: fix generation of empty providerConfigs

wainersm:
The providerConfigs for libvirt and docker has all the values commented and sync-chart-values generates `libvirt: {}` and `docker: {}` entries. This is not syntactically wrong, however, if user uncomment a value (e.g. `LIBVIRT_URI`) and forget to delete `{}` then it becomes an invalid yaml.

In reality the extract `{}` is not needed, so let's remove it.

### DIFF
--- a/src/cloud-api-adaptor/install/charts/peerpods/providers/docker.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/providers/docker.yaml
@@ -10,7 +10,7 @@ image:
   tag: "latest"
 
 providerConfigs:
-  docker: {}
+  docker:
     # CA cert file
     # (default: "")
     # CACERT_FILE: ""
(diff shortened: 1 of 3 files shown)


wainersm:
> Oh - it looks like the e2e isn't happy with this change?

Unfortunately it isn't :(

The error is like:

```
Error: INSTALLATION FAILED: failed parsing --set-literal data: unable to parse key: interface conversion: interface {} is nil, not map[string]interface {}
F0211 18:39:13.245423   20211 env.go:369] Setup failure: failed to install helm chart: exit status 1, output: Error: INSTALLATION FAILED: failed parsing --set-literal data: unable to parse key: interface conversion: interface {} is nil, not map[string]interface {}
```

If I call helm with `libvirt.yaml` untouched and I do *not* pass any `--set-*` parameter then it gets installed. On the other hand if any have the same untouched yaml and pass, for example, `--set-literal providerConfigs.libvirt.LIBVIRT_URI=blah` then the error above. It seems a bug on helm command that parses `providerConfigs.libvirt` as `nil` instead of an empty map. That problem was reported to helm years ago but closed unfixed (https://github.com/hashicorp/terraform-provider-helm/issues/773).

I think the best I can do is to add a comment in the yaml file warning users to remove the `{}` in case they want to set values in the `providerConfigs.PROVIDER` section... does it work?

PR #2850, opened 2026-02-12
test/provisioner: fix bugs on Azure provisioner for helm install

wainersm:
The first commit fixes a silly mistake that breaks the deploy completely.

The second commit ensures the AZURE_TENANT_ID will be in the peer-pods-secrets; that's needed for the workload identity.


mkulke (approved):
fwiw: the docs claim that tenant_id should not be required for workload identity: https://github.github.io/gh-aw/blog/2026-01-13-meet-the-workflows-documentation/

cc @pawelpros

PR #2860, opened 2026-02-19
helm: Move ext-network flags to FlagRegistrar

beraldoleal:
EXTERNAL_NETWORK_VIA_PODVM and POD_SUBNET_CIDRS were registered as raw flags without env var binding, relying on entrypoint.sh for conversion. Move them to FlagRegistrar so the config-extractor picks them up and they appear in generated helm values.

Thanks to @wainersm for finding this.


wainersm (approved):
Thanks @beraldoleal !

stevenhorsman (approved):
LGTM. Thanks

PR #2867, opened 2026-02-25
workflow: Add artifact-metadata: write

stevenhorsman:
The azure nightly e2e tests are showing the following warning:
```
 build-podvm-image / build-podvm-image
Please check that the "artifact-metadata:write" permission has been included
```
which seems to be a new permission added recently, so try and address this.


stevenhorsman:
Let's see if this helps with the warning and funky github ui tonight 🤞

PR #2869, opened 2026-02-26
podvm: remove fetching commit hash approach using curl

pawelpros:
Removed Curl and Github token from makefile to simplify building process.


mkulke (approved):
small nits, otherwise lgtm. thanks!

PR #2870, opened 2026-02-26
ci: add artifact-metadata to az image build job

mkulke:
it probably also need to be propagated to the individal job, not just on the call site.


stevenhorsman:
This is a dup of #2868, but that gives me confidence that I got it right this time!

PR #2879, opened 2026-03-02
test: remove kustomize-based CAA install from e2e

wainersm:
First removed all traces of kustomize install from the providers then from the base provider file. CI and users won't need to export `INSTALL_METHOD=helm` after these changes.

Notice that InstallOverlay and kustomize module are still kept as they are using by the KBS installation.


stevenhorsman:
The golangci-lint issue:
```
src/cloud-api-adaptor/src/cloud-api-adaptor/test/provisioner/gcp/utils.go:24:6: func copyFile is unused 
```
seems helpful here

PR #2903, opened 2026-03-09
workflows: Use IBM Actions Z runners for s390x CI jobs

BbolroC:
As a follow-up to #2673, switch the remaining self-hosted runners labeled 's390x' to IBM Actions Z runners.

IBM Actions Z runners currently do not support nested virtualization, so keep the existing configuration for the runner labeled 's390x-large'.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,6 +6,5 @@
 self-hosted-runner:
   # Labels of self-hosted runner that linter should ignore
   labels:
-    - S390X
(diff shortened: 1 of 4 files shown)


BbolroC:
### Note to reviewers

This change will likely require follow-up PRs to fix CI issues that cannot be predicted before merging. Since we are close to a new release, please merge this after the release. Thanks!

wainersm (approved):
Thanks @BbolroC !

PR #2914, opened 2026-03-11
mkosi: Bump to Fedora 43

BbolroC:
As Fedora 41 reached EOL in May 2025, it seems the GPG online bundle now only contains keys for Fedora 42 and above, not 41. 

Bump to Fedora 43.


BbolroC:
An error was observed on s390x: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/22942801078/job/66587941401

PR #2916, opened 2026-03-11
Gpu

yousef-cohere:


bpradipt:
@yousef-cohere we are progressing towards switching from Fedora to Ubuntu based mkosi and one of the reasons is stable support for NVIDIA GPUs. Take a look at - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi.ubuntu

Can you target your work towards Ubuntu mkosi. Also wondering if we can use systemd extensions for such additions. 

cc @mkulke

mkulke:
> @yousef-cohere we are progressing towards switching from Fedora to Ubuntu based mkosi and one of the reasons is stable support for NVIDIA GPUs. Take a look at - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi.ubuntu
> 
> Can you target your work towards Ubuntu mkosi. Also wondering if we can use systemd extensions for such additions.
> 
> cc @mkulke

I don't think sysext is a good fit here, we own the image building process fully and the point of migration to ubuntu was to have supported nvidia drivers from the distro's repository, no?

bpradipt:
> > @yousef-cohere we are progressing towards switching from Fedora to Ubuntu based mkosi and one of the reasons is stable support for NVIDIA GPUs. Take a look at - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi.ubuntu
> > Can you target your work towards Ubuntu mkosi. Also wondering if we can use systemd extensions for such additions.
> > cc @mkulke
> 
> I don't think sysext is a good fit here, we own the image building process fully and the point of migration to ubuntu was to have supported nvidia drivers from the distro's repository, no?

I was wondering whether we should have a single image or a base image + extensions?  Is having extensions even worth it for our use case? Not sure honestly

mkulke:
> > > @yousef-cohere we are progressing towards switching from Fedora to Ubuntu based mkosi and one of the reasons is stable support for NVIDIA GPUs. Take a look at - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.mkosi.ubuntu
> > > Can you target your work towards Ubuntu mkosi. Also wondering if we can use systemd extensions for such additions.
> > > cc @mkulke
> > 
> > 
> > I don't think sysext is a good fit here, we own the image building process fully and the point of migration to ubuntu was to have supported nvidia drivers from the distro's repository, no?
> 
> I was wondering whether we should have a single image or a base image + extensions? Is having extensions even worth it for our use case? Not sure honestly

I wouldn't start with that without a good reason, it'll add complexity.

PR #2923, opened 2026-03-13
gha: publish the podvm-byom-binaries-image

Amulyam24:
Build and publish the byom binaries image.

### DIFF
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -320,11 +320,10 @@ jobs:
       github.event_name == 'schedule' ||
       github.event_name == 'workflow_dispatch' ||
       contains(github.event.pull_request.labels.*.name, 'test_e2e_byom')
-    needs: [podvm_ubuntu_amd64, caa_image_amd64, peerpod_ctrl_image_amd64, webhook_image]
+    needs: [caa_image_amd64, peerpod_ctrl_image_amd64, webhook_image]
     uses: ./.github/workflows/e2e_byom.yaml
     with:
       caa_image: ${{ inputs.registry }}/cloud-api-adaptor:${{ inputs.caa_image_tag }}-dev-amd64
-      podvm_image: ${{ needs.podvm_ubuntu_amd64.outputs.byom_e2e_image }}
       git_ref: ${{ inputs.git_ref }}
       peerpod_ctrl_image: ${{ inputs.registry }}/peerpod-ctrl:${{ inputs.caa_image_tag }}-amd64
       webhook_image: ${{ inputs.registry }}/peer-pods-webhook:${{ inputs.caa_image_tag }}
--- a/.github/workflows/podvm_byom_binaries_publish.yaml
+++ b/.github/workflows/podvm_byom_binaries_publish.yaml
@@ -1,0 +1,103 @@
+# Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Build and push the BYOM binaries image.
+---
+name: Publish BYOM Binaries Image
+on:
+  workflow_dispatch:
+    inputs:
+      git_ref:
+        description: 'Git ref to build from'
+        required: false
+        type: string
+        default: ''
+      image_tag:
+        description: 'Image tag to use'
+        required: false
+        type: string
+        default: ''
+      arch:
+        description: 'Architecture to build for (e.g., amd64, arm64, s390x)'
+        required: false
+        type: string
+        default: 'amd64'
+      registry:
+        description: 'Container registry to push images to (e.g., quay.io/your-username)'
+        required: false
+        type: string
+        default: ''
+  workflow_call:
+    inputs:
+      git_ref:
+        description: 'Git ref to build from'
+        required: false
+        type: string
+        default: ''
+      image_tag:
+        description: 'Image tag to use'
+        required: false
+        type: string
+        default: ''
+      arch:
+        description: 'Architecture to build for (e.g., amd64, arm64, s390x)'
+        required: false
+        type: string
+        default: 'amd64'
+      registry:
+        description: 'Container registry to push images to (e.g., quay.io/your-username)'
+        required: false
+        type: string
+        default: ''
+    secrets:
+      QUAY_PASSWORD:
+        required: true
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    working-directory: src/cloud-api-adaptor
+
+jobs:
+  build-and-push:
+    name: Build and push BYOM binaries image
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read  # Required for checkout
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.git_ref || github.sha }}
+          persist-credentials: false
+
+      - name: Login to Container Registry
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: quay.io
+          username: ${{ vars.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build BYOM binaries image
+        working-directory: src/cloud-api-adaptor/podvm
+        run: make podvm-byom-binaries-image
+        env:
+          REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
+          PODVM_TAG: ${{ inputs.image_tag || github.sha }}
+          ARCH: ${{ inputs.arch }}
+          PODVM_DISTRO: ubuntu
+
+      - name: Push BYOM binaries image
+        working-directory: src/cloud-api-adaptor/podvm
+        run: make push-podvm-byom-binaries-image
+        env:
+          REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
+          PODVM_TAG: ${{ inputs.image_tag || github.sha }}
+          ARCH: ${{ inputs.arch }}
+          PODVM_DISTRO: ubuntu
--- a/.github/workflows/podvm_mkosi_ubuntu.yaml
+++ b/.github/workflows/podvm_mkosi_ubuntu.yaml
@@ -57,12 +57,6 @@ on:
       qcow2_oras_image:
         description: The location of the qcow2 oras container this workflow pushed
         value: ${{ jobs.build-image.outputs.qcow2_oras_image }}
-      docker_oci_image:
-        description: The location of the docker oci container image this workflow pushed
-        value: ${{ jobs.build-image.outputs.docker_oci_image }}
-      byom_e2e_image:
-        description: The location of the BYOM e2e container image
-        value: ${{ jobs.build-image.outputs.byom_e2e_image }}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-mkosi-ubuntu-${{ toJSON(inputs) }}
@@ -86,8 +80,6 @@ jobs:
       artifact-metadata: write # Required by actions/attest to write attestation metadata
     outputs:
       qcow2_oras_image: ${{ steps.publish_oras_qcow2.outputs.image }}:${{ steps.publish_oras_qcow2.outputs.tag }}
-      docker_oci_image: ${{ steps.build_docker_oci.outputs.image }}
-      byom_e2e_image: ${{ steps.build_byom_e2e.outputs.image }}
     steps:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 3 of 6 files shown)


.github/workflows/podvm_binaries.yaml
```diff
@@ -72,7 +72,9 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Build and push
-      run: make podvm-binaries
+      run: |
+        make podvm-binaries
+        make podvm-byom-binaries-image
```
stevenhorsman: I'm not sure if you are aware that this is part of the packer image workflow which is deprecated?

Amulyam24: Hi @stevenhorsman, can you point me to where would be an appropriate place to enable publishing the byom-binaries-image? which would be dependent on the original binaries image being published by `make podvm-binaries` 

I see that this workflow is being called in three different actions files and added it here after referring to `podvm_publish.yaml` and `release.yaml `

Amulyam24:
I will move this to podvm-mkosi-ubuntu

.github/workflows/podvm_mkosi_ubuntu.yaml
```diff
@@ -156,6 +156,20 @@ jobs:
           ARCH: ${{ inputs.arch }}
           PODVM_DISTRO: ubuntu
           VERIFY_PROVENANCE: yes
+          PUSH: "true"
+          REGISTRY: ${{ inputs.registry }}
+          PODVM_TAG: ${{ inputs.image_tag }}
+
+      - name: Build and push BYOM binaries image
```
mkulke: Should we move this into its own workflow? It feels a bit out-of-place here and the additional binary + image build will extend build time for all generic e2e test runs. the e2e_byom workflow might also be a good place to run this step

stevenhorsman: Good point. I think having it as part of the BYOM e2e tests definitely makes sense, there is then a question of integrating it into the release as well, so it being a separate workflow would enable that more easily I think.

stevenhorsman: I think we would want to move out:
```
- name: Clean up some space for the docker podvm image build
        working-directory: src/cloud-api-adaptor/podvm
        run: |
          sudo du --max-depth=2 /home/runner || true
          sudo du --max-depth=2 /var/lib || true
          sudo rm -rf ./build
          sudo rm -rf ./mkosi.cache

      - name: Build and push image for docker podvm image
        id: build_docker_oci
        working-directory: src/cloud-api-adaptor/podvm
        run: |
          tag=$(git rev-parse --short HEAD)
          PODVM_TAG=${tag} make image-container
          PODVM_TAG=${tag} make push-image-container
          arch=$(uname -m)
          arch=${arch/x86_64/amd64}
          echo "image=${REGISTRY}/podvm-docker-image-${arch}:${tag}" >> "$GITHUB_OUTPUT"
        env:
          REGISTRY: ${{ inputs.registry }}
          PODVM_DISTRO: ubuntu

      - name: Build and push BYOM e2e image
        id: build_byom_e2e
        working-directory: src/cloud-api-adaptor/podvm
        run: |
          # Build BYOM e2e image (depends on docker image built in previous step)
          PODVM_TAG=${IMAGE_TAG} make image-byom-e2e-container
          PODVM_TAG=${IMAGE_TAG} make push-image-byom-e2e-container
          arch=$(uname -m)
          arch=${arch/x86_64/amd64}
          echo "image=${REGISTRY}/podvm-byom-e2e-image-${arch}:${IMAGE_TAG}" >> "$GITHUB_OUTPUT"
        env:
          IMAGE_TAG: ${{ inputs.image_tag }}
          REGISTRY: ${{ inputs.registry }}
          PODVM_DISTRO: ubuntu
```
if the docker ones are even still needed? as well to the same reasons. These should probably be called by podvm_publish.yaml for the release flow.

stevenhorsman:
Sorry - really basic question (feel free to point me towards the docs that explains this). What is this image for and how does it relate to the `podvm-byom-e2e-image` image. Is that just the "docker container" version of this image and this is the "proper" qcow2? (How does that related to podvm-docker-image?) If so how do we test this image?

.github/workflows/podvm_mkosi_ubuntu.yaml
Amulyam24: The docker and byom e2e images are only needed by the BYOM e2e workflow, I will move them to it.

The podvm byom binaries image would ideally need to be published for every release, does it make sense to add it to a separate workflow and call it from the release.yaml?

Amulyam24:
> Sorry - really basic question (feel free to point me towards the docs that explains this). What is this image for and how does it relate to the `podvm-byom-e2e-image` image. Is that just the "docker container" version of this image and this is the "proper" qcow2? (How does that related to podvm-docker-image?) If so how do we test this image?

The `byom-e2e-container` image is the docker version of podvm with added BYOM support(SFTP) that is only consumed by the BYOM e2e.


The `podvm-byom-binaries-image` contains the necessary changes which would turn an existing VM into a pool VM to be used with BYOM provider. This gives flexibility alongside using the prebuilt podvm images to create the pool VMs.

It is primarily consumed by the setup script - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/hack/setup-podvm-byom.sh

Amulyam24:
Addressed the flow and moved the steps accordingly, PTAL!

.github/workflows/e2e_run_all.yaml
```diff
@@ -320,11 +320,10 @@ jobs:
       github.event_name == 'schedule' ||
       github.event_name == 'workflow_dispatch' ||
       contains(github.event.pull_request.labels.*.name, 'test_e2e_byom')
-    needs: [podvm_ubuntu_amd64, caa_image_amd64, peerpod_ctrl_image_amd64, webhook_image]
```
mkulke, line 323: seems to be an unrelated change?

mkulke (commented):
generally looks good, the zizmor errors need to be addressed

.github/workflows/e2e_run_all.yaml
Amulyam24: @mkulke, for the byom e2e, we no longer depend on the `podvm_ubuntu_amd64` step since we moved the docker and byom-e2e image builds to the byom-e2e workflow and `podvm_ubuntu_amd64` can be removed.

src/cloud-api-adaptor/podvm/Makefile
stevenhorsman, line 1: It would be good to get an explanation of this change in the commit message body - Like the explanation that this image containers the binaries that can turn an existing VM into a BYOM pool VM?

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -0,0 +1,105 @@
...
+        working-directory: src/cloud-api-adaptor/podvm
+        run: |
+          # Build using the podvm-binaries image for the specified architecture
+          docker build \
+            --build-arg "BINARIES_IMG=${REGISTRY}/podvm-binaries-ubuntu-${ARCH}:${IMAGE_TAG}" \
+            -t "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:${IMAGE_TAG}" \
+            -t "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:latest" \
+            -f Dockerfile.podvm_byom_binaries \
+            ..
+
```
stevenhorsman: Could you use the make target added in the previous commit rather than duplicating it?

.github/workflows/podvm_mkosi_ubuntu.yaml
```diff
@@ -156,6 +156,20 @@ jobs:
...
+        working-directory: src/cloud-api-adaptor/podvm
+        run: |
+          make podvm-byom-binaries-image
+          make push-podvm-byom-binaries-image
+        env:
+          ARCH: ${{ inputs.arch }}
+          PODVM_DISTRO: ubuntu
+          REGISTRY: ${{ inputs.registry }}
+          PODVM_TAG: ${{ inputs.image_tag }}
```
stevenhorsman: Is there an opportunity to use the other workflow here as otherwise we aren't running the same code in our CI as our release process?

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -0,0 +1,105 @@
...
+          docker build \
+            --build-arg "BINARIES_IMG=${REGISTRY}/podvm-binaries-ubuntu-${ARCH}:${IMAGE_TAG}" \
+            -t "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:${IMAGE_TAG}" \
+            -t "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:latest" \
+            -f Dockerfile.podvm_byom_binaries \
+            ..
+          
+          # Push both tags
+          docker push "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:${IMAGE_TAG}"
+          docker push "${REGISTRY}/podvm-byom-binaries-ubuntu-${ARCH}:latest"
```
stevenhorsman: Could this run other than on merged PRs - if we runs on PRs then we don't want to push to latest?

stevenhorsman (commented):
Some comments/questions - also proper git commits messages are essential to help understand the explanation for changes.

src/cloud-api-adaptor/podvm/Makefile
Amulyam24: Sure, added it.

.github/workflows/podvm_mkosi_ubuntu.yaml
Amulyam24: @stevenhorsman , If we want to test this in our CI, should we add it to `e2e_run_all.yaml`?

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: @stevenhorsman, I didn't get you here? are you suggesting we push this image on every PR merge?

stevenhorsman: So podvm_mkosi_ubuntu is run on every PR and after every merge, so if we might ever integrate this in, should we gate the `latest` tag. Does that make sense?

.github/workflows/podvm_mkosi_ubuntu.yaml
stevenhorsman: I don't understand the question. I want to understand why you have 

```
  - name: Build and push BYOM binaries image
        working-directory: src/cloud-api-adaptor/podvm
        run: |
          make podvm-byom-binaries-image
          make push-podvm-byom-binaries-image
        env:
          ARCH: ${{ inputs.arch }}
          PODVM_DISTRO: ubuntu
          REGISTRY: ${{ inputs.registry }}
          PODVM_TAG: ${{ inputs.image_tag }}
```
here, when there is also a workflow `podvm_byom_binaries_publish.yaml` that seems to do this exact thing. If we use that workflow here we will test is on each PR and merge, if we don't and stick with your code, then we only will find issues during the release workflow, which is not good practice. 

Does that help clarify?

Amulyam24: There is some confusion here, these changes have been removed from this workflow.

stevenhorsman: Ok, so we have no testing for this image, or workflow?

Amulyam24: Currently no, we can only test if the build works fine.

stevenhorsman: so this is only run at release time and we don't have any testing, so this is an experimental feature still, or is there are plan to address this. I guess I was original expecting the BYOM tests to use this image, rather than creating their own container.

Amulyam24: I agree that this should be tested, I will explore how the e2e tests can consume this image as a follow up, does that make sense?

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: yes, addressed it.

.github/workflows/podvm_mkosi_ubuntu.yaml
stevenhorsman: Sure. Thanks

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Build BYOM binaries image
+        working-directory: src/cloud-api-adaptor/podvm
+        run: make podvm-byom-binaries-image
+        env:
+          REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
+          PODVM_TAG: ${{ inputs.image_tag || github.sha }}
+          ARCH: ${{ inputs.arch }}
+          PODVM_DISTRO: ubuntu
+
+      - name: Push BYOM binaries image
```
stevenhorsman, line 96: This still isn't calling `make push-podvm-byom-binaries-image`?

stevenhorsman (commented):
Thanks for all the updates, just one follow up on a previous review.

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: The make target pushes images with both `latest` and `podvm_tag` tags. Since we would want to avoid pushing the latest tag, I added docker push directly instead of make.

stevenhorsman: I'm still not sure that the makefile should have the `latest` tag hardcoded to always happen in it as it's something people should be able to run in development, but I've made this point multiple times, so it's not worth continuing and blocking

stevenhorsman (approved):

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: Sorry, I was in a different understanding and initially followed the push approach of 2 tags which was already there. I agree to the point and fixed it.

Amulyam24:
Hi @bpradipt @mkulke, can you PTAL?

mkulke (approved):

PR #2933, opened 2026-03-16
podvm.hack: Add serial console output on failure

ldoktor:
The smoke-test has been failing quite often recently, let's add serial console to get at least some information. While on it add fall-back to "qemu" user in case "libvirt-qemu" doesn't exists (or fails).


ldoktor:
> should we maybe log only when the test failed?

Oh I meant to do that but it got lost when I moved it to cleanup. Fixed.

PR #2935, opened 2026-03-17
workflows: align peerpods-chart concurrency group with parent workflow

wainersm:
Update the concurrency group pattern in peerpods-chart_image.yaml to match the parent workflow (publish_images_on_push.yaml) to prevent deadlock errors when called via workflow_call.

Fix commit 40bb0a763cf1c1146adbcfc0d1421973d14dabc9
Assisted-by: Claude

### DIFF
--- a/.github/workflows/peerpods-chart_image.yaml
+++ b/.github/workflows/peerpods-chart_image.yaml
@@ -33,7 +33,7 @@ on:
 permissions: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ inputs.chart_version || github.ref }}-peerpods-chart-image
   cancel-in-progress: true
 
 jobs:


.github/workflows/peerpods-chart_image.yaml
```diff
@@ -33,7 +33,7 @@ on:
 permissions: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
ldoktor: My knowledge about concurrency groups are limited, but my AI and googling says that the `github.event.pull_request.number` is not passed to workflow_calls automatically and suggest we should pass it manually, did you experiment with that?

Also looking at where we call it, it's either on release or via `publish_images_on_push.yaml`. So IMO suitable concurrency group would be `chart_version` as that is the product and when someone triggers new job with the same version we don't need to continue with the previous one, what do you think?

Issue #1534, opened 2023-10-23
RFC: Simpler management of VM instances and PeerPod objects


yoheiueda:
`cloud-api-adaptor` (CAA) and `peerpod-ctrl` have some similar functionality regarding VM instance management, and we can simplify CAA by moving some of its functionality to `peerpod-ctrl`.

## Current design

When CAA creates a cloud VM instance using a cloud API as a peer pod VM, it also creates a `PeerPod` CRD object with a finalizer. A PeerPod object also has a owner reference to the corresponding Pod object. PeerPod objects were introduced to avoid dangling VM instances.

CAA runs as a DaemonSet, so a CAA pod runs on each worker node. CAA manages a mapping between sandbox ID and VM instance ID, and forwards kata agent API calls from kata shim to a corresponding peer pod VM using it.

When a user deletes the Pod object, CAA receives a `StopVM` request from the kata shim. Then, CAA deletes a cloud VM instance using a cloud API, and it also deletes the finalizer of the PeerPod object. The PeerPod object is eventually deleted since it has owner reference to the deleted Pod object.

This mechanism of VM deletion does not work when CAA crashes or is restarted.  `peerpod-ctrl` handles such dangling VM instances. When it detects deleted PeerPod object, it checks whether the PeerPod object has a finalizer or not. If it does not have a finalizer, it probably means that CAA did not delete the VM instance, so  `peerpod-ctrl` deletes the VM instance on behalf of CAA.

We also need [a special care](https://github.com/confidential-containers/cloud-api-adaptor/pull/1517) when CAA restarts. The mapping information between sandbox ID and VM instance ID is stored in the memory of the CAA process, so it will be lost when CAA restarts. Kubelet fails to monitor status of the containers running in the pod when CAA restarts, and then it tries to recreate a sandbox and containers in it. In this case, the Pod object is kept and not recreated. When CAA creates a new peer pod VM, the old peer pod VM is still alive since the old PeerPod object is also still alive. `peerpod-ctrl` monitors PeerPod objects, and delete the stale PeerPod objects as well as old peer pod VM instances. In this current design, peer pod VMs needs to be recreated when CAA restarts.

The following table summarizes which component creates/deletes/monitors which entity. Note that both CAA and peerpod-ctrl manage VM instances.

|                | Created by | Deleted By   | Monitored by | 
|--------------- | ---------- | ------------ |------------- |
| Pod object     | User       |  User        |              | 
[...]

PR #2981, opened 2026-04-07
peerpod-ctrl: Add orphan VM garbage collector

Hru-s:
## Problem

Orphan VM instances are created when the CAA crashes after `CreateInstance()` succeeds but before the corresponding PeerPod CR is created:

```go
instance, err := s.provider.CreateInstance(ctx, sandbox.podName, string(sid), sandbox.cloudConfig, sandbox.spec)
// ← crash here = orphan VM
if s.ppService != nil {
    if ownErr := s.ppService.OwnPeerPod(sandbox.podName, sandbox.podNamespace, instance.ID); ownErr != nil {
        logger.Printf("failed to create PeerPod: %v", ownErr)
    }
}
```

When this happens, no PeerPod object exists to track the VM, so the existing controller-based deletion never fires. The instance remains running indefinitely, consuming cloud resources.

## Solution

Add a periodic garbage collector to `peerpod-ctrl` that:

1. **Tags instances at creation** — the CAA injects the `kube-system` namespace UID as a `caa-cluster-uid` tag into `TAGS` at startup, so every VM is tagged with its cluster identity.
2. **Lists cloud instances** — via a new optional `InstanceLister` interface that providers can implement. Implemented for AWS using `DescribeInstances` filtered by the cluster UID tag. The cluster UID is passed explicitly via a `ListInstancesInput` struct rather than relying on environment variables.
3. **Compares against PeerPod CRs** — any cloud instance with no matching PeerPod CR is considered an orphan and deleted.
4. **Grace period** — instances are only deleted after the GC has observed them as orphan candidates for at least 10 minutes (configurable) across consecutive cycles, using only the controller's local clock (no cross-clock dependency with the cloud provider — safe for disconnected/air-gapped environments). The 10-minute default provides a safety margin for large GPU instances that can take 5+ minutes to boot.

The GC is **enabled by default** via `ENABLE_GC` (defaults to `true`) and `GC_INTERVAL` (default 30m) in the `peer-pods-cm` ConfigMap. All settings are hot-reloadable — no restart needed. Providers that do not implement `InstanceLister` are gracefully skipped — no impact to existing providers.

## Changes

- `cloud-providers/types.go` — Add `InstanceLister` interface, `ListInstancesInput` struct, `ClusterUIDTagKey` constant, and `CreatedAt` field on `Instance` (informational; not used for grace period decisions)
- `cloud-providers/aws/provider.go` — Implement `ListInstances` for AWS, accepting `ListInstancesInput` with explicit cluster UID; populate `CreatedAt` from EC2 `LaunchTime`
- `cloud-providers/aws/list_instances_test.go` — Unit tests for `ListInstances` including filter verification and `CreatedAt` assertion
- `cloud-api-adaptor/cmd/main.go` — Inject cluster UID tag at CAA startup with GC-contextualized log messages
- `cloud-api-adaptor/charts/peerpods/rbac.yaml` — RBAC for CAA to read namespaces
- `peerpod-ctrl/controllers/gc.go` — The garbage collector (package-level logger, lazy init with retry, clean cluster UID resolution, local-clock-only "first seen" grace period tracking, configurable grace period)
- `peerpod-ctrl/controllers/peerpod_controller.go` — Extract shared `loadCloudConfigs` helper to eliminate code duplication between GC and reconciler
- `peerpod-ctrl/main.go` — Register GC with the controller manager, disable cache for Namespace objects
- `peerpod-ctrl/chart/templates/rbac.yaml` — RBAC for peerpod-ctrl to read namespaces
- `peerpod-ctrl/README.md` — Documentation for orphan VM garbage collection

## Configuration

| ConfigMap key | Default | Description |
|---|---|---|
| `ENABLE_GC` | `true` | Set to `false` to disable orphan VM garbage collection |
| `GC_INTERVAL` | `30m` | How often the GC runs (Go duration string, must be > 0) |
| `GC_GRACE_PERIOD` | `10m` | How long an orphan candidate must be observed before deletion (Go duration string, must be > 0) |

**Timing:** An orphan is deleted on the first GC cycle after both (a) it has been discovered and (b) the grace period has elapsed since discovery. With defaults (`GC_INTERVAL=30m`, `GC_GRACE_PERIOD=10m`), deletion occurs on the cycle after first discovery (~30m), since 30m > 10m. Worst-case time from VM creation to deletion is approximately 2x `GC_INTERVAL` (one interval to discover, one to delete).

## Design decisions

- **Enabled by default** — GC runs automatically; operators can disable with `ENABLE_GC=false`
- **Provider opt-out** — Providers that don't implement `InstanceLister` are silently skipped
- **Cluster-scoped tagging** — Using `kube-system` namespace UID ensures multi-cluster environments don't interfere with each other
- **Explicit input** — `ListInstances` receives a `ListInstancesInput` struct with the cluster UID, keeping the interface clean and decoupled from environment variables
- **Clean UID resolution** — The GC resolves the cluster UID directly from the kube-system namespace without manipulating the TAGS env var (TAGS is only used by the CAA's CreateInstance path)
- **Dynamic config** — GC interval, grace period, and enabled state are re-read from the ConfigMap every cycle, no restart needed
- **Lazy initialization with retry** — Provider is only initialized when GC is enabled; transient failures (e.g., ConfigMap not ready) are retried on the next cycle instead of permanently disabling GC
- **Package-level logger** — Uses `ctrl.Log.WithName("gc")` at package level for consistent logging without parameter passing
- **Shared config loading** — `loadCloudConfigs` extracted as a shared helper used by both the PeerPod reconciler and the GC, eliminating code duplication
- **Local-clock grace period** — Uses a "first seen" tracker (local clock only) instead of comparing cloud provider timestamps against the controller clock. This avoids cross-clock dependency and works reliably in disconnected/air-gapped environments where NTP may not be available. Instances must be observed as orphan candidates for the configurable grace period (default 10m) before deletion

## Test plan

- [x] Unit tests for AWS `ListInstances` pass (including filter verification and `CreatedAt` population)
- [x] `go vet`, `golangci-lint`, `go mod tidy` clean
- [x] Deploy with `ENABLE_GC=true`, verify GC logs show periodic runs
- [x] Create orphan VM (tagged with cluster UID, no PeerPod CR), verify GC detects and deletes it after grace period
- [x] Verify GC **skips recently discovered orphan candidates** (< grace period) with `"skipping recently discovered orphan candidate"` log
- [x] Verify GC does **not** delete VMs with matching PeerPod CRs
- [x] Hot-reload `ENABLE_GC=false` -> GC tears down provider and stops
- [x] Hot-reload `ENABLE_GC=true` -> GC reinitializes and resumes
- [x] Hot-reload `GC_INTERVAL` change -> GC picks up new interval
- [x] Configurable `GC_GRACE_PERIOD` — set to `2m` via ConfigMap, verified orphan skipped for 2m then deleted
- [x] Shared `loadCloudConfigs` works for both reconciler and GC initialization

## Cluster test results (AWS EKS, us-east-2)

Tested on a live AWS EKS cluster with `GC_INTERVAL=30s` and `GC_GRACE_PERIOD=2m` for faster cycle iteration.

### Test 1 — Orphan VM deleted after configurable grace period

Created `i-0ca40918c22418615` (`gc-test-orphan-v2`) tagged with cluster UID but **no** PeerPod CR.
`GC_GRACE_PERIOD` set to `2m` via ConfigMap.

```
09:37:38  ListInstances: found 2 instances for cluster UID 35a596a9-...
09:37:38  gc  skipping recently discovered orphan candidate  instanceID=i-0ca40918c22418615  firstSeen=838ns
09:38:08  gc  skipping recently discovered orphan candidate  instanceID=i-0ca40918c22418615  firstSeen=29.9s
09:38:38  gc  skipping recently discovered orphan candidate  instanceID=i-0ca40918c22418615  firstSeen=59.9s
09:39:08  gc  skipping recently discovered orphan candidate  instanceID=i-0ca40918c22418615  firstSeen=1m29s
09:39:38  gc  skipping recently discovered orphan candidate  instanceID=i-0ca40918c22418615  firstSeen=1m59s
09:40:08  gc  deleting orphan instance  instanceID=i-0ca40918c22418615  instanceName=gc-test-orphan-v2
09:40:09  gc  garbage collection cycle complete  orphansDeleted=1  totalCloudInstances=2
```

AWS confirmed: `i-0ca40918c22418615` → **terminated**

### Test 2 — Non-orphan VM NOT deleted

Created `i-001efcfb1e4403967` (`gc-test-nonorphan-v2`) with a matching PeerPod CR (`spec.instanceID: i-001efcfb1e4403967`).

- Instance **never** appeared as an orphan candidate in any GC cycle
- AWS confirmed: `i-001efcfb1e4403967` → **running** (throughout all cycles)

### Test 3 — Configurable grace period (GC_GRACE_PERIOD=2m)

The orphan instance was tracked across consecutive 30s cycles using only the controller's local clock, with the grace period set to `2m` via ConfigMap:

| Cycle | firstSeen | Action |
|---|---|---|
| 09:37:38 | 838ns | skipped |
| 09:38:08 | 29.9s | skipped |
| 09:38:38 | 59.9s | skipped |
| 09:39:08 | 1m29s | skipped |
| 09:39:38 | 1m59s | skipped |
| 09:40:08 | >2m | **deleted** |

No dependency on cloud provider `CreatedAt` timestamps — uses `time.Now()` / `time.Since()` only.

### Test 4 — Hot-reload toggle

```
# Set ENABLE_GC=false in peer-pods-cm (no restart)
09:40:38  gc  GC disabled, tearing down provider

# Set ENABLE_GC=true in peer-pods-cm (no restart)
(GC reinitializes on next cycle)
```

GC tears down cleanly on disable and reinitializes on re-enable — no pod restart required.

### Summary

| Test | Instance | Expected | Actual |
|---|---|---|---|
| Orphan deleted after grace | `i-0ca40918c22418615` | terminated | ✅ terminated |
| Non-orphan not deleted | `i-001efcfb1e4403967` | running | ✅ running |
| Configurable grace (2m) | skip then delete | skip then delete | ✅ skip then delete |
| Hot-reload disable | GC stops | GC stops | ✅ tears down provider |
| Hot-reload re-enable | GC resumes | GC resumes | ✅ reinitializes |

### DIFF
--- a/src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
+++ b/src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
@@ -25,6 +25,10 @@ import (
 
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/podnetwork"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/probe"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 const (
@@ -99,6 +103,25 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		printHelp(os.Stderr)
 		cmd.Exit(1)
 	}
+
+	// Resolve the cluster UID (kube-system namespace UID) so providers can
+	// tag VMs for orphan GC discovery. Each provider applies the tag in its
+	// own CreateInstance using provider.ClusterUID. Without this tag, the GC
+	// cannot discover VMs belonging to this cluster, so failure is fatal.
+	kubeConfig, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load in-cluster config for GC tagging: %w", err)
+	}
+	clientset, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kubernetes client for GC tagging: %w", err)
+	}
+	ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get kube-system namespace UID for GC tagging: %w", err)
+	}
+	provider.ClusterUID = string(ns.UID)
+	fmt.Printf("%s: resolved cluster UID for GC tagging: %s\n", programName, ns.UID)
 
 	var (
 		disableTLS      bool
--- a/src/cloud-api-adaptor/install/charts/peerpods/templates/rbac.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/templates/rbac.yaml
@@ -24,6 +24,9 @@ rules:
 - apiGroups: [""]
   resources: ["serviceaccounts"]
   verbs: ["get", "list"]
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/src/cloud-providers/aws/list_instances_test.go
+++ b/src/cloud-providers/aws/list_instances_test.go
@@ -1,0 +1,208 @@
+// (C) Copyright Confidential Containers Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package aws
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	provider "github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers"
+)
+
+type listInstancesMockEC2Client struct {
+	mockEC2Client
+	instances []types.Instance
+	lastInput *ec2.DescribeInstancesInput
+	callCount int
+}
+
+func (m *listInstancesMockEC2Client) DescribeInstances(ctx context.Context,
+	params *ec2.DescribeInstancesInput,
+	optFns ...func(*ec2.Options)) (*ec2.DescribeInstancesOutput, error) {
+
+	m.lastInput = params
+	m.callCount++
+	return &ec2.DescribeInstancesOutput{
+		Reservations: []types.Reservation{
+			{Instances: m.instances},
+		},
+	}, nil
+}
+
+func TestListInstances(t *testing.T) {
+	tests := []struct {
+		name       string
+		clusterUID string
+		instances  []types.Instance
+		wantCount  int
+		wantErr    bool
+	}{
+		{
+			name:       "empty cluster UID returns error",
+			clusterUID: "",
+			wantErr:    true,
+		},
+		{
+			name:       "no instances found",
+			clusterUID: "test-uid-123",
+			instances:  nil,
+			wantCount:  0,
+		},
+		{
+			name:       "returns matching instances",
+			clusterUID: "test-uid-123",
+			instances: []types.Instance{
+				{
+					InstanceId: aws.String("i-aaa"),
+					Tags: []types.Tag{
+						{Key: aws.String("Name"), Value: aws.String("vm-1")},
+						{Key: aws.String(provider.ClusterUIDTagKey), Value: aws.String("test-uid-123")},
+					},
+				},
+				{
+					InstanceId: aws.String("i-bbb"),
+					Tags: []types.Tag{
+						{Key: aws.String("Name"), Value: aws.String("vm-2")},
+						{Key: aws.String(provider.ClusterUIDTagKey), Value: aws.String("test-uid-123")},
+					},
+				},
+			},
+			wantCount: 2,
+		},
+		{
+			name:       "instance without Name tag",
+			clusterUID: "test-uid-123",
+			instances: []types.Instance{
+				{
+					InstanceId: aws.String("i-ccc"),
+					Tags: []types.Tag{
+						{Key: aws.String(provider.ClusterUIDTagKey), Value: aws.String("test-uid-123")},
+					},
+				},
+			},
+			wantCount: 1,
+		},
+		{
+			name:       "skips instance with nil InstanceId",
+			clusterUID: "test-uid-123",
+			instances: []types.Instance{
+				{InstanceId: nil},
+				{
+					InstanceId: aws.String("i-ddd"),
+					Tags: []types.Tag{
+						{Key: aws.String("Name"), Value: aws.String("vm-valid")},
+					},
+				},
+			},
+			wantCount: 1,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p := &awsProvider{
+				ec2Client:     &listInstancesMockEC2Client{instances: tt.instances},
+				serviceConfig: &Config{},
+			}
+
+			got, err := p.ListInstances(context.Background(), provider.ListInstancesInput{ClusterUID: tt.clusterUID})
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("ListInstances() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.wantErr {
+				return
+			}
+			if len(got) != tt.wantCount {
+				t.Errorf("ListInstances() returned %d instances, want %d", len(got), tt.wantCount)
+			}
+
+			for _, inst := range got {
+				if inst.ID == "" {
+					t.Error("ListInstances() returned instance with empty ID")
+				}
+			}
+		})
+	}
+}
+
+func TestListInstances_VerifyFilters(t *testing.T) {
+	mockClient := &listInstancesMockEC2Client{}
+	p := &awsProvider{
+		ec2Client:     mockClient,
+		serviceConfig: &Config{},
+	}
+
+	_, _ = p.ListInstances(context.Background(), provider.ListInstancesInput{ClusterUID: "uid-filter-test"})
+
+	if mockClient.lastInput == nil {
+		t.Fatal("DescribeInstances was not called")
+	}
+
+	filters := mockClient.lastInput.Filters
+	if len(filters) != 2 {
+		t.Fatalf("expected 2 filters, got %d", len(filters))
+	}
+
+	expectedTagFilter := "tag:" + provider.ClusterUIDTagKey
+	if aws.ToString(filters[0].Name) != expectedTagFilter {
+		t.Errorf("expected filter name %q, got %q", expectedTagFilter, aws.ToString(filters[0].Name))
+	}
+	if len(filters[0].Values) != 1 || filters[0].Values[0] != "uid-filter-test" {
+		t.Errorf("expected filter value [uid-filter-test], got %v", filters[0].Values)
+	}
+
+	if aws.ToString(filters[1].Name) != "instance-state-name" {
+		t.Errorf("expected filter name %q, got %q", "instance-state-name", aws.ToString(filters[1].Name))
+	}
+	expectedStates := map[string]bool{"pending": true, "running": true, "stopping": true, "stopped": true}
+	for _, v := range filters[1].Values {
+		if !expectedStates[v] {
+			t.Errorf("unexpected instance state filter value: %s", v)
+		}
+	}
+}
+
+func TestListInstances_VerifyIDAndName(t *testing.T) {
+	launchTime := time.Date(2026, 5, 19, 10, 0, 0, 0, time.UTC)
+	mockClient := &listInstancesMockEC2Client{
+		instances: []types.Instance{
+			{
+				InstanceId: aws.String("i-test123"),
+				LaunchTime: aws.Time(launchTime),
+				Tags: []types.Tag{
+					{Key: aws.String("Name"), Value: aws.String("my-pod-vm")},
+					{Key: aws.String(provider.ClusterUIDTagKey), Value: aws.String("uid-abc")},
+				},
+			},
+		},
+	}
+
+	p := &awsProvider{
+		ec2Client:     mockClient,
+		serviceConfig: &Config{},
+	}
+
+	instances, err := p.ListInstances(context.Background(), provider.ListInstancesInput{ClusterUID: "uid-abc"})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if len(instances) != 1 {
+		t.Fatalf("expected 1 instance, got %d", len(instances))
+	}
+
+	if instances[0].ID != "i-test123" {
+		t.Errorf("expected ID i-test123, got %s", instances[0].ID)
+	}
+	if instances[0].Name != "my-pod-vm" {
+		t.Errorf("expected Name my-pod-vm, got %s", instances[0].Name)
+	}
+	if !instances[0].CreatedAt.Equal(launchTime) {
+		t.Errorf("expected CreatedAt %v, got %v", launchTime, instances[0].CreatedAt)
+	}
+}
--- a/src/cloud-providers/aws/provider.go
+++ b/src/cloud-providers/aws/provider.go
@@ -225,6 +225,13 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 		})
 	}
 
+	if provider.ClusterUID != "" {
+		instanceTags = append(instanceTags, types.Tag{
+			Key:   aws.String(provider.ClusterUIDTagKey),
+			Value: aws.String(provider.ClusterUID),
+		})
+	}
+
 	// Create TagSpecifications for the instance
 	tagSpecifications := []types.TagSpecification{
 		{
@@ -888,4 +895,59 @@ func (p *awsProvider) getDeviceNameAndSize(imageID string) (string, int32, error
 	logger.Printf("image %s: device name=[%s], size=[%d]", imageID, *deviceName, *deviceSize)
 
 	return *deviceName, *deviceSize, nil
+}
+
+func (p *awsProvider) ListInstances(ctx context.Context, input provider.ListInstancesInput) ([]*provider.Instance, error) {
+	if input.ClusterUID == "" {
+		return nil, fmt.Errorf("cluster UID is required")
+	}
+
+	describeInput := &ec2.DescribeInstancesInput{
+		Filters: []types.Filter{
+			{
+				Name:   aws.String("tag:" + provider.ClusterUIDTagKey),
+				Values: []string{input.ClusterUID},
+			},
+			{
+				Name:   aws.String("instance-state-name"),
+				Values: []string{"pending", "running", "stopping", "stopped"},
+			},
+		},
+	}
+
+	var instances []*provider.Instance
+	for {
+		result, err := p.ec2Client.DescribeInstances(ctx, describeInput)
+		if err != nil {
+			return nil, fmt.Errorf("failed to describe instances: %w", err)
+		}
+
+		for _, reservation := range result.Reservations {
+			for _, inst := range reservation.Instances {
+				if inst.InstanceId == nil {
+					continue
+				}
+				name := ""
+				for _, tag := range inst.Tags {
+					if tag.Key != nil && *tag.Key == "Name" && tag.Value != nil {
+						name = *tag.Value
+						break
+					}
+				}
+				instances = append(instances, &provider.Instance{
+					ID:        *inst.InstanceId,
+					Name:      name,
+					CreatedAt: aws.ToTime(inst.LaunchTime),
+				})
+			}
+		}
+
+		if result.NextToken == nil {
+			break
+		}
+		describeInput.NextToken = result.NextToken
+	}
+
+	logger.Printf("ListInstances: found %d instances for cluster UID %s", len(instances), input.ClusterUID)
+	return instances, nil
 }
--- a/src/cloud-providers/cmd/config-extractor/main.go
+++ b/src/cloud-providers/cmd/config-extractor/main.go
@@ -127,15 +127,20 @@ func parseFile(path string) ([]FlagInfo, error) {
 func parsePackageConstants(dir string, fset *token.FileSet) map[string]string {
 	constants := make(map[string]string)
 
-	pkgs, err := parser.ParseDir(fset, dir, nil, parser.ParseComments)
+	entries, err := os.ReadDir(dir)
 	if err != nil {
 		return constants
 	}
 
-	for _, pkg := range pkgs {
-		for _, file := range pkg.Files {
-			extractConstants(file, constants)
+	for _, entry := range entries {
+		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".go") {
+			continue
 		}
+		file, err := parser.ParseFile(fset, filepath.Join(dir, entry.Name()), nil, parser.ParseComments)
+		if err != nil {
+			continue
+		}
+		extractConstants(file, constants)
 	}
 
 	return constants
--- a/src/peerpod-ctrl/controllers/gc.go
+++ b/src/peerpod-ctrl/controllers/gc.go
@@ -1,0 +1,251 @@
+/*
+Copyright Confidential Containers Contributors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	provider "github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers"
+	confidentialcontainersorgv1alpha1 "github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl/api/v1alpha1"
+)
+
+const defaultGCInterval = 30 * time.Minute
+const defaultGCGracePeriod = 10 * time.Minute
+
+var logger = ctrl.Log.WithName("gc")
+
+type GarbageCollector struct {
+	KubeClient client.Client
+	Namespace  string
+	firstSeen  map[string]time.Time
+}
+
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get
+
+func (gc *GarbageCollector) Start(ctx context.Context) error {
+	if !gc.isGCEnabled(ctx) {
+		logger.Info("GC is disabled via ENABLE_GC=false")
+		return nil
+	}
+
+	var (
+		cloudProvider string
+		clusterUID    string
+		p             provider.Provider
+		lister        provider.InstanceLister
+		initialized   bool
+	)
+
+	interval, prevGracePeriod := gc.readGCDynamicConfig(ctx)
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("stopping garbage collector")
+			return nil
+		case <-ticker.C:
+			newInterval, gracePeriod := gc.readGCDynamicConfig(ctx)
+			if newInterval != interval {
+				logger.Info("GC interval changed", "old", interval, "new", newInterval)
+				interval = newInterval
+				ticker.Reset(interval)
+			}
+			if gracePeriod != prevGracePeriod {
+				logger.Info("GC grace period changed, resetting firstSeen tracking",
+					"old", prevGracePeriod, "new", gracePeriod)
+				gc.firstSeen = nil
+				prevGracePeriod = gracePeriod
+			}
+			if !initialized {
+				var err error
+				cloudProvider, p, lister, clusterUID, err = gc.initialize(ctx)
+				if err != nil {
+					logger.Error(err, "failed to initialize garbage collector, will retry next cycle")
+					continue
+				}
+				if lister == nil {
+					return nil
+				}
+				initialized = true
+				logger.Info("starting garbage collector", "provider", cloudProvider)
+			}
+			gc.collect(ctx, p, lister, cloudProvider, clusterUID, gracePeriod)
+		}
+	}
+}
+
+func (gc *GarbageCollector) initialize(ctx context.Context) (string, provider.Provider, provider.InstanceLister, string, error) {
+	if err := loadCloudConfigs(ctx, gc.KubeClient, gc.Namespace); err != nil {
+		return "", nil, nil, "", fmt.Errorf("failed to load cloud configs: %w", err)
+	}
+	logger.Info("cloud config loaded for GC")
+
+	uid, err := gc.getClusterUID(ctx)
+	if err != nil {
+		return "", nil, nil, "", fmt.Errorf("failed to get cluster UID: %w", err)
+	}
+
+	cloudProvider := os.Getenv("CLOUD_PROVIDER")
+	if cloudProvider == "" {
+		return "", nil, nil, "", fmt.Errorf("CLOUD_PROVIDER is not set in peer-pods-cm")
+	}
+
+	p, err := GetProvider(cloudProvider)
+	if err != nil {
+		return "", nil, nil, "", fmt.Errorf("failed to create provider %s: %w", cloudProvider, err)
+	}
+
+	lister, ok := p.(provider.InstanceLister)
+	if !ok {
+		logger.Info("provider does not implement InstanceLister, GC disabled", "provider", cloudProvider)
+		return cloudProvider, p, nil, uid, nil
+	}
+
+	return cloudProvider, p, lister, uid, nil
+}
+
+func (gc *GarbageCollector) getClusterUID(ctx context.Context) (string, error) {
+	ns := corev1.Namespace{}
+	if err := gc.KubeClient.Get(ctx, types.NamespacedName{Name: "kube-system"}, &ns); err != nil {
+		return "", fmt.Errorf("failed to get kube-system namespace: %w", err)
+	}
+	logger.Info("cluster UID resolved", "uid", string(ns.UID))
+	return string(ns.UID), nil
+}
+
+func parseDurationConfig(data map[string]string, key string, defaultVal time.Duration) time.Duration {
+	v, ok := data[key]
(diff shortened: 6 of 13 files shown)


src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,222 @@
...
+			logger.Info("stopping garbage collector")
+			return nil
+		case <-ticker.C:
+			newInterval, enabled := gc.readGCConfig(ctx, logger)
+			if newInterval != interval {
+				logger.Info("GC interval changed", "old", interval, "new", newInterval)
+				interval = newInterval
+				ticker.Reset(interval)
+			}
+			if !enabled {
```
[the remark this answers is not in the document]
bpradipt: @Hru-s ptal

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,222 @@
...
+	logger := ctrl.Log.WithName("gc")
+
+	cloudProvider, p, lister, err := gc.initialize(ctx, logger)
+	if err != nil {
+		logger.Error(err, "failed to initialize garbage collector")
+		return nil
+	}
+	if lister == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
bpradipt: @Hru-s ptal

src/cloud-providers/aws/provider.go
```diff
@@ -208,7 +208,6 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 		},
 	}
 
-	// Add custom tags (k=v) from serviceConfig.Tags to the instance
```
snir911, line 211: why this comment have been removed?

src/cloud-providers/aws/provider.go
```diff
@@ -730,3 +729,58 @@ func (p *awsProvider) getDeviceNameAndSize(imageID string) (string, int32, error
 
 	return *deviceName, *deviceSize, nil
 }
+
+func (p *awsProvider) ListInstances(ctx context.Context) ([]*provider.Instance, error) {
+	clusterUID := p.serviceConfig.Tags[provider.ClusterUIDTagKey]
```
snir911, line 734: nit: some things aren't defined yet (ClusterUIDTagKey) maybe this commit should be moved forward?

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,24 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		cmd.Exit(1)
 	}
 
+	// Fetch the kube-system namespace UID and append it to the TAGS env var
+	// so it gets parsed into Config.Tags by the existing flag logic.
+	if kubeConfig, err := rest.InClusterConfig(); err == nil {
+		if clientset, err := kubernetes.NewForConfig(kubeConfig); err == nil {
+			if ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err == nil {
+				tag := provider.ClusterUIDTagKey + "=" + string(ns.UID)
+				if existing := os.Getenv("TAGS"); existing != "" {
```
snir911, line 94: The current approach is pragmatic and works well.  have you considered adding the cluster UID tag directly in each provider's CreateInstance() function? it will surly be more intrusive, however, there will be a clearer separation from the user configured tags.

src/cloud-api-adaptor/install/charts/peerpods/templates/rbac.yaml
```diff
@@ -20,6 +20,9 @@ rules:
 - apiGroups: [""]
   resources: ["serviceaccounts"]
   verbs: ["get", "list"]
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get"]
```
snir911, line 25: i wonder if we can avoid this RBAC requirement and use some other unique identifier instead, however, I'd leave it as is for now

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,222 @@
...
+	interval := defaultGCInterval
+	enabled := false
+
+	cm := corev1.ConfigMap{}
+	if err := gc.KubeClient.Get(ctx, types.NamespacedName{Name: ppConfigMap, Namespace: gc.Namespace}, &cm); err != nil {
+		logger.Error(err, "failed to read peer-pods-cm for GC config, using defaults")
+		return interval, enabled
+	}
+
+	if v, ok := cm.Data["ENABLE_GC"]; ok && v == "true" {
```
snir911, line 168: Does it make sense to enable it by default ? for disabling GC_INTERVAL=0 can be used, WDYT?

snir911 (commented):
Thank you for your contribution @Hru-s ! i added couple of comments

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,24 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		cmd.Exit(1)
 	}
 
+	// Fetch the kube-system namespace UID and append it to the TAGS env var
+	// so it gets parsed into Config.Tags by the existing flag logic.
+	if kubeConfig, err := rest.InClusterConfig(); err == nil {
+		if clientset, err := kubernetes.NewForConfig(kubeConfig); err == nil {
+			if ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err == nil {
+				tag := provider.ClusterUIDTagKey + "=" + string(ns.UID)
```
thejasn: I would recommend having some reference to GC in the tag key. Makes the purpose of the tag clearer to users.

src/peerpod-ctrl/controllers/gc.go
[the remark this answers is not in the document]
Hru-s: addressed

src/peerpod-ctrl/controllers/gc.go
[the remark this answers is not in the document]
Hru-s: addressed

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,222 @@
...
+		enabled = true
+	}
+
+	if v, ok := cm.Data["GC_INTERVAL"]; ok && v != "" {
+		if parsed, err := time.ParseDuration(v); err == nil {
+			interval = parsed
+		} else {
+			logger.Error(err, "invalid GC_INTERVAL, using default", "value", v)
+		}
+	}
```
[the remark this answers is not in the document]
Hru-s: addressed

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,24 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
+				if existing := os.Getenv("TAGS"); existing != "" {
+					os.Setenv("TAGS", existing+","+tag)
+				} else {
+					os.Setenv("TAGS", tag)
+				}
+				fmt.Printf("%s: cluster UID (kube-system ns): %s\n", programName, ns.UID)
+			} else {
+				fmt.Printf("%s: warning: failed to get kube-system namespace UID: %v\n", programName, err)
+			}
+		}
```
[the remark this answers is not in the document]
Hru-s: addressed

src/cloud-providers/aws/list_instances_test.go
```diff
@@ -0,0 +1,165 @@
...
+func (m *listInstancesMockEC2Client) DescribeInstances(ctx context.Context,
+	params *ec2.DescribeInstancesInput,
+	optFns ...func(*ec2.Options)) (*ec2.DescribeInstancesOutput, error) {
+
+	return &ec2.DescribeInstancesOutput{
+		Reservations: []types.Reservation{
+			{Instances: m.instances},
+		},
+	}, nil
+}
```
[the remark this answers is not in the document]
Hru-s: addressed

src/cloud-providers/aws/provider.go
Hru-s: addressed

src/cloud-providers/aws/provider.go
Hru-s: addressed

src/peerpod-ctrl/controllers/gc.go
Hru-s: i do agree keeping the gc enabled by default makes sense. i think having a flag to enable or disable adds better clarity.

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
Hru-s: Good point — I did consider this. Went with the env-var approach for a few reasons:

Single point of change — one addition in main.go covers all providers, vs modifying CreateInstance() in each one (AWS, Azure, GCP, IBM Cloud, etc.) which all have different tagging surfaces.
No interface changes — per-provider injection would need a ClusterUID field plumbed into every provider's config or the base Provider interface.
New providers get it for free — any provider added in the future automatically inherits the tag without having to remember to include it.

snir911:
Thanks! pls check out the failing lint & vulns checks also

bpradipt:
@Hru-s the govulncheck is known and you can ignore it. lint errors must be fixed.

src/cloud-providers/aws/provider.go
```diff
@@ -730,3 +730,58 @@ func (p *awsProvider) getDeviceNameAndSize(imageID string) (string, int32, error
 
 	return *deviceName, *deviceSize, nil
 }
+
+func (p *awsProvider) ListInstances(ctx context.Context) ([]*provider.Instance, error) {
```
snir911: Have you considered passing the cluster UID explicitly to ListInstances rather than using the TAGS env var? instead of TAGS serve two purposes

Hru-s: i agree with your idea, passing cluster UID to ListInstances is cleaner than passing it as env var, i have updated the pr

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,245 @@
...
+	if v, ok := cm.Data["ENABLE_GC"]; ok && v == "false" {
+		enabled = false
+	}
+
+	if v, ok := cm.Data["GC_INTERVAL"]; ok && v != "" {
+		if parsed, err := time.ParseDuration(v); err == nil {
+			if parsed > 0 {
+				interval = parsed
+			} else {
+				logger.Info("GC_INTERVAL must be positive, using default", "value", v)
```
thejasn: nit: confusing log if we want to call out a miss in config. Maybe call out as a warning and you could log the default value that is being used.

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	cloudProvider := os.Getenv("CLOUD_PROVIDER")
+	if cloudProvider == "" {
+		return "", nil, nil, "", fmt.Errorf("CLOUD_PROVIDER is not set in peer-pods-cm")
+	}
+
+	p, err := GetProvider(cloudProvider)
+	if err != nil {
+		return "", nil, nil, "", fmt.Errorf("failed to create provider %s: %w", cloudProvider, err)
```
thejasn, line 119:
```suggestion
		return "", nil, nil, "", fmt.Errorf("failed to get provider %s: %w", cloudProvider, err)
```

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,245 @@
...
+				}
+				initialized = true
+				logger.Info("starting garbage collector", "provider", cloudProvider)
+			}
+			gc.collect(ctx, logger, p, lister, cloudProvider, clusterUID)
+		}
+	}
+}
+
+func (gc *GarbageCollector) initialize(ctx context.Context, logger logr.Logger) (string, provider.Provider, provider.InstanceLister, string, error) {
```
thejasn: Ideally we want context logging or package level logger, no need to pass loggers across all funcs.

thejasn (commented):
Seeing how the bulk of the operations are duplicated in this controller and the peerpods controller (fetch cm/secret, list peerpod, cleaning of VMs etc). I think the central logic of cleanup should still be added on that controller. 

Since you've implemented a `Runnable()` here anyways, I'm wondering if this could just be a periodic controller that triggers reconcile events for the peer pod controller to do the clean up. This avoids splitting of responsibility across 2 different controllers.

src/peerpod-ctrl/controllers/gc.go
thejasn: actually ignore this, since internally it does a `create`

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,28 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
+			if ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err == nil {
+				tag := provider.ClusterUIDTagKey + "=" + string(ns.UID)
+				if existing := os.Getenv("TAGS"); existing != "" {
+					os.Setenv("TAGS", existing+","+tag)
+				} else {
+					os.Setenv("TAGS", tag)
+				}
+				fmt.Printf("%s: cluster UID (kube-system ns): %s\n", programName, ns.UID)
+			} else {
+				fmt.Printf("%s: warning: failed to get kube-system namespace UID: %v\n", programName, err)
```
snir911: Also here, i'd add some prefix to the warnings to give some context of where these warning comes from, like "Failed to tag VM with UID" or something similar..

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,28 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
...
+	if kubeConfig, err := rest.InClusterConfig(); err == nil {
+		if clientset, err := kubernetes.NewForConfig(kubeConfig); err == nil {
+			if ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err == nil {
+				tag := provider.ClusterUIDTagKey + "=" + string(ns.UID)
+				if existing := os.Getenv("TAGS"); existing != "" {
+					os.Setenv("TAGS", existing+","+tag)
+				} else {
+					os.Setenv("TAGS", tag)
+				}
+				fmt.Printf("%s: cluster UID (kube-system ns): %s\n", programName, ns.UID)
```
snir911: I think it worth adding some context like : "VMs TAGS will be updated with UID for GC" or something like that

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,244 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	provider "github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers"
+	confidentialcontainersorgv1alpha1 "github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl/api/v1alpha1"
+)
+
+const defaultGCInterval = 12 * time.Hour
```
snir911: Could be even shorter IMHO, for some instance types 12h would cost a fortune :D

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,244 @@
...
+		return interval, enabled
+	}
+
+	if v, ok := cm.Data["ENABLE_GC"]; ok && v == "false" {
+		enabled = false
+	}
+
+	if v, ok := cm.Data["GC_INTERVAL"]; ok && v != "" {
+		if parsed, err := time.ParseDuration(v); err == nil {
+			if parsed > 0 {
```
snir911: Consider checking for minimal value to avoid extensive calls

snir911 (commented):
Thanks LGTM overall, pls also update the documentation here https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/peerpod-ctrl/README.md

bpradipt:
Let's get some additional feedback from @stevenhorsman or @mkulke if they can spare some review cycles

mkulke:
> Let's get some additional feedback from @stevenhorsman or @mkulke if they can spare some review cycles

I haven't looked deeply yet, but to me this looks like another band aid for a structural problem in CAA that adds a considerable chunk of code and spreads management (ownership) for cloud resources further across different components.

> Orphan VM instances are created when the CAA crashes after CreateInstance() succeeds but before the corresponding PeerPod CR is created:

IMO we should address this problem. This should never happen, I guess the CAA daemonset should only create/delete CRs while peerpod-ctrl should create, delete, garbage-collect the backing resources, taking over the resource ownership (and privileges) from the daemonset.

bpradipt:
> > Let's get some additional feedback from @stevenhorsman or @mkulke if they can spare some review cycles
> 
> I haven't looked deeply yet, but to me this looks like another band aid for a structural problem in CAA that adds a considerable chunk of code and spreads management (ownership) for cloud resources further across different components.
> 
> > Orphan VM instances are created when the CAA crashes after CreateInstance() succeeds but before the corresponding PeerPod CR is created:
> 
> IMO we should address this problem. This should never happen, I guess the CAA daemonset should only create/delete CRs while peerpod-ctrl should create, delete, garbage-collect the backing resources, taking over the resource ownership (and privileges) from the daemonset.

Imho we have to ideally plan for a redesign spread across multiple releases. Probably piggybacking on byom provider to claim from a pool of VMs. The pool creation and maintenance being offloaded to specific cloud providers. Just a thought.

snir911:
I agree and AFAICT, the redesign and resource management work tracked in issue #1534 is still being explored by @Hru-s.
The intention there is to address a dangling resources issue that still occurs despite peerpod-ctrl

Hru-s:
Thanks for the feedback, @mkulke! I completely agree that the broader redesign being explored in #1534 is the right long-term direction. My hope with this PR is just to provide a safety net for orphaned VMs left behind by crashes or partial failures—something that might still be useful even when peerpod-ctrl owns the full lifecycle. I tried to keep this GC narrowly scoped (~200 lines and opt-in) so it doesn't disrupt current boundaries and can easily adapt to whatever architecture we land on. Let me know if you think this makes sense as a complementary step!

src/cloud-providers/cmd/config-extractor/main.go
```diff
@@ -127,15 +127,20 @@ func parseFile(path string) ([]FlagInfo, error) {
 func parsePackageConstants(dir string, fset *token.FileSet) map[string]string {
 	constants := make(map[string]string)
 
-	pkgs, err := parser.ParseDir(fset, dir, nil, parser.ParseComments)
+	entries, err := os.ReadDir(dir)
```
thejasn, line 130: This seems to be a unrelated change, please move to separate commit.

thejasn (commented):
> Seeing how the bulk of the operations are duplicated in this controller and the peerpods controller (fetch cm/secret, list peerpod, cleaning of VMs etc). I think the central logic of cleanup should still be added on that controller.
> 
> Since you've implemented a Runnable() here anyways, I'm wondering if this could just be a periodic controller that triggers reconcile events for the peer pod controller to do the clean up. This avoids splitting of responsibility across 2 different controllers.

Wondering if you gave this a thought?

Hru-s:
> > Seeing how the bulk of the operations are duplicated in this controller and the peerpods controller (fetch cm/secret, list peerpod, cleaning of VMs etc). I think the central logic of cleanup should still be added on that controller.
> > Since you've implemented a Runnable() here anyways, I'm wondering if this could just be a periodic controller that triggers reconcile events for the peer pod controller to do the clean up. This avoids splitting of responsibility across 2 different controllers.
> 
> Wondering if you gave this a thought?

I definitely considered this @thejasn, but decided to keep the GC separate to keep the reconciler clean and focused.

The Reconciler expects a CR: Triggering it for orphan VMs (via fake events or temporary CRs) requires messy workarounds, since it expects a real Custom Resource with proper owner references and deletion timestamps.

Unnecessary complexity: The discovery logic is exactly the same either way. Forcing the final delete through the reconciler just adds extra steps for no real benefit.

I've extracted loadCloudConfigs into a shared helper to eliminate the duplication, and GetProvider was already a shared package-level function that both components use.

what do you think about this?

src/cloud-providers/aws/provider.go
```diff
@@ -730,3 +730,58 @@ func (p *awsProvider) getDeviceNameAndSize(imageID string) (string, int32, error
...
+				for _, tag := range inst.Tags {
+					if tag.Key != nil && *tag.Key == "Name" && tag.Value != nil {
+						name = *tag.Value
+						break
+					}
+				}
+				instances = append(instances, &provider.Instance{
+					ID:        *inst.InstanceId,
+					Name:      name,
+					CreatedAt: aws.ToTime(inst.LaunchTime),
```
snir911, line 940: Wouldn't comparing the cloud provider's clock vs the controller clock be risky?

Hru-s: cloud APIs and Go's time.Since() both use UTC, so we don't have to worry about timezones.

cloud-hosted clusters (EKS, AKS, GKE, etc.) sync to the same NTP infrastructure as the cloud APIs, keeping skew well under 1 second.
The only real risk is NTP clock drift. With our 5-minute grace period, a standard drift of a few seconds is completely fine. The only way this breaks is if a cluster's clock is off by several minutes—but at that point, core Kubernetes features (like certs and leases) would already be failing anyway!

I have updated the doc mentioning this NTP assumption

snir911:
Thanks for this! I have a preference for the configurable grace period approach, a few points on the implementation:   
1. A suspected orphan is deleted gracePeriod + GC_INTERVAL after first discovery (2h 5m by default), did i understand this right? if yes, i'd try to think if this can be avoided or at least documenting this clearly (and reduce default interval to much less..)
2. Consider increasing the grace period to 10 minutes. Large GPU instances can take ~5 minutes to boot, leaving minimal safety margin with the current 5-minute default

Hru-s:
> Thanks for this! I have a preference for the configurable grace period approach, a few points on the implementation:
> 
> 1. A suspected orphan is deleted gracePeriod + GC_INTERVAL after first discovery (2h 5m by default), did i understand this right? if yes, i'd try to think if this can be avoided or at least documenting this clearly (and reduce default interval to much less..)
> 2. Consider increasing the grace period to 10 minutes. Large GPU instances can take ~5 minutes to boot, leaving minimal safety margin with the current 5-minute default

The default GC_INTERVAL is now 30 minutes. Because deletion delay is simply the longer of the two values (grace period vs. interval), they don't stack. This means the worst-case time an orphan lingers drops from ~4 hours to ~1 hour, all while safely respecting the grace period.

i have also made grace period configurable through configMap

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (gc *GarbageCollector) collect(ctx context.Context, p provider.Provider, lister provider.InstanceLister, cloudProvider, clusterUID string, gracePeriod time.Duration) {
+	cloudInstances, err := lister.ListInstances(ctx, provider.ListInstancesInput{ClusterUID: clusterUID})
+	if err != nil {
+		logger.Error(err, "failed to list cloud instances, skipping GC cycle")
+		return
+	}
+
+	if len(cloudInstances) == 0 {
```
thejasn, line 191: If VMs are manually cleaned and won't the early exit lead to having stale ids in `firstSeen`?

snir911:
@Hru-s LGTM, thanks, pls remove the merge commit

src/peerpod-ctrl/controllers/peerpod_controller.go
```diff
@@ -136,29 +136,32 @@ func (r *PeerPodReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		for k, v := range cm.Data {
 			os.Setenv(k, v)
 		}
 	}
 
 	var secretErr error
-	if secretErr = r.Get(context.TODO(), types.NamespacedName{Name: ppSecret, Namespace: ns}, &peerpodssecret); secretErr == nil {
-		for k, v := range peerpodssecret.Data {
+	if secretErr = k.Get(ctx, types.NamespacedName{Name: ppSecret, Namespace: namespace}, &secret); secretErr == nil {
+		for k, v := range secret.Data {
```
thejasn: `k` seems to be shadowed here. The linter should have flagged this though

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
```diff
@@ -81,6 +85,28 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		cmd.Exit(1)
 	}
 
+	// Fetch the kube-system namespace UID and append it to the TAGS env var
+	// so it gets parsed into Config.Tags by the existing flag logic.
+	if kubeConfig, err := rest.InClusterConfig(); err == nil {
+		if clientset, err := kubernetes.NewForConfig(kubeConfig); err == nil {
+			if ns, err := clientset.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{}); err == nil {
+				tag := provider.ClusterUIDTagKey + "=" + string(ns.UID)
+				if existing := os.Getenv("TAGS"); existing != "" {
```
thejasn: I think we should reconsider using env variables to pass configurations silently across various providers. Yes it does make it easier now but re-using this env variable for this purpose doesn't seem right.

As you brought up, each provider parses and allows tags in different ways and we should not be adding a value to it outside of the providers. The `TAGS` env should still be owned by each provider.

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,247 @@
...
+		if parsed, err := time.ParseDuration(v); err == nil {
+			if parsed > 0 {
+				gracePeriod = parsed
+			} else {
+				logger.Info("WARNING: GC_GRACE_PERIOD must be positive, falling back to default", "value", v, "default", defaultGCGracePeriod)
+			}
+		} else {
+			logger.Error(err, "invalid GC_GRACE_PERIOD, using default", "value", v)
+		}
+	}
```
thejasn: Very similar blocks of logic, please simplify it to make it more readable.
```
parseDuration(data, "GC_INTERVAL", defaultGCInterval)
```

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,247 @@
...
+				if err != nil {
+					logger.Error(err, "failed to initialize garbage collector, will retry next cycle")
+					continue
+				}
+				if lister == nil {
+					return nil
+				}
+				initialized = true
+				logger.Info("starting garbage collector", "provider", cloudProvider)
+			}
```
thejasn: I'm wondering why do we need `ENABLE_GC` to be a toggle? I'm not yet convinced making it hot-reloadable.

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,247 @@
...
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("stopping garbage collector")
+			return nil
+		case <-ticker.C:
+			newInterval, gracePeriod, enabled := gc.readGCConfig(ctx)
```
thejasn: Making `gracePeriod` also hot-reloadable could have unintended effect if the user reduces the grace period say from 10m to 2m. Since `firstSeen` is processed against this, we could be deleting VMs that were first observed with 10m grace period. 

For this config update, it might be better to have restart if need so the all firstSeen vms can also be reset.

src/peerpod-ctrl/controllers/gc.go
```diff
@@ -0,0 +1,247 @@
...
+				initialized = true
+				logger.Info("starting garbage collector", "provider", cloudProvider)
+			}
+			gc.collect(ctx, p, lister, cloudProvider, clusterUID, gracePeriod)
+		}
+	}
+}
+
+func (gc *GarbageCollector) initialize(ctx context.Context) (string, provider.Provider, provider.InstanceLister, string, error) {
+	if err := loadCloudConfigs(ctx, gc.KubeClient, gc.Namespace); err != nil {
```
thejasn, line 102: I think this can be simplified by just building the provider once in main, right? Its not like new provider or provider changes can happen on the fly. 

@snir911 Thoughts on this? Not sure if the same issue with `peerpod-ctrl` would be faced here where provider is not present. I'm just trying to see if can avoid using the same pattern that is used across providers where we are using environmental variables as a data bus implicitly across components.

snir911: in some cases (OSC) the controller is registered as standalone controller so it would work IIRC, however, maybe having it in a separate utility file would work

thejasn: If this could work, then it would simplify a lot for this controller. But I saw a few issues where Provider/CM was not available at the start up of CAA (not sure if the issue is still pertinent).

```
// main.go

...
// readCM()
// build provider()
...
...
...
// inject provider in GarbageCollector() controller
```

src/peerpod-ctrl/controllers/gc.go
Hru-s: Fixed. When zero cloud instances are returned, firstSeen is now cleared before the early return.

src/peerpod-ctrl/controllers/peerpod_controller.go
Hru-s: renamed the parameter from k to reader to eliminate the shadowing.

src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
Hru-s: The TAGS env var injection has been removed. The cluster UID is now set via provider.ClusterUID (a package-level variable), and each provider reads it directly in its own CreateInstance. This keeps tag ownership within each provider.

src/peerpod-ctrl/controllers/gc.go
Hru-s: Done -- extracted parseDurationConfig(data, key, defaultVal) helper. Both GC_INTERVAL and GC_GRACE_PERIOD now use it.

src/peerpod-ctrl/controllers/gc.go
Hru-s: Agreed -- ENABLE_GC is now read once at startup. If disabled, the GC exits immediately without entering the loop. To re-enable, operators restart the pod. This removes the teardown/re-init state machine and simplifies Start(). GC_INTERVAL and GC_GRACE_PERIOD remain hot-reloadable since they don't involve provider lifecycle changes.

src/peerpod-ctrl/controllers/gc.go
Hru-s: Good point. When GC_GRACE_PERIOD changes at runtime, firstSeen is now cleared entirely so all orphan candidates must be re-observed for the full new grace period before deletion. This prevents the scenario where a reduced grace period causes premature deletion of instances tracked under the old (longer) grace. Documented this behavior in the README as well.

src/peerpod-ctrl/controllers/gc.go
Hru-s: I've extracted loadCloudConfigs and GetProvider into a dedicated cloud_util.go file so they're clearly shared infrastructure rather than being embedded in the reconciler. Both the PeerPod reconciler and the GC now import from the same utility.

I agree the env-var-based config loading pattern isn't ideal. Happy to explore that as a follow-up to keep this PR focused.

PR #3007, opened 2026-04-21
podvm: Run iptables-wrapper directly

BbolroC:
iptables-wrapper-installer.sh was removed in
https://github.com/kubernetes-sigs/iptables-wrappers/pull/14, so call the binary directly.

Assisted-by: IBM Bob

### DIFF
--- a/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
+++ b/src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
@@ -4,10 +4,14 @@ ARG BASE_IMAGE=registry.fedoraproject.org/fedora:43
 
 FROM $BASE_IMAGE AS iptables
 
+COPY ./versions.yaml /versions.yaml
+
 RUN echo "Building iptables-wrapper... " \
     && dnf install -y git golang \
     && git clone https://github.com/kubernetes-sigs/iptables-wrappers.git /iptables \
     && cd /iptables \
+    && version=$(grep 'iptables-wrapper:' /versions.yaml | awk '{print $2}') \
+    && git checkout $version \
     && make BIN_DIR=. build
 
 FROM $BASE_IMAGE AS base
@@ -32,18 +36,19 @@ RUN echo "Enabling / Disabling services ... " \
     && systemctl enable systemd-logind dbus.socket
 
 RUN --mount=type=cache,target=/iptables,from=iptables,source=/iptables,readonly \
-    cd /iptables && ./iptables-wrapper-installer.sh --no-sanity-check --no-cleanup
+    cp /iptables/iptables-wrapper /usr/sbin \
+    && /usr/sbin/iptables-wrapper install
 
 # Add podvm resources
-COPY ./resources/binaries-tree/etc/ /etc/
-COPY ./resources/binaries-tree/usr/ /usr/
-COPY ./resources/binaries-tree/pause_bundle/ /pause_bundle/
+COPY ./podvm-mkosi/resources/binaries-tree/etc/ /etc/
+COPY ./podvm-mkosi/resources/binaries-tree/usr/ /usr/
+COPY ./podvm-mkosi/resources/binaries-tree/pause_bundle/ /pause_bundle/
 
 RUN curl -LO https://raw.githubusercontent.com/confidential-containers/cloud-api-adaptor/main/src/cloud-api-adaptor/podvm/qcow2/misc-settings.sh
 
 RUN PODVM_DISTRO=ubuntu CLOUD_PROVIDER=generic bash ./misc-settings.sh
 
-COPY --chmod=0755 entrypoint.sh  /usr/local/bin/
+COPY --chmod=0755 ./podvm-mkosi/entrypoint.sh  /usr/local/bin/
 
 # https://systemd.io/CONTAINER_INTERFACE/
 ENV container=docker
--- a/src/cloud-api-adaptor/versions.yaml
+++ b/src/cloud-api-adaptor/versions.yaml
@@ -29,7 +29,7 @@ cloudimg:
 tools:
   cert-manager: v1.15.3
   bats: 1.10.0
-  iptables-wrapper: v0.0.0-20240819165702-06cad2ec6cb5
+  iptables-wrapper: bfef9e5087a198b50a4124bb9ce9d2c7c99025dd
   golang: 1.25.9
   kcli: 99.0.202507200957
   mkosi: v22
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -12,6 +12,7 @@ ARG BASE=registry.fedoraproject.org/fedora@sha256:893f7eeffce8e3e2bb2bc62786ba56
 FROM --platform=$BUILDPLATFORM $BUILDER_BASE AS builder-release
 ARG YQ_VERSION
 RUN go install github.com/mikefarah/yq/v4@$YQ_VERSION
+RUN dnf install -y git && dnf clean all
 
 # For `dev` builds due to CGO constraints we have to emulate the target platform
 # instead of using Go's cross-compilation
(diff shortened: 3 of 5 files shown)


src/cloud-api-adaptor/versions.yaml
```diff
@@ -29,7 +29,7 @@ cloudimg:
 tools:
   cert-manager: v1.15.3
   bats: 1.10.0
-  iptables-wrapper: v0.0.0-20240819165702-06cad2ec6cb5
+  iptables-wrapper: v0.0.0-20260420133543-bfef9e5087a1
```
stevenhorsman: Should us pinning the image have prevented us for being impacted by this, or are we missing this somewhere?

stevenhorsman: Okay - I dug into this. it seems to be because  `v0.0.0-20240819165702-06cad2ec6cb5` is a Go pseudo-version, not a Git tag. So when we curl from the url `https://raw.githubusercontent.com/kubernetes-sigs/iptables-wrappers/v0.0.0-20240819165702-06cad2ec6cb5/...` is doesn't return stable-ly and probably just gives us back the `main` branch version.

I think you have effectively fixed this as the external curl has gone and we are just using the go binary, which should be stable based on the go pseudo-version

stevenhorsman (approved):
Thanks for the fix!

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
```diff
@@ -0,0 +1,85 @@
...
+        echo "Error: No supported package manager found (dnf or apt-get)"
+        exit 1
+    fi
+else
+    echo "All required dependencies are already installed."
+fi
+
+# Clone the repository
+echo "Cloning iptables-wrappers repository..."
+git clone https://github.com/kubernetes-sigs/iptables-wrappers.git "${OUTPUT_DIR}"
```
stevenhorsman: Could we do a `--depth 1` shallow clone as we are later doing a checkout as it will be quicker and we don't need the other code?

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
```diff
@@ -0,0 +1,85 @@
...
+git checkout "${VERSION}"
+
+# Build the binary
+echo "Building iptables-wrapper..."
+if [[ -n "${GOARCH}" ]]; then
+    GOARCH="${GOARCH}" make BIN_DIR=. build
+else
+    make BIN_DIR=. build
+fi
+
```
stevenhorsman, line 105: Can we do a clean up of the git repo to reduce the layer size once the build it done?

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
BbolroC: Yep, I will update the PR.

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
BbolroC: Good catch! I will introduce the trap function. Thanks!

stevenhorsman (approved):

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
```diff
@@ -32,18 +31,19 @@ RUN echo "Enabling / Disabling services ... " \
...
 RUN --mount=type=cache,target=/iptables,from=iptables,source=/iptables,readonly \
-    cd /iptables && ./iptables-wrapper-installer.sh --no-sanity-check --no-cleanup
+    cp /iptables/iptables-wrapper /usr/sbin \
+    && /usr/sbin/iptables-wrapper install
 
 # Add podvm resources
-COPY ./resources/binaries-tree/etc/ /etc/
-COPY ./resources/binaries-tree/usr/ /usr/
-COPY ./resources/binaries-tree/pause_bundle/ /pause_bundle/
+COPY ./podvm-mkosi/resources/binaries-tree/etc/ /etc/
```
mkulke, line 43: why are those paths changed?

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
```diff
@@ -0,0 +1,106 @@
...
+    if [[ ${#INSTALLED_PACKAGES[@]} -gt 0 ]]; then
+        echo "Removing installed packages: ${INSTALLED_PACKAGES[*]}"
+        if command -v dnf &> /dev/null; then
+            dnf remove -y "${INSTALLED_PACKAGES[@]}" || true
+            dnf clean all
+        elif command -v apt-get &> /dev/null; then
+            apt-get purge -y "${INSTALLED_PACKAGES[@]}" || true
+            apt-get autoremove -y
+            apt-get clean
+            rm -rf /var/lib/apt/lists/*
```
mkulke: While I understand we want to do this to keep the daemonset image slim, I wouldn't want a script to do stuff like this on my development machine (I'd have to invoke it with sudo).

We could add a stage to the daemonset dockerfile that just builds the wrapper (and consume it via `COPY --from=iptables-wrapper-builder ...`) we wouldn't have to care about the leftovers and an elaborate build script

mkulke (commented):
other go tools like umoci we build as part of the binaries and then we copy the artifacts to the required place, why aren't we following the same approach

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
BbolroC: Because a location where `docker build` is called has been changed to a parent directory to access `versions.yaml` and `build-iptables-wrapper.sh`. Yes, we can copy these files via URL if we don't want to change the path. 😉

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
BbolroC: Good idea! I will update the PR. Thanks for the feedback! 👍

BbolroC:
> other go tools like umoci we build as part of the binaries and then we copy the artifacts to the required place, why aren't we following the same approach

@stevenhorsman I need your input for this. Thanks a lot!

stevenhorsman:
> > other go tools like umoci we build as part of the binaries and then we copy the artifacts to the required place, why aren't we following the same approach
> 
> @stevenhorsman I need your input for this. Thanks a lot!

I think we use iptables in both the docker provider image and the CAA image, so it's a bit different from the Makefile.inc binaries that are just for the podvm image as the CAA image doesn't use the podvm_binaries in it's build and I don't think we want to mix them?

Using a separate stage in CAA image build for this code sounds good and means my comment about clean up and Magnus' objections wouldn't be required.

It's worth noting that the docker provider might be heading towards the end of life soon, in case that changes the equation here as I think we should focus on getting the CAA image build as clean/ideal as we can and if the docker provider is less ideal then it's less important.

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
mkulke: ah, I am asking because the /podvm/resources folder has been generic for both mkosi and packer images. so, the packer images wouldn't have the fix?

mkulke:
> > > other go tools like umoci we build as part of the binaries and then we copy the artifacts to the required place, why aren't we following the same approach
> > 
> > 
> > @stevenhorsman I need your input for this. Thanks a lot!
> 
> I think we use iptables in both the docker provider image and the CAA image, so it's a bit different from the Makefile.inc binaries that are just for the podvm image as the CAA image doesn't use the podvm_binaries in it's build and I don't think we want to mix them?
> 
> Using a separate stage in CAA image build for this code sounds good and means my comment about clean up and Magnus' objections wouldn't be required.
> 
> It's worth noting that the docker provider might be heading towards the end of life soon, in case that changes the equation here as I think we should focus on getting the CAA image build as clean/ideal as we can and if the docker provider is less ideal then it's less important.

so, the "podvm" image for the docker provider needs it and it's also required on the daemonset image. maybe we can live it with a bit of duplication (2x the same stage) rather than introducing a new builder script as an abstraction for both to use?

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
BbolroC: Oh, I wasn't aware of it. Let me check. Thanks for the info. 👍

BbolroC: Based on my quick search, the packer-based VM images 
- podvm/qcow2
- aws/image
- azure/image

do not use iptables-wrapper. If this is not what you are looking for, I would be grateful if you could point exactly (and concretely) where your concern goes. Thanks!

BbolroC:
> so, the "podvm" image for the docker provider needs it and it's also required on the daemonset image. maybe we can live it with a bit of duplication (2x the same stage) rather than introducing a new builder script as an abstraction for both to use?

Do you mean by going with only 1st commit?

src/cloud-api-adaptor/podvm-mkosi/Dockerfile.podvm_docker_provider
stevenhorsman: Yeah - as I mentioned iptables is not a podvm image things really, it just is used by the docker variant of the podvm

src/cloud-api-adaptor/hack/build-iptables-wrapper.sh
BbolroC: I've confirmed all Dockerfiles use a separate stage for the binary build and use it via the cache mount. Let me remove the cleanup function from the script. Thanks!

mkulke (approved):
thx for accomodating!

PR #3009, opened 2026-04-21
libvirt: make podvm deletion robust to disk XML variations

### LINKED ISSUE #3008: libvirt: make PodVM deletion robust to disk XML variations
`DeleteDomain()` currently assumes a fixed disk layout during cleanup by indexing the first two disks from the libvirt domain XML.

Today the normal libvirt PodVM flow creates two disks, but the delete path is still brittle because it depends on that exact XML shape. Cleanup can become fragile if a disk source is missing, if the XML differs across cases, or if teardown is retried after a partial cleanup.

This small hardening change would improve reliability now, and would also make the code safer if libvirt gains support for additional attached disks in the future.

Suggested scope:

- avoid assuming exactly two disks during cleanup
- safely handle missing or non-file-backed disk sources
- make repeated cleanup attempts more robust where practical

ajaypvictor:
DeleteDomain currently assumes the domain XML always contains exactly two file-backed disks and indexes them directly during cleanup.

This is brittle when the XML is incomplete or the disk layout differs from the expected boot and cloud-init pair, and it can cause cleanup to fail before all removable volumes are processed.

Iterate over the file-backed disks discovered in the domain XML instead of relying on fixed indexes, and collect cleanup errors while still attempting to undefine the domain. Add a focused unit test for disk path extraction edge cases.

Fixes: #3008

### DIFF
--- a/src/cloud-providers/libvirt/libvirt.go
+++ b/src/cloud-providers/libvirt/libvirt.go
@@ -8,8 +8,10 @@ package libvirt
 import (
 	"context"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"net/netip"
+	"strings"
 	"time"
 
 	retry "github.com/avast/retry-go/v4"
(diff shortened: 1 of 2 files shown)


stevenhorsman:
Hi @ajaypvictor - can you rebase to pick up the podvm build fix please

stevenhorsman (approved):
LGTM as long as the CI passes. Thanks @ajaypvictor!

BbolroC (approved):
Looks reasonable to me. Thanks @ajaypvictor !

PR #3012, opened 2026-04-23
workflows: correctly publish mkosi images

wainersm:
The non-debug images have been published as debug by mistake. inputs.debug is a boolean type and it should be handled as boolean.

### DIFF
--- a/.github/workflows/podvm_mkosi.yaml
+++ b/.github/workflows/podvm_mkosi.yaml
@@ -149,12 +149,12 @@ jobs:
           VERIFY_PROVENANCE: yes
 
       - name: Build mkosi debug image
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):

mkulke (approved):

wainersm:
Not just publishing correctly but also build the debug flavor (was non-debug) on CI: https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/24843247928

stevenhorsman:
@wainersm - The s390x runners are having performance problems (which have been reported), and you fork, which is a better test as it includes the workflow changes) so any reasons to not merge this now?

PR #3013, opened 2026-04-24
Kata 3.29.0 bump

stevenhorsman:
Bump our versions of guest-components and runtime to match the kata 3.29.0 release
a) so that we have an earlier check that we are still compatible
b) to pick up part of the [GHSA-q49m-57vm-c8cc](https://github.com/kata-containers/kata-containers/security/advisories/GHSA-q49m-57vm-c8cc) fix
c) In case we need a quick release due to b


stevenhorsman:
The docker e2e seems to be failing on multiple e2e tests with `  Error from server (BadRequest): container "cloud-api-adaptor-con" in pod "cloud-api-adaptor-daemonset-68q56" is waiting to start: trying and failing to pull image`, and it seems to be on the way out, so I think we just ignore it for now as it's still passing in nightlies?

bpradipt:
> The docker e2e seems to be failing on multiple e2e tests with ` Error from server (BadRequest): container "cloud-api-adaptor-con" in pod "cloud-api-adaptor-daemonset-68q56" is waiting to start: trying and failing to pull image`, and it seems to be on the way out, so I think we just ignore it for now as it's still passing in nightlies?

Yeah lets ignore for now. I'll start taking a look in parallel.

fidencio (approved):
lgtm, thanks @stevenhorsman!

fidencio:
cc @mkulke

bpradipt (approved):

PR #3020, opened 2026-04-26
podvm: fix IMDS timeout counter

HernanGatta:
The script runs with `set -e`, which causes the script to exit on first error. The line `((counter++))` causes Bash to evaluate the expression, which on first iteration returns `0`, which is regarded as an error, causing the script to exit with `1` prematurely.

Use the prefix `++` operator instead.

**Note:** The failing [lint / govulncheck (pull_request)](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/24968315133/job/73125344842?pr=3020) check is a red herring, at least as far as the change in this PR goes.

---

I've observed the following behavior on my Azure PodVMs with the Fedora (debug) image on CoCo v0.19.0:

image

Notice how the script exits one second after starting, indicating that it did not wait.

I did also try this:

```bash
#!/bin/bash
set -euo pipefail
counter=0
((counter++))
```

```
❯ ./test.sh

❯ echo $?
1
```

Whereas:

```bash
#!/bin/bash
set -euo pipefail
counter=0
((++counter))
```

```
❯ ./test.sh

❯ echo $?
0
```

So I figure this change should do the trick.

## Update

A coworker created a new image, and the script works well now:

image

### DIFF
--- a/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh
+++ b/src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh
@@ -56,7 +56,7 @@ while :; do
 		exit 1
 	fi
 	sleep 1
-	((counter++))
+	((++counter))
 done
 
 # Execute functions


src/cloud-api-adaptor/podvm/files/usr/local/bin/setup-nat-for-imds.sh
```diff
@@ -56,7 +56,7 @@ while :; do
 		exit 1
 	fi
 	sleep 1
-	((counter++))
+  ((++counter))
```
wainersm: Hi @HernanGatta ,

I could reproduce the error and I learned something new today ;)

The fix is correct but you also need to adjust the indentation (from space to tab).

HernanGatta: Woops, fixed!

snir911 (approved):
LGTM, thanks!

wainersm (approved):
Thanks @HernanGatta !

PR #3023, opened 2026-04-27
podvm-mkosi: Fix podvm Ubuntu image build on s390x

BbolroC:
`24.04` was passed to `mkosi --tools-tree-release` which led to `404` error like:

```
Err:4 http://ports.ubuntu.com 24.04 Release
404  Not Found [IP: 91.189.91.104 80]
```

Use the code name `noble` instead. 

Plus, bump mkosi from v22 to v26 because it is too old and the default tools tree for v22 includes the following:

```
grub2 systemd-boot pesign sbsigntool
```

which are not available for s390x and this can be resolved by the version update.

Assisted-by: IBM Bob


BbolroC:
## Major Structural Changes from v22 to v26 

### 1. Directory Reorganization

mkosi v26 introduced a new directory structure using `mkosi.images/` instead of `mkosi.presets/` and `mkosi.profiles/`:

**Old structure (v22)**:
```
mkosi.presets/
  system/
    mkosi.conf
    mkosi.conf.d/
    mkosi.repart/
    mkosi.finalize.chroot
mkosi.profiles/
  debug.conf
  production.conf
  sftp.conf
mkosi.postinst          # Applied to ALL sub-images
mkosi.skeleton/         # Applied to ALL sub-images
```

**New structure (v26)**:
```
mkosi.images/
  initrd/
    mkosi.conf
    mkosi.conf.d/
    mkosi.postinst      # Specific to initrd image only
  system/
    mkosi.conf
    mkosi.conf.d/
    mkosi.repart/
    mkosi.finalize.chroot
    mkosi.postinst      # Specific to system image only
    mkosi.profiles/
      debug.conf
      production.conf
      sftp.conf
    mkosi.skeleton/     # Specific to system image only
```

#### Critical Behavior Change: Global vs Image-Specific Files

In mkosi v22, global `mkosi.skeleton/` and `mkosi.postinst` files at the root level were automatically applied to **all sub-images** (presets). In mkosi v26, this behavior changed significantly:

| File/Directory | mkosi v22 (mkosi.presets/) | mkosi v26 (mkosi.images/) |
|----------------|----------------------------|---------------------------|
| Global `mkosi.skeleton/` | Applied to all sub-images | Applied to top-level only (NOT sub-images) |
| Global `mkosi.postinst` | Run for all sub-images | Run for top-level only (NOT sub-images) |

**Impact**: In the migration to v26, files that were previously in global locations had to be moved into image-specific directories:
- `mkosi.postinst` → Split into `mkosi.images/initrd/mkosi.postinst` and `mkosi.images/system/mkosi.postinst`
- `mkosi.skeleton/` → Moved to `mkosi.images/system/mkosi.skeleton/`
- Network configuration files that were in the root skeleton are now in `mkosi.images/system/mkosi.skeleton/`

This ensures each sub-image (initrd and system) has its own specific configuration and files, rather than inheriting from a global location.

### 2. Configuration File Changes

#### KernelModules Configuration (Critical Breaking Change)

**This is the most important breaking change in mkosi v26.** The `KernelModules` setting is now **required** in the `[Content]` section for initrd images. Without this setting, mkosi v26 will not include kernel modules in the initrd, leading to boot failures.

**Before (v22)**:
```ini
[Content]
Packages=kernel-core

[Host]
ToolsTree=default
```

**After (v26)**:
```ini
[Content]
Packages=kernel-core
KernelModules=default
```

The `KernelModules=default` setting was added to:
- `mkosi.images/initrd/mkosi.conf` (base initrd configuration)
- `mkosi.images/initrd/mkosi.conf.d/fedora-s390x.conf`
- `mkosi.images/initrd/mkosi.conf.d/ubuntu-s390x.conf`

#### Profile Configuration Syntax

The `Profile` field was renamed to `Profiles` (plural) in match conditions:

**Before (v22)**:
```ini
[Match]
Distribution=fedora
Profile=debug
```

**After (v26)**:
```ini
[Match]
Distribution=fedora
Profiles=debug
```

#### Environment Variables Section

Profile configuration files moved environment variables from `[Content]` to `[Build]` section:

**Before (v22)**:
```ini
[Content]
Environment=VARIANT_ID=debug
```

**After (v26)**:
```ini
[Build]
Environment=VARIANT_ID=debug
```

### 3. Distribution Configuration Cleanup

Removed redundant `[Distribution]` sections from architecture-specific configuration files:

**Removed from files like `fedora-s390x.conf` and `ubuntu-s390x.conf`**:
```ini
[Distribution]
Distribution=fedora
Release=43
```

These settings are now inherited from the base distribution configuration files (`mkosi.conf.fedora` and `mkosi.conf.ubuntu`).

### 4. Ubuntu Release Configuration

Fixed Ubuntu 24.04 configuration to use the proper codename:

**Before (v22)**:
```ini
[Distribution]
Distribution=ubuntu
Release=24.04
```

**After (v26)**:
```ini
[Distribution]
Distribution=ubuntu
Release=noble
```

Together with the change in Makefile, this change prevents 404 errors when mkosi tries to fetch packages from Ubuntu repositories.

### 5. Tools Tree Configuration

Removed explicit `ToolsTree=default` settings from s390x configuration files, as this is now handled more intelligently by mkosi v26.

**Removed from `fedora-s390x.conf` and `ubuntu-s390x.conf`**:
```ini
[Host]
ToolsTree=default
```

## New Dependencies

Added new package dependencies in `Dockerfile.mkosi`:

```dockerfile
# New packages for mkosi v26
btrfs-progs
distribution-gpg-keys
```

These packages are required by mkosi v26 for proper image building functionality.

## Script and Configuration Updates

### 1. Initrd Configuration

Added new `mkosi.images/initrd/mkosi.conf` with base configuration for the initrd image.

### 2. System Finalization Script

Enhanced `mkosi.finalize.chroot` with additional functionality:

- **Resolv.conf setup**: Ensures `/etc/resolv.conf` symlink is properly configured
- **TPM PCR masking**: Masks systemd TPM PCR measurement units that shouldn't run in the image
- **Cloud-init handling**: Improved cloud-init service enablement for Ubuntu s390x
- **SFTP conditional logic**: Better handling of SFTP-specific configuration

### 3. System Post-installation Script

Created new `mkosi.images/system/mkosi.postinst` to add image metadata to `/etc/os-release`:

```bash
{
  echo "IMAGE_ID=\"podvm\""
  echo "IMAGE_VERSION=\"${IMAGE_VERSION-v0.0.0}\""
  echo "VARIANT_ID=\"${VARIANT_ID}\""
} >> "${BUILDROOT}/etc/os-release"
```

### 4. s390x Boot Process

Updated `hack/build-s390x-image.sh` to handle boot artifacts more carefully:

```bash
# Place boot artifacts after system extraction to avoid /boot content from
# the tar stream replacing or masking the expected files.
# Ensure that the extracted system tree does not leave behind an initramfs
# or vmlinuz in /boot, which may interfere with the expected boot artifacts.
rm -f "${dst_mnt}/boot/initrd.img" "${dst_mnt}/boot/vmlinuz"
cp initrd.cpio.zst "${dst_mnt}"/boot/initrd.img
cp system.vmlinuz "${dst_mnt}"/boot/vmlinuz
```

stevenhorsman (approved):
So I will start with the disclaimer that I'm not an expert in mkosi, but thanks to a great write-up I think the changes make sense, the e2e tests are all passing and bumping us up to a newer version of mkosi is a great bonus, so I think we definitely want this. I'm slightly concerned that the smoke-tests seem to be reliably failing on this PR, but I think we've discussed that they don't provide as much value as the e2e tests, and might be replaced, so I don't think we need to block this right now, though finding a fix would be good, so I'll have a play with it

stevenhorsman:
Ok, Bob has suggested a fix, so I've pushed to see if that helps with the smoke tests...

stevenhorsman:
I also re-based to resolve a merge conflict with a recent change

stevenhorsman:
Another commit to try and fix the smoke test issue by adding dm-verity kernel module

stevenhorsman:
OK - it took a long time, but Bob seems to have fixed the smoke test. The scratch space one still seems to have some out of resources failures - maybe caused by the changes, so I'll carry on trying...

The update didn't work, but I think we were planning to replace scratch-space by the kata approach, so maybe it's not a big issue?

mkulke:
> OK - it took a long time, but Bob seems to have fixed the smoke test. The scratch space one still seems to have some out of resources failures - maybe caused by the changes, so I'll carry on trying...
> 
> The update didn't work, but I think we were planning to replace scratch-space by the kata approach, so maybe it's not a big issue?

after glancing at it, I am not sure kata's and CAA's scratch space feature are similar, the scratch space will afaik just extend the /run partition to allow downloading images instead of consuming ram for tmpfs. the ephemeral disk feature will only apply to empty-dirs volumens that a user explicitly adds to their pods.

BbolroC:
Thanks @stevenhorsman for taking care of the PR. I will look into the smoke test and fix the issue.
Thanks @mkulke for the comment.

BbolroC:
### Issue on the smoke test

- mkosi v26 auto-generates a separate full kernel module archive and appends it to the base initrd when building the UKI. As a result, the initrd grew from 200 MB (v22) to 330 MB, which caused an out of resources error while the EFI firmware was loading the initrd.

### Resolution

- Required kernel modules should be declared via KernelInitrdModules in the system image where the kernel resides. It turns out that kernel module configuration in the initrd image has no effect. With the 3rd commit, the initrd size dropped to 68MB.

stevenhorsman (approved):
LGTM. Thanks for the initrd size trimming work for the smoke test!

wainersm (approved):
Hey @BbolroC , amazing work! Thank you so much!

PR #3032, opened 2026-04-30
ci: add versions cli

mkulke:
yq has been a constant source of problems in this repo, either because
we end up confusing v3 and v4 on runners, which do not share the same
syntax, or we have problems installing it via snap source.

the introduction of a project-local versions cli tool addresses this
issue, but also tightens and enforces a schema on the values that we put
into versions.yaml, currently we miss e.g. sha256 prefixes and use
floating point numbers as version strings.

```bash
$ ./versions -c
versions.yaml: ok
```

will check the schema, we can use this as part of CI. values can be
queried like this:

```
$ ./versions -o kata-containers -R
8dccf4cf37aeea4b6c2caacf3e61510d6eef2f71
$ ./versions --help
Query and validate versions.yaml

Usage:
  versions [flags]

Flags:
  -c, --check          Validate versions.yaml against the embedded JSON schema
  -f, --file string    Path to versions.yaml (default "versions.yaml")
  -h, --help           help for versions
  -o, --oci string     Select an oci.<name> entry (use with -r or -R)
  -R, --reference      With -o, print oci.<name>.reference
  -r, --registry       With -o, print oci.<name>.registry
  -t, --tool string    Select a tools.<name> entry (use -v/-d for object entries)
  -d, --tool-digest    With -t, print the tool's sha256 digest
  -v, --tool-version   With -t, print the tool's version
      --version        Print the version
```

The PR also adds a workflow to verify versions.yaml changes

### DIFF
--- a/.github/workflows/podvm_smoketest_ubuntu.yaml
+++ b/.github/workflows/podvm_smoketest_ubuntu.yaml
@@ -33,16 +33,15 @@ jobs:
           sudo apt-get install -y \
             genisoimage \
             qemu-utils
-          sudo snap install yq
 
       - name: Read properties from versions.yaml
         working-directory: src/cloud-api-adaptor
         run: |
           {
-            echo "MKOSI_VERSION=$(yq -e '.tools.mkosi' versions.yaml)";
-            echo "ORAS_VERSION=$(yq -e '.tools.oras' versions.yaml)";
-            echo "KATA_REF=$(yq -e '.oci.kata-containers.reference' versions.yaml)";
-            echo "KATA_REG=$(yq -e '.oci.kata-containers.registry' versions.yaml)";
+            echo "MKOSI_VERSION=$(./hack/versions.py -q tools.mkosi)";
+            echo "ORAS_VERSION=$(./hack/versions.py -q tools.oras)";
+            echo "KATA_REF=$(./hack/versions.py -q oci.kata-containers.reference)";
+            echo "KATA_REG=$(./hack/versions.py -q oci.kata-containers.registry)";
           } >> "$GITHUB_ENV"
 
       - uses: oras-project/setup-oras@38de303aac69abb66f3e6255b7198bff35f323e3 # v2.0.0
@@ -101,19 +100,23 @@ jobs:
             qemu-utils\
             socat \
             virt-manager
-          sudo snap install yq
 
       - name: Read properties from versions.yaml
         working-directory: src/cloud-api-adaptor
         run: |
           {
-            echo "KATA_REF=$(yq -e '.oci.kata-agent-ctl.reference' versions.yaml)";
-            echo "KATA_REG=$(yq -e '.oci.kata-agent-ctl.registry' versions.yaml)";
+            echo "ORAS_VERSION=$(./hack/versions.py -q tools.oras)";
+            echo "KATA_AGENT_CTL_REF=$(./hack/versions.py -q oci.kata-agent-ctl.reference)";
+            echo "KATA_AGENT_CTL_REG=$(./hack/versions.py -q oci.kata-agent-ctl.registry)";
           } >> "$GITHUB_ENV"
+
+      - uses: oras-project/setup-oras@38de303aac69abb66f3e6255b7198bff35f323e3 # v2.0.0
+        with:
+          version: ${{ env.ORAS_VERSION }}
 
       - name: Install kata-agent-ctl
         run: |
-          oras pull "${KATA_REG}/agent-ctl:${KATA_REF}-x86_64"
+          oras pull "${KATA_AGENT_CTL_REG}/agent-ctl:${KATA_AGENT_CTL_REF}-x86_64"
           tar --zstd -xf kata-static-agent-ctl.tar.zst
           cp opt/kata/bin/kata-agent-ctl /usr/local/bin
 
--- a/src/cloud-api-adaptor/hack/versions.py
+++ b/src/cloud-api-adaptor/hack/versions.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+"""Query and validate versions.yaml."""
+
+import argparse
+import json
+import sys
+
+import yaml
+
+
+def lookup(data, path):
+    cur = data
+    for key in path.split("."):
+        if not isinstance(cur, dict) or key not in cur:
+            sys.exit(f"error: '{path}' not found")
+        cur = cur[key]
+    if isinstance(cur, dict):
+        sys.exit(f"error: '{path}' is an object, not a scalar")
+    return cur
+
+
+def validate(data, schema_path):
+    import jsonschema
+
+    with open(schema_path) as f:
+        schema = json.load(f)
+    errors = list(jsonschema.Draft7Validator(schema).iter_errors(data))
+    if errors:
+        for e in errors:
+            print(f"  - {e.message}", file=sys.stderr)
+        sys.exit(1)
+    print("ok")
+
+
+def main():
+    p = argparse.ArgumentParser(description="Query and validate versions.yaml")
+    p.add_argument("-f", "--file", default="versions.yaml")
+    p.add_argument("-q", "--query", help="Dot path, e.g. tools.golang, oci.pause.tag")
+    p.add_argument(
+        "-c", "--check", metavar="SCHEMA", help="Validate against JSON schema"
+    )
+    args = p.parse_args()
+
+    with open(args.file) as f:
+        data = yaml.safe_load(f)
+
+    if args.check:
+        validate(data, args.check)
+    elif args.query:
+        print(lookup(data, args.query))
+    else:
+        p.print_help()
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/workflows/versions.yaml
+++ b/.github/workflows/versions.yaml
@@ -1,0 +1,47 @@
+# (C) Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Validate src/cloud-api-adaptor/versions.yaml against the JSON schema
+# embedded in the `versions` CLI tool. Runs on PRs and main pushes that
+# touch the file, the schema, or the tool's source.
+---
+name: versions
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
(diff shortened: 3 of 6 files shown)


.github/workflows/podvm_smoketest_ubuntu.yaml
```diff
@@ -101,14 +108,23 @@ jobs:
...
         working-directory: src/cloud-api-adaptor
         run: |
+          make versions
           {
-            echo "KATA_REF=$(yq -e '.oci.kata-agent-ctl.reference' versions.yaml)";
-            echo "KATA_REG=$(yq -e '.oci.kata-agent-ctl.registry' versions.yaml)";
+            echo "MKOSI_VERSION=$(./versions -t mkosi)";
+            echo "ORAS_VERSION=$(./versions -t oras)";
+            echo "KATA_REF=$(./versions -o kata-containers -R)";
+            echo "KATA_REG=$(./versions -o kata-containers -r)";
```
stevenhorsman: This was reading kata-agent-ctl?

stevenhorsman (commented):
I don't love having to homebuild a tool that should exist, but I recognise that yq hasn't been kind to us, so hopefully the maintenance of this is lower than having to manage yq. One read that seems to have been changed though?

.github/workflows/podvm_smoketest_ubuntu.yaml
mkulke: oh, indeed. thx. we were also missing oras setup in this job apparently

mkulke:
> I don't love having to homebuild a tool that should exist, but I recognise that yq hasn't been kind to us, so hopefully the maintenance of this is lower than having to manage yq. One read that seems to have been changed though?

I'm usually on this side of the fence, but yq (and yaml, tbh) is just a particular unfortunate pick for this purpose. it's a tool that is not part of distro repositories, has ambiguous behaviour acroos versions, and the versions differ from kata and caa repos (which often share runners), so we'll keep running into issues.

Another option is to convert versions.yaml to json, jq is available everywhere and unambigious. the json schema still makes sense, it can be checked by some inline python I guess. a downside: we have to adapt the version extraction everywhere.

stevenhorsman (approved):
Thanks for the fix. My comments weren't meant to block this - I'm happy to see how this goes and adapt if it causes issues later, but let's hope it's not needed 😄

wainersm:
Hi @mkulke ,

Mixed feelings here.

It will require the conversion of all remaining 15 workflows that still use yq. Is it in your plan to use the new tool everywhere?

Did you consider writing the tool in Python but figured it would have its limitations too? I'm asking because my friend Claude generated a very simply script and told me the pyyaml is pre-installed in the gh runners.

The script for reference:

```python
#!/usr/bin/env python3
  """Query and validate versions.yaml."""                                                                                                         
                                                                  
  import argparse
  import json
  import sys

  import yaml

                                                                                                                                                  
  def lookup(data, path):
      cur = data                                                                                                                                  
      for key in path.split("."):                                 
          if not isinstance(cur, dict) or key not in cur:                                                                                         
              sys.exit(f"error: '{path}' not found")
          cur = cur[key]                                                                                                                          
      if isinstance(cur, dict):                                                                                                                   
          sys.exit(f"error: '{path}' is an object, not a scalar")
      return cur                                                                                                                                  
                                                                  

  def validate(data, schema_path):
      import jsonschema
                                                                                                                                                  
      with open(schema_path) as f:
          schema = json.load(f)                                                                                                                   
      errors = list(jsonschema.Draft7Validator(schema).iter_errors(data))                                                                         
      if errors:                                                                                                                                  
          for e in errors:                                                                                                                        
              print(f"  - {e.message}", file=sys.stderr)                                                                                          
          sys.exit(1)                                                                                                                             
      print("ok")
                                                                                                                                                  
                                                                  
  def main():
      p = argparse.ArgumentParser(description="Query and validate versions.yaml")
      p.add_argument("-f", "--file", default="versions.yaml")
      p.add_argument("-q", "--query", help="Dot path, e.g. tools.golang, oci.pause.tag")                                                          
      p.add_argument("-c", "--check", metavar="SCHEMA", help="Validate against JSON schema")
      args = p.parse_args()                                                                                                                       
                                                                  
      with open(args.file) as f:                                                                                                                  
          data = yaml.safe_load(f)                                

      if args.check:
          validate(data, args.check)
      elif args.query:                                                                                                                            
          print(lookup(data, args.query))
      else:                                                                                                                                       
          p.print_help()                                          
          sys.exit(1)


  if __name__ == "__main__":
      main()
```

Use would be like:

```
 # Query any path generically — no special flags per section                                                                                     
  ./versions.py -q tools.golang        # → 1.25.9                                                                                                 
  ./versions.py -q tools.helm.version  # → v4.0.4                                                                                                 
  ./versions.py -q oci.pause.tag       # → 3.9                                                                                                    
  ./versions.py -q git.kbs.reference   # → 22788122...                                                                                            
                                                                                                                                                  
  # Schema validation                                                                                                                             
  ./versions.py -c pkg/versions/schema.json
```

mkulke:
> Hi @mkulke ,
> 
> Mixed feelings here.
> 
> It will require the conversion of all remaining 15 workflows that still use yq. Is it in your plan to use the new tool everywhere?

yq will keep working. it can be gradually adopted, once yq is blocking stuff again b/c a snap "repo" is offline or some worker has the wrong version installed. that was the idea. the other idea would be to convert it to version.json, which would require adopting everything at once.
> 
> Did you consider writing the tool in Python but figured it would have its limitations too? I'm asking because my friend Claude generated a very simply script and told me the pyyaml is pre-installed in the gh runners.
> 
> The script for reference:
> 
> ```python
> #!/usr/bin/env python3
>   """Query and validate versions.yaml."""                                                                                                         
>                                                                   
>   import argparse
>   import json
>   import sys
> 
>   import yaml
> 
>                                                                                                                                                   
>   def lookup(data, path):
>       cur = data                                                                                                                                  
>       for key in path.split("."):                                 
>           if not isinstance(cur, dict) or key not in cur:                                                                                         
>               sys.exit(f"error: '{path}' not found")
>           cur = cur[key]                                                                                                                          
>       if isinstance(cur, dict):                                                                                                                   
>           sys.exit(f"error: '{path}' is an object, not a scalar")
>       return cur                                                                                                                                  
>                                                                   
> 
>   def validate(data, schema_path):
>       import jsonschema
>                                                                                                                                                   
>       with open(schema_path) as f:
>           schema = json.load(f)                                                                                                                   
>       errors = list(jsonschema.Draft7Validator(schema).iter_errors(data))                                                                         
>       if errors:                                                                                                                                  
>           for e in errors:                                                                                                                        
>               print(f"  - {e.message}", file=sys.stderr)                                                                                          
>           sys.exit(1)                                                                                                                             
>       print("ok")
>                                                                                                                                                   
>                                                                   
>   def main():
>       p = argparse.ArgumentParser(description="Query and validate versions.yaml")
>       p.add_argument("-f", "--file", default="versions.yaml")
>       p.add_argument("-q", "--query", help="Dot path, e.g. tools.golang, oci.pause.tag")                                                          
>       p.add_argument("-c", "--check", metavar="SCHEMA", help="Validate against JSON schema")
>       args = p.parse_args()                                                                                                                       
>                                                                   
>       with open(args.file) as f:                                                                                                                  
>           data = yaml.safe_load(f)                                
> 
>       if args.check:
>           validate(data, args.check)
>       elif args.query:                                                                                                                            
>           print(lookup(data, args.query))
>       else:                                                                                                                                       
>           p.print_help()                                          
>           sys.exit(1)
> 
> 
>   if __name__ == "__main__":
>       main()
> ```
> 
> Use would be like:
> 
> ```
>  # Query any path generically — no special flags per section                                                                                     
>   ./versions.py -q tools.golang        # → 1.25.9                                                                                                 
>   ./versions.py -q tools.helm.version  # → v4.0.4                                                                                                 
>   ./versions.py -q oci.pause.tag       # → 3.9                                                                                                    
>   ./versions.py -q git.kbs.reference   # → 22788122...                                                                                            
>                                                                                                                                                   
>   # Schema validation                                                                                                                             
>   ./versions.py -c pkg/versions/schema.json
> ```

I used golang because we use it elsewhere for tooling in the repo. python might bring its own dependency issues. I don't know whether pyyaml + jsonschema can be assumed to be installed everywhere (also on more esoteric self-hosted runners). _but_ the python version would still be a net improvement over where we are now. we don't need to encode the hierarchies a CLI as long as the config sprawl is limited by a jsonschema, so "some.thing.ref" should be ok.

should we adopt the python version and hope it works everywhere?

ldoktor:
> I used golang because we use it elsewhere for tooling in the repo. python might bring its own dependency issues. I don't know whether pyyaml + jsonschema can be assumed to be installed everywhere (also on more esoteric self-hosted runners). _but_ the python version would still be a net improvement over where we are now. we don't need to encode the hierarchies a CLI as long as the config sprawl is limited by a jsonschema, so "some.thing.ref" should be ok.
> 
> should we adopt the python version and hope it works everywhere?

I'd vote for python, because it's less boilerplate code and still in use by most modern distros. But I'm biased as I love python :-)

mkulke:
ok, switched to the python impl.

src/cloud-api-adaptor/hack/versions.py
```diff
@@ -0,0 +1,57 @@
...
+        if not isinstance(cur, dict) or key not in cur:
+            sys.exit(f"error: '{path}' not found")
+        cur = cur[key]
+    if isinstance(cur, dict):
+        sys.exit(f"error: '{path}' is an object, not a scalar")
+    return cur
+
+
+def validate(data, schema_path):
+    import jsonschema
```
stevenhorsman, line 23: Is the theory that all our runners should have jsonschema pre-installed, so we don't need `pip install jsonschema` in the workflow?

mkulke: yes, this is indeed the theory, backed by 2 engineers and 1 agent. 🤞 if we need juggle python deps then this will no less painful than conjuring yq 🙄. in this case I will make it my personal cause to convert versions.yaml to versions.json

stevenhorsman: Cool - thanks for confirming. Sounds like a reasonable plan to me

wainersm (approved):
Thanks @mkulke for tackling the problem and accepting Python in your heart :D

PR #3037, opened 2026-05-03
Add cloud-native volume passthrough for Peer Pods

mohammedadnan21:
## Summary

Enable CSI-provisioned block volumes to be attached to PodVMs as cloud-managed disks and mounted inside the guest. This PR contains only the CAA-side changes; the CAA CSI block driver is maintained in a separate repository.

Ref: https://github.com/confidential-devhub/caa-csi-block-driver

### Key Changes

**1. mountInfo.json handling in CAA proxy** (`pkg/adaptor/proxy/service.go`)
- Reads `mountInfo.json` from the shared direct-volumes directory (written by the CSI block driver) for each CSI volume mount
- Constructs a `cloud_volumes` annotation with volume metadata (disk ID, fs type, mount point) and deterministic LUN index, forwarded to the PodVM interceptor

**2. Device detection and mounting in APF interceptor** (`pkg/forwarder/interceptor/interceptor.go`)
- Parses `cloud_volumes` annotation to identify volumes
- Auto-detects the cloud provider and uses provider-specific device paths:
  - **Azure:** `/dev/disk/azure/data/by-lun/N`, Hyper-V by-path symlinks, sysfs HCTL fallback
  - **AWS:** `/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*` with EBS volume ID matching for multi-disk correctness
  - **Libvirt:** `/dev/vd*` (virtio)
- Safe `formatAndMount`: tries mount first (preserves existing data), retries blkid with exit-code awareness, only formats genuinely empty disks
- Retry loop with configurable attempts/delay for disk detection (hot-added disks may take time to appear)

**3. Azure provider disk attachment** (`cloud-providers/azure/provider.go`)
- Attaches CSI volumes as Azure Managed Disk data disks during VM creation with LUN-based indexing
- Deterministic LUN assignment via sorted volume keys

**4. AWS provider disk attachment** (`cloud-providers/aws/provider.go`)
- Attaches EBS volumes after instance creation using `AttachVolume` API
- Parallel attachment with consistent device naming (`/dev/xvd[b-z]`)
- Detaches volumes on instance termination via `DetachEBSVolumes`

**Supporting changes:**
- `cloud-providers/types.go`: Added `Volumes` field to `InstanceTypeSpec`
- `pkg/adaptor/cloud/cloud.go`: Passes CSI volumes to cloud provider `CreateInstance`
- `pkg/util/cloud.go`: `GetCSIVolumesForPod` scans direct-volumes directory; exported `KataDirectVolumesDir` as single source of truth
- `podvm-mkosi/mkosi.images/system/mkosi.postinst`: Azure udev rules for `/dev/disk/azure/data/by-lun/N` symlinks

### Testing

| Cloud | Scenario | Result |
|-------|----------|--------|
| Azure (AKS) | Single disk persistence | PASSED |
| Azure (AKS) | Multi disk persistence | PASSED |
| AWS (EKS) | Single disk persistence | PASSED |
| AWS (EKS) | Multi disk persistence | PASSED |

**Unit tests:**
- `go test ./pkg/adaptor/proxy/...` — proxy volume annotation (single/multi-disk, LUN skipping, pod-UID filtering)
- `go test ./pkg/util/...` — CSI volume discovery
- `go test ../cloud-providers/azure/...` — Azure resource ID parsing, disk attachment
- `go test ../cloud-providers/aws/...` — EBS attachment, device naming

### Notes

- CAA CSI block driver is maintained separately: https://github.com/confidential-devhub/caa-csi-block-driver
- AWS does not require PodVM udev rules — EBS NVMe devices are enumerated automatically by the kernel NVMe driver

### DIFF
--- a/src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
+++ b/src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
@@ -6,11 +6,14 @@ package proxy
 import (
 	"context"
 	b64 "encoding/base64"
+	"encoding/json"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/util"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/util/agentproto"
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
 	"google.golang.org/protobuf/types/known/emptypb"
@@ -22,14 +25,12 @@ type proxyService struct {
 }
 
 const (
-	defaultPauseImage            = "registry.k8s.io/pause:3.7"
-	kataDirectVolumesDir         = "/run/kata-containers/shared/direct-volumes"
-	volumeTargetPathKey          = "io.confidentialcontainers.org.peerpodvolumes.target_path"
-	csiPluginEscapeQualifiedName = "kubernetes.io~csi"
-	imageGuestPull               = "image_guest_pull"
-	cdiAnnotationKey             = "cdi.k8s.io/peer-pods"
-	defaultCDIType               = "nvidia.com/gpu=all"
-	defaultGPUsAnnotation        = "io.katacontainers.config.hypervisor.default_gpus"
+	defaultPauseImage     = "registry.k8s.io/pause:3.7"
+	volumeTargetPathKey   = "io.confidentialcontainers.org.peerpodvolumes.target_path"
+	imageGuestPull        = "image_guest_pull"
+	cdiAnnotationKey      = "cdi.k8s.io/peer-pods"
+	defaultCDIType        = "nvidia.com/gpu=all"
+	defaultGPUsAnnotation = "io.katacontainers.config.hypervisor.default_gpus"
 )
 
 func newProxyService(dialer func(context.Context) (net.Conn, error), pauseImage string) *proxyService {
@@ -47,12 +48,16 @@ func newProxyService(dialer func(context.Context) (net.Conn, error), pauseImage 
 func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContainerRequest) (*emptypb.Empty, error) {
 	var pullImageInGuest bool
 	logger.Printf("CreateContainer: containerID:%s", req.ContainerId)
+	if req.OCI.Annotations == nil {
+		req.OCI.Annotations = make(map[string]string)
+	}
+
 	if len(req.OCI.Mounts) > 0 {
 		logger.Print("    mounts:")
 		for i, m := range req.OCI.Mounts {
 			logger.Printf("        destination:%s source:%s type:%s", m.Destination, m.Source, m.Type)
 
-			if isNodePublishVolumeTargetPath(m.Source, kataDirectVolumesDir) {
+			if isNodePublishVolumeTargetPath(m.Source, util.KataDirectVolumesDir) {
 				if i > 0 {
 					req.OCI.Annotations[volumeTargetPathKey] += ","
 				}
@@ -94,6 +99,101 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 		logger.Printf("Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.")
 	}
 
+	// Detect cloud volumes by scanning the direct-volumes directory in canonical
+	// order. The scan order matches GetCSIVolumesForPod (os.ReadDir sorts by
+	// name), so the LUN index here is consistent with the cloud provider's
+	// disk attachment order.
+	cloudVolumes := make(map[string]util.CloudVolumeAnnotation)
+	podUID := req.OCI.Annotations["io.kubernetes.cri.sandbox-uid"]
+
+	dirEntries, dirErr := os.ReadDir(util.KataDirectVolumesDir)
+	if dirErr == nil {
+		canonicalIdx := 0
+		for _, entry := range dirEntries {
+			if !entry.IsDir() {
+				continue
+			}
+			decodedBytes, err := b64.URLEncoding.DecodeString(entry.Name())
+			if err != nil {
+				continue
+			}
+			decodedPath := string(decodedBytes)
+
+			if !strings.Contains(decodedPath, "/volumes/"+util.CSIPluginEscapeQualifiedName+"/") {
+				continue
+			}
+			if podUID != "" && !strings.Contains(decodedPath, "/pods/"+podUID+"/") {
+				continue
+			}
+
+			mountInfoPath := filepath.Join(util.KataDirectVolumesDir, entry.Name(), "mountInfo.json")
+			data, err := os.ReadFile(mountInfoPath)
+			if err != nil {
+				logger.Printf("could not read mountInfo.json for %s: %v", decodedPath, err)
+				continue
+			}
+			var mountInfo map[string]interface{}
+			if err := json.Unmarshal(data, &mountInfo); err != nil {
+				logger.Printf("could not parse mountInfo.json for %s: %v", decodedPath, err)
+				continue
+			}
+
+			diskID := ""
+			if md, ok := mountInfo["metadata"].(map[string]interface{}); ok {
+				if cp, ok := md["cloud-volume-path"].(string); ok && cp != "" {
+					diskID = cp
+				}
+			}
+			if diskID == "" {
+				if d, ok := mountInfo["device"].(string); ok {
+					diskID = d
+				}
+			}
+			if diskID == "" {
+				logger.Printf("cloud volume at %s has no disk ID, skipping", decodedPath)
+				continue
+			}
+
+			fsType := "ext4"
+			if ft, ok := mountInfo["fstype"].(string); ok && ft != "" {
+				fsType = ft
+			}
+
+			mountDest := ""
+			for _, m := range req.OCI.Mounts {
+				if m.Source == decodedPath {
+					mountDest = m.Destination
+					break
+				}
+			}
+			if mountDest == "" {
+				logger.Printf("cloud volume disk %s has no matching mount in container spec, skipping", diskID)
+				canonicalIdx++
+				continue
+			}
+
+			volKey := fmt.Sprintf("vol-%d", canonicalIdx)
+			cloudVolumes[volKey] = util.CloudVolumeAnnotation{
+				MountPoint: mountDest,
+				FSType:     fsType,
+				LUN:        fmt.Sprintf("%d", canonicalIdx),
+				DiskID:     diskID,
+			}
+			logger.Printf("Detected cloud volume %s -> %s (lun=%d, disk=%s, fs=%s)", volKey, mountDest, canonicalIdx, diskID, fsType)
+			canonicalIdx++
+		}
+	}
+
+	if len(cloudVolumes) > 0 {
+		cvJSON, err := json.Marshal(cloudVolumes)
+		if err != nil {
+			logger.Printf("failed to marshal cloud_volumes annotation: %v", err)
+		} else {
+			req.OCI.Annotations[util.CloudVolumesAnnotationKey] = string(cvJSON)
+			logger.Printf("Set cloud_volumes annotation: %s", string(cvJSON))
+		}
+	}
+
 	res, err := s.Redirector.CreateContainer(ctx, req)
 
 	if err != nil {
@@ -104,7 +204,7 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 }
 
 func isNodePublishVolumeTargetPath(volumePath, directVolumesDir string) bool {
-	if !strings.Contains(filepath.Clean(volumePath), "/volumes/"+csiPluginEscapeQualifiedName+"/") {
+	if !strings.Contains(filepath.Clean(volumePath), "/volumes/"+util.CSIPluginEscapeQualifiedName+"/") {
 		return false
 	}
 
--- a/src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
+++ b/src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
@@ -5,11 +5,17 @@ package interceptor
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"net"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	retry "github.com/avast/retry-go/v4"
@@ -18,14 +24,22 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"google.golang.org/protobuf/types/known/emptypb"
 
+	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/util"
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/util/agentproto"
 )
 
 const (
-	volumeTargetPathKey = "io.confidentialcontainers.org.peerpodvolumes.target_path"
-	volumeCheckInterval = 5 * time.Second
-	volumeCheckTimeout  = 3 * time.Minute
+	volumeTargetPathKey  = "io.confidentialcontainers.org.peerpodvolumes.target_path"
+	cloudVolumeMountBase = "/run/cloud-volumes"
+	volumeCheckInterval  = 5 * time.Second
+	volumeCheckTimeout   = 3 * time.Minute
 )
+
+var allowedFSTypes = map[string]bool{
+	"ext4": true,
+	"ext3": true,
+	"xfs":  true,
+}
 
 var logger = log.New(log.Writer(), "[forwarder/interceptor] ", log.LstdFlags|log.Lmsgprefix)
 
@@ -36,7 +50,22 @@ type Interceptor interface {
 type interceptor struct {
 	agentproto.Redirector
 
-	nsPath string
+	nsPath          string
+	cloudMountPaths []string
+}
+
+// unmountCloudVolumes unmounts all cloud volume mount points in reverse order.
+// Called during DestroySandbox to clean up before the PodVM is terminated.
+func (i *interceptor) unmountCloudVolumes() {
+	for idx := len(i.cloudMountPaths) - 1; idx >= 0; idx-- {
+		mp := i.cloudMountPaths[idx]
+		if err := syscall.Unmount(mp, 0); err != nil {
+			logger.Printf("WARNING: failed to unmount cloud volume %s: %v", mp, err)
+		} else {
+			logger.Printf("Unmounted cloud volume %s", mp)
+		}
+	}
+	i.cloudMountPaths = nil
 }
 
 func dial(ctx context.Context, agentSocket string) (net.Conn, error) {
@@ -93,6 +122,95 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 	logger.Printf("    namespaces:")
 	for _, ns := range req.OCI.Linux.Namespaces {
 		logger.Printf("    %s: %q", ns.Type, ns.Path)
+	}
+
+	// Handle cloud volumes: detect device, format if needed, mount, and
+	// bind-mount into the container's mount namespace.
+	if cvJSON, ok := req.OCI.Annotations[util.CloudVolumesAnnotationKey]; ok && cvJSON != "" {
+		var cloudVolumes map[string]util.CloudVolumeAnnotation
+		if err := json.Unmarshal([]byte(cvJSON), &cloudVolumes); err != nil {
+			return nil, fmt.Errorf("corrupt cloud_volumes annotation (pod would start without volumes): %w", err)
+		}
+
+		volNames := make([]string, 0, len(cloudVolumes))
+		for k := range cloudVolumes {
+			volNames = append(volNames, k)
+		}
+		sort.Strings(volNames)
+
+		for _, volName := range volNames {
+			volInfo := cloudVolumes[volName]
+			mountPoint := volInfo.MountPoint
+			fsType := volInfo.FSType
+			lunStr := volInfo.LUN
+			if mountPoint == "" || lunStr == "" {
+				return nil, fmt.Errorf("cloud volume %s missing required mount_point or lun field", volName)
+			}
+
+			safeName := filepath.Base(volName)
+			if safeName != volName || safeName == "." || safeName == ".." {
+				return nil, fmt.Errorf("cloud volume %q has unsafe name", volName)
+			}
+
+			if fsType == "" {
+				fsType = "ext4"
+			}
+			if !allowedFSTypes[fsType] {
+				return nil, fmt.Errorf("cloud volume %s requests unsupported filesystem type %q (allowed: ext4, ext3, xfs)", volName, fsType)
+			}
+
+			lunIdx, err := strconv.Atoi(lunStr)
+			if err != nil {
+				return nil, fmt.Errorf("cloud volume %s has invalid lun %q: %w", volName, lunStr, err)
+			}
+
+			diskID := volInfo.DiskID
+			device, err := findDataDiskDevice(lunIdx, diskID)
+			if err != nil {
+				return nil, fmt.Errorf("cloud volume %s: %w", volName, err)
+			}
+			logger.Printf("cloud volume %s: LUN %d -> device %s", volName, lunIdx, device)
+
+			hostMountPoint := filepath.Join(cloudVolumeMountBase, safeName)
+			if err := os.MkdirAll(hostMountPoint, 0o755); err != nil {
+				return nil, fmt.Errorf("creating mount point for %s: %w", volName, err)
+			}
+
+			if err := waitForDevice(device); err != nil {
+				return nil, fmt.Errorf("cloud volume %s device %s not available: %w", volName, device, err)
+			}
+
+			if err := formatAndMount(device, hostMountPoint, fsType); err != nil {
+				return nil, fmt.Errorf("failed to mount cloud volume %s at %s: %w", volName, hostMountPoint, err)
+			}
+			i.cloudMountPaths = append(i.cloudMountPaths, hostMountPoint)
+
+			if fsGroupStr := volInfo.FSGroup; fsGroupStr != "" {
+				if gid, err := strconv.Atoi(fsGroupStr); err == nil {
+					logger.Printf("cloud volume %s: applying fsGroup %d to %s", volName, gid, hostMountPoint)
+					if err := os.Chown(hostMountPoint, -1, gid); err != nil {
+						logger.Printf("WARNING: failed to chown %s to gid %d: %v", hostMountPoint, gid, err)
+					}
+					if err := os.Chmod(hostMountPoint, 0o2775); err != nil {
+						logger.Printf("WARNING: failed to chmod %s: %v", hostMountPoint, err)
+					}
+				}
+			}
+
+			rewrote := false
+			for idx, m := range req.OCI.Mounts {
+				if m.Destination == mountPoint {
+					req.OCI.Mounts[idx].Source = hostMountPoint
+					req.OCI.Mounts[idx].Type = "bind"
+					logger.Printf("cloud volume %s: rewrote mount source to %s", volName, hostMountPoint)
+					rewrote = true
+					break
+				}
+			}
+			if !rewrote {
+				logger.Printf("WARNING: cloud volume %s mount_point %q not found in container mounts", volName, mountPoint)
+			}
+		}
 	}
 
 	volumeTargetPath := req.OCI.Annotations[volumeTargetPathKey]
@@ -221,6 +339,8 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
 
 	logger.Printf("DestroySandbox")
 
+	i.unmountCloudVolumes()
+
 	res, err := i.Redirector.DestroySandbox(ctx, req)
 
 	if err != nil {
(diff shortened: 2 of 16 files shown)


src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +288,160 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
 
 	return res, err
 }
+
+// findDataDiskDevice locates the block device for the given LUN index.
+// It tries Azure-specific symlinks first, then falls back to scanning sysfs.
+func findDataDiskDevice(lunIdx int) (string, error) {
```
bpradipt: What happens for other providers?
Do you need to detect the platform and accordingly handle the symlinks ?

mohammedadnan21: Good point. Right now the fallback heuristic (scanning sysfs for partition-free devices) handles other providers, but it's not production-safe.

I'll pass the cloud provider name through the cloud_volumes annotation so the interceptor can use provider-specific device detection paths:
- Azure: `/dev/disk/azure/data/by-lun/N` and Hyper-V by-path symlinks
- AWS: `/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*` (NVMe)
- Libvirt: `/dev/vd*` (virtio)
- Generic fallback with a safety check (refuse to format if only heuristic was used)

Will update in this PR.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,51 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+				continue
+			}
+			var mountInfo map[string]interface{}
+			if err := json.Unmarshal(data, &mountInfo); err != nil {
+				logger.Printf("could not parse mountInfo.json for %s: %v", m.Source, err)
+				continue
+			}
+			volInfo := map[string]string{
+				"mount_point": m.Destination,
+				"fs_type":     "ext4",
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. The proxy now reads `fstype` from `mountInfo.json` and only falls back to `ext4` when the field is absent or empty.

src/cloud-api-adaptor/pkg/util/cloud.go
```diff
@@ -114,3 +119,65 @@ func Contains(slice []string, s string) bool {
...
+// mountInfo.json files written by the CSI block driver. Each file
+// describes a cloud volume that should be attached to the PodVM.
+func GetCSIVolumesForPod(annotations map[string]string) []provider.CloudVolume {
+	var volumes []provider.CloudVolume
+
+	entries, err := os.ReadDir(kataDirectVolumesDir)
+	if err != nil {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. Both `GetCSIVolumesForPod` and the proxy now filter volumes by pod UID (`io.kubernetes.cri.sandbox-uid` annotation). The decoded base64 path contains the pod UID, so only volumes belonging to the current pod are processed.

src/cloud-api-adaptor/pkg/util/cloud.go
```diff
@@ -114,3 +119,65 @@ func Contains(slice []string, s string) bool {
...
+const (
+	kataDirectVolumesDir         = "/run/kata-containers/shared/direct-volumes"
+	csiPluginEscapeQualifiedName = "kubernetes.io~csi"
+)
+
+// GetCSIVolumesForPod scans the shared direct-volumes directory for
+// mountInfo.json files written by the CSI block driver. Each file
+// describes a cloud volume that should be attached to the PodVM.
+func GetCSIVolumesForPod(annotations map[string]string) []provider.CloudVolume {
+	var volumes []provider.CloudVolume
```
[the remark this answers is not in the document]
mohammedadnan21: Acknowledged. Unit tests for `GetCSIVolumesForPod` will be added in a follow-up. The function now includes pod-UID-based filtering and validation warnings for invalid/missing data.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,51 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+				"lun":         fmt.Sprintf("%d", cloudVolIdx),
+			}
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. The proxy now scans `kataDirectVolumesDir` directly in canonical order (using `os.ReadDir` which sorts by name), matching the same ordering used by `GetCSIVolumesForPod`. This ensures the LUN index assigned in the annotation is consistent with the cloud provider's disk attachment order.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -95,6 +101,59 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
...
+
+				device, err := findDataDiskDevice(lunIdx)
+				if err != nil {
+					return nil, fmt.Errorf("cloud volume %s: %w", volName, err)
+				}
+				logger.Printf("cloud volume %s: LUN %d -> device %s", volName, lunIdx, device)
+
+				hostMountPoint := filepath.Join(cloudVolumeMountBase, volName)
+				if err := os.MkdirAll(hostMountPoint, 0o755); err != nil {
+					return nil, fmt.Errorf("creating mount point for %s: %w", volName, err)
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. Volume names are now validated with `filepath.Base()` — names containing path separators, `.`, or `..` are rejected before the mount point path is constructed.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +288,160 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
...
+func formatAndMount(device, mountPoint, fsType string) error {
+	if fsType == "" {
+		fsType = "ext4"
+	}
+
+	// Try mounting first — if the disk already has a valid filesystem, this
+	// avoids any risk of accidentally reformatting it.
+	mountCmd := exec.Command("mount", "-t", fsType, device, mountPoint)
+	if mountOut, mountErr := mountCmd.CombinedOutput(); mountErr == nil {
+		logger.Printf("Mounted existing filesystem on %s at %s (type=%s)", device, mountPoint, fsType)
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. `formatAndMount` now checks `mountinfo.Mounted(mountPoint)` at the top and returns success immediately if already mounted.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +288,160 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
...
+
+	if !needsFormat {
+		return fmt.Errorf("cannot determine filesystem state of %s after retries; refusing to format to protect data", device)
+	}
+
+	logger.Printf("No filesystem on %s, formatting as %s", device, fsType)
+	mkfsCmd := exec.Command("mkfs."+fsType, device)
+	if mkfsOut, mkfsErr := mkfsCmd.CombinedOutput(); mkfsErr != nil {
+		return fmt.Errorf("mkfs.%s on %s failed: %s: %w", fsType, device, strings.TrimSpace(string(mkfsOut)), mkfsErr)
+	}
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. Added an `allowedFSTypes` allowlist (`ext4`, `ext3`, `xfs`). The fsType is validated both in `CreateContainer` (before device detection) and in `formatAndMount` (defense in depth). Any unrecognized type is rejected.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +288,160 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
...
+	for attempt := 0; attempt < 30; attempt++ {
+		if strings.HasPrefix(device, "/dev/disk/") {
+			if _, err := os.Lstat(device); err == nil {
+				return nil
+			}
+		} else {
+			devName := filepath.Base(device)
+			if _, err := os.Stat(filepath.Join("/sys/block", devName)); err == nil {
+				return nil
+			}
```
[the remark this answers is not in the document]
mohammedadnan21: The current implementation checks both symlink paths (via `os.Lstat`) and sysfs entries. Since cloud volume devices are always whole disks (not partitions), the sysfs check is valid. The symlink path check (`/dev/disk/...`) handles the resolved-to-partition case via `Lstat`. Will add a comment clarifying this assumption.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +288,160 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
...
+		if err != nil || len(partEntries) > 0 {
+			continue
+		}
+		candidates = append(candidates, "/dev/"+name)
+	}
+
+	if lunIdx < len(candidates) {
+		logger.Printf("Found data disk via sysfs heuristic: %s (lun=%d)", candidates[lunIdx], lunIdx)
+		return candidates[lunIdx], nil
+	}
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. The sysfs fallback now reads the actual SCSI HCTL (Host:Channel:Target:Lun) from the sysfs device path instead of relying on directory listing order. This correctly matches the LUN assigned by the cloud provider.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -95,6 +101,59 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
 		logger.Printf("    %s: %q", ns.Type, ns.Path)
 	}
 
+	// Handle cloud volumes: detect device, format if needed, mount, and
+	// bind-mount into the container's mount namespace
+	if cvJSON, ok := req.OCI.Annotations[cloudVolumesKey]; ok && cvJSON != "" {
+		var cloudVolumes map[string]map[string]string
+		if err := json.Unmarshal([]byte(cvJSON), &cloudVolumes); err != nil {
```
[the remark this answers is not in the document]
mohammedadnan21: Acknowledged. Unit tests for the cloud volume handling will be added in a follow-up commit. The current round focuses on correctness and safety fixes.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,51 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+			data, err := os.ReadFile(mountInfoPath)
+			if err != nil {
+				logger.Printf("could not read mountInfo.json for %s: %v", m.Source, err)
+				continue
+			}
+			var mountInfo map[string]interface{}
+			if err := json.Unmarshal(data, &mountInfo); err != nil {
+				logger.Printf("could not parse mountInfo.json for %s: %v", m.Source, err)
+				continue
+			}
```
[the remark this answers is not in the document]
mohammedadnan21: The filtering now uses two checks: (1) the decoded path must contain the CSI plugin qualifier `kubernetes.io~csi`, and (2) the path must contain the current pod's UID. This prevents processing non-CSI volumes or volumes belonging to other pods.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,51 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+			logger.Printf("Detected cloud volume %s -> %s (lun=%s)", volKey, m.Destination, volInfo["lun"])
+		}
+	}
+	if len(cloudVolumes) > 0 {
+		cvJSON, _ := json.Marshal(cloudVolumes)
+		if req.OCI.Annotations == nil {
+			req.OCI.Annotations = make(map[string]string)
+		}
+		req.OCI.Annotations["io.confidentialcontainers.org.cloud_volumes"] = string(cvJSON)
+		logger.Printf("Set cloud_volumes annotation: %s", string(cvJSON))
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. The `json.Marshal` error is now handled — on failure, a warning is logged and the annotation is not set.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -95,6 +101,59 @@ func (i *interceptor) CreateContainer(ctx context.Context, req *pb.CreateContain
...
+
+				// Rewrite mount source to point to the host mount point
+				for idx, m := range req.OCI.Mounts {
+					if m.Destination == mountPoint {
+						req.OCI.Mounts[idx].Source = hostMountPoint
+						req.OCI.Mounts[idx].Type = "bind"
+						logger.Printf("cloud volume %s: rewrote mount source to %s", volName, hostMountPoint)
+						break
+					}
+				}
```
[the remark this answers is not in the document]
mohammedadnan21: Fixed. The interceptor now tracks whether the mount rewrite succeeded and logs a warning when no matching mount destination is found.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,51 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 		logger.Printf("Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.")
 	}
 
+	// Detect cloud volumes from mountInfo.json and pass as annotation
+	cloudVolumes := make(map[string]map[string]string)
+	cloudVolIdx := 0
+	if len(req.OCI.Mounts) > 0 {
+		for _, m := range req.OCI.Mounts {
```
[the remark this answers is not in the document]
mohammedadnan21: Acknowledged. Unit tests for the cloud volume annotation logic will be added in a follow-up.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -94,6 +96,106 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+			mountDest := ""
+			for _, m := range req.OCI.Mounts {
+				if m.Source == decodedPath {
+					mountDest = m.Destination
+					break
+				}
+			}
+			if mountDest == "" {
+				logger.Printf("cloud volume disk %s has no matching mount in container spec, skipping", diskID)
+				continue
```
bpradipt, line 172: Shouldn't the `canonicalIdx` be incremented here?
Say the directory has three volumes in sorted order: vol-a, vol-b, vol-c.
- vol-a has oci mount, so lun index is 0
- vol-b doesn't have oci mount, it's skipped but canonicalIdx is still 0
- vol-c has oci mount, gets canonicalIdx as 1 which is wrong. It should be 2

You can test by creating such a scenario. Also important to have these test scenarios added.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -229,3 +313,254 @@ func (i *interceptor) DestroySandbox(ctx context.Context, req *pb.DestroySandbox
...
+	}
+	if lunIdx >= len(candidates) {
+		return "", fmt.Errorf("AWS EBS disk index %d out of range (have %d)", lunIdx, len(candidates))
+	}
+	target, err := filepath.EvalSymlinks(candidates[lunIdx])
+	if err != nil {
+		return "", err
+	}
+	logger.Printf("Found AWS EBS disk: %s -> %s", candidates[lunIdx], target)
+	return target, nil
```
bpradipt, line 517: This mapping LUN index to the NVME device seems brittle. Have you tried this on AWS?
Especially what happens when multiple device is added?

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
mohammedadnan21: Good catch. Fixed in 1be451c — `canonicalIdx` is now incremented before the `continue`, so the index stays aligned with `GetCSIVolumesForPod` even when a volume has no matching OCI mount.

Will add test coverage for this scenario in the follow-up unit tests PR.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
mohammedadnan21: You're right, this hasn't been tested on AWS yet. The current NVMe-to-LUN mapping via glob order is a placeholder — on AWS, NVMe device naming doesn't correspond to attachment order reliably, especially with multiple devices.

The proper approach for AWS would be to read the volume ID from the NVMe controller metadata (via `nvme id-ctrl /dev/nvmeXn1` which returns the EBS volume ID) and match against the disk ID in the cloud_volumes annotation, rather than relying on index-based matching.

For now, this is documented as a known limitation in the PR description. I'll address it properly when adding the AWS provider-side disk attachment code in a follow-up PR, along with end-to-end testing on EC2.

mohammedadnan21:
## Test Results

### Azure (AKS)
| Test | Result |
|------|--------|
| Single Disk Persistence | PASSED |
| Multi Disk Persistence | PASSED |

- Cluster: AKS with Kata + CAA + caa-csi-block-driver
- PodVM image: Custom mkosi image with udev rules and patched interceptor
- Verified: PVC create → attach to PeerPod → write data → delete pod → reattach → data persisted

### AWS (EKS)
| Test | Result |
|------|--------|
| Single Disk Persistence | PASSED |
| Multi Disk Persistence | PASSED |

- Cluster: EKS with Kata + CAA + caa-csi-block-driver
- Verified: EBS volume create → attach via NVMe passthrough → write data → delete pod → reattach → data persisted

### Unit Tests
```
go test ./pkg/adaptor/proxy/... → PASS
go test ./pkg/util/... → PASS
go test ./src/cloud-providers/azure/... → PASS
go test ./src/cloud-providers/aws/... → PASS
```

src/cloud-api-adaptor/podvm-mkosi/mkosi.postinst
```diff
@@ -0,0 +1,40 @@
...
+rm -f "${BUILDROOT}"/etc/systemd/system/{run-image,run-kata\x2dcontainers}.mount
+rm -f "${BUILDROOT}"/etc/systemd/system/multi-user.target.wants/{run-image,run-kata\x2dcontainers}.mount
+
+# mask unwanted sytemd units that measure a bunch of stuff into the vTPM
+# Set up /etc/resolv.conf symlink if not already present (Ubuntu only)
+if [ ! -e "${BUILDROOT}/etc/resolv.conf" ]; then
+    ln -s /run/systemd/resolve/stub-resolv.conf "${BUILDROOT}/etc/resolv.conf" || true
+fi
+
+# Install Azure storage udev rules so /dev/disk/azure/data/by-lun/N symlinks
```
bpradipt: No changes required for AWS ?

mohammedadnan21: No, AWS doesn't need udev rules here. EBS NVMe devices are enumerated automatically by the kernel NVMe driver and appear as /dev/nvmeXn1 with symlinks under /dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_*. The interceptor uses those symlinks directly for device detection. Only Azure needs the custom udev rules because Hyper-V SCSI disks don't get the /dev/disk/azure/ symlinks without them.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
mohammedadnan21: Addressed in the latest force-push. The AWS device detection now matches by EBS volume ID instead of relying on glob order. findAWSDataDisk normalizes the disk ID and matches it against the NVMe by-id symlink names. This is reliable for multi-disk because each volume has a unique ID regardless of attachment order. AttachEBSVolumes and DetachEBSVolumes have been added to the AWS provider with parallel attachment and consistent device naming. Unit tests cover single and multi-disk attachment, device naming, and volume ID matching.

mohammedadnan21:
Rebased on latest main and addressed review feedback (signal handler fix, data loss prevention, improved cloud detection, input validation, shared structs). All unit tests pass locally. Could a maintainer please approve the workflow runs? Thanks!

mohammedadnan21:
The s390x e2e failures are unrelated to this PR — cert-manager fails to deploy on the s390x runner because kube-controller-manager is in Error state. All other 25 checks pass. Could someone please re-run the failed jobs?

bpradipt:
@mohammedadnan21 sorry this will need a rebase

bpradipt:
I tested this on AWS and found the following issues

- The CAA container has no hostPath mount for /run/kata-containers/. It can't see the mountInfo.json files written by the CSI driver on the host. You need to add to the DaemonSet template:
```
  volumes:
  - name: kata-direct-volumes
    hostPath:
      path: /run/kata-containers/shared/direct-volumes
      type: DirectoryOrCreate
  containers:
  - volumeMounts:
    - name: kata-direct-volumes
      mountPath: /run/kata-containers/shared/direct-volumes
```

- VM and the volume must be in the same AZ otherwise attachment will not work. Thus must be included in a doc. Probably as a comment in code and also in https://github.com/confidential-devhub/caa-csi-block-driver

mohammedadnan21:
Made the kata-direct-volumes-dir hostPath mount unconditional in the Helm chart DaemonSet 
Added AZ requirement documentation on attachEBSVolumes.

bpradipt (approved):
/lgtm
Thanks @mohammedadnan21. This is a much needed attention to enable proper CSI support for peer-pods.

bpradipt:
@snir911 @stevenhorsman @mkulke ptal

bpradipt:
@mohammedadnan21 can you please resolve the conflict and update this PR. With latest removal of docker provider, the govulncheck should be green as well

mkulke:
I think I need more context to make an assessment on this PR, maybe a demo + q&a on a coco community call?

From a code health perspective, we probably want to be conversative when it comes to merging such large comprehensive, features as they incur a real maintenance burden. We have to consider aspect like e2e test coverage and long-term ownership.

On a high-level, it's a bit concerning that this is coupled to on an out-of-tree component, especially if that component is CAA specific. I do like the idea of experimenting with new ideas out-ot-tree though, to learn how useful they prove to be in practice before integrating them into CAA and become a single, sanctioned solution.

So maybe we should be thinking about providing integration/extension points in CAA instead? What abstractions would we have to provide for something like this PR if it was developed out-of-tree?

mohammedadnan21:
> I think I need more context to make an assessment on this PR, maybe a demo + q&a on a coco community call?
> 
> From a code health perspective, we probably want to be conversative when it comes to merging such large comprehensive, features as they incur a real maintenance burden. We have to consider aspect like e2e test coverage and long-term ownership.
> 
> On a high-level, it's a bit concerning that this is coupled to on an out-of-tree component, especially if that component is CAA specific. I do like the idea of experimenting with new ideas out-ot-tree though, to learn how useful they prove to be in practice before integrating them into CAA and become a single, sanctioned solution.
> 
> So maybe we should be thinking about providing integration/extension points in CAA instead? What abstractions would we have to provide for something like this PR if it was developed out-of-tree?

I presented this at the May 20 community call — here's the meeting agenda: https://docs.google.com/document/d/1QtiOpSavz177Nq3jwzjQ5vvZjquOT7tFs9XBplNeC-o/edit?tab=t.0 and the slides: https://mohammedadnan21.github.io/volume-passthrough-presentation/

@bpradipt @stevenhorsman could you share the recording link?

Happy to do a follow-up Q&A at a future call if needed. On the extension point question — the CAA-side changes are intentionally minimal (reading mountInfo.json from the existing Kata direct-volumes path + a single annotation). The caa-csi-block-driver is out-of-tree by design so it can iterate independently without impacting CAA's core.

stevenhorsman:
> @bpradipt @stevenhorsman could you share the recording link?

I don't have access to the recordings I'm afraid.

bpradipt:
> > @bpradipt @stevenhorsman could you share the recording link?
> 
> I don't have access to the recordings I'm afraid.

I'm checking..

src/cloud-providers/azure/provider.go
```diff
@@ -412,6 +431,201 @@ func (p *azureProvider) updateInstanceSizeSpecList() error {
...
+	}
+
+	logger.Printf("WARNING: could not find max disk count for VM size %s, skipping validation", instanceSize)
+	return nil
+}
+
+// validateDiskZones checks that all CSI volumes are in the same region as the
+// target VM. Managed disks are regional (not zonal) unless they were explicitly
+// created in a zone, so we log a warning if zones are set.
+func (p *azureProvider) validateDiskZones(ctx context.Context, volumes []provider.CloudVolume) error {
```
mkulke: the fn name (and comment) is not entirely accurate, we also check for existence and attachment state here, i'd call it validateDisk()

src/cloud-providers/azure/provider.go
```diff
@@ -283,7 +286,23 @@ func (p *azureProvider) CreateInstance(ctx context.Context, podName, sandboxID s
 		imageID = spec.Image
 	}
 
-	vmParameters, err := p.getVMParameters(instanceSize, diskName, cloudConfigData, sshBytes, instanceName, nicName, imageID)
+	if len(spec.Volumes) > 0 {
+		if err := p.validateDiskCount(ctx, instanceSize, len(spec.Volumes)); err != nil {
+			return nil, err
+		}
+		detachErr := p.forceDetachStaleDisks(ctx, spec.Volumes)
```
mkulke: we probably don't want to do this kind of self-healing here, implicitly at instance creation. it obscures a real problem with leaking virtual machines. If the disk is not 100% in the state that we expect it to be, VM creation should fail, IMO. There are other components (peerpod-ctrl) that attempt to garbage collect leftover resources.

src/cloud-providers/aws/provider.go
```diff
@@ -339,6 +351,12 @@ func (p *awsProvider) CreateInstance(ctx context.Context, podName, sandboxID str
 		ips[0] = publicIPAddr
 	}
 
+	if len(spec.Volumes) > 0 {
+		if err := p.attachEBSVolumes(ctx, instanceID, spec.Volumes); err != nil {
```
mkulke, line 358: in the azure provider we perform some sanity checks before creating the VM, should similar logic be in the AWS provider?

mkulke:
> I presented this at the May 20 community call — here's the meeting agenda: https://docs.google.com/document/d/1QtiOpSavz177Nq3jwzjQ5vvZjquOT7tFs9XBplNeC-o/edit?tab=t.0 and the slides: https://mohammedadnan21.github.io/volume-passthrough-presentation/
> 
> @bpradipt @stevenhorsman could you share the recording link?
> 
> Happy to do a follow-up Q&A at a future call if needed. On the extension point question — the CAA-side changes are intentionally minimal (reading mountInfo.json from the existing Kata direct-volumes path + a single annotation). The caa-csi-block-driver is out-of-tree by design so it can iterate independently without impacting CAA's core.

thanks 👍  the presentation provided a lot of necessary context. I understand this is built on an existing facility in kata containers, so we implement a functionality in CAA that existed in kata before and attempt to map it to cloud providers, is that correct?

mohammedadnan21:
> > I presented this at the May 20 community call — here's the meeting agenda: https://docs.google.com/document/d/1QtiOpSavz177Nq3jwzjQ5vvZjquOT7tFs9XBplNeC-o/edit?tab=t.0 and the slides: https://mohammedadnan21.github.io/volume-passthrough-presentation/
> > @bpradipt @stevenhorsman could you share the recording link?
> > Happy to do a follow-up Q&A at a future call if needed. On the extension point question — the CAA-side changes are intentionally minimal (reading mountInfo.json from the existing Kata direct-volumes path + a single annotation). The caa-csi-block-driver is out-of-tree by design so it can iterate independently without impacting CAA's core.
> 
> thanks 👍 the presentation provided a lot of necessary context. I understand this is built on an existing facility in kata containers, so we implement a functionality in CAA that existed in kata before and attempt to map it to cloud providers, is that correct?

Yes, exactly. Kata Containers already defines the direct-volumes protocol (mountInfo.json at a well-known path) for block device passthrough. This PR implements the CAA side — reading that metadata, attaching cloud disks (EBS/Managed Disks) to the PodVM, and mounting them inside the guest. The CSI driver (out-of-tree) is the component that writes the mountInfo.json, bridging the Kubernetes PVC workflow with the existing Kata protocol.

src/cloud-providers/azure/provider.go
mohammedadnan21: Good point — renamed to validateDisks and updated the comment to reflect that it checks existence, attachment state, and region.

src/cloud-providers/azure/provider.go
mohammedadnan21: Good point — removed the force-detach logic from both Azure and AWS providers. If a volume is still attached to a stale VM, creation now fails immediately. Cleanup of orphaned VMs is handled by peerpod-ctrl anyway, so self-healing here just hides real issues.

mkulke (approved):

src/cloud-providers/aws/provider.go
mohammedadnan21: Good call — added a validateVolumes step in the AWS provider that runs before instance creation. It calls DescribeVolumes to check all requested volumes exist and are in available state. If any volume is still attached, creation fails immediately with a clear error.

bpradipt:
@mohammedadnan21 you will need to rebase and remove the merge commit. I updated the PR which added the merge commits to run the tests..

bpradipt:
All the tests passed - https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/26933165654?pr=3037
But looks like you'll need another rebase due to merging of other PRs. Sorry about it..

Please rebase and push and again, hopefully this time we can merge it.

cc @stevenhorsman @mkulke

PR #3038, opened 2026-05-05
libvirt: validate required provider configuration earlier

### LINKED ISSUE #3024: libvirt: validate required provider configuration earlier
## Problem

The libvirt provider currently accepts some incomplete or inconsistent configuration at startup and only fails later when the first PodVM is created or deleted.

Today, [`ConfigVerifier()`](src/cloud-providers/libvirt/provider.go) only checks whether `VolName` is non-empty. Other required settings are effectively validated much later by runtime operations in [`NewProvider()`](src/cloud-providers/libvirt/provider.go), [`CreateInstance()`](src/cloud-providers/libvirt/provider.go), or lower-level helpers in [`libvirt.go`](src/cloud-providers/libvirt/libvirt.go).

That means operator mistakes can survive initialization and show up only as delayed runtime failures, which makes troubleshooting harder than necessary.

## Examples of the gap

Some examples of configuration that should likely fail fast:

- missing or empty libvirt URI
- missing storage pool / network configuration required by the selected flow
- invalid memory / vCPU defaults
- inconsistent launch-security settings
- image / volume settings that are known to be unusable before any VM is created

## Proposed change

Expand [`ConfigVerifier()`](src/cloud-providers/libvirt/provider.go) so the libvirt provider validates more of its required configuration during startup rather than waiting for first use.

Suggested scope:

- keep the change limited to validation and error reporting
- return clear, actionable error messages for missing or invalid fields
- avoid changing successful runtime behavior for already-valid configurations
- add focused unit tests for the newly validated cases

## Expected benefit

This should make libvirt deployments fail fast with clearer errors, reduce time spent debugging first-PodVM failures, and bring libvirt closer to the stronger upfront validation patterns used by some other providers.

ajaypvictor:
Extend ConfigVerifier() to fail fast on missing required libvirt settings and invalid default CPU or memory values.

Add focused unit tests for the new validation paths.

Fixes: #3024

### DIFF
--- a/src/cloud-providers/libvirt/libvirt_test.go
+++ b/src/cloud-providers/libvirt/libvirt_test.go
@@ -207,3 +207,99 @@ func TestGetDeletableDiskPaths(t *testing.T) {
 		})
 	}
 }
+
+func TestConfigVerifier(t *testing.T) {
+	newConfig := func(overrides func(*Config)) *Config {
+		cfg := &Config{
+			URI:         "qemu:///system",
+			PoolName:    "default",
+			NetworkName: "default",
+			VolName:     "podvm-base.qcow2",
+			CPU:         1,
+			Memory:      512,
+		}
+		if overrides != nil {
+			overrides(cfg)
+		}
+		return cfg
+	}
+
+	tests := []struct {
+		name          string
+		provider      *libvirtProvider
+		expectedError string
+	}{
+		{
+			name: "empty URI fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.URI = ""
+				}),
+			},
+			expectedError: "URI is empty",
+		},
+		{
+			name: "empty pool name fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.PoolName = ""
+				}),
+			},
+			expectedError: "PoolName is empty",
+		},
+		{
+			name: "empty network name fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.NetworkName = ""
+				}),
+			},
+			expectedError: "NetworkName is empty",
+		},
+		{
+			name: "empty volume name fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.VolName = ""
+				}),
+			},
+			expectedError: "VolName is empty",
+		},
+		{
+			name: "zero CPU fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.CPU = 0
+				}),
+			},
+			expectedError: "CPU must be greater than zero",
+		},
+		{
+			name: "zero memory fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.Memory = 0
+				}),
+			},
+			expectedError: "Memory must be greater than zero",
+		},
+		{
+			name: "valid config passes",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(nil),
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.provider.ConfigVerifier()
+			if tc.expectedError == "" {
+				assert.NoError(t, err)
+				return
+			}
+
+			assert.EqualError(t, err, tc.expectedError)
+		})
+	}
+}
--- a/src/cloud-providers/libvirt/provider.go
+++ b/src/cloud-providers/libvirt/provider.go
@@ -158,9 +158,31 @@ func (p *libvirtProvider) Teardown() error {
 }
 
 func (p *libvirtProvider) ConfigVerifier() error {
-	VolName := p.serviceConfig.VolName
-	if len(VolName) == 0 {
+	config := p.serviceConfig
+
+	if config.URI == "" {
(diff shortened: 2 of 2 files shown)


src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -207,3 +207,128 @@ func TestGetDeletableDiskPaths(t *testing.T) {
...
+			err := tc.provider.ConfigVerifier()
+			if tc.expectedError == "" {
+				assert.NoError(t, err)
+				return
+			}
+
+			assert.EqualError(t, err, tc.expectedError)
+		})
+	}
+}
```
stevenhorsman, line 305: Can you reduce the duplication, but only specifying what is different from default with something like (disclaimer generated-by Bob at my prompting):
```
func TestConfigVerifier(t *testing.T) {
	// Helper function to create a valid config with optional overrides
	newConfig := func(overrides func(*Config)) *Config {
		cfg := &Config{
			URI:         "qemu:///system",
			PoolName:    "default",
			NetworkName: "default",
			VolName:     "podvm-base.qcow2",
			CPU:         1,
			Memory:      512,
		}
		if overrides != nil {
			overrides(cfg)
		}
		return cfg
	}

	tests := []struct {
		name          string
		provider      *libvirtProvider
		expectedError string
	}{
		{
			name:          "nil provider config fails",
			provider:      &libvirtProvider{},
			expectedError: "libvirt config is nil",
		},
		{
			name: "empty URI fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.URI = ""
				}),
			},
			expectedError: "URI is empty",
		},
		{
			name: "empty pool name fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.PoolName = ""
				}),
			},
			expectedError: "PoolName is empty",
		},
		{
			name: "empty network name fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.NetworkName = ""
				}),
			},
			expectedError: "NetworkName is empty",
		},
		{
			name: "empty volume name fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.VolName = ""
				}),
			},
			expectedError: "VolName is empty",
		},
		{
			name: "zero CPU fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.CPU = 0
				}),
			},
			expectedError: "CPU must be greater than zero",
		},
		{
			name: "zero memory fails",
			provider: &libvirtProvider{
				serviceConfig: newConfig(func(c *Config) {
					c.Memory = 0
				}),
			},
			expectedError: "Memory must be greater than zero",
		},
		{
			name: "valid config passes",
			provider: &libvirtProvider{
				serviceConfig: newConfig(nil),
			},
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			err := tc.provider.ConfigVerifier()
			if tc.expectedError == "" {
				assert.NoError(t, err)
				return
			}

			assert.EqualError(t, err, tc.expectedError)
		})
	}
}
```

src/cloud-providers/libvirt/provider.go
```diff
@@ -158,9 +158,35 @@ func (p *libvirtProvider) Teardown() error {
 }
 
 func (p *libvirtProvider) ConfigVerifier() error {
-	VolName := p.serviceConfig.VolName
-	if len(VolName) == 0 {
+	if p == nil || p.serviceConfig == nil {
```
stevenhorsman: I don't think we need this defensive code as if p was nil, when `p.ConfigVerifier()` was called it would have already failed?

stevenhorsman (commented):
A couple of suggestions. As understanding of the background info - are the provider files somehow overwriting the defaults which can result in the config values being empty/0?

src/cloud-providers/libvirt/libvirt_test.go
ajaypvictor: this looks better, updated.. thanks..

src/cloud-providers/libvirt/provider.go
ajaypvictor: makes sense, thanks for pointing it out, removed the nil check and deleted the respective testcase too..

stevenhorsman (commented):
Can you update the commit to append `Assisted-by: IBM Bob <noreply@ibm.com>` in line with the Linux Foundations best practices on AI-attribution. I know that the email doesn't necessarily make sense, but it looks like it's to ensure compliance with tooling. I've never used it so far and not hit issues, but just learnt about it now.

ajaypvictor:
> Can you update the commit to append `Assisted-by: IBM Bob <noreply@ibm.com>` in line with the Linux Foundations best practices on AI-attribution. I know that the email doesn't necessarily make sense, but it looks like it's to ensure compliance with tooling. I've never used it so far and not hit issues, but just learnt about it now.

done, thanks for pointing it out..

stevenhorsman (approved):
LGTM. Thanks

wainersm:
Hi @ajaypvictor ,

This works but doesn't fix a major problem which is: flags definition are handled three places:  `manager.go`, helm charts and `ConfigVerifier()`. We have already a mechanism (`make sync-chart-values` + workflow) to keep manager.go and helm in sync, but `ConfigVerifier()` don't. I've a half-baked solution that I could be sharing but first I will discuss a little bit with @beraldoleal (he has worked to improve that part of the code). So in meanwhile I am okay to have your fix in.

wainersm (approved):
Thanks @ajaypvictor !

PR #3043, opened 2026-05-07
Switch AWS CI to use mkosi podvm image

wainersm:
The packer-based podvm images are deprecated and will be removed eventually. The AWS CI still use those images and this PR is about making the switch to mkosi. Because life is not easy, I had to make some adjustments here and there for tests to continue passing. Most notably, mkosi image needs UEFI boot mode and currently we are using legacy mode.

Tested locally a single test case and ran full CI in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/25454939108/job/74684191291

### DIFF
--- a/.github/workflows/e2e_run_all.yaml
+++ b/.github/workflows/e2e_run_all.yaml
@@ -192,18 +192,12 @@ jobs:
       github.event_name == 'schedule' ||
       github.event_name == 'workflow_dispatch' ||
       contains(github.event.pull_request.labels.*.name, 'test_e2e_aws')
-    needs: [podvm, caa_image_amd64]
+    needs: [podvm_ubuntu_amd64, caa_image_amd64]
     strategy:
       fail-fast: false
       matrix:
         container_runtime:
           - crio
-        os:
-          - ubuntu
-        provider:
-          - generic
-        arch:
-          - amd64
     permissions:
       id-token: write # Required by aws-actions/configure-aws-credentials
       contents: read  # Required by aws-actions/configure-aws-credentials
@@ -211,9 +205,8 @@ jobs:
     with:
       caa_image: ${{ inputs.registry }}/cloud-api-adaptor:${{ inputs.caa_image_tag }}-dev-amd64
       container_runtime: ${{ matrix.container_runtime }}
-      podvm_image: ${{ inputs.registry }}/podvm-${{ matrix.provider }}-${{ matrix.os }}-${{ matrix.arch }}:${{ inputs.podvm_image_tag }}
+      podvm_image: ${{ needs.podvm_ubuntu_amd64.outputs.qcow2_oras_image }}
       git_ref: ${{ inputs.git_ref }}
-      oras: false
     secrets:
       AWS_IAM_ROLE_ARN: ${{ secrets.AWS_IAM_ROLE_ARN }}
 
--- a/src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
+++ b/src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
@@ -68,6 +68,7 @@ type Vpc struct {
 	Client            *ec2.Client
 	ID                string
 	InternetGatewayID string
+	PodvmInstanceType string
 	Region            string
 	RouteTableID      string
 	SecurityGroupID   string
@@ -323,7 +324,7 @@ func (a *AWSProvisioner) GetProperties(ctx context.Context, cfg *envconf.Config)
 		"pause_image":          a.PauseImage,
 		"podvm_launchtemplate": "",
 		"podvm_ami":            a.Image.ID,
-		"podvm_instance_type":  "t3.medium",
+		"podvm_instance_type":  a.Vpc.PodvmInstanceType,
 		"sg_ids":               a.Vpc.SecurityGroupID, // TODO: what other SG needed?
 		"subnet_id":            a.Vpc.SubnetID,
 		"ssh_kp_name":          a.SSHKpName,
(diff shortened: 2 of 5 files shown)


wainersm:
@stevenhorsman could you send the link to your branch that contain similar changes? I lost it, sorry, but I would like to reuse some changes you did in the workflow.

stevenhorsman:
> @stevenhorsman could you send the link to your branch that contain similar changes? I lost it, sorry, but I would like to reuse some changes you did in the workflow.

Here is the branch: https://github.com/confidential-containers/cloud-api-adaptor/tree/e2e-aws-switch-to-mkosi take, or leave any thing you like 😁

.github/workflows/e2e_run_all.yaml
```diff
@@ -192,28 +192,22 @@ jobs:
...
       contents: read  # Required by aws-actions/configure-aws-credentials
     uses: ./.github/workflows/e2e_aws.yaml
     with:
       caa_image: ${{ inputs.registry }}/cloud-api-adaptor:${{ inputs.caa_image_tag }}-dev-amd64
       container_runtime: ${{ matrix.container_runtime }}
-      podvm_image: ${{ inputs.registry }}/podvm-${{ matrix.provider }}-${{ matrix.os }}-${{ matrix.arch }}:${{ inputs.podvm_image_tag }}
+      podvm_image: ${{ needs.podvm_ubuntu_amd64.outputs.qcow2_oras_image }}
       git_ref: ${{ inputs.git_ref }}
-      oras: false
+      oras: true
```
stevenhorsman: Can you update the aws workflow to not take in the oras input as it's not needed now.

stevenhorsman (commented):
I guess this is a breaking change as UEFI and the different instance type is needed, so we should remember to mention it in the release notes

.github/workflows/e2e_run_all.yaml
wainersm: done

wainersm:
Hi @stevenhorsman ,

I addressed your comments but when I run the workflow again the tests started all to fail. Maybe because I rebased and now it is getting the new version of mkosi? Don't know...so set this PR to draft so I can have a look.

stevenhorsman (approved):
Code looks good to me if the CI can pass.

src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
```diff
@@ -1017,9 +1069,11 @@ func (i *AMIImage) registerImage(imageName string) error {
 				SnapshotId:          aws.String(i.EBSSnapshotID),
 			},
 		}},
+		BootMode:           ec2types.BootModeValuesUefi,
 		Description:        aws.String(i.Description),
 		EnaSupport:         aws.Bool(true),
 		RootDeviceName:     aws.String(i.RootDeviceName),
+		TpmSupport:         ec2types.TpmSupportValuesV20,
```
mkulke, line 1076: is this TPM required for UEFI on AWS or would it be used for attestation?

mkulke (approved):

src/cloud-api-adaptor/test/provisioner/aws/provision_common.go
wainersm: Hi @mkulke , it would be used for attestation only. It is not strictly need for this PR... If another PR for attestation on aws that will need it.

wainersm:
> Hi @stevenhorsman ,
> 
> I addressed your comments but when I run the workflow again the tests started all to fail. Maybe because I rebased and now it is getting the new version of mkosi? Don't know...so set this PR to draft so I can have a look.

Bingo! I reverted https://github.com/confidential-containers/cloud-api-adaptor/pull/3023 and tests passed. I will manage to find some time to debug it. We were close to get it switched (/me cry).

wainersm:
Ok, I've found the problem: it was missing the NVMe kernel drivers in the generated image. I pushed a new commit to fix in mkosi and re-ran the tests (https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/25823406255/job/75872503719).

@mkulke according to Claude mkosi image is missing the drivers for Hyper-v. is the Azure CI using the Ubuntu mkosi image too?

wainersm:
> Ok, I've found the problem: it was missing the NVMe kernel drivers in the generated image. I pushed a new commit to fix in mkosi and re-ran the tests (https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/25823406255/job/75872503719).
>

FYI @BbolroC ^^

It seems that I got lucky :)

> @mkulke according to Claude mkosi image is missing the drivers for Hyper-v. is the Azure CI using the Ubuntu mkosi image too?

mkulke:
> @mkulke according to Claude mkosi image is missing the drivers for Hyper-v. is the Azure CI using the Ubuntu mkosi image too?

The azure image is still built on fedora. I wanted to wait until the migration for libvirt is stabilized before switching the azure images to ubuntu to avoid working on a moving target.

would you say this a good point to consider switching or should we remove the fedora podvm support first? (the azure fedora podvm images are currently broken anyway, afaict)

wainersm:
> > @mkulke according to Claude mkosi image is missing the drivers for Hyper-v. is the Azure CI using the Ubuntu mkosi image too?
> 
> The azure image is still built on fedora. I wanted to wait until the migration for libvirt is stabilized before switching the azure images to ubuntu to avoid working on a moving target.
> 

Fair. I worked this PR in a moving target (packer -> Ubuntu mkosi) and it was not so nice.

> would you say this a good point to consider switching or should we remove the fedora podvm support first? (the azure fedora podvm images are currently broken anyway, afaict)

afaict libvirt CI with Ubuntu images is stable now. Having this PR merged we will have it more exercised. Maybe in one week from now, given that aws CI keeps stable, I would make the switch & remove fedora completely.

mkulke:
> afaict libvirt CI with Ubuntu images is stable now. Having this PR merged we will have it more exercised. Maybe in one week from now, given that aws CI keeps stable, I would make the switch & remove fedora completely.

that makes sense, I'll take a look then.

PR #3054, opened 2026-05-11
cloud-api-adaptor: fix unpacking image error

### LINKED ISSUE #3033: GCP: Error unpacking image, failed to extract layer sha256:82c60ccaf916322916d16bcdb4223f93acc1f68e2087dba4ddf64990b1dc27fb
### Describe the bug

The error we got
```
spec.initContainers{_service_}: Error: failed to create containerd container: error unpacking image: apply layer error for "_place_holder_for_our_container_image_": 
failed to extract layer sha256:82c60ccaf916322916d16bcdb4223f93acc1f68e2087dba4ddf64990b1dc27fb: 
failed to get reader from content store: content digest sha256:fa8ae93e2b3a7478248483e942ff665efa7219c6cd72d7a03c775372076e98dc: not found
```

We had a patched version of `kata-deploy` which adds `discard_unpacked_layers = false` to the drop-in config file(`/opt/kata/containerd/config.d/kata-deploy.toml`) deployed on GKE, the initial test looked promising with images like ubuntu, nginx, postgres, alpine, etc. However,  we saw the above error when deploying our own images.

After some debugging, I narrowed the problem down to `distroless` based images. Why `distroless`? That seems arbitrary and mysterious. We spent more time, and the answer is peer-pods-webhook https://github.com/confidential-containers/cloud-api-adaptor/blob/ab0b4f93d11a9de586642f220abf81a597915f71/src/webhook/Dockerfile#L26 It is based on distroless which shares some common layers with our image.

So basically kata-deploy and peer-pods-webhook are deployed to a node simultaneously, the latter is much smaller which means that its image can get pulled and unpacked first, before kata-deploy updates the containerd config and restarts it. That is a problem because image layers can get discarded since the `discard_unpacked_layers` is still true.

More broadly speaking, this is not about peer-pods-webhook. If there are other damonsets in the GKE cluster, or CoCo components are deployed to a long standing node, it is possible that there are images pulled already and some of their layers have been discarded before kata-deploy or we manually update(s) the containerd config. We'd like to get some guidance here, thank you.

### How to reproduce

Deploy CoCo `v0.19.0` to a GKE cluster. 
```
# ssh into the worker node
gcloud compute ssh ${vm_name} --project={your_project} --zone={zone_name}

sudo su - root
cd /var/lib/containerd/io.containerd.content.v1.content/blobs/sha256

# Grab the image digest and get the digest for amd64 linux
# ctr -n k8s.io image ls | grep peer-pods-webhook
# ctr -n k8s.io image ls | grep distroless

# 6f3f2123de90d2e7998b8161a2838433ec32560a827d07bcab339dacbf0cf16f is the digest for distroless image
cat 6f3f2123de90d2e7998b8161a2838433ec32560a827d07bcab339dacbf0cf16f  |jq '.layers [].digest' | tr -d '"' | awk -F':' '{print $2}' | xargs ls -l
ls: cannot access 'fa8ae93e2b3a7478248483e942ff665efa7219c6cd72d7a03c775372076e98dc': No such file or directory
ls: cannot access 'c172f21841dff4c8cf45cde46589c1c2616cefe7e819965e92e6d3475c428aa0': No such file or directory
ls: cannot access 'b4e6f1bfce0a1fba2b5421041552f4a897aada9cd5680926580f9e2c6247a7ae': No such file or directory
ls: cannot access 'b4242723c53fe4e094eb78569a2c15b6aafb8eb42aa9c3c2666130654a316ae2': No such file or directory
ls: cannot access 'd6b1b89eccacc15c2420b2776d72c1dae334a00805ed9af54bf2f71e4d536f28': No such file or directory
ls: cannot access 'b839dfae01f66e15c6a8b63520557ed315bdfe036342fa7a0c537259f10d7a9a': No such file or directory
-r--r--r-- 1 root root     67 Apr 29 22:32 2780920e5dbfbe103d03a583ed75345306e572ec5a48cb10361f046767d9f29a
-r--r--r-- 1 root root    123 Apr 29 22:32 3214acf345c0cc6bbdb56b698a41ccdefc624a09d6beb0d38b5de0b2303ecaf4
-r--r--r-- 1 root root    162 Apr 29 22:32 52630fc75a18675c530ed9eba5f55eca09b03e91bd5bc15307918bbc1a7e7296
-r--r--r-- 1 root root    188 Apr 29 22:32 7c12895b777bcaa8ccae0605b4de635b68fc32d60fa08f421dc3818bf55ee212
-r--r--r-- 1 root root 136993 Apr 29 22:32 bdfd7f7e5bf6fc27e70b59101db21c3d8284d283884419dd5fe7020583bb79ca
-r--r--r-- 1 root root     80 Apr 29 22:32 dd64bf2dd177757451a98fcdc999a339c35dee5d9872d8f4dc69c8f3c4dd0112
-r--r--r-- 1 root root    311 Apr 29 22:32 ebddc55facdc6b1f7e0f30816a5fc7cc62f38abdf76c0a8b0a0ce52085754795
```

We can see some of the layers are already gone. This particular layer `fa8ae93e2b3a7478248483e942ff665efa7219c6cd72d7a03c775372076e98dc` is the one we saw in the error message mentioned in the beginning.

### CoCo version information

CoCo v0.19.0

### What TEE are you seeing the problem on

Tdx

### Failing command and relevant log output

```shell

```

Qiang-Xu:
Fixes #3033 

This PR addresses the error `unpacking image` by disabling `nydus` for CAA.

We've seen this error on both Azure and GCP(even after the `discard_unpacked_layers` is set to false).  Run 
```
ctr -n k8s.io image check
```
to see whether there are incomplete images that have been unpacked.

The code does two things
- Use `overlayfs` as the snapshotter for `remote` runtime.
- Add a storage to `CreateContainerRequest` which carries the image information that kata agent will pick up and use for pulling the image.

A image is not required to be complete any more(e.g, `discard_unpacked_layers=true` works fine) because nydus is not used.

We've tested the changes on both Azure and GCP with incomplete images, it's been working well so far.

### DIFF
--- a/src/cloud-api-adaptor/install/charts/peerpods/values.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/values.yaml
@@ -179,5 +179,11 @@ kata-deploy:
       enabled: false
     qemu-coco-dev-runtime-rs:
       enabled: false
+    qemu-snp-runtime-rs:
+      enabled: false
+    qemu-tdx-runtime-rs:
+      enabled: false
     remote:
       enabled: true
+      containerd:
+        snapshotter: "overlayfs"
--- a/src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
+++ b/src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
@@ -6,6 +6,8 @@ package proxy
 import (
 	"context"
 	b64 "encoding/base64"
+	"encoding/json"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
@@ -13,6 +15,7 @@ import (
 
 	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/pkg/util/agentproto"
 	pb "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
+	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/types"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
@@ -91,7 +94,24 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
 	}
 
 	if !pullImageInGuest {
-		logger.Printf("Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.")
+		// There is some issue with nydus(error unpacking image) when the image layers are missing due to
+		//  - discard_unpacked_layers set to true
+		//  - other reasons we don't know yet
+		// Run: ctr -n k8s.io image check
+		// to see whether the image is complete(all layers are present)
+		//
+		// nydus adds one mount that carries the image information which is then picked up
+		// by kata shim, then kata shim passes it to kata agent in the PodVM. Without nydus, we
+		// have to add the mount point manually.
+		vol, err := handleVirtualVolumeStorageObject(req)
+		if err != nil {
+			return nil, err
+		}
+
+		req.Storages = append(req.Storages, vol)
+		s := req.Storages[len(req.Storages)-1]
+		logger.Print("    storages added for guest_image_pull:")
+		logger.Printf("        mount_point:%s source:%s fstype:%s driver:%s", s.MountPoint, s.Source, s.Fstype, s.Driver)
 	}
 
 	res, err := s.Redirector.CreateContainer(ctx, req)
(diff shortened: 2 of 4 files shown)


bpradipt:
@Qiang-Xu can you please look into the test failures.

Qiang-Xu:
> @Qiang-Xu can you please look into the test failures.

Just saw it, I will take a look, thanks.

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
```diff
@@ -179,5 +179,11 @@ kata-deploy:
...
     qemu-coco-dev-runtime-rs:
       enabled: false
+    qemu-snp-runtime-rs:
+      enabled: false
+    qemu-tdx-runtime-rs:
+      enabled: false
     remote:
       enabled: true
+      containerd:
+        snapshotter: "overlayfs"
```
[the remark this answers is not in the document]
Qiang-Xu: `overlayfs` is only used for `remote` runtime, `nydus` is still the default for other runtimes.

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -91,7 +94,24 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
+		vol, err := handleVirtualVolumeStorageObject(req)
+		if err != nil {
+			return nil, err
+		}
+
+		req.Storages = append(req.Storages, vol)
+		s := req.Storages[len(req.Storages)-1]
+		logger.Print("    storages added for guest_image_pull:")
+		logger.Printf("        mount_point:%s source:%s fstype:%s driver:%s", s.MountPoint, s.Source, s.Fstype, s.Driver)
 	}
```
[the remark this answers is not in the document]
Qiang-Xu: The current test actually covers this branch because no storage is included, `pullImageInGuest` is always false.

Qiang-Xu:
@bpradipt I've fixed the test and addressed the copilot review comments, please take a look again, thank you.

beraldoleal (commented):
This looks valid. @wainersm @stevenhorsman do you guys see any problem with using overlayfs here?

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
```diff
@@ -91,7 +94,24 @@ func (s *proxyService) CreateContainer(ctx context.Context, req *pb.CreateContai
...
 
 	if !pullImageInGuest {
-		logger.Printf("Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.")
+		// There is some issue with nydus(error unpacking image) when the image layers are missing due to
```
bpradipt, line 97: @Qiang-Xu Overall looks good to me.
I'm wondering if it makes sense to have this change in Kata itself for the remote hypervisor?
Or that's not a technical possibility ?

cc @stevenhorsman @wainersm @beraldoleal wdyt ?

stevenhorsman:
> This looks valid. @wainersm @stevenhorsman do you guys see any problem with using overlayfs here?

Sorry for being late to see this. Is the proposal is to skip using the nydus-snapshotter? Is that in general, or just if there are issues? The negative with not using nydus is you get the host pull as well as guest pull, so things like encrypted images aren't supported. I know nydus has been made more reliable recently, so I'm not sure if that would help, but if not I would suggest looking into the `experimental-force-guest-pull` option in kata containers as that set the `image_guest_pull` in storage.driver with the image without using nydus, and if we can do that from the kata side rather than duplicating that feature here it seems like it might be a better approach?

src/cloud-api-adaptor/pkg/adaptor/proxy/service.go
stevenhorsman: I think this is just to use `experimental-force-guest-pull` in the kata configuration?

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
stevenhorsman, line 189: We would lose support for encryption image support here by using overlayfs?

stevenhorsman (commented):
I don't think that hard-coding overlayfs is the right approach for peer pods as it breaks a strong part of confidential containers in the ability to support encrypted images, please consider using experimental-force-guest-pull on the kata config if you want to skip nydus.

Qiang-Xu:
Thanks guys for the feedback, I'll close the PR for now.

PR #3055, opened 2026-05-12
adaptor: add sandbox persistence and recovery across CAA restarts

thejasn:
## Problem

CAA stores all sandbox state in an in-memory map. When the CAA pod restarts, this state is lost. The new CAA instance has no record of existing sandboxes and does not recreate the agent proxy. The shim's ttrpc connection to agent.ttrpc goes stale, the shim's health monitor detects the broken connection and exits, and the pod restarts. A routine CAA daemonset update disrupts every peer pod workload on the node.

## Solution

This PR persists sandbox state to disk so that a restarted CAA can reconnect to existing pod VMs instead of destroying them. After recovery, the shim's ttrpc requests resume flowing through the new agent proxy to the original kata-agent — containers continue running with zero restarts.

State is saved at each lifecycle transition: CreateVM persists initial metadata, StartVM updates instance info and network config, StopVM deletes state. On startup, CAA scans `/run/peerpod/pods/*` for state files and reconnects to sandboxes that were in the Running phase. Sandboxes in any other phase are cleaned up.

### Additional Changes

Several subsystems assumed fresh-start semantics and needed fixes:

- TLS material: ephemeral CA + client keys regenerated on each start break mTLS to existing forwarders. Persist on first run, reload on subsequent starts.
- VXLAN pod index: counter resets to zero causing ID collisions with recovered sandboxes. Advance past all restored indices before creating new sandboxes.
- Network teardown: extracted VXLAN params from pod netns which may be inaccessible after restart. Persist destination IP so teardown works from state alone.
- Forwarder listener: old ttrpc Serve goroutine stays blocked after reconnect, routing requests to the stale session. Close the previous connection when a new one arrives.
- Agent proxy shutdown: immediate close severs in-flight requests. Drain pending requests before closing.
- Startup probe: restart-time check blocks indefinitely for surviving pods. Socket bind fails due to TIME_WAIT. Remove the check, add SO_REUSEPORT.
- Extended resources: RemoveExtendedResources on shutdown drops node capacity. Remove the call since resources are re-advertised on startup.

## Testing

Tested on libvirt with a two-node kcli cluster (control-plane + worker) using a debug podvm image with a custom kata-agent and shim (includes changes from https://github.com/kata-containers/kata-containers/pull/12945)

Test flow:
1. Deploy CAA daemonset with sandbox persistence changes
2. Create a peer pod running a counter process (`i=0; while true; do echo counter=$i; i=$((i+1)); sleep 1; done`)
3. Verify kubectl exec works and counter is incrementing
4. Delete the CAA pod (kubectl delete pod/cloud-api-adaptor-daemonset-xxxxx)
5. Wait for new CAA pod to reach Ready
6. Verify kubectl exec works — counter continues from where it left off (not reset to 0)
7. Verify kubectl get pod shows RESTARTS=0

### Unit and e2e Tests
- Unit tests added
- E2E test: Planning on raising a separate PR that adds a optional e2e test that tests end to end sandbox persistence  

Fixes: rhjira#KATA-4915

### DIFF
--- a/src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
+++ b/src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go
@@ -99,6 +99,7 @@ func (cfg *daemonConfig) Setup() (cmd.Starter, error) {
 		// Common flags with environment variable support
 		reg.StringWithEnv(&cfg.serverConfig.SocketPath, "socket", adaptor.DefaultSocketPath, "REMOTE_HYPERVISOR_ENDPOINT", "Unix domain socket path of remote hypervisor service")
 		reg.StringWithEnv(&cfg.serverConfig.PodsDir, "pods-dir", adaptor.DefaultPodsDir, "PODS_DIR", "base directory for pod directories")
+		reg.StringWithEnv(&cfg.serverConfig.TLSMaterialPath, "tls-material-path", "/run/peerpod/tls-material.json", "TLS_MATERIAL_PATH", "Path to persist TLS material across CAA restarts")
 		reg.StringWithEnv(&cfg.serverConfig.PauseImage, "pause-image", "", "PAUSE_IMAGE", "pause image to be used for the pods")
 		reg.StringWithEnv(&cfg.serverConfig.ForwarderPort, "forwarder-port", daemon.DefaultListenPort, "FORWARDER_PORT", "port number of agent protocol forwarder")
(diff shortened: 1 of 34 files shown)


thejasn:
rebased and pushed CI failure fixes (lint issues)

bpradipt:
This got slipped. @thejasn where does this stand ?

thejasn:
> This got slipped. @thejasn where does this stand ?

Hi @bpradipt , this slipped (de-prioritized) on our side as well. This will need to be retested with the kata rust shim (since 4.0 is out) so still has unknowns as I don't know the behavior of the rust shim.

PR #3056, opened 2026-05-12
cloud-api-adaptor: support Workload Identity on GKE

PeteE:
# GCP: entrypoint.sh blocks Workload Identity on GKE

## Summary
When running `cloud-api-adaptor` on GKE with Workload Identity, there is no service-account JSON key to provide — the pod's KSA is bound to a GSA and the Google SDK obtains tokens via the node's metadata server (Application Default Credentials). However, `src/cloud-api-adaptor/entrypoint.sh` hard-requires `GCP_CREDENTIALS` and exits before launching the binary, making the WI path unusable without bypassing the entrypoint entirely.

The Go provider already handles this correctly — see `src/cloud-providers/gcp/provider.go`
when `GcpCredentials` is empty it falls through to `compute.NewInstancesRESTClient(context.TODO())`, which uses ADC. So this is purely a gap in the shell wrapper (and, by extension, the Helm chart, which pins command to the entrypoint).

## Steps to reproduce
- Create a GKE cluster with Workload Identity enabled.
- Bind the cloud-api-adaptor KSA to a GSA that has the required compute roles.
- Install the peerpods Helm chart with `provider=gcp` and `GCP_PROJECT_ID`, `GCP_ZONE`, `PODVM_IMAGE_NAME` set, but no `GCP_CREDENTIALS`.
- The `DaemonSet` pod logs: `$GCP_CREDENTIALS is NOT set` and the container exits 1 without ever invoking `cloud-api-adaptor gcp`.

## Expected behavior
`entrypoint.sh` should treat `GCP_CREDENTIALS` as optional. When unset, leave `GOOGLE_APPLICATION_CREDENTIALS` unset so the Google SDK falls back to ADC.

## Proposed change
In `src/cloud-api-adaptor/entrypoint.sh`, drop `GCP_CREDENTIALS` from the required-vars list:
```
  gcp() {
      test_vars GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME

      if [[ -n "$GCP_CREDENTIALS" ]]; then
          # Static service-account key: write to disk and point the SDK at it.
          echo "$GCP_CREDENTIALS" > /tmp/gcp-creds.json
          export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp-creds.json
      fi
      # If GCP_CREDENTIALS is unset, the Google SDK uses Application Default
      # Credentials — i.e. the GKE metadata server, which is how Workload
      # Identity delivers tokens for the pod's bound service account.

      set -x
      exec cloud-api-adaptor gcp ${optionals}
  }
```

This is fully backward-compatible: every existing deployment that sets `GCP_CREDENTIALS` continues to work identically.

### DIFF
--- a/src/cloud-api-adaptor/entrypoint.sh
+++ b/src/cloud-api-adaptor/entrypoint.sh
@@ -67,11 +67,17 @@ alibabacloud() {
 }
 
 gcp() {
-    test_vars GCP_CREDENTIALS GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
+    test_vars GCP_PROJECT_ID GCP_ZONE PODVM_IMAGE_NAME
 
-    # Avoid using node's metadata service credentials for GCP authentication
-    echo "$GCP_CREDENTIALS" > /tmp/gcp-creds.json
-    export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp-creds.json
+    if [[ -n "$GCP_CREDENTIALS" ]]; then
+        # Static service-account key: write to disk and point the SDK at it.
+        # This bypasses the node's metadata service.
+        echo "$GCP_CREDENTIALS" > /tmp/gcp-creds.json
+        export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp-creds.json
+    fi
+    # If GCP_CREDENTIALS is unset, the Google SDK falls back to Application
+    # Default Credentials — i.e. the GKE metadata server, which is how
+    # Workload Identity delivers a token for the pod's bound service account.
 
     set -x
     exec cloud-api-adaptor gcp ${optionals}
--- a/src/cloud-api-adaptor/install/charts/peerpods/values.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/values.yaml
@@ -92,6 +92,11 @@ limit: "10"
 
 # DaemonSet configuration
 daemonset:
+  # Command to run in the CAA container. Default uses the entrypoint script
+  # which validates required env vars per provider before exec'ing the binary.
+  # Override to bypass the entrypoint.
+  command:
+    - /usr/local/bin/entrypoint.sh
   # Update strategy controls how the CAA DaemonSet performs rolling updates
   updateStrategy:
     # maxUnavailable: Maximum number of pods that can be unavailable during update
--- a/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
@@ -25,7 +25,7 @@ spec:
       serviceAccountName: cloud-api-adaptor
       containers:
       - command:
-        - /usr/local/bin/entrypoint.sh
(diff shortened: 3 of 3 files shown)


src/cloud-api-adaptor/entrypoint.sh
```diff
@@ -67,11 +67,17 @@ alibabacloud() {
...
+
+    if [[ -n "$GCP_CREDENTIALS" ]]; then
+        # Static service-account key: write to disk and point the SDK at it.
+        # This bypasses the node's metadata service.
+        echo "$GCP_CREDENTIALS" > /tmp/gcp-creds.json
+        export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp-creds.json
+    fi
+    # If GCP_CREDENTIALS is unset, the Google SDK falls back to Application
+    # Default Credentials — i.e. the GKE metadata server, which is how
+    # Workload Identity delivers a token for the pod's bound service account.
```
beraldoleal, line 80: GCP_CREDENTIALS was made mandatory in #2345 because without it the SDK silently uses the node's metadata service credentials, which have OAuth scopes too restrictive for some operations. The pod starts fine but fails later on specific API calls.

If I understand correctly, with this change, clusters without Workload Identity would land back on that path. Maybe we could have an automatic WI detection to trigger the right path, so we can still keep GCP_CREDENTIALS required for clusters without WI?

src/cloud-api-adaptor/install/charts/peerpods/values.yaml
```diff
@@ -92,6 +92,11 @@ limit: "10"
 
 # DaemonSet configuration
 daemonset:
+  # Command to run in the CAA container. Default uses the entrypoint script
+  # which validates required env vars per provider before exec'ing the binary.
+  # Override to bypass the entrypoint.
+  command:
+    - /usr/local/bin/entrypoint.sh
```
beraldoleal, line 99: iiuc, if we accept/merge the entrypoint fix, this would not be needed. Maybe I'm missing the reason for changing the command. (Just trying to keep the help api clean)

beraldoleal (commented):
hey @PeteE , thanks a lot your PR, we need this.

I just left a few comments for you.

src/cloud-api-adaptor/entrypoint.sh
snir911: I think using a separate environment variable for the direct WIF as @PeteE suggested in #3077 (see my [comment](https://github.com/confidential-containers/cloud-api-adaptor/pull/3077#issuecomment-5158333016) there also) will allow us to check in the entrypoint for **either GCP_CREDENTIALS or GOOGLE_APPLICATION_CREDENTIALS** existence and avoid silent usage in the node's IMDS credentials (BTW in that case maybe it's better to switch naming to something like GCP_WIF_CREDENTIALS..)

snir911:
@PeteE could you pls rebase (actually maybe worth adding the changes from #3077 to this pr also, WDYT?)

bpradipt:
@snir911 @beraldoleal may be one of you can pick this up and take it to closure ?

beraldoleal:
I'm closing this PR, since #3294 was merged. Feel to reopen if needed.

PR #3057, opened 2026-05-12
charts: support imagePullSecrets on peerpods and peerpod-ctrl

PeteE:
Expose imagePullSecrets as a configurable Helm value on both charts:

* peerpods: attaches to the cloud-api-adaptor DaemonSet pods so the CAA image can be pulled from a private registry.
* peerpod-ctrl: attaches to the controller-manager pod so its image (and kube-rbac-proxy sidecar when enabled) can be pulled from a private registry.

Defaults to an empty list, so existing deployments are unaffected.

### DIFF
--- a/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
+++ b/src/cloud-api-adaptor/install/charts/peerpods/templates/daemonset.yaml
@@ -23,6 +23,10 @@ spec:
         {{- end }}
     spec:
       serviceAccountName: cloud-api-adaptor
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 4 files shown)


mkulke (commented):
Can you associate an imagePullSecret with the serviceAccount? afaiu, enumerating imagePullSecrets explicitly on a pod spec isn't idiomatic in modern kubernetes anymore, you would rather do that via serviceAccount or better, node managed identity on nodes to access private registries.

PeteE:
> Can you associate an imagePullSecret with the serviceAccount? afaiu, enumerating imagePullSecrets explicitly on a pod spec isn't idiomatic in modern kubernetes anymore, you would rather do that via serviceAccount or better, node managed identity on nodes to access private registries.

Fair enough! I'll close this PR.

Issue #2466, opened 2025-06-10
Agent Protocol Forwarder Fails to Start in podVM boot process


ben-pty:
### Describe the bug

I am trying to follow the AWS example from [CoCo documents](https://confidentialcontainers.org/docs/examples/aws-simple/). However, in the [deploy-workload](https://confidentialcontainers.org/docs/examples/aws-simple/#deploy-workload) part, the sample nginx pod fails to start correctly in my experiment.

In the podVM, it looks like the service **agent-protocol-forwarder** and **setup-nat-for-imds** fail to start during the boot process, as shown in the boot log below. Meanwhile, in the cloud-api-adaptor, it is trying to establish a connection with the podVM, but it never succeeds. Eventually, the CAA shut down the podVM and recreated a new one, which also has the same issue.

As a result, the status of the pod is always in `ContainerCreating`. And from the EC2 console, I can see that podVM instances are created and destroyed repetitively.

From the log below, it looks like the **agent-protocol-forwarder** eventually starts successfully in the third try. Yet, it looks like the CAA still fails to connect to the podVM.

### How to reproduce

Follow the guidance from [here](https://confidentialcontainers.org/docs/examples/aws-simple/) to run a demo on AWS.
In "CAA pod VM image" part, export PODVM_AMI_ID with the AMI ID found in this [link](https://us-east-2.console.aws.amazon.com/ec2/home?region=us-east-2#ImageDetails:imageId=ami-04142703a1d6f0d95) manually. (i.e. export PODVM_AMI_ID=ami-04142703a1d6f0d95)

### CoCo version information

CSP: AWS, CAA: v0.14.0, podVM image: fedora-mkosi-debug-tee-amd-0-14-0

### What TEE are you seeing the problem on

Snp

### Failing command and relevant log output

```shell
Relevant portion in podVM boot log, in the order it happened:

[FAILED] Failed to start agent-protocol-for…service - Agent Protocol Forwarder.
See 'systemctl status agent-protocol-forwarder.service' for details.
         Starting setup-nat-for-imds.service - Setup NAT for IMDS...
         Starting systemd-tmpfiles-setup-de… Device Nodes in /dev gracefully...
         Starting systemd-userdbd.service - User Database Manager...
[FAILED] Failed to start setup-nat-for-imds.service - Setup NAT for IMDS.
See 'systemctl status setup-nat-for-imds.service' for details.
[...]

PR #3058, opened 2026-05-13
userdata: use IMDSv2 token for AWS metadata with IMDSv1 fallback

Agrek11:
AWSUserDataProvider currently issues a bare GET to /latest/user-data (IMDSv1). On EC2 instances configured with MetadataOptions.HttpTokens=required (IMDSv2-only), this returns 401 and peer-pod boot fails before kata-agent starts. Many enterprise AWS organizations enforce IMDSv2 via an SCP, so the bare IMDSv1 path is unusable in those environments, and AWS now defaults new EC2 launches to v2-only as well.

This change adds an IMDSv2 token PUT before the user-data GET and attaches the returned session token via the X-aws-ec2-metadata-token header. If the token PUT fails for any reason (network policy blocks PUT, legacy IMDSv1-only configuration, transient error), the helper returns nil headers so the existing IMDSv1 GET path is preserved as a fallback. No existing flow regresses.

Validated on an AWS organization with SCP-enforced HttpTokens=required: peer-pod boots end to end, /dev/sev-guest reachable inside the SEV-SNP guest, attestation report retrieved.

Unit tests cover the success path, non-200 token response, the returned headers shape, and an end-to-end fallback where the token endpoint 401s and the user-data GET succeeds without the token header.

Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html

### DIFF
--- a/src/cloud-api-adaptor/pkg/userdata/provision.go
+++ b/src/cloud-api-adaptor/pkg/userdata/provision.go
@@ -23,6 +23,9 @@ const (
 	// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
 	AWSImdsURL         = "http://169.254.169.254/latest/dynamic/instance-identity/document"
 	AWSUserDataImdsURL = "http://169.254.169.254/latest/user-data"
+	// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-use-IMDSv2.html
+	AWSIMDSv2TokenURL = "http://169.254.169.254/latest/api/token"
+	AWSIMDSv2TokenTTL = "60"
 	// Ref: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service
 	AzureImdsURL         = "http://169.254.169.254/metadata/instance/compute?api-version=2021-01-01"
 	AzureUserDataImdsURL = "http://169.254.169.254/metadata/instance/compute/userData?api-version=2021-01-01&format=text"
@@ -92,7 +95,7 @@ func (a AWSUserDataProvider) GetUserData(ctx context.Context) ([]byte, error) {
 	url := AWSUserDataImdsURL
 	logger.Printf("provider: AWS, userDataUrl: %s\n", url)
 	// aws user data is not base64 encoded
-	return imdsGet(ctx, url, false, nil)
+	return imdsGet(ctx, url, false, awsIMDSHeaders(ctx, AWSIMDSv2TokenURL))
 }
 
 type GCPUserDataProvider struct{ DefaultRetry }
--- a/src/cloud-api-adaptor/pkg/userdata/imds.go
+++ b/src/cloud-api-adaptor/pkg/userdata/imds.go
@@ -6,11 +6,48 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 )
 
 type kvPair struct {
 	k string
 	v string
+}
+
+const awsIMDSv2TokenFetchTimeout = 5 * time.Second
+
+func awsIMDSv2Token(ctx context.Context, tokenURL string) (string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, tokenURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create IMDSv2 token request: %w", err)
+	}
+	req.Header.Set("X-aws-ec2-metadata-token-ttl-seconds", AWSIMDSv2TokenTTL)
+
+	client := &http.Client{Timeout: awsIMDSv2TokenFetchTimeout}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to send IMDSv2 token request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("IMDSv2 token endpoint returned %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to read IMDSv2 token response: %w", err)
+	}
+	return string(body), nil
+}
+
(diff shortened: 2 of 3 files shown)


walidshaari:
Independent reproduction confirms this is the right fix.

**Environment:** OSC 1.12.0 + Trustee 1.1.0 on OCP 4.21.15 (self-managed, AWS eu-west-1), CAA midstream `v0.8.0-1.6`, pod-VM image `osc-dm-verity-image@sha256:9731adbb...` (RHEL 10.1 base, systemd 257), AWS account with `HttpTokens=required` set as account-level default.

**Symptom matches exactly:** `process-user-data.service` runs for ~30s, then `[FAILED] Failed to start process-user-data.service`. `agent-protocol-forwarder.service` starts without `apf.json`, never binds 15150. CAA gets `connection refused` looping until `PROXY_TIMEOUT`. Worker-side VXLAN interface shows zero RX traffic from CVM (confirms `podNode.Setup()` never runs because config is missing).

**Verified in our environment:**

1. CAA explicitly sets `MetadataOptions.HttpTokens=required` in `RunInstances` regardless of account-level default — confirmed via `aws ec2 describe-instances --instance-ids <cvm-id> --query 'Reservations[*].Instances[*].MetadataOptions'`. Even with the account default cleared to `no-preference`, the instance launches with `HttpTokens=required`.

2. `process-user-data` cannot reach IMDS without a session token — `pkg/userdata/imds.go:imdsGet()` does a plain GET with no `X-aws-ec2-metadata-token` header.

3. ben-pty's case in #2466 succeeded only because his account allowed both v1 and v2. Any account or organization that enforces v2 hits this bug.

**Tracking:** filed #3068 for the symptom (forwarder bind failure on OSC 1.12 + AWS SEV-SNP). Will update that issue to point at this PR as the root-cause fix.

For maintainers: would also be useful to know whether Red Hat plans to backport this into OSC midstream for the 1.12.x / 1.13 release train. The bug is effectively a blocker for any AWS deployment of OSC peer-pods in security-conscious accounts.

bpradipt:
Thanks for your PR @Agrek11.
You will have to remove the merge commit.

Agrek11:
@bpradipt 

rebased onto main — merge commit removed.
 
re: govulncheck failure — pre-existing on main (GO-2026-4887, GO-2026-4883 in docker/docker, Fixed in: N/A). Same job failing on last 5+ main commits. My PR doesn't touch go.mod.

src/cloud-api-adaptor/pkg/userdata/provision.go
```diff
@@ -23,6 +23,9 @@ const (
 	// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
 	AWSImdsURL         = "http://169.254.169.254/latest/dynamic/instance-identity/document"
 	AWSUserDataImdsURL = "http://169.254.169.254/latest/user-data"
+	// Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-use-IMDSv2.html
+	AWSIMDSv2TokenURL = "http://169.254.169.254/latest/api/token"
+	AWSIMDSv2TokenTTL = "21600"
```
snir911: nit, could be much shorter no?

snir911 (approved):
LGTM overall thanks, i added a minor comment and lets see checks are passing

bpradipt (approved):
/lgtm
Thanks @Agrek11 
Triggered an e2e CI on AWS. Let's see
cc @wainersm

src/cloud-api-adaptor/pkg/userdata/provision.go
Agrek11: Noted — kept named constants for testability; happy to inline if it's blocking.

snir911: I meant to the TTL value :)

Agrek11: Good catch — shortened to 60s. Token's only used for the immediate user-data GET, so 6h was too large. Rebased onto main and amended the change in. Thanks!

wainersm:
Hi @bpradipt , @Agrek11 ,

> /lgtm Thanks @Agrek11 Triggered an e2e CI on AWS. Let's see cc @wainersm

The current CI isn't going to test IMDS v2-only but it will exercise the fallback to v1 path. I've an idea how to change the CI to test this PR; please give me one day at most that I can return with the result.

bpradipt:
> Hi @bpradipt , @Agrek11 ,
> 
> > /lgtm Thanks @Agrek11 Triggered an e2e CI on AWS. Let's see cc @wainersm
> 
> The current CI isn't going to test IMDS v2-only but it will exercise the fallback to v1 path. I've an idea how to change the CI to test this PR; please give me one day at most that I can return with the result.

The fallback path is working fine. So there is no regression.
We'll merge this once we hear back from @wainersm.

Thanks again @Agrek11 for your work on this one. 
Helps @walidshaari as well.

wainersm:
> > Hi @bpradipt , @Agrek11 ,
> > > /lgtm Thanks @Agrek11 Triggered an e2e CI on AWS. Let's see cc @wainersm
> > 
> > 
> > The current CI isn't going to test IMDS v2-only but it will exercise the fallback to v1 path. I've an idea how to change the CI to test this PR; please give me one day at most that I can return with the result.
> 
> The fallback path is working fine. So there is no regression. We'll merge this once we hear back from @wainersm.
> 
> Thanks again @Agrek11 for your work on this one. Helps @walidshaari as well.

I made this change (https://github.com/wainersm/cc-cloud-api-adaptor/commit/ea9633020ba24441deab09964fc000c34ba3f5cb) to test code to publish the podvm image AMI with IMDS v2-only support. Then I ran the CI without the fix from this PR and could confirm all tests failed (https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/26119502898/job/76819844445). Then I put this PR on top and re-run CI, now this time it passed (https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/26126918431/job/76844795415). Hence the fix is fine.

I will send a follow up PR with my changes on CI.

PR #3061, opened 2026-05-14
fix: Intermittent libvirt connection failures

rafsal-rahim:
Intermittent libvirt connection failures occur when cloud-api-adaptor runs as a non-root user in Kubernetes/OpenShift environments. The errors manifest as:

[adaptor/cloud/libvirt] Unable to create libvirt connection: virError(Code=38, Domain=7, 
Message='Cannot recv data: Could not create directory '/.ssh' (Permission denied). 
Failed to add the host to the list of known hosts (/.ssh/known_hosts).')

[adaptor/cloud/libvirt] failed to create an instance : Error in checking instance: 
virError(Code=1, Domain=7, Message='internal error: client socket is closed')

These failures occur intermittently (approximately 1 in 20 pod creations) and are difficult to reproduce consistently, making them particularly problematic in production environments.

Root Cause
When the container runs as a non-root user without a properly set HOME environment variable, the libvirt SSH client (qemu+ssh:// URI) attempts to create the ~/.ssh directory and known_hosts file in the root directory (/), which fails due to permission restrictions.

The intermittent nature occurs because:

Sometimes ~/.ssh exists from previous successful connections
Race conditions during pod initialization
Connection reuse may temporarily mask the problem

Solution:
Set HOME=/tmp in the Dockerfile to ensure SSH has a writable location for creating .ssh directory and known_hosts file.

While the issue was difficult to reproduce consistently, the fix addresses a well-known pattern in containerized applications using SSH. The solution has been validated through:

- Code analysis of libvirt SSH connection flow
- Understanding of SSH client behavior when HOME is not set
- Alignment with Kubernetes security best practices
- Review of similar fixes in other containerized SSH applications

created with assistance from bob

### DIFF
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -72,4 +72,5 @@ RUN apt-get update && \
 
 FROM base-${BUILD_TYPE}
 COPY --from=builder /work/cloud-api-adaptor/cloud-api-adaptor /work/cloud-api-adaptor/entrypoint.sh /usr/local/bin/
+ENV HOME=/tmp
 CMD ["entrypoint.sh"]


src/cloud-api-adaptor/Dockerfile
stevenhorsman, line 1: Can you add information to the commit message to explain the change and why it works please

stevenhorsman (commented):
IBM Bob has suggested that this fix isn't correct:
> Critical Analysis
❌ MAJOR ISSUE: This fix does NOT address the actual problem
After reviewing the deployment configuration, I found that:
>
> 
> The daemonset does NOT run as non-root - Line 44-48 of daemonset.yaml shows:
> 
> securityContext:
>   capabilities:
>     add:
>     - NET_ADMIN
>     - SYS_ADMIN
> 
> There is no runAsNonRoot: true or runAsUser specified, meaning the container runs as root (UID 0) by default.
> 
> SSH keys are mounted to /root/.ssh/ - Lines 59-63 show:
> 
> volumeMounts:
> - mountPath: /root/.ssh/
>   name: ssh
>   readOnly: true
> 
> This explicitly uses /root/.ssh/, which is only accessible to root.
> 
> The problem description is inconsistent - The PR states the issue occurs "when cloud-api-adaptor runs as a non-root user", but the deployment manifests show it runs as root.
> 
> Root Cause Analysis
> The actual issue is likely:
> 
> Race condition during SSH connection initialization - When libvirt tries to establish SSH connections, it may attempt to create ~/.ssh/known_hosts before the volume mount is ready or accessible
> HOME environment variable not set - If HOME is unset or empty, SSH client defaults to / for home directory
> Intermittent timing - The 1-in-20 failure rate suggests a race condition, not a consistent permissions issue
> Why This Fix Appears to Work (But Shouldn't)
> Setting HOME=/tmp would:
> 
> Make SSH client use /tmp/.ssh/ instead of /root/.ssh/ or /.ssh/
> /tmp is always writable, so known_hosts creation succeeds
> BUT this bypasses the intentional SSH key mount at /root/.ssh/
> Security and Functional Concerns
> SSH keys ignored - The fix causes SSH to look in /tmp/.ssh/ instead of /root/.ssh/ where keys are mounted
> Host key verification broken - known_hosts in /tmp won't persist, defeating host key verification
> Authentication may fail - If SSH private keys are mounted to /root/.ssh/, they won't be found in /tmp/.ssh/
> Misleading problem description - The PR claims this is about non-root users, but the deployment runs as root
> 


Any thoughts on this? Under what cirumstances does the daemonset run as non-root? Is it a downstream only issue? If so it might be more reliable to get it fixed in the downstream code/installer?

PR #3063, opened 2026-05-14
ci: build peerpod-ctrl and webhook images in nightly e2e

wainersm:
After the kustomize-to-helm migration, nightly e2e tests deploy peerpod-ctrl and webhook as sub-charts but use stale `latest` images from quay.io. Since peerpod-ctrl shares the cloud-providers Go module with caa, version skew can mask breaking changes.

Build both images alongside caa/podvm in the nightly pipeline and wire their references through to the helm install via PEERPOD_CTRL_IMAGE and WEBHOOK_IMAGE environment variables.

The subchart image override logic is centralized in Helm.ConfigureSubchartImages() to avoid duplicating code across all provider implementations.

Assisted-by: Claude

### DIFF
--- a/src/cloud-api-adaptor/test/provisioner/helm.go
+++ b/src/cloud-api-adaptor/test/provisioner/helm.go
@@ -14,6 +14,37 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 )
 
+// ConfigureSubchartImages sets helm override values for sub-chart images
+// (peerpod-ctrl and webhook) from environment variables. These overrides are
+// provider-agnostic and applied centrally from Deploy().
+func (h *Helm) ConfigureSubchartImages() {
+	if img := os.Getenv("PEERPOD_CTRL_IMAGE"); img != "" {
+		parts := strings.SplitN(img, ":", 2)
+		if parts[0] == "" {
+			log.Warnf("PEERPOD_CTRL_IMAGE has empty repository: %q, skipping", img)
+		} else {
+			h.OverrideValues["resourceCtrl.image.repository"] = parts[0]
+			if len(parts) == 2 && parts[1] != "" {
+				h.OverrideValues["resourceCtrl.image.tag"] = parts[1]
+			}
+			log.Infof("Configuring helm: peerpod-ctrl image override (%s)", img)
+		}
+	}
+
+	if img := os.Getenv("WEBHOOK_IMAGE"); img != "" {
+		parts := strings.SplitN(img, ":", 2)
+		if parts[0] == "" {
+			log.Warnf("WEBHOOK_IMAGE has empty repository: %q, skipping", img)
+		} else {
+			h.OverrideValues["webhook.image.repository"] = parts[0]
+			if len(parts) == 2 && parts[1] != "" {
+				h.OverrideValues["webhook.image.tag"] = parts[1]
+			}
+			log.Infof("Configuring helm: webhook image override (%s)", img)
+		}
+	}
+}
+
 // Helm represents a Helm chart installer
 type Helm struct {
 	ChartPath               string            // path to the chart directory
--- a/.github/workflows/e2e_aws.yaml
+++ b/.github/workflows/e2e_aws.yaml
@@ -18,6 +18,16 @@ on:
         description: Git ref to checkout the cloud-api-adaptor repository. Defaults to main.
         required: false
         type: string
+      peerpod_ctrl_image:
+        description: The peerpod-ctrl OCI image (including tag)
+        required: false
+        default: ''
+        type: string
+      webhook_image:
+        description: The peer-pods-webhook OCI image (including tag)
+        required: false
+        default: ''
+        type: string
       cluster_type:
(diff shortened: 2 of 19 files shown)


wainersm:
Hi @stevenhorsman , the PR as promised.

There is one caveat: peerpods-ctrl build workflow is only multi-arch, which makes now the aws, libvirt (x86_64)...etc also depend on non-x86_64 builds (and runners). As a result, I could not test this in my fork (I don't have s390x runners). I was thinking in split the workflow in two: one per-arch and another multi-arch just like you did for caa image build. What do you think?

Regardless, pushed this PR to a topic branch and being tested in https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/25882245927

stevenhorsman:
> I was thinking in split the workflow in two: one per-arch and another multi-arch just like you did for caa image build. What do you think?

Yeah - that's makes sense. I was tempted to do it in the rework PR and in fact created both workflows, but then decided not to over-engineer and just solve the performance issue first.

wainersm:
> > I was thinking in split the workflow in two: one per-arch and another multi-arch just like you did for caa image build. What do you think?
> 
> Yeah - that's makes sense. I was tempted to do it in the rework PR and in fact created both workflows, but then decided not to over-engineer and just solve the performance issue first.

Cool. Feel free to send your changes that I can rebase this PR on top.

stevenhorsman:
> > > I was thinking in split the workflow in two: one per-arch and another multi-arch just like you did for caa image build. What do you think?
> > 
> > 
> > Yeah - that's makes sense. I was tempted to do it in the rework PR and in fact created both workflows, but then decided not to over-engineer and just solve the performance issue first.
> 
> Cool. Feel free to send your changes that I can rebase this PR on top.

I don't have the changes as I lost them when I merged into a single workflow I'm afraid. Claude can probably write them quickly again though!

wainersm:
> > > > I was thinking in split the workflow in two: one per-arch and another multi-arch just like you did for caa image build. What do you think?
> > > 
> > > 
> > > Yeah - that's makes sense. I was tempted to do it in the rework PR and in fact created both workflows, but then decided not to over-engineer and just solve the performance issue first.
> > 
> > 
> > Cool. Feel free to send your changes that I can rebase this PR on top.
> 
> I don't have the changes as I lost them when I merged into a single workflow I'm afraid. Claude can probably write them quickly again though!

Ah, I thought you had them still. Ok, I will talk to my new best friend to implement it.

wainersm:
Hi @stevenhorsman , I've just pushed the commit that split peerpods-ctrl build in two workflows for handling multi-arch. Tested in https://github.com/wainersm/cc-cloud-api-adaptor/actions/runs/25940549101/job/76258717043 , you gonna see in the logs a message like:

```
time="2026-05-15T21:11:57Z" level=info msg="Configuring helm: peerpod-ctrl image override (ghcr.io/wainersm/peerpod-ctrl:latest-amd64)"
time="2026-05-15T21:11:57Z" level=info msg="Configuring helm: webhook image override (ghcr.io/wainersm/peer-pods-webhook:latest)"

```

wainersm:
Pushed some updates to make zizmor happy again.

stevenhorsman:
Sorry if this seems unreasonable, but could you flip the commit order so that the peerpod-ctrl split is first rather than having the workflow to introduce the old name and then all the edits related to that as a separate PR?

mkulke (approved):

wainersm:
> Sorry if this seems unreasonable, but could you flip the commit order so that the peerpod-ctrl split is first rather than having the workflow to introduce the old name and then all the edits related to that as a separate PR?

Just inverted the order of the commits

src/cloud-api-adaptor/test/provisioner/helm.go
```diff
@@ -14,6 +14,29 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 )
 
+// ConfigureSubchartImages sets helm override values for sub-chart images
+// (peerpod-ctrl and webhook) from environment variables. These overrides are
+// provider-agnostic and applied centrally from Deploy().
+func (h *Helm) ConfigureSubchartImages() {
+	if img := os.Getenv("PEERPOD_CTRL_IMAGE"); img != "" {
+		parts := strings.SplitN(img, ":", 2)
+		h.OverrideValues["resourceCtrl.image.repository"] = parts[0]
```
stevenhorsman: Minor nit - it might be worth validating that that `parts[0]` is non-zero incase we got something like `:tag` submitted due to mixing up env creation?

stevenhorsman (approved):
One optional suggestion, but looks good

src/cloud-api-adaptor/test/provisioner/helm.go
wainersm: Hi @stevenhorsman , 

The new version strengthen the parsing of the environment variables. Looks better now?

stevenhorsman: Looks good. Thanks!

Issue #2387, opened 2025-04-24
Tracking OpenStack provider support


tarumizu:
This issue is created to track the development progress. I have listed the development items below to add support for the OpenStack provider. For each development item submitted as a PR, a corresponding Issue will be created. If you have any tasks related to adding support for the OpenStack provider, please feel free to leave a comment.

## Motivation

By supporting OpenStack, I expect to enable secure container execution in more areas. I am eager to work on the necessary tasks to achieve this goal, particularly focusing on Arm CCA. I welcome any insights or collaboration from others who are also exploring this area.

## Development Item

I am planning to incrementally add support for the OpenStack provider. The OpenStack community is currently working on SEV-ES support[1], which is still in progress. My approach is to begin with non-CVM support and gradually integrate CVM support as development progresses.

[1] [Confidential VM and TDX support in upstream Openstack](https://lists.openstack.org/archives/list/openstack-discuss@lists.openstack.org/thread/T4FQAJE4DOVHGG5IIQL36RCX73U5RXDA/)

### Implementation

* [ ] Adding initial support for OpenStack
  * Add support for a new built-in provider by following the `addnewprovider.md`
* Integrate CVM support (such as SNP, TDX and CCA)
  * These items will be considered once the dependent features in OpenStack are completed.

### Testing

* [ ] E2E tests for the OpenStack provider

### Documentation

* [ ] How to build a Pod VM image

tarumizu:
@zvonkok Thanks for your interest.

We're currently facing a resource bottleneck, slowing down progress. We plan to add developers in June/July should improve it.

Interested in presenting. We would appreciate the opportunity to consider presenting once the team expands and development progresses further.

tarumizu:
The patch submission planned for December will be delayed, as development is taking longer than expected. I'm now aiming for mid-to-late January. Apologies for the delay.

PR #2805, opened 2026-01-26
Add support for OpenStack

cw-kojima1003:
Summary:
This pull request allows cloud-api-adaptor(CAA) to support OpenStack-based clouds.

Background:
There is currently no provider in CAA that supports clouds built with standard OpenStack.
By implementing an OpenStack provider, we can offer secure pod execution environments in a wider range of fields.

As stated in the link below, we are focusing on implementations related to ARM CCA.
This PR provides an architecture-independent implementation.
Moving forward, we aim to execute PeerPods on OpenStack clouds that support ARM CCA.

Links:
This PR is for the purpose and related work described in the following Issue.
https://github.com/confidential-containers/cloud-api-adaptor/issues/2387

Task:
Adding an built-in Provider for OpenStack, referring to the following documentation.

https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md

- Add and initialize the OpenStack provider manager
- Add a definition for the configuration struct
- Add cloud interfaces
- Add provider interfaces
- Add additional files to modularize the code
- Add relevant unit tests
- Update entrypoint.sh and Makefile

Implemented:
- [x] Launch Peer Pod with standard VM
- [x] Unit tests
- [x] Manual connectivity verification  

Future implementation:
- []  Validate user-data reception on the Pod VM image
      The user-data related features are currently foundational, with  validation and testing for operational integrity still pending.
- []  Create documentation for CAA deployment and Pod VM creation steps
- []  Implement end-to-end (E2E) tests (CI) for the new provider 
- []  Enable CVM (Confidential Virtual Machine) launch for CCA support
- []  Other features: pause-image, VXLAN, TLS, etc

PR #3064, opened 2026-05-15
Add support for OpenStack

cw-kojima1003:
Summary:  
This pull request allows cloud-api-adaptor(CAA) to support OpenStack-based clouds.

Background:  
There is currently no provider in CAA that supports clouds built with standard OpenStack.  
By implementing an OpenStack provider, we can offer secure pod execution environments in a wider range of fields.
As stated in the link below, we are focusing on implementations related to ARM CCA.  
This PR provides an architecture-independent implementation.  
Moving forward, we aim to execute PeerPods on OpenStack clouds that support ARM CCA.

Links:  
This PR is for the purpose and related work described in the following Issue.  
[#2387](https://github.com/confidential-containers/cloud-api-adaptor/issues/2387)

Task:  
Adding an built-in Provider for OpenStack, referring to the following documentation.
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md

- Add and initialize the OpenStack provider manager
- Add a definition for the configuration struct
- Add cloud interfaces
- Add provider interfaces
- Add additional files to modularize the code
- Add relevant tests (unit tests and e2e tests)
- Add documentation on how to build a Pod VM image
- Update entrypoint.sh and Makefile

Implemented:
[x] Launch Peer Pod with standard VM
[x] Unit tests and E2E tests
[x] Manual connectivity verification


cw-kojima1003:
Resubmission:  
This PR has been resubmitted. Due to a change in our company's internal policy, the repository was moved from a personal account to this organization. 

The original discussion can be found here: #2805.

Changes since #2805: 
- Added e2e tests for the OpenStack provider 
- Added a README covering CAA deployment on OpenStack, Pod VM creation, and e2e test steps 
- Switched deployment to a Helm-based approach and updated compatibility to CAA v0.18 and later (Removed the previous kustomize-based deployment)

mkulke (commented):
I have only glanced at the code so far (but it looks comprehensive and well documented👍). 

I have a couple of abstract questions:

- I noticed in your instructions that you use the packer-based flow to create podvm images. The project has been moving towards podvm images that are built with mkosi (./podvm-mkosi). It has different security characteristics (integrity-protected rootfs, no cloud-init and other hardening measures). Did you test the Openstack provider with mkosi images?

- There is e2e test code in the PR, but I don't see github action workflows that would exercise them. I assume a public github-runner is not big enough to host an OS and a kubernetes deployment. Where do you plan to run those tests?
 
- I don't see any mention of attestation in the docs, I assume the OS provider is agnostic to the TEE that is being used? Do our current attestation-agent builds support all conceivable TEEs or do we need to adjust that?

cw-kojima1003:
@mkulke, thank you for your detailed comments and questions.

- About whether we have tested the OpenStack provider with mkosi images

  For this validation, we used the following image:

- quay.io/confidential-containers/podvm-generic-ubuntu-amd64: v0.20.0  
  https://quay.io/repository/confidential-containers/podvm-generic-ubuntu-amd64?tab=tags

  We believe that the above image was not built with mkosi.
  We would like to test with an mkosi-based image and add links to the corresponding steps in the README.


- Where the tests are run

  We are considering where to run the e2e tests for this provider. As you noted, a public GitHub runner is too small to host both OpenStack and Kubernetes.
  Since this work is also part of adding support for an arm64 CCA environment, we want to add workflows for these e2e tests.
  We are currently working out where we can realistically run them.


- Attestation agent

  The implementation of this provider itself does not depend on a specific TEE; instead, we expect that TEEs will be configured on the OpenStack side, for example through flavors.
  We plan to further consider how to build the PodVM image and configure the attestation agent to be bundled in it.

  According to the current attestation-agent “Supported Platforms”, the TEEs we are targeting are supported. We will provide feedback if any issues come up in future testing.
  https://github.com/confidential-containers/guest-components/tree/main/attestation-agent#supported-platforms

stevenhorsman:
Hi, we had a discussion on the community call about the proliferation of new providers and the challenges of having code dropped into the repo and then not maintained adding burden to a small group of maintainers, as such we are looking to move to an approach of keeping provider out of tree. We have some documentation for this approach in the new providers page: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md#memo-adding-support-for-a-new-external-provider. We admit that it might be outdated and might require more clarification, but if you could consider following this approach, rather than a built-in provider then we'd love the feedback. Thanks

PR #3065, opened 2026-05-15
libvirt: add cpuset support for CPU pinning

### LINKED ISSUE #2397: libvirt: Add support for CPUSET field while on VM creation
The `--cpuset=CPUSET` (reference from https://linux.die.net/man/1/virt-install) can be used to enhance the AI inferencing usecases on a libvirt based VM by ensuring a set of cores belongs to different sockets.

For example:
If there are 32 CPUs with 3 sockets, `--cpuset=0,2,4,6,14,16,18,28,30` can be used to ensure that the set of cores belongs to the three sockets like `([0,2,4,6], [14,16,18,20], [28,30])` whereas the different set of CPUs mentioned on the square brackets belong to different sockets.

ajaypvictor:
Add --cpuset flag and LIBVIRT_CPUSET env var to pin vCPUs to specific physical CPUs. Useful for AI inferencing where cores from different sockets improve performance.

Changes:
- Add CPUSet field to Config and vmConfig structures
- Add --cpuset CLI flag and LIBVIRT_CPUSET environment variable
- Embed CPUSet attribute in Domain XML for all architectures
- Expose LIBVIRT_CPUSET in Helm chart values for deployment
- Add unit tests for CPUSet functionality

Revives PR #2744 with Helm chart adaptation (replaces kustomization.yaml).

Fixes: #2397

### DIFF
--- a/src/cloud-providers/libvirt/libvirt_test.go
+++ b/src/cloud-providers/libvirt/libvirt_test.go
@@ -284,6 +284,33 @@ func TestConfigVerifier(t *testing.T) {
 			expectedError: "Memory must be greater than zero",
 		},
 		{
+			name: "invalid cpuset fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.CPUSet = "ABC"
+				}),
+			},
+			expectedError: "invalid CPUSet format",
+		},
+		{
+			name: "invalid cpuset with mixed alphanumeric fails",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.CPUSet = "0,a,2"
+				}),
+			},
+			expectedError: "invalid CPUSet format",
+		},
+		{
+			name: "valid cpuset passes",
+			provider: &libvirtProvider{
+				serviceConfig: newConfig(func(c *Config) {
+					c.CPUSet = "0,2,4-7"
+				}),
+			},
+			expectedError: "",
+		},
+		{
 			name: "valid config passes",
 			provider: &libvirtProvider{
 				serviceConfig: newConfig(nil),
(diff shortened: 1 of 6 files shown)


src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -303,3 +303,108 @@ func TestConfigVerifier(t *testing.T) {
...
+			cpuset:         "0-3",
+			expectedCPUSet: "0-3",
+		},
+		{
+			name:           "with empty cpuset",
+			cpuset:         "",
+			expectedCPUSet: "",
+		},
+	}
+
```
stevenhorsman, line 445: Is it worth adding a negative test case e.g. setting cpuset to something invalid like "ABC". Then we can see if either a) libvirtxml.DomainVCPU handles this proper, or b) we should adding something like `validateCPUSet` to can use to check it?

ajaypvictor: thanks Steve for the suggestion, makes sense.. updated the changes..

stevenhorsman:
It looks like you've added the provider value in a place that the automation doesn't match with?

ajaypvictor:
> It looks like you've added the provider value in a place that the automation doesn't match with?

Sorry, missed that and added it manually.. Executed:
```
make sync-chart-values CLOUD_PROVIDER=libvirt
```
re-generated the file.
Thanks for spotting.

stevenhorsman (approved):
Looks okay to me. Thanks!

wainersm:
Hi @ajaypvictor ,

In the original PR you made a pertinent comment in https://github.com/confidential-containers/cloud-api-adaptor/pull/2744#discussion_r2667840406 . This cpuset configuration being global will make all podvm be pinned to the same set, which might exhaust the host as well. IMHO adding a way to overwrite the global configuration (i.e. via annotation) is a welcome idea. I was thinking in open an issue to track that, what do you think @ajaypvictor ?

src/cloud-providers/libvirt/manager.go
```diff
@@ -43,6 +43,7 @@ func (*Manager) ParseCmd(flags *flag.FlagSet) {
 	reg.StringWithEnv(&libvirtcfg.Firmware, "firmware", defaultFirmware, "LIBVIRT_EFI_FIRMWARE", "Path to OVMF")
 	reg.UintWithEnv(&libvirtcfg.CPU, "cpu", 2, "LIBVIRT_CPU", "Number of processors allocated")
 	reg.UintWithEnv(&libvirtcfg.Memory, "memory", 8192, "LIBVIRT_MEMORY", "Amount of memory in MiB")
+	reg.StringWithEnv(&libvirtcfg.CPUSet, "cpuset", "", "LIBVIRT_CPUSET", "CPU set for pinning vCPUs to physical CPUs (e.g., \"0,2,4,6\" or \"0-3\")")
```
wainersm: Individual CPUs ("0,2,4,6"), range ("0-3") and mixed ("0-3,4,6,8") are valid specifications. It would be nice to have an example of mixed spec in this comment as it will end up visible to users in the helm chart (in src/cloud-api-adaptor/install/charts/peerpods/providers/libvirt.yaml)

ajaypvictor:
> Hi @ajaypvictor ,
> 
> In the original PR you made a pertinent comment in [#2744 (comment)](https://github.com/confidential-containers/cloud-api-adaptor/pull/2744#discussion_r2667840406) . This cpuset configuration being global will make all podvm be pinned to the same set, which might exhaust the host as well. IMHO adding a way to overwrite the global configuration (i.e. via annotation) is a welcome idea. I was thinking in open an issue to track that, what do you think @ajaypvictor ?

Thanks for the suggestion Wainer.. I think you are right about the resource contention concern with global CPUSet.

Per-pod annotation override would be great, but as Steve replied previously to [the original suggestion](https://github.com/confidential-containers/cloud-api-adaptor/pull/2744#discussion_r2667861254), passing remote cloud provider specific annotations through the runtime might be slightly challenging if there are no aft annotation on runtime..

I will create a follow up issue to track this enhancement and explore possible solutions. I guess for now, the global configuration addresses the primary use case for dedicated workloads.

src/cloud-providers/libvirt/manager.go
ajaypvictor: updated, thanks for the suggestion..

snir911 (approved):
Thanks, LGTM, some tests are failing for s390, seems unrelated though..

wainersm (commented):
thanks @ajaypvictor !

wainersm (approved):

PR #3071, opened 2026-05-19
e2e: move the cluster provisioning using kind to byom

Amulyam24:
Move the cluster provisioning to BYOM to remove the dependency on the deprecated Docker provider.

### DIFF
--- a/src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
+++ b/src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# Ref: https://stackoverflow.com/questions/299728/how-do-you-use-newgrp-in-a-script-then-stay-in-that-group-when-the-script-exits
+newgrp docker <<EOF
+
+# Accept two arguments: create and delete
+# create: creates a kind cluster
+# delete: deletes a kind cluster
+
+CLUSTER_NAME="${CLUSTER_NAME:-peer-pods}"
+KIND_CONFIG_FILE="${KIND_CONFIG_FILE:-kind-config.yaml}"
+
+if [ "$1" == "create" ]; then
+    # Check if kind is installed
+    if [ ! -x "$(command -v kind)" ]; then
+        echo "kind is not installed"
+        exit 0
+    fi
+    echo "Check if the cluster \$CLUSTER_NAME already exists"
+    if kind get clusters | grep -q "\$CLUSTER_NAME"; then
+        echo "Cluster \$CLUSTER_NAME already exists"
+        exit 0
+    fi
+    # Set some sysctls
+    # Ref: https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files
+    sudo sysctl fs.inotify.max_user_watches=524288
+    sudo sysctl fs.inotify.max_user_instances=512
+
+    # Create a kind cluster
+    echo "runtime: " "\$CONTAINER_RUNTIME"
+
+    # Create a kind cluster
+    echo "Creating a kind cluster"
+    kind create cluster --name "\$CLUSTER_NAME" --config "\$KIND_CONFIG_FILE" || exit 1
+
+    # Deploy calico
+    kubectl apply -f https://raw.githubusercontent.com/projectcalico/calico/v3.28.0/manifests/calico.yaml || exit 1
+
+    exit 0
+fi
+
+if [ "$1" == "delete" ]; then
+    # Check if kind is installed
+    if [ ! -x "$(command -v kind)" ]; then
+        echo "kind is not installed"
+        exit 0
+    fi
+
+    # Delete the kind cluster
+    echo "Deleting the kind cluster"
+    kind delete cluster --name "\$CLUSTER_NAME" || exit 1   
+
+    exit 0
+fi
+EOF
--- a/src/cloud-api-adaptor/test/provisioner/byom/provision_common.go
+++ b/src/cloud-api-adaptor/test/provisioner/byom/provision_common.go
@@ -14,14 +14,12 @@ import (
 	"time"
 
 	pv "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/provisioner"
-	"github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/provisioner/docker"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 )
 
-// ByomProvisioner uses DockerProvisioner for BYOM-specific functionality
+// ByomProvisioner implements BYOM e2e provisioning
 type ByomProvisioner struct {
-	*docker.DockerProvisioner
 	provisionerCreatedVMs []string // Track VMs created by this provisioner instance
 }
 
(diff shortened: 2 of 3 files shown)


src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+    # Check if kind is installed
+    if [ ! -x "$(command -v kind)" ]; then
+        echo "kind is not installed"
+        exit 0
+    fi
+    echo "Check if the cluster \$CLUSTER_NAME already exists"
+    if kind get clusters | grep -q "\$CLUSTER_NAME"; then
+        echo "Cluster \$CLUSTER_NAME already exists"
+        exit 0
+    fi
```
stevenhorsman, line 23: Are the backslashes here correct, or will that just escape and cause a search for the literal string `$CLUSTER_NAME`?

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+    # Set some sysctls
+    # Ref: https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files
+    sudo sysctl fs.inotify.max_user_watches=524288
+    sudo sysctl fs.inotify.max_user_instances=512
+
+    # Create a kind cluster
+    echo "runtime: " "\$CONTAINER_RUNTIME"
```
stevenhorsman, line 30: This isn't used anywhere?

src/cloud-api-adaptor/test/provisioner/byom/provision_common.go
```diff
@@ -125,10 +111,24 @@ func (b *ByomProvisioner) CreateCluster(ctx context.Context, cfg *envconf.Config
...
 	}
 
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to get user home directory: %w", err)
+	}
+	cfg.WithKubeconfigFile(filepath.Join(home, ".kube/config"))
+
+	if err := pv.AddNodeRoleWorkerLabel(context.Background(), ByomProps.ClusterName, cfg); err != nil {
+
```
stevenhorsman, line 128: nit: Remove blank line please

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+
+    # Create a kind cluster
+    echo "Creating a kind cluster"
+    kind create cluster --name "\$CLUSTER_NAME" --config "\$KIND_CONFIG_FILE" || exit 1
+
+    # Deploy calico
+    kubectl apply -f https://raw.githubusercontent.com/projectcalico/calico/v3.28.0/manifests/calico.yaml || exit 1
```
stevenhorsman, line 37: 3.28.0 is a pretty old version and hidden in this file, should be read this from versions.yaml instead?

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ "$1" == "delete" ]; then
+    # Check if kind is installed
+    if [ ! -x "$(command -v kind)" ]; then
+        echo "kind is not installed"
+        exit 0
+    fi
+
+    # Delete the kind cluster
+    echo "Deleting the kind cluster"
+    kind delete cluster --name "\$CLUSTER_NAME" || exit 1
```
stevenhorsman, line 51: nit: Trailing whitespace at the end of this line.

src/cloud-api-adaptor/test/provisioner/byom/provision_common.go
```diff
@@ -453,3 +474,37 @@ func (b *ByomProvisioner) destroyContainer(containerName string) error {
 	log.Infof("Container %s destroyed successfully", containerName)
 	return nil
 }
+
+func (b *ByomProvisioner) createKindCluster(workingDir string) error {
+	// Create kind cluster by executing the script on the node
+	cmd := exec.Command("/bin/bash", "-c", "./kind_cluster.sh create")
+	cmd.Dir = workingDir
+	cmd.Stdout = os.Stdout
+	// TODO: better handle stderr. Messages getting out of order.
```
stevenhorsman, line 483: Do we need a tracking issue to reference here?

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# Ref: https://stackoverflow.com/questions/299728/how-do-you-use-newgrp-in-a-script-then-stay-in-that-group-when-the-script-exits
+newgrp docker <<EOF
```
mkulke, line 4: I understand this is a requirement for github runners?

I wouldn't recommend putting it in this script. If the whole script logic is a heredoc string shellcheck will not consider it. 

Can you elaborate why this is required?

If it's really required, you can quote the EOF and then we don't have to use escaping I think:

```
$ newgrp docker << "EOF"
echo This is running as group $(id -gn)
EOF
This is running as group docker
```

Amulyam24: Hi @mkulke, 

> Can you elaborate why this is required?

I'm not sure I have an answer for this, the script was moved from the docker folder to byom.

I will check on handling this better as you suggested, Thanks.

mkulke: I see. if this is in fact a move of an existing file and no new logic then it we don't need to fix it here, since apparently it has been acked before.

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
mkulke: this is a verbatim copy of https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docker/kind_cluster.sh afaui

mkulke (approved):

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
stevenhorsman: It is, but I still think it's broken, so we should fix it.

mkulke: but does it have to be fixed in scope of this PR?

it would be sort of a drive-by fix that is unrelated to the intent of this PR. At worst it's no regression to what exists currently and it unblocks the removal of the docker-provider code that has been causing CI to fail for all PRs.

stevenhorsman: Surely this discussion is wasting way more time that it would be to fix the issue!? It's not like it's the only problem in the commit?

mkulke: I'm not sure. I suspect the whole script has been wrapped in a heredoc string out of desperation to make it work as an unprivileged user on github runners.

you cannot trivially change the group of the user within the scope of a run, because the actions process inherits the original gid. fixing that can be a rabbit hole.

that's why the escaped \$ is there, I think. otherwise, the heredoc would evaluate it at parsing time, not at execution time. there are other ways to fix that, but I think the premise is wrong: just run the whole thing with sudo (hoping that it works on the private runners).

stevenhorsman (approved):

src/cloud-api-adaptor/byom/e2e/kind_cluster.sh
stevenhorsman: fine. let's merge it

Amulyam24:
Thank you @mkulke and @stevenhorsman  for the pointers, I can take refining/revisiting the script as a follow up task.

PR #3078, opened 2026-05-19
cloud-api-adaptor: populate machine type specs on GCP

Qiang-Xu:
`SelectInstanceTypeToUse` is called for determining the right size of VM when resource requests and limits are given.

However, the `MachineTypeSpecList` is always empty in GCP provider which causes `SelectInstanceTypeToUse` to fail. 

I implemented `updateInstanceSizeSpecList` to populate the machine specs. We've tested it on GKE,

### DIFF
--- a/src/cloud-providers/gcp/provider.go
+++ b/src/cloud-providers/gcp/provider.go
@@ -61,6 +61,11 @@ func NewProvider(config *Config) (provider.Provider, error) {
 			return nil, fmt.Errorf("NewInstancesRESTClient error: %s", err)
 		}
 	}
+
+	if err := provider.updateInstanceSizeSpecList(); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm
Thanks @Qiang-Xu

beraldoleal (approved):
LGTM. Please rebase.

govulncheck seems to be failing on main as well, so iirc, safe to ignore? @stevenhorsman ?

wainersm:
> LGTM. Please rebase.
> 
> govulncheck seems to be failing on main as well, so iirc, safe to ignore? @stevenhorsman ?

Hi @beraldoleal , yes, govulncheck check has failed for a while now due to a security fix on docker libraries that unfortunately was not addressed up to now. As we gonna get ride of docker provider soon, it won't be a problem soon either.

PR #3092, opened 2026-05-25
podvm: fail loud when oras resolve returns empty digest

Agrek11:
## Summary

  `pull_agent_artifact` and `pull_gc_artifact` in `src/cloud-api-adaptor/podvm/Makefile.inc` fail with a misleading error far from the root cause when the registry does not have the tag they ask for. This PR makes
  them fail loud at the point of failure instead.

  Both functions do `OCI_DIGEST := $(shell oras resolve $OCI_IMAGE:$tag)`, then `OCI_REF := $(OCI_IMAGE)@$(OCI_DIGEST)`. When `oras resolve` exits non-zero (missing tag), `$(shell)` swallows the exit code and
  captures empty stdout, so `OCI_REF` becomes `$OCI_IMAGE@` (trailing `@`). The subsequent `oras pull` then fails with a generic usage error (`Error: "<image>@": no tag or digest specified`) — two steps removed
  from the actual problem, with nothing pointing at the missing tag or the `versions.yaml` pin that caused it.

  This change adds `$(if $(OCI_DIGEST),,$(error ...))` between the resolve and the `OCI_REF` construction. Empty digest → immediate `make` error naming the exact `image:tag` that failed to resolve.

  ## Why this matters (real-world impact)

  I hit this while iterating on `versions.yaml` pins during a CAA v0.20.0 SEV-SNP peer-pod POC on AWS: pinning a guest-components SHA whose artifact tag had not yet been published by the publish workflow makes the
  build die on an oras usage error that names neither the missing tag nor the `versions.yaml` field to fix. With this guard, the same mistake is a single self-explanatory error line at the resolve step.

  ## Reproduction (without this PR)

  1. In `src/cloud-api-adaptor/versions.yaml`, set `oci.guest-components.reference` to a 40-char SHA for which no `<sha>-snp_amd64` (or equivalent) tag has been published — for example, a fresh commit on
  `guest-components` `main` that hasn't yet been built and pushed by the publish workflow.
  2. Run `TEE_PLATFORM=amd make image-debug` from `src/cloud-api-adaptor/podvm-mkosi/`.
  3. Observe: the build fails with `Error: "ghcr.io/confidential-containers/guest-components/attestation-agent@": no tag or digest specified` from `oras pull` — a usage error that does not mention the missing tag.

  With this PR applied, step 2 instead exits with:
```
Makefile.inc:119: *** oras resolve returned empty digest for ghcr.io/confidential-containers/guest-components/attestation-agent:<sha>-snp_amd64.  Stop.
```

  ## What this PR does NOT do

  - Does not change anything for builds whose `versions.yaml` pins resolve correctly — no-op in the happy path.
  - Does not suppress or alter `oras resolve`'s own stderr — registry error details still print as before; the `$(error)` adds the make-level context on top.
  - Does not add a CI workflow to pre-validate all `versions.yaml` SHAs at PR / release time. That would be a useful follow-up but kept out of this PR to keep the diff minimal.

  ## Test plan

  - [x] Diff is minimal — both functions get the same guard, no other code path touched
  - [x] No-op in the happy path (existing successful builds unchanged)
  - [x] Error message names the exact `image:tag` that failed to resolve
  - [ ] CI: existing `podvm_mkosi_ubuntu` smoke test + AWS e2e workflow should pass unchanged on this PR
  - [ ] Reviewer can verify the failure mode by setting `GUEST_COMPONENTS_REF` to any 40-char SHA that lacks published artifacts and running `make binaries`

Happy to follow this up with a CI workflow that pre-validates all `versions.yaml::oci.*.reference` SHAs resolve in their registries, if maintainers see value.

### DIFF
--- a/src/cloud-api-adaptor/podvm/Makefile.inc
+++ b/src/cloud-api-adaptor/podvm/Makefile.inc
@@ -100,7 +100,8 @@ PAUSE_BUNDLE  ?= pause_bundle
 define pull_agent_artifact
 	$(eval $(call generate_tag,tag,$(KATA_REF),$(ARCH)))
 	$(eval OCI_IMAGE := $(KATA_REGISTRY)/agent)
-	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag}))
+	$(eval OCI_DIGEST := $(strip $(shell oras resolve $(OCI_IMAGE):${tag})))
+	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}))
 	$(eval OCI_REF := $(OCI_IMAGE)@$(OCI_DIGEST))
 	$(if $(filter yes,$(VERIFY_PROVENANCE)),$(ROOT_DIR)hack/verify-provenance.sh \
 		-a $(OCI_REF) \
@@ -114,7 +115,8 @@ endef
 define pull_gc_artifact
 	$(eval $(call generate_tag,tag,$(GUEST_COMPONENTS_REF),$(2)))
 	$(eval OCI_IMAGE := $(GUEST_COMPONENTS_REGISTRY)/$(1))
-	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag}))
+	$(eval OCI_DIGEST := $(strip $(shell oras resolve $(OCI_IMAGE):${tag})))
+	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}))
 	$(eval OCI_REF := $(OCI_IMAGE)@$(OCI_DIGEST))
 	$(if $(filter yes,$(VERIFY_PROVENANCE)),$(ROOT_DIR)hack/verify-provenance.sh \
 		-a $(OCI_REF) \


src/cloud-api-adaptor/podvm/Makefile.inc
```diff
@@ -100,7 +100,8 @@ PAUSE_BUNDLE  ?= pause_bundle
 define pull_agent_artifact
 	$(eval $(call generate_tag,tag,$(KATA_REF),$(ARCH)))
 	$(eval OCI_IMAGE := $(KATA_REGISTRY)/agent)
-	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag}))
+	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag} 2>/dev/null))
```
mkulke: why mask errors here? stderr shouldn't end up in the variable, I think, so we can still check for empty ness, and maybe $strip to be sure:

```
(eval oci_digest := $(shell oras resolve $(oci_image):${tag}))
ifeq ($(strip $(oci_digest)),)
    $(error oras resolve failed for $(oci_image):${tag})
endif
```

Agrek11: Both taken:
  - Dropped `2>/dev/null` — stderr never enters the $(shell) capture anyway; leaving oras's output visible.
  - Added `$(strip ...)` for whitespace safety.

  Kept `$(if ...,,$(error ...))` over `ifeq` because the define is invoked via `$(call)` at recipe position — directives don't fire there, function form does. Same end effect. Happy to hoist the check outside the
  define if you'd prefer.

src/cloud-api-adaptor/podvm/Makefile.inc
```diff
@@ -100,7 +100,8 @@ PAUSE_BUNDLE  ?= pause_bundle
 define pull_agent_artifact
 	$(eval $(call generate_tag,tag,$(KATA_REF),$(ARCH)))
 	$(eval OCI_IMAGE := $(KATA_REGISTRY)/agent)
-	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag}))
+	$(eval OCI_DIGEST := $(strip $(shell oras resolve $(OCI_IMAGE):${tag})))
+	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}. The tag is missing from the registry; check KATA_REF=$(KATA_REF) in versions.yaml against published tags at $(KATA_REGISTRY). Failing the build instead of silently pulling a stale layer.))
```
mkulke:
```suggestion
	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}))
```

src/cloud-api-adaptor/podvm/Makefile.inc
```diff
@@ -114,7 +115,8 @@ endef
 define pull_gc_artifact
 	$(eval $(call generate_tag,tag,$(GUEST_COMPONENTS_REF),$(2)))
 	$(eval OCI_IMAGE := $(GUEST_COMPONENTS_REGISTRY)/$(1))
-	$(eval OCI_DIGEST := $(shell oras resolve $(OCI_IMAGE):${tag}))
+	$(eval OCI_DIGEST := $(strip $(shell oras resolve $(OCI_IMAGE):${tag})))
+	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}. The tag is missing from the registry; check GUEST_COMPONENTS_REF=$(GUEST_COMPONENTS_REF) in versions.yaml against published tags at $(GUEST_COMPONENTS_REGISTRY). Failing the build instead of silently pulling a stale layer.))
```
mkulke:
```suggestion
	$(if $(OCI_DIGEST),,$(error oras resolve returned empty digest for $(OCI_IMAGE):${tag}))
```

Agrek11: Applied both, thanks @mkulke

Agrek11:
@mkulke both suggestions applied and threads resolved. Could you run ok-to-test when you have a moment? Thanks!

mkulke:
> pull_agent_artifact and pull_gc_artifact in src/cloud-api-adaptor/podvm/Makefile.inc silently produce stale PodVM images when the registry does not have the tag they ask for. This PR makes them fail loud instead.

@Agrek11 can you elaborate under which circumstances the Makefile.inc will produce "stale PodVM images". that should not happen, either with or without the change in the PR, so we might have a serious bug in the Makefile chain. My understanding from looking at the changes, the build should always fail hard, because "oras pull $image@" should result in a failure.

AFAIU this PR improves diagnostics, but if we end up with an image that contains undesired (stale) artifacts, then we need to fix more.

Agrek11:
@mkulke you're right,thanks for pushing on this, i re-verified. With a missing tag,oras resolve fails inside $(shell)            
(code swallowed, empty digest),  and  the subsequent oras pull $(OCI_IMAGE)@ fails hard with Error: "...@": no tag or digest specified. So the build does abort on this path — no stale image possible from the empty digest alone. I've updated the PR description: this change is a diagnostics improvement (fail at resolve, naming the missing image:tag and the versions.yaml field, instead of a usage error two steps removed from the cause).
The stale-binary incident that motivated this was real, but I now believe I misattributed it. Two candidate vectors I can see, both independent of this PR: (1) the binaries file targets depend on $(FORCE_TARGET), which is empty unless FORCE is set — so a pre-existing binary in FILES_DIR short-circuits the pull entirely; (2) in the podvm-mkosi flow, the binaries stage runs inside docker buildx build, so a cached layer can be reused even when the registry content behind a mutable tag has changed. I'd like to reproduce these properly and file a separate issue with evidence rather than widen this PR — does that sound reasonable?

PR #3097, opened 2026-05-26
webhook: retry cert-manager apply on transient failures

BbolroC:
The `kubectl apply` of the cert-manager manifest occasionally fails with an etcd timeout ([log](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/26434517527/job/77815218201)):

```
Error from server: error when creating ".../cert-manager.yaml":
etcdserver: request timed out
```

The manifest creates a large batch of resources (6 CRDs, RBAC, deployments, webhook configs) in a single apply, and individual writes can occasionally exceed the API server's request timeout under transient cluster load.

Wrap the apply in a retry loop (3 attempts, 10s backoff). `kubectl apply` is idempotent, so retrying after a partial failure simply creates the missing resources without disturbing the ones already applied. On the observed failing run, everything except (likely) the last webhook config got created, and the retry will fill in the gap.

For `kubectl wait`, the Endpoints object isn't created by the cert-manager manifest directly. It's created by the kube-controller-manager's endpoints controller after the Service is created and once matching pods exist.
Right after `kubectl apply` returns, there's a brief window where the Service exists but its Endpoints object hasn't been populated yet. The apply just finished, so `kubectl wait` raced ahead of the controller.

Assisted-by: IBM Bob

### DIFF
--- a/src/webhook/Makefile
+++ b/src/webhook/Makefile
@@ -165,11 +165,19 @@ kind-deploy: docker-build kind-load deploy ## deploy the webhook in the local ki
 deploy-cert-manager: ## Deploy cert-manager for webhook.
 	curl -fsSL -o cmctl https://github.com/cert-manager/cmctl/releases/latest/download/cmctl_${OS}_${ARCH}
 	chmod +x cmctl
-	# Deploy cert-manager
-	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml
+	# Deploy cert-manager (retry to tolerate transient etcd timeouts)
+	@for i in 1 2 3; do \
+	  echo "Applying cert-manager manifest (attempt $$i)..."; \
+	  kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml && { \
+	    echo "cert-manager applied successfully"; break; }; \
+	  if [ $$i -eq 3 ]; then \
+	    echo "cert-manager apply failed after 3 attempts"; exit 1; \
+	  fi; \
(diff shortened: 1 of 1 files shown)


src/webhook/Makefile
```diff
@@ -165,8 +165,15 @@ kind-deploy: docker-build kind-load deploy ## deploy the webhook in the local ki
 deploy-cert-manager: ## Deploy cert-manager for webhook.
 	curl -fsSL -o cmctl https://github.com/cert-manager/cmctl/releases/latest/download/cmctl_${OS}_${ARCH}
 	chmod +x cmctl
-	# Deploy cert-manager
-	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml
+	# Deploy cert-manager (retry to tolerate transient etcd timeouts)
+	@for i in 1 2 3; do \
+	  echo "Applying cert-manager manifest (attempt $$i)..."; \
+	  kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml && break; \
```
stevenhorsman: Optional: Adding a success message might be helpful if it is sometimes flakey?
```suggestion
	  kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml  && { echo "cert-manager applied successfully"; break; }
```

stevenhorsman (approved):

src/webhook/Makefile
BbolroC: Yep, I will update the PR. Thanks a lot!

beraldoleal (approved):
LGTM, thanks.

PR #3099, opened 2026-05-27
Podvm release build fixes

stevenhorsman:
Attempts to fix the release podvm builds

### DIFF
--- a/.github/workflows/podvm_mkosi_ubuntu.yaml
+++ b/.github/workflows/podvm_mkosi_ubuntu.yaml
@@ -163,6 +163,7 @@ jobs:
         run: make image-debug
         env:
           PODVM_DISTRO: ubuntu
+          MKOSI_VERSION: ${{ env.MKOSI_VERSION }}
 
       - name: Build mkosi image
         if: ${{ inputs.debug == false }}
@@ -170,6 +171,7 @@ jobs:
         run: make image
         env:
           PODVM_DISTRO: ubuntu
+          MKOSI_VERSION: ${{ env.MKOSI_VERSION }}
 
       - name: Upload the qcow2 with oras
         id: publish_oras_qcow2
--- a/src/cloud-api-adaptor/podvm/hack/cross-build-extras.sh
+++ b/src/cloud-api-adaptor/podvm/hack/cross-build-extras.sh
@@ -6,8 +6,9 @@
 # If ARCH is not set, exit
 [[ -z $ARCH ]] && exit 0
 
-# If ARCH is equal to HOST, exit
-[[ $ARCH = $(uname -m) ]] && exit 0
+# If ARCH is equal to HOST, exit (no cross-compilation needed)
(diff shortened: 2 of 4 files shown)


src/cloud-api-adaptor/podvm/hack/cross-build-extras.sh
```diff
@@ -6,8 +6,15 @@
...
-[[ $ARCH = $(uname -m) ]] && exit 0
+# Normalize architecture names for comparison
+HOST_ARCH=$(uname -m)
+# Convert arm64 to aarch64 for comparison
+NORMALIZED_ARCH=$ARCH
+[[ $ARCH == "arm64" ]] && NORMALIZED_ARCH="aarch64"
+[[ $HOST_ARCH == "arm64" ]] && HOST_ARCH="aarch64"
+
+# If ARCH is equal to HOST, exit (no cross-compilation needed)
+[[ $NORMALIZED_ARCH = $HOST_ARCH ]] && exit 0
```
mkulke: I think replacement patterns work well here. do we have to normalize the arch? afaik linux reports always as aarch64.

```bash
[[ "$ARCH/arm64/aarch64" = "$(uname -m)" ]] && exit 0
```

stevenhorsman: Yeah, that's much neater - updated and attributed you. Thank you

.github/workflows/podvm_mkosi_ubuntu.yaml
```diff
@@ -163,13 +163,15 @@ jobs:
         run: make image-debug
         env:
           PODVM_DISTRO: ubuntu
+          MKOSI_VERSION: ${{ env.MKOSI_VERSION }}
```
mkulke, line 166: nit: that should be a noop, no?

mkulke (approved):

.github/workflows/podvm_mkosi_ubuntu.yaml
stevenhorsman: That is my opinion, but two AIs disagreed with me. From gemini:
> The change is not pointless. It fixes a hidden scoping issue in GitHub Actions
>
> When a step declares its own local env: block, GitHub Actions merges it with the global environment. While it should inherit the global ones during this merge, explicit declaration is a defensive engineering practice in complex CI pipelines. It completely eliminates any risk of a merge failure or context-masking error, ensuring MKOSI_VERSION is definitively available alongside PODVM_DISTRO.

I am happy to remove the commit, but it seemed harmless, so I left it in as it might be considered good practice, but equally could be slop! Maybe you could ask copilot to review and see what it thinks?

mkulke: as a defensive measure that's ok I guess

BbolroC (approved):
LGTM, thanks @stevenhorsman for raising the fix!

PR #3102, opened 2026-05-27
Arm podvm build support

stevenhorsman:
Make the arm podvm build use a native runner, and also add a fix to the cross-build in case of local cross-build runs


mkulke (approved):

fidencio (approved):
lgtm, thanks @stevenhorsman!

PR #3106, opened 2026-05-30
Dockerfile: add ca-certificates to release image

bpradipt:
debian:trixie-slim ships without CA certificates. The dev build got them transitively via openssh-client, but the release build only installed iptables, causing Go TLS verification to fail against ec2.<region>.amazonaws.com with x509: certificate signed by unknown authority.

### DIFF
--- a/src/cloud-api-adaptor/Dockerfile
+++ b/src/cloud-api-adaptor/Dockerfile
@@ -59,7 +59,7 @@ RUN echo "Building iptables-wrapper... " \
 FROM $BASE AS base-release
 
 RUN apt-get update && \
-    apt-get install -y iptables && \
+    apt-get install -y ca-certificates iptables && \
     rm -rf /var/lib/apt/lists/*
 RUN --mount=type=cache,target=/iptables,from=iptables,source=/iptables,readonly \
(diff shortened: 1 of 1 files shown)


bpradipt:
@stevenhorsman @wainersm it will be important to figure out how to catch such issues via CI. I'll think about it as well.

stevenhorsman (approved):

mkulke (approved):

jack1902:
When would the next release be? (guess i could use `-dev` build or something when deploying the helm-chart `0.21.1` ?

@bpradipt thanks for the fix (i'm close to using CVM in azure hence raising https://github.com/confidential-containers/cloud-api-adaptor/issues/3136)

stevenhorsman:
The next release is planned for late July. You can find out latest dev charts at `ghcr.io/confidential-containers/cloud-api-adaptor/charts/peerpods:0.0.0-dev` in the meantime though

jack1902:
> The next release is planned for late July. You can find out latest dev charts at `ghcr.io/confidential-containers/cloud-api-adaptor/charts/peerpods:0.0.0-dev` in the meantime though

Cheers, I'll give that a shot next week once I'm back in work to try the peerpods approach 🙌

PR #3120, opened 2026-06-01
entrypoint: fix AWS auth validation logic

snir911:
The previous implementation used test_vars with && and || operators, which caused incorrect behavior because test_vars exits on failure. This has caused CI failure.

Changed to use one_of to verify at least one auth method is available, then use if/else to validate the specific method's required variables. This properly handles both access key and web identity token auth flows.

Assisted-by: Claude AI

### DIFF
--- a/src/cloud-api-adaptor/entrypoint.sh
+++ b/src/cloud-api-adaptor/entrypoint.sh
@@ -42,7 +42,16 @@ one_of() {
 }
 
 aws() {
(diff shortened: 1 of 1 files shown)


wainersm:
Hi @snir911 , thanks for the quick fix, this problem made the aws CI unusable.

test_vars cannot be used in logical statements because it returns error code when the test succeeds. Fixing the function is also a possibility but the way you did here looks cleaner IMO.

wainersm (approved):

beraldoleal (approved):

PR #3122, opened 2026-06-01
csi-wrapper: support volumeMode: Block PVCs

tarassh:
## Summary

Adds block-mode (`volumeMode: Block`) support to the CSI node-side wrapper.
Filesystem-mode PVCs are unchanged; the new code path activates only when
kubelet hands the wrapper a `/volumeDevices/` target path.

Today every example under `src/csi-wrapper/examples/{aws,azure,ibm}/`
omits `volumeMode` (default = Filesystem) and the AWS README walks through
a filesystem-mode workload only. Trying to use a block PVC fails because:

- `getPodUIDandVolumeName` only knows the filesystem path layout, so it
  can't correlate the CSI call with a pod UID.
- The worker-side placeholder is a directory, not a device, which
  containerd rejects with `not a device node`.
- The kata-direct-volume `mountInfo` registers `fstype=ext4`, which
  causes kata-agent to attempt a filesystem mount on a raw device.

This PR fixes all three:

1. **Path parsing.** Recognise `/volumeDevices/publish/<vol>/<UID>` and
   `/volumeDevices/staging/<vol>` in addition to the filesystem layout.

2. **Worker-side device placeholder.** Create a 1 MiB sparse file, attach
   via `losetup -f --show`, bind-mount `/dev/loopN` over the publish path
   so `stat()` reports a block special file. The actual block device lives
   in the peer-pod VM and is wired up through kata-direct-volume; the
   worker-side placeholder is discarded once the guest takes over.

3. **kata-direct-volume `mountInfo`.** Split into a filesystem variant
   (existing `{"Device": "/dev/zero", "fstype": "ext4"}`) and a new block
   variant (`{"volume-type": "block", "device": "/dev/zero", "fstype": ""}`).

`NodeUnpublishVolume` tears down in the reverse order: umount,
`losetup -d`, `RemoveAll`.

## Tested

AWS EKS + `aws-ebs-csi-driver` v1.37, peer-pods on `m6a.xlarge` SEV-SNP.
After the patch the worker side produces the expected state:

```
attached /var/lib/kubelet/.../publish/<pvc>/<UID> to /dev/loop8
stat <publishpath>: block special file
losetup -l: /dev/loop8 -> <publishpath>
direct-volumes/<base64>/mountInfo.json:
  {"volume-type": "block", "device": "/dev/zero", "fstype": ""}
```

## Out of scope

End-to-end block-mode also needs the kata-runtime / CAA proxy side to
bypass `HotplugAddDevice` when a kata-direct-volume entry exists for a
`/volumeDevices/`-style path. That bypass currently only fires on
filesystem-mode mount paths (see
`pkg/adaptor/proxy/service.go: isNodePublishVolumeTargetPath`). Happy
to follow up with a second PR for that once there is agreement on the
wrapper-side change here.

## Filing as draft

Filing as draft to gauge whether block-mode support is desired upstream.
If maintainers prefer a different approach (e.g. a real block device on
the worker side via `dd if=/dev/zero` rather than a sparse file, or
registering the placeholder differently), happy to revise.

### DIFF
--- a/src/csi-wrapper/pkg/wrapper/nodeservice.go
+++ b/src/csi-wrapper/pkg/wrapper/nodeservice.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"strings"
 
 	podvminfo "github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/proto/podvminfo"
@@ -31,6 +32,10 @@ const (
 	DefaultKubeletLibDir  = "/var/lib/kubelet"
 	DefaultKubeletDataDir = "/var/data/kubelet"
 	DefaultMountInfo      = "{\"Device\": \"/dev/zero\", \"fstype\": \"ext4\"}"
+	// Block-mode mountInfo. volume-type=block tells kata-agent to
+	// passthrough the device without mounting a filesystem; the workload
+	// receives a raw block device at the bind-mount target.
+	BlockMountInfo = "{\"volume-type\": \"block\", \"device\": \"/dev/zero\", \"fstype\": \"\"}"
 )
 
 type NodeService struct {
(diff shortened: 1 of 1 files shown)


tarassh:
For visibility: the kata-runtime side of `volumeMode: Block` support (the one mentioned under "Out of scope" above) is now also filed as a draft PR upstream:

https://github.com/kata-containers/kata-containers/pull/13148

That PR makes `remoteHypervisor.HotplugAddDevice` a silent no-op for block devices, which is what's needed for this csi-wrapper change to get a pod to `Running` end-to-end on AWS peer-pods. The two PRs are independent (this one stands on its own as a strict superset of existing behaviour for filesystem-mode), but reviewing them together gives the full story.

stevenhorsman:
Hi @tarassh, thanks for the contribution, however the csi-wrapper is deprecated and the removal is in progress in #3105, so we are not accepting changes at this point. If you are interested in the replacement then #3037 has the code. Thanks

tarassh:
@stevenhorsman Thank you for quick response. Does the replacement support LUSK. This PR was built for a use case when the disk is encrypted in k8s mode `volumeMode: Block`. Disk is open after attestation.

bpradipt:
> @stevenhorsman Thank you for quick response. Does the replacement support LUSK. This PR was build for a use case when the disk is encrypted in k8s mode `volumeMode: Block`. Disk is open after attestation.

It's in plan. Very happy to collaborate to get this quickly. Take a look at the CSI driver - https://github.com/confidential-devhub/caa-csi-block-driver

And the CAA side changes are here as Steve already pointed out https://github.com/confidential-containers/cloud-api-adaptor/pull/3037

Issue #3124, opened 2026-06-02
Tracking KubeVirt provider support


tarumizu:
This issue tracks the development progress. We have listed the development items below to add support for the KubeVirt provider. Each development item are submitted as a PR. If you have any suggestions or tasks related to adding support for the KubeVirt provider, please feel free to comment.

## Motivation

By supporting KubeVirt, we expect to enable secure container execution in more areas. We are eager to work on the necessary tasks to achieve this goal, particularly focusing on Arm CCA. We welcome any insights or collaboration from others who are also exploring this area.

## Development Item

We are planning to incrementally add support for the KubeVirt provider. The KubeVirt community is actively working on confidential computing support (SEV-SNP, TDX)[1], though both are currently registered as Alpha feature gates[2]. Our approach is to begin with non-CVM support and gradually integrate CVM support as development progresses.

[1] <https://kubevirt.io/user-guide/cluster_admin/confidential_computing/>
[2] <https://github.com/kubevirt/kubevirt/blob/v1.8.2/pkg/virt-config/featuregate/active.go#L250-L251>

### Implementation

* [ ] Adding initial support for KubeVirt
  * Add support for a new built-in provider by following the `addnewprovider.md`
* [ ] Integrate CVM support (such as SNP, TDX and SEV)
  * These items will be considered once the dependent features in KubeVirt are completed.

### Testing

* [ ] E2E tests for the KubeVirt provider

### Documentation

* [ ] How to build a Pod VM image

PR #3133, opened 2026-06-04
Add support for KubeVirt

Satoshi-Hirai371:
Summary:
This pull request allows cloud-api-adaptor(CAA) to support KubeVirt-based clouds.

Background:
There is currently no provider in CAA that supports clouds built with standard KubeVirt.  
By implementing an KubeVirt provider, we can offer secure pod execution environments in a wider range of fields.
As stated in the link below, we are focusing on implementations related to Arm CCA (Confidential Compute Architecture).  
This PR provides an architecture-independent implementation.  
Moving forward, we aim to execute PeerPods on KubeVirt clouds that support Arm CCA.

Links:
This PR is for the purpose and related work described in the following Issue.  
[#3124](https://github.com/confidential-containers/cloud-api-adaptor/issues/3124)

Task:
Refer to the following documentation to add a Provider for KubeVirt.
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md

- Add and initialize the KubeVirt provider manager
- Add a definition for the configuration struct
- Add cloud interfaces
- Add provider interfaces
- Add additional files to modularize the code
- Add configuration to the Helm chart configmap to include the contents of a specified file
- Add relevant unit tests
- Add relevant E2E tests
- Update entrypoint.sh and Makefile
- Added a README covering CAA deployment on KubeVirt, Pod VM creation, and e2e test steps 

Implemented:
- [x] Launch Peer Pod with standard VM
- [x] Unit tests and E2E tests
- [x] Manual connectivity verification  

Out of scope: Future implementation:
- How to Connect to PodVM Using an External IP Address
- Support for KubeVirt with Arm CCA support


bpradipt:
@Satoshi-Hirai371 can you please take a look at the copilot review comments

Satoshi-Hirai371:
Hi maintainers and reviewers,
Correcting feedback from AI/bot.

- Fixed the package imports in `test/tools/provisioner-cli/kubevirt.go`, which is used in e2e tests.
- Corrected the return value in `test/provisioner/kubevirt/provision_common.go`, which is used in e2e tests, when an error occurs.
- Ensured that the context passed from the `CreateInstance` method is included when issuing an API call to KubeVirt.
- We revised the function names in `cloud-providers/kubevirt/kubevirt.go` to better match their intended purposes.
- We corrected an incorrect namespace listed in `install/charts/peerpods/templates/configmap-filelist.yaml`.

Note that we determined no corrections were necessary for the other issues raised, for the following reasons:

- It was noted that the `providerConfigs` in `kubevirt.yaml` are invalid YAML and cause a failure, but since they are all commented out, there are no operational issues.
- It was noted that the test fails when the configuration file name passed to the kubevirt provider during e2e testing is changed because it is fixed; however, since the README specifies that the file name must be provided, this is not an issue.
- It was noted that `NewProvider` in `provider.go` does not perform a `nil` check; however, since `manager.go` is designed to pass the contents of `Config` as a constant, this is not an issue.
- It is noted that it would be safer not to use `tpl`, but we will retain the `tpl` conversion process to ensure that template syntax can continue to be expanded in the future.
- In `kubevirt.go`, the configuration file path is defined as a constant; however, since this is an internal value within the pod and serves the purpose of standardizing the placement of configuration files under `/etc/config/caa/kubevirt`, this is not an issue.

stevenhorsman:
Hi, we had a discussion on the community call about the proliferation of new providers and the challenges of having code dropped into the repo and then not maintained adding burden to a small group of maintainers, as such we are looking to move to an approach of keeping provider out of tree. We have some documentation for this approach in the new providers page: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/addnewprovider.md#memo-adding-support-for-a-new-external-provider. We admit that it might be outdated and might require more clarification, but if you could consider following this approach, rather than a built-in provider then we'd love the feedback. Thanks

Issue #2214, opened 2024-12-17
Security considerations of pull_request_target github action workflows


mkulke:
There are reports about recent [supply-chain attacks](https://blog.pypi.org/posts/2024-12-11-ultralytics-attack-analysis/) that use pull_request_target as an entry vector to comprise an open source project's binaries in ways that are very hard to prevent or even detect. 

There is clear advice against using foreign, untrusted code in a pull_request_target workflow that is running with the privileges and secrets of the project's repo. Our CI use case makes it hard to consider this advice, we want/have to run foreign code on private runners for architectures (s390x atm, maybe more later) not covered in github-hosted runners.

However, we can maybe carefully quarantine those s390x jobs into simple and heavily scrutinized workflows that will run in a pull_request_target context.

The amd64 jobs that we run on a PR should not require special privileges, they should be able to run in the context of the PR opener's fork. At the moment this is not possible:

## PodVM

artifacts are passed around as OCI images. you need write access to a registry for that, but I suspect that's an implementation detail that can be reworked to use the GH action artifact store.

## CAA Image

that's harder. since we need an actual OCI image to launch as daemonset. maybe we can use a [k8s-hosted](https://github.com/kyma-project/docker-registr) registry. problem: the test provisioner code wants to create the cluster and deploy CAA. the installation of the registry and the pushing of the CAA daemonset image would have to be somehow injected between both.

mkulke:
> - The token having write access means that if we execute arbitrary code (i.e. a pull request) then we could leak that secret and then an attacker could use it to cause issues
- The write token permission means that actions like poisoning caches, deleting artifacts? etc could be possible.

yes, executing untrusted code in the context of the project repo is just a large attack vector. there are ways to siphon secrets from those runs through various means that are hard to spot in a review (like string extrapolation). see [here](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)

![image](https://github.com/user-attachments/assets/5deba1e3-af54-4422-92c6-6c52efd53ef1)

We use one risky practice (pull_request_target) as a remedy for another [insecure practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#hardening-for-self-hosted-runners) (operating self-hosted runners on a public repo)

![image](https://github.com/user-attachments/assets/4a192f8e-4a58-4839-9c56-9da56a5a40cb)

[...]

PR #3143, opened 2026-06-10
fix(security): add security controls for pull_request_target in e2e workflow

rafsal-rahim:
Fixes dangerous pull_request_target trigger in e2e_on_pull.yaml workflow
by implementing comprehensive security gates that validate PR source and
require maintainer approval for high-risk scenarios.

Security Controls Added:
- Pre-execution security gate with fork detection
- Workflow file change detection
- Maintainer approval verification via GitHub API
- 4-tier security policy enforcement
- Comprehensive audit logging

The solution allows non-maintainers to create workflow PRs while ensuring
malicious code cannot execute with elevated permissions or access secrets.

Security Policy Matrix:
- Fork PRs: Always require maintainer approval
- Workflow changes: Always require maintainer approval
- Same-repo PRs (no workflow changes): Proceed safely

Assisted by Bob

### DIFF
--- a/.github/workflows/e2e_on_pull.yaml
+++ b/.github/workflows/e2e_on_pull.yaml
@@ -1,19 +1,27 @@
 # (C) Copyright Confidential Containers Contributors 2023.
 # SPDX-License-Identifier: Apache-2.0
 #
-# Run end-to-end (e2e) tests on pull request.
+# Privileged E2E tests requiring secrets and private runners
+# This workflow uses pull_request_target and requires maintainer approval
+#
+# For basic validation without secrets, see e2e_basic_validation.yaml
 ---
-name: e2e tests
+name: Privileged E2E tests (requires approval)
 
 on:
-  # Note on repository checkout: pull_request_target sets `GITHUB_SHA` to the
-  # "last commit on the PR base branch", meaning that by default `actions/checkout`
-  # is going to checkout the repository main branch. In order to pick up the pull
-  # request code, this workflow uses the `github.event.pull_request.head.sha`
-  # property to get the last commit on the HEAD branch. One limitation of this approach
-  # is that, unlike the `pull_request` event, the checked pull request isn't necessarily
-  # rebased to main (so it is up to users ensure the pull request is rebased **before*
-  # triggering this workflow).
+  # SECURITY NOTE: This workflow uses pull_request_target which runs with write
+  # permissions and access to secrets. It is designed to run code from PRs on
+  # project infrastructure with the following security measures:
+  #
+  # 1. Security gate validates PR source and checks for maintainer approval
+  # 2. Fork PRs ALWAYS require maintainer approval before running
+  # 3. Same-repo PRs with workflow changes require maintainer approval
+  # 4. Code is rebased atop main branch before execution
+  #
+  # For basic validation tests that run without approval, see e2e_basic_validation.yaml
+  #
+  # Note: pull_request_target sets GITHUB_SHA to the base branch commit.
+  # We explicitly checkout the PR head SHA to test the actual PR code.
   pull_request_target:
     types:
       # This workflow will be run if the pull request is labeled test_e2e_libvirt,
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e_on_pull.yaml
```diff
@@ -33,9 +33,131 @@ concurrency:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Get PR reviews from maintainers
+          REVIEWS=$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/reviews)
+          
+          # Check if any maintainer has approved
```
stevenhorsman: How is this checking that the approver is a maintainer?

.github/workflows/e2e_on_pull.yaml
```diff
@@ -33,9 +33,131 @@ concurrency:
...
+          if [ "${{ needs.security-gate.outputs.is-fork }}" == "true" ] && \
+             [ "${{ needs.security-gate.outputs.has-workflow-changes }}" == "true" ]; then
+            if [ "${{ needs.security-gate.outputs.approved }}" != "true" ]; then
+              echo "::error::❌ SECURITY: Fork PR with workflow changes requires maintainer approval"
+              exit 1
+            fi
+            echo "✅ Fork PR with workflow changes has maintainer approval"
+          fi
+          
+          # Case 2: PR from fork without workflow changes - MUST have approval
```
stevenhorsman: This is a superset of case 1, so what is case 1 providing?

.github/workflows/e2e_on_pull.yaml
```diff
@@ -33,9 +33,131 @@ concurrency:
...
 
 jobs:
+  # Security gate: Validates PR source and checks for maintainer approval
+  # This job runs BEFORE any code checkout to prevent malicious code execution
+  security-gate:
+    name: security-gate
+    runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: read
+      contents: read
```
stevenhorsman, line 51: Do we need the `contents: read` permission?

.github/workflows/e2e_on_pull.yaml
rafsal-rahim: yes, you are correct steve, this case can be optimized let me check how i can do that.

.github/workflows/e2e_on_pull.yaml
rafsal-rahim: We are using contents: read
- checkout the repository
- fetch git objects
- git diff --name-only - Needs the repository content to compare changes

mkulke (commented):
IMO we should split workflows that depend on pull-request-target (i.e. the ones that need private runners or secrets from the repo) from the ones that do not need it (i.e. the ones that can run on forks independently).

I understand it's a conscious choice to run code from PRs on the project infra, but we should limit the surface. afaict, there is no reason why common e2e tests are not executed in the context of the fork repo. apart from the security hygiene it's also constant source of confusion, because changes in the workflow itself cannot be reflected in pull-request-target runs (the workflow code will always be pulled from on main).

.github/workflows/e2e_on_pull.yaml
stevenhorsman: Ok, I thought it was only needed to read content from another repo, but maybe I'm mistaken

rafsal-rahim:
## Changes Summary

### 1. **New: Basic Validation Workflow** (`e2e_basic_validation.yaml`)

- **Runs automatically** on all PRs without approval
- **No secrets required** - builds locally, no registry push
- **Fast feedback** - validates builds and runs basic libvirt smoke tests
- **Trigger:** `pull_request` (safe for forks)

### 2. **New: Privileged-Only Workflow** (`e2e_privileged_only.yaml`)

- **Contains all 16 privileged jobs** (AWS, image builds, full E2E tests)
- **Requires secrets** - AWS credentials, registry passwords
- **Not directly triggerable** - only called by other workflows
- **Trigger:** `workflow_call` only

### 3. **Updated: PR Workflow** (`e2e_on_pull.yaml`)

- Now calls `e2e_privileged_only.yaml` instead of old workflow
- Security gates unchanged - still requires approval for fork PRs
- Privileged tests still need maintainer approval + E2E labels

### 4. **Updated: Daily Tests** (`daily-e2e-tests.yaml`)

- Runs **both workflows sequentially**:
  1. Basic validation first (fast validation)
  2. Privileged tests second (full E2E suite)
  
Does this changes look good or any update needed?

.github/workflows/e2e_on_pull.yaml
```diff
@@ -33,17 +41,207 @@ concurrency:
...
+    outputs:
+      is-fork: ${{ steps.check-fork.outputs.is-fork }}
+      has-workflow-changes: ${{ steps.check-workflow-changes.outputs.has-changes }}
+      approved: ${{ steps.check-approval.outputs.approved }}
+      approved-by: ${{ steps.check-approval.outputs.approved_by }}
+    steps:
+      - name: Check if PR is from a fork
```
stevenhorsman, line 58: What's the significance of checking a fork? Does this help with security.

.github/workflows/e2e_on_pull.yaml
```diff
@@ -33,17 +41,207 @@ concurrency:
...
+            echo "has-changes=true" >> $GITHUB_OUTPUT
+            echo "::warning::PR modifies workflow files - maintainer approval required"
+            echo "Changed workflow files:"
+            echo "$CHANGED_FILES" | grep "^\.github/workflows/" || true
+          else
+            echo "has-changes=false" >> $GITHUB_OUTPUT
+            echo "::notice::No workflow files modified"
+          fi
+
+      - name: Check for maintainer approval
```
stevenhorsman, line 94: Does the e2e_ labels not provide assurance that a committer has authorised it? Maybe we could try and add robustness that labels get removed on update?

.github/workflows/e2e_run_all.yaml
```diff
@@ -57,45 +56,46 @@ jobs:
       arch: amd64
       debug: true
     permissions:
-      contents: read  # Required to checkout code
-      packages: write  # Required to push container images
-      id-token: write  # Required for OIDC authentication
-      attestations: write  # Required to write attestations
-      artifact-metadata: write  # Required by podvm_mkosi.yaml to write attestation metadata
```
stevenhorsman, line 64: Why have you removed these zizmor required comments?

.github/workflows/e2e_run_all.yaml
```diff
@@ -253,55 +253,54 @@ jobs:
...
-      podvm_image: ${{ needs.podvm_ubuntu_amd64.outputs.qcow2_oras_image }}
+      runner: s390x-large
+      caa_image: ${{ inputs.registry }}/cloud-api-adaptor:${{ inputs.caa_image_tag }}-dev-s390x
+      podvm_image: ${{ needs.podvm_mkosi_s390x.outputs.qcow2_oras_image }}
       git_ref: ${{ inputs.git_ref }}
-      peerpod_ctrl_image: ${{ inputs.registry }}/peerpod-ctrl:${{ inputs.caa_image_tag }}-amd64
+      peerpod_ctrl_image: ${{ inputs.registry }}/peerpod-ctrl:${{ inputs.caa_image_tag }}-s390x
       webhook_image: ${{ inputs.registry }}/peer-pods-webhook:${{ inputs.caa_image_tag }}
     secrets:
       REGISTRY_CREDENTIAL_ENCODED: ${{ secrets.REGISTRY_CREDENTIAL_ENCODED }}
```
stevenhorsman, line 294: Why have you re-ordered jobs that has created a huge diff?

stevenhorsman (commented):
Rafsal - I appreciate you trying to help this problem, but I think we need to try and work through a proper agreed design for this in https://github.com/confidential-containers/cloud-api-adaptor/issues/2214 rather than just throwing AI at it and hoping for the best or we are going to have too many rounds of iteration on this. If you'd like to propose a design in there that would be great and it sounds like @mkulke has some ideas you could base it on like you'd started here. Thanks

PR #3145, opened 2026-06-11
ci: Temporarily ignore the pull_request_target warning

stevenhorsman:
As this warning is the last zizmor issue we have, should we just ignore it as we have an issue to track this and that means if there are any other issues introduced we are more likely to notice them rather than the zizmor check always being failing?


wainersm (approved):
I agree with you @stevenhorsman !

mkulke (approved):

Issue #3147, opened 2026-06-12
PodVM | TDX RTMR Support


Xynnn007:
## Background

Now the podvm turns to be ubuntu. I test myself and find that current ubuntu does not support TDX RTMR for ubuntu base because I met the failure when using ASR's rtmr api, with AA enabling eventlog via initdata. I suspect it's caused by ubuntu's guest kernel version is lower than 6.16 and does not have backport patches.

## What do we want?

We alibaba cloud try to use a podvm that supports TDX RTMR. And also hopes that the CI would cover latest changes upon the code.

## Options and pro/cons

There might be two ways

#### Way 1: Upgrade to `ubuntu 26.04` for the repo.

Thanks to @stevenhorsman for the idea. ~~Let me confirm if `ubuntu 26.04` with `7.0` kernel version has support for TDX RTMR.~~

It works on ubuntu 26.04.

But needs folks to check if the updation is acceptable.

#### Way 2: Introduce Alinux4 to upstream

Alinux4 is an OS released by alibaba cloud and now with a kernel that supports TDX RTMR. 
The embarrasing thing is that I do not think other CSPs would use it tbo, and the maintainer would be alibaba cloud.

So way 2 means the repo needs to accept

1. bring alinux4 mkosi to caa official repo

Xynnn007:
Update: It works for ubuntu 26.04.

PR #3154, opened 2026-06-15
podvm: Switch Ubuntu  image to 26.04 (Resolute)

stevenhorsman:
Updated mkosi configuration and documentation to support Ubuntu 26.04 (Resolute) as the base image for podvm builds. This enables building podvm images on the latest Ubuntu LTS release.

Generated-By: IBM Bob

### DIFF
--- a/src/cloud-api-adaptor/podvm/hack/smoke_test.sh
+++ b/src/cloud-api-adaptor/podvm/hack/smoke_test.sh
@@ -258,7 +258,7 @@ fi
 
 sudo virt-install \
 	--name "${VM_NAME}" \
-	--ram 1024 \
+	--ram 2048 \
 	--vcpus 2 \
 	--disk "path=${IMAGE},format=${FMT}" \
 	--disk "path=${WORKDIR}/cloud-init.iso,device=cdrom" \
--- a/src/cloud-api-adaptor/podvm/hack/smoke_test.sh
+++ b/src/cloud-api-adaptor/podvm/hack/smoke_test.sh
@@ -237,8 +237,16 @@ chmod a+rwx "${WORKDIR}"
 sudo chown -R libvirt-qemu "${WORKDIR}" || sudo chown -R qemu "${WORKDIR}" || true
 sudo chmod +x "${WORKDIR}"
 
-# Resize the VM disk image to add free space
+# Resize the VM disk image to add free space, then fix the GPT backup
+# header which qemu-img resize leaves at the wrong position. Without
+# this, systemd-repart silently skips partition creation (no trusted_store
+# partition appears) and scratch-storage.service cannot find its device.
+# sgdisk requires a block device, so expose the qcow2 via NBD first.
 sudo qemu-img resize -f "${FMT}" "${IMAGE}" +1G
+sudo modprobe nbd max_part=8
+sudo qemu-nbd --connect=/dev/nbd0 "${IMAGE}"
+sudo sgdisk -e /dev/nbd0
+sudo qemu-nbd --disconnect /dev/nbd0
 
 # Start the VM
 echo "::debug:: Starting VM for test mode: $TEST_MODE"
--- a/src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.repart/20-root-verity.conf
+++ b/src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.repart/20-root-verity.conf
@@ -2,5 +2,5 @@
 Type=root-verity
 Verity=hash
 VerityMatchKey=root
-SizeMinBytes=64M
-SizeMaxBytes=64M
+SizeMinBytes=72M
+SizeMaxBytes=72M
--- a/.github/workflows/podvm_smoketest.yaml
+++ b/.github/workflows/podvm_smoketest.yaml
@@ -76,6 +76,7 @@ jobs:
     needs: build
 
     strategy:
+      fail-fast: false
       matrix:
         test-mode:
           - name: podvm-ubuntu
--- a/.github/workflows/podvm_smoketest.yaml
+++ b/.github/workflows/podvm_smoketest.yaml
@@ -92,8 +92,9 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y \
+            gdisk \
             genisoimage \
-            qemu-utils\
+            qemu-utils \
             socat \
             virt-manager
 
--- a/src/cloud-api-adaptor/podvm/mkosi.conf
+++ b/src/cloud-api-adaptor/podvm/mkosi.conf
@@ -10,5 +10,5 @@ OutputDirectory=build
 
 [Distribution]
 Distribution=ubuntu
-Release=noble
+Release=resolute
 Repositories=universe
--- a/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
+++ b/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
@@ -14,3 +14,7 @@ Packages=systemd
          util-linux
          kmod
          tpm2-tools
+         systemd-cryptsetup
+KernelInitrdModules=default
+                    dm-verity
+                    dm-crypt
--- a/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
+++ b/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
@@ -14,5 +14,7 @@ Packages=systemd
          util-linux
          kmod
 KernelInitrdModules=default
+                    dm-mod
+                    dm-bufio
                     dm-verity
                     dm-crypt
--- a/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
+++ b/src/cloud-api-adaptor/podvm/mkosi.images/initrd/mkosi.conf
@@ -13,8 +13,6 @@ Packages=systemd
          udev
          util-linux
          kmod
-         tpm2-tools
-         systemd-cryptsetup
 KernelInitrdModules=default
                     dm-verity
                     dm-crypt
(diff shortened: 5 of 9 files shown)


mkulke (dismissed):

src/cloud-api-adaptor/podvm/files/etc/tmpfiles.d/kata-containers.conf
```diff
@@ -0,0 +1,3 @@
+# Create /kata-containers directory for bind mount to /run/kata-containers
+# This directory is used by run-kata-containers.mount systemd unit
+d /kata-containers 0755 root root -
```
mkulke: can you elaborate why this is required as part of the update?

src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.repart/20-root-verity.conf
```diff
@@ -2,5 +2,5 @@
 Type=root-verity
 Verity=hash
 VerityMatchKey=root
-SizeMinBytes=64M
-SizeMaxBytes=64M
+SizeMinBytes=72M
+SizeMaxBytes=72M
```
mkulke, line 6: how come we have to increase the verity partition?

src/cloud-api-adaptor/podvm/files/etc/tmpfiles.d/kata-containers.conf
stevenhorsman: I'm not sure it this - this is still WIP - but the issue we were seeing is that the scratch-space test failed with `Failed to mount run-kata-containers.mount` so this is to try and explicity create it at boot time to see if that makes a difference

src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.repart/20-root-verity.conf
stevenhorsman: We have the failure:
```
Partition 2's contents (66.1M) don't fit in the partition (64.0M)
```
so it seemed like the verity data in 26.04 has grown due to a larger dm-verity has tree due to larger file system?

mkulke: that makes sense

mkulke:
@stevenhorsman sorry, copilot went ahead and created a commit on top of yours. I guess it does that because the source branch is not on a fork. I'm not convinced it's the right fix, sounds handwavy. We can test and revert if it doesn't help.

stevenhorsman:
> @stevenhorsman sorry, copilot went ahead and created a commit on top of yours. I guess it does that because the source branch is not on a fork. I'm not convinced it's the right fix, sounds handwavy. We can test and revert if it doesn't help.

Thanks for highlighting. I originally made it as a branch as I thought we might end up needed a workflow change, but it doesn't look like it. Good to know when copilot will jump in!

mkulke:
> > @stevenhorsman sorry, copilot went ahead and created a commit on top of yours. I guess it does that because the source branch is not on a fork. I'm not convinced it's the right fix, sounds handwavy. We can test and revert if it doesn't help.
> 
> Thanks for highlighting. I originally made it as a branch as I thought we might end up needed a workflow change, but it doesn't look like it. Good to know when copilot will jump in!

i'm pretty sure the agent fix is based overconfidence and unproven assumptions, I think we can remove the commit

stevenhorsman:
> i'm pretty sure the agent fix is based overconfidence and unproven assumptions, I think we can remove the commit

The problem seems to have got worse since I reverted it, so I will try and re-add it (which some extras)

stevenhorsman:
OK - it took about five days to get the smoke tests passing with 26.04, but I think it's stable now. A big disclaimer in that I don't fully understand mkosi, or all the kernel modules we are using, so this was a lot of trial-and-error from me and AI, so if someone has actual expertise then a careful review would be appreciated.

src/cloud-api-adaptor/podvm/hack/smoke_test.sh
```diff
@@ -237,8 +237,16 @@ chmod a+rwx "${WORKDIR}"
 sudo chown -R libvirt-qemu "${WORKDIR}" || sudo chown -R qemu "${WORKDIR}" || true
 sudo chmod +x "${WORKDIR}"
 
-# Resize the VM disk image to add free space
+# Resize the VM disk image to add free space, then fix the GPT backup
```
mkulke, line 240: odd that this was not required on 24.04 or fedora?

I am not sure we can use this feature then for cloud storage, we cannot fixup the headers there the disks will just be increased and the guest os is responsible for claiming that space.

mkulke (commented):
do we know that the scratch-space feature is working on the current head with ubuntu 24.04? at several points in the past I attempted to use it and never had success, might be me, but there have been always errors in the startup log. that is maybe something that we should assert in the smoketest (systemctl list-units --failed should be empty or something).

If it is currently not functional in in 24.04 I would opt to remove it from the tests and the code and go back to the drawing board.

stevenhorsman:
> do we know that the scratch-space feature is working on the current head with ubuntu 24.04? at several points in the past I attempted to use it and never had success, might be me, but there have been always errors in the startup log. that is maybe something that we should assert in the smoketest (systemctl list-units --failed should be empty or something).
> 
> If it is currently not functional in in 24.04 I would opt to remove it from the tests and the code and go back to the drawing board.

Oh - I had assumed that given that the CI tests were passing then it was working, if not that would have saved a lot of time trying to get it passing on 26.04 😞

src/cloud-api-adaptor/podvm/hack/smoke_test.sh
stevenhorsman: Let me try and revert in case the later fix allows it to get past this, but IIRC I tried to revert it in my local test yesterday and it failed without this change.

stevenhorsman: Ok, looks like it isn't needed after all. I'm not sure what I did wrong yesterday then 🤷

Xynnn007:
btw this PR can also close https://github.com/confidential-containers/cloud-api-adaptor/issues/3147

bpradipt:
> OK - it took about five days to get the smoke tests passing with 26.04, but I think it's stable now. A big disclaimer in that I don't fully understand mkosi, or all the kernel modules we are using, so this was a lot of trial-and-error from me and AI, so if someone has actual expertise then a careful review would be appreciated.

I'll try locally and update the results.

src/cloud-api-adaptor/podvm/hack/smoke_test.sh
```diff
@@ -166,7 +166,7 @@ if [ -z "$KATACTL" ]; then
 	KATA_REG=$(yq -e '.oci.kata-containers.registry' ${SCRIPTDIR}/../../versions.yaml)
 	echo "::debug:: Pulling kata-ctl from ${KATA_REG}/agent-ctl:${KATA_REF}-x86_64"
 	oras pull "${KATA_REG}/agent-ctl:${KATA_REF}-x86_64"
-	tar --ztsd -xvf kata-static-agent-ctl.tar.zst ./opt/kata/bin/kata-agent-ctl --transform='s/opt\/kata\/bin\/kata-agent-ctl/kata-agent-ctl/'
+	tar --zstd -xvf kata-static-agent-ctl.tar.zst ./opt/kata/bin/kata-agent-ctl --transform='s/opt\/kata\/bin\/kata-agent-ctl/kata-agent-ctl/'
```
Xynnn007: Looks like this is only re-oreder?

stevenhorsman: Yes - you are right - this is some AI slop, so I will tidy that up! Sorry I missed it first time around

mkulke (approved):
lgtm

bpradipt (approved):

Xynnn007 (approved):

stevenhorsman:
Thanks for the reviews. I think that, given we are in the release process, that it's worth completely that first rather then (he small risk given the testing, pushing this mid-release?

bpradipt:
> Thanks for the reviews. I think that, given we are in the release process, that it's worth completely that first rather then (he small risk given the testing, pushing this mid-release?

Didn't fully understand the statement.

stevenhorsman:
> Didn't fully understand the statement.

I want to wait until 0.22 has finalised to merge this, rather than merging during the release period

stevenhorsman:
Lastest code changes run through daily e2e tests as a branch: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/30443154883 AWS is failing, with:
```
E0729 10:43:34.830816   21544 env.go:414] Setup failure: operation error EC2: ImportSnapshot, https response error StatusCode: 400, RequestID: b5a631b6-fc5d-488a-a6f2-d641dcca3b0e, api error InvalidParameter: The service role ci-caa-30443154883-1-vmimport provided does not exist or does not have sufficient permissions.
```
but that's the same in repeated PRs and dailies, so ignoring it :(

src/cloud-api-adaptor/podvm/hack/smoke_test.sh
```diff
@@ -237,8 +237,16 @@ chmod a+rwx "${WORKDIR}"
...
+# Resize the VM disk image to add free space, then fix the GPT backup
+# header which qemu-img resize leaves at the wrong position. Without
+# this, systemd-repart silently skips partition creation (no trusted_store
+# partition appears) and scratch-storage.service cannot find its device.
+# sgdisk requires a block device, so expose the qcow2 via NBD first.
 sudo qemu-img resize -f "${FMT}" "${IMAGE}" +1G
+sudo modprobe nbd max_part=8
+sudo qemu-nbd --connect=/dev/nbd0 "${IMAGE}"
+sudo sgdisk -e /dev/nbd0
+sudo qemu-nbd --disconnect /dev/nbd0
```
ldoktor, line 249: Interesting, this sounds like a `qemu-img resize` bug to me. Have you tried raising a bug against it?

ldoktor: OK I tried asking our v2v expert Rich and he said virt-resize might or might not help, then he pointed me to https://manpages.debian.org/testing/systemd/systemd-growfs.8.en.html which should. Not sure if I find some time to look into it since you already find a workaround...

Xynnn007:
One question: do we need to update guest-components to release on the new base OS?

stevenhorsman:
> One question: do we need to update guest-components to release on the new base OS?

We didn't seem to need to, so maybe the glibc versions are compatible?

PR #3155, opened 2026-06-19
Add LUKS encryption support for cloud volumes via CDH

mohammedadnan21:
## Summary

Adds LUKS2 encryption support for cloud volumes passed through to Peer Pods. When a StorageClass specifies `encrypt-type: LUKS` and `kbs-key-id`, the interceptor inside the PodVM calls CDH's `secure_mount` TTRPC API instead of raw `formatAndMount`. CDH handles LUKS formatting/opening and fetches the encryption key from KBS after remote attestation.

This builds on the volume passthrough merged in #3037.

## Changes

- **pkg/util/cloud.go** — Extend `CloudVolumeAnnotation` with `EncryptType` and `KeyID` fields
- **pkg/adaptor/proxy/service.go** — Read `encrypt-type` and `kbs-key-id` from `mountInfo.json` metadata and pass in OCI annotation
- **pkg/forwarder/interceptor/cdh_client.go** — Lightweight CDH TTRPC client (uses existing `containerd/ttrpc` dependency)
- **pkg/forwarder/interceptor/interceptor.go** — Branch on `EncryptType`: encrypted volumes go through CDH, plain volumes use existing `formatAndMount`; `isLuks` detection for first-mount vs re-mount; LUKS cleanup via `cryptsetup luksClose`
- **Tests** — Unit tests for annotation flow, CDH request construction, and encrypted mount tracking

## How it works

```
StorageClass (encrypt-type, kbs-key-id)
  → CSI Driver writes mountInfo.json with metadata
    → CAA Proxy reads metadata, sets cloud_volumes annotation
      → Interceptor in PodVM receives annotation
        → if encrypt_type set: call CDH secure_mount (LUKS + KBS key)
        → else: formatAndMount (existing path)
```

CDH determines whether to format a fresh disk or open an existing LUKS container by checking `cryptsetup isLuks`.

## What doesn't change

- CSI driver `NodePublishVolume` — encryption params already flow to `mountInfo.json`
- Cloud providers (AWS/Azure) — disk attachment is the same
- PodVM image — CDH and cryptsetup already installed
- KBS/attestation — existing provisioning handles this

## Test plan

- [x] Unit tests pass for annotation serialization with encryption fields
- [x] Unit tests pass for CDH request format verification
- [x] Unit tests pass for encrypted mount cleanup tracking
- [x] `go build ./...` passes
- [x] `gofmt` clean
- [ ] E2e: create PVC with encrypted StorageClass, verify data persists across pod restarts

Closes: confidential-devhub/caa-csi-block-driver#16


bpradipt:
@mohammedadnan21 this will need a rebase

mohammedadnan21:
Rebased onto latest `main` to pick up dependency fixes and resolve the `govulncheck` CI failure. All other checks were already passing.

@bpradipt — ready for review when you get a chance. Would appreciate e2e test trigger once CI is green.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -779,3 +830,69 @@ func formatAndMount(device, mountPoint, fsType string) error {
 	logger.Printf("Mounted %s at %s (type=%s)", device, mountPoint, fsType)
 	return nil
 }
+
+// isLuks checks if a device already has a LUKS header.
+func isLuks(device string) bool {
+	err := exec.Command("cryptsetup", "isLuks", device).Run()
```
bpradipt: No native go func for this check ?
Additionally should any errors be logged here ?

mohammedadnan21: Good catch. There's no standard Go library just for LUKS detection — the existing ones are full disk management libraries which would be overkill here. But LUKS has a fixed 6-byte magic signature (`LUKS\xba\xbe`) at the start of the device that hasn't changed since 2004, so I'll replace the exec call with a direct read of those bytes. This also lets me return a proper error so failures are logged instead of silently treated as "not LUKS". Will push the fix.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -820,9 +820,32 @@ func formatAndMount(device, mountPoint, fsType string) error {
...
 }
 
-func isLuks(device string) bool {
-	err := exec.Command("cryptsetup", "isLuks", device).Run()
-	return err == nil
+// luksMagic is the 6-byte signature at the start of every LUKS1/LUKS2 volume.
+// Defined in the LUKS on-disk format specification (unchanged since 2004).
+var luksMagic = []byte{'L', 'U', 'K', 'S', 0xba, 0xbe}
+
+func isLuks(device string) (bool, error) {
```
bpradipt, line 828: nit:  this commit should get squashed with the original one in this series

bpradipt (approved):
I think we can merge this for initial support with CI in place.

src/cloud-api-adaptor/pkg/forwarder/interceptor/cdh_client.go
```diff
@@ -21,8 +21,6 @@ type cdhClient struct {
 	client *ttrpc.Client
 }
 
-// newCDHClient connects to CDH's TTRPC socket with retries, since CDH
-// may still be starting when the interceptor first needs it.
```
snir911: the containerd bump doesn't seems to be a prt of this commit.

Also IMO you can either keep the comments or just squash it with the relevant commit

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -779,3 +830,94 @@ func formatAndMount(device, mountPoint, fsType string) error {
...
+func secureMount(ctx context.Context, device, mountPoint, fsType, encryptType, keyID, mapperName string) error {
+	normalized, err := validateEncryptParams(encryptType, keyID)
+	if err != nil {
+		return err
+	}
+
+	sourceType := "empty"
+	luks, err := isLuks(device)
+	if err != nil {
+		logger.Printf("WARNING: LUKS detection failed for %s, assuming empty: %v", device, err)
```
snir911: is it possible to wrongly set empty sourceType because an encrypted volume will return some other error?

Actually, i'd except the CDH to check the luks header instead of here, as it's the one in charge of encryption, I'm assuming it's not the case..

src/cloud-api-adaptor/pkg/forwarder/interceptor/cdh_client.go
```diff
@@ -0,0 +1,73 @@
...
+
+	client := ttrpc.NewClient(conn)
+	return &cdhClient{conn: conn, client: client}, nil
+}
+
+func (c *cdhClient) close() {
+	if err := c.client.Close(); err != nil {
+		logger.Printf("WARNING: closing CDH ttrpc client: %v", err)
+	}
+	if err := c.conn.Close(); err != nil {
```
snir911: isn't this always call always fails? as the ttrpc connection is already closed..

snir911 (commented):
Thanks @mohammedadnan21, i have added few minor comments

src/cloud-api-adaptor/pkg/forwarder/interceptor/cdh_client.go
mohammedadnan21: Fair point — the containerd bump was just from the rebase, not an actual change in my diff. Squashed everything into 2 commits now: proto generation + feature.

src/cloud-api-adaptor/pkg/forwarder/interceptor/cdh_client.go
mohammedadnan21: Yep, you're right. Checked the ttrpc source — Client.Close() already closes the underlying conn internally. Removed the redundant conn.Close() and the field itself.

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
mohammedadnan21: Good catch. Verified in CDH source — it doesn't do LUKS detection itself, trusts sourceType directly. Changed this to a hard error so we never proceed with an uncertain state.

src/cloud-api-adaptor/pkg/forwarder/interceptor/cdh_client.go
snir911: This is valid of course, however, actually I'd consider just switching the order of calls

mohammedadnan21: Agreed, switched the order — conn.Close() first, then client.Close(). Checked the ttrpc source and Client.Close() handles the already-closed conn fine (its internal closed() is just a context cancel, doesn't need the conn open).

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
```diff
@@ -779,3 +819,88 @@ func formatAndMount(device, mountPoint, fsType string) error {
...
+	}
+	defer f.Close()
+
+	header := make([]byte, 6)
+	n, err := f.Read(header)
+	if err != nil {
+		return false, fmt.Errorf("reading LUKS header from %s: %w", device, err)
+	}
+	if n < 6 {
+		return false, nil
```
snir911: nit: does it necessarily mean it isn't LUKS? I guess it's unlikely you'll reach here, however, if you want to make it more robust, I think there should be a ReadAll function that should either read whole 6 or fail.

snir911 (approved):
LGTM , thanks @mohammedadnan21 , I have added a minor additional comment for your consideration.
make sure to rebase

src/cloud-api-adaptor/pkg/forwarder/interceptor/interceptor.go
mohammedadnan21: switched to io.ReadFull so a short read is a hard error instead of silently returning false.

bpradipt:
Given tests are all green, I'm merging this one.

PR #3163, opened 2026-06-24
libvirt: add configurable root disk size

### LINKED ISSUE #3162: libvirt: add configurable root disk size for podVM
## Problem

The libvirt provider hardcodes the root disk size to 10 GiB in `CreateDomain()`:

```go
v.rootDiskSize = uint64(10)  // libvirt.go:562
```

The internal `rootDiskSize` field exists in `vmConfig` but is never wired to configuration, so there is no way to override it at runtime.

Every other provider already exposes this:

| Provider | Flag | Env var | Default |
|---|---|---|---|
| AWS | `--root-volume-size` | `ROOT_VOLUME_SIZE` | 30 GiB |
| Azure | `--root-volume-size` | `ROOT_VOLUME_SIZE` | 0 (image default) |
| GCP | `--root-volume-size` | `ROOT_VOLUME_SIZE` | 10 GiB |
| Alibaba | `--system-disk-size` | `SYSTEM_DISK_SIZE` | 40 GiB |

## Proposed change

- Add `RootDiskSize int` to the `Config` struct in `types.go`
- Register `--root-disk-size` / `LIBVIRT_ROOT_DISK_SIZE` in `manager.go` (default: `10`, units: GiB)
- Wire it through `CreateInstance` in `provider.go` into `vmConfig.rootDiskSize`
- If unset or 0, fall back to 10 GiB (no behaviour change for existing deployments)
- The existing `createVolume` logic in `volume.go` already handles the case where the backing image is larger than the requested size — no change needed there

## Expected benefit

Operators can size the root disk to fit their workloads without rebuilding the podVM image. This brings libvirt to parity with all other supported providers.

ajaypvictor:
The libvirt provider hardcodes the root disk size to 10 GiB with no way to override it at runtime. Every other provider (AWS, Azure, GCP, Alibaba) already exposes a configurable root volume size. This brings libvirt to parity by adding a --root-disk size flag (env: LIBVIRT_ROOT_DISK_SIZE, default: 10 GiB). Deployments that do not set the variable get the same behaviour as before.

This also fixes a pre-existing unit mismatch in createVolume(): the volume size parameter was compared directly against the backing image capacity in bytes without converting GiB to bytes first, making any non-default size produce the wrong result. The parameter is now explicitly converted using math/bits.Mul64, which also detects overflow for pathologically large inputs and returns a clean error instead of silent integer wraparound.

The defaulting and capacity sizing logic is extracted into pure helper functions (normalizeRootDiskSize, volumeCapacityBytes) so they can be unit tested without a live libvirt connection. The config-extractor AST tool is updated to recognise Uint64WithEnv registrations and to unwrap typed numeric constants like uint64(10) so that libvirt.yaml is regenerated correctly.

Assisted-by: IBM Bob \<noreply@ibm.com\>
Fixes: #3162

### DIFF
--- a/src/cloud-providers/libvirt/volume.go
+++ b/src/cloud-providers/libvirt/volume.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/bits"
 	"time"
 
 	libvirt "libvirt.org/go/libvirt"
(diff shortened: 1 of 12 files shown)


src/cloud-providers/libvirt/volume.go
```diff
@@ -182,7 +183,37 @@ func newCopier(conn *libvirt.Connect, volume *libvirt.StorageVol, size uint64) f
...
+// guard can be unit-tested independently of libvirt.
+func volumeCapacityBytes(volSizeGiB, backingCapacityBytes uint64) (uint64, error) {
+	hi, requested := bits.Mul64(volSizeGiB, bytesPerGiB)
+	if hi != 0 {
+		return 0, fmt.Errorf("root-disk-size %d GiB overflows uint64 when converted to bytes", volSizeGiB)
+	}
+	if backingCapacityBytes > requested {
+		return backingCapacityBytes, nil
+	}
+	return requested, nil
```
stevenhorsman: nit: I think using `max` would be cleaner?
```suggestion
	return max(requested, backingCapacityBytes), nil
```

stevenhorsman (commented):
One suggestion and the go fmt issues needs fixing. Thanks

src/cloud-providers/libvirt/volume.go
ajaypvictor: makes total sense, thanks for the suggestion..

stevenhorsman (changes requested):
The go fmt error has still not been addressed?

ajaypvictor:
> The go fmt error has still not been addressed?

apologies, something was wrong on my setup, I think it's fixed now.. thanks..

stevenhorsman (approved):
LGTM. Thanks

ajaypvictor:
> LGTM. Thanks

Would you mind adding a libvirt e2e label on this?

BbolroC (approved):
LGTM as far as the CI is green. Thanks @ajaypvictor !

PR #3174, opened 2026-07-02
ci: add govulncheck ignore list support

stevenhorsman:
Adds .govulncheck-ignore.toml (matching the format of osv-scanner's ignore list) to suppress known false-positive vulnerability IDs, and extends hack/govulncheck.sh to apply it via hack/govulncheck-filter.py when the file is present.

Each [[IgnoredVulns]] entry requires an id and reason. An optional ignoreUntil date causes the entry to warn rather than silently suppress once it has passed.

Seeded with three containerd CRI checkpoint/restore vulns with no upstream fix that are not exploitable in CAA:
- GO-2026-5622: arbitrary host CRI log file read via symlink
- GO-2026-5338: CRI checkpoint import image tag poisoning
- GO-2026-5064: CRI checkpoint restore CDI annotation smuggling

Generated-By: IBM Bob

### DIFF
--- a/.govulncheck-ignore.toml
+++ b/.govulncheck-ignore.toml
@@ -1,0 +1,35 @@
+# .govulncheck-ignore.toml
+#
+# Ignore list for hack/govulncheck.sh.
+# Add [[IgnoredVulns]] entries to suppress specific govulncheck vulnerability
+# IDs that are known false-positives or have no upstream fix available yet.
+#
+# Format:
+#
+#   [[IgnoredVulns]]
+#   id = "GO-YYYY-NNNN"
+#   reason = "Why this is suppressed"
+#
+# Optionally add ignoreUntil (TOML local date). An expired entry still suppresses
+# the finding but prints a warning so the team knows to review it:
+#
+#   [[IgnoredVulns]]
+#   id = "GO-YYYY-NNNN"
+#   reason = "Waiting for upstream patch"
+#   ignoreUntil = 2026-12-31
+
+# All three vulns below affect github.com/containerd/containerd@v1.7.33 via the
+# CRI checkpoint/restore feature. CAA does not expose that feature, so they are
+# not exploitable in this codebase. No fix is available upstream at this time.
+
+[[IgnoredVulns]]
+id = "GO-2026-5622"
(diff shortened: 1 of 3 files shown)


stevenhorsman:
Hi folks, I tried implementing this in python (only stdlibs) and plain shell, but shell needed yq/tomlq and jq, so I picked the python, but I'm happy to switch to shell approach, or golang if people would prefer.

mkulke:
> Hi folks, I tried implementing this in python (only stdlibs) and plain shell, but shell needed yq/tomlq and jq, so I picked the python, but I'm happy to switch to shell approach, or golang if people would prefer.

thanks! python is a fair choice here, IMO. Maybe we should have a second, non-json run if the first one w/ json output fails (i.e. there are vulns that we do not ignore), so it's easier to read the report?

stevenhorsman:
> > Hi folks, I tried implementing this in python (only stdlibs) and plain shell, but shell needed yq/tomlq and jq, so I picked the python, but I'm happy to switch to shell approach, or golang if people would prefer.
> 
> thanks! python is a fair choice here, IMO. Maybe we should have a second, non-json run if the first one w/ json output fails (i.e. there are vulns that we do not ignore), so it's easier to read the report?

Ok, this is implemented and shows something interesting:
```
./hack/govulncheck.sh
[IGNORED] GO-2026-5064 — containerd CRI checkpoint restore CDI annotation smuggling. Not exploitable in CAA: CRI checkpoint/restore is not exposed.
[IGNORED] GO-2026-5338 — CRI checkpoint import allows local image tag poisoning in containerd. Not exploitable in CAA: CRI checkpoint/restore is not exposed.
[IGNORED] GO-2026-5622 — Arbitrary host CRI log file read via symlink in containerd. Not exploitable in CAA: CRI checkpoint/restore is not exposed.
Vulnerability: GO-2026-4762
Vulnerability: GO-2026-4762

Re-running govulncheck for human-readable output (non-ignored findings in ./src/webhook):

=== Symbol Results ===

No vulnerabilities found.

Your code is affected by 0 vulnerabilities.
```

so it looks like the json output shows more than the normal one. Looking into this more...

stevenhorsman:
So it looks like the json mode reports package and module findings not just the callable code, which is why we get more results, so I will try and update to filter that out too... Done

mkulke (approved):
thanks!

bpradipt (approved):
nice..

PR #3182, opened 2026-07-06
PodVM | Upgrade ubuntu to 26.04

### LINKED ISSUE #3147: PodVM | TDX RTMR Support

Xynnn007:
This PR updates base image of ubuntu from 24.04 to 26.04.

Fixes #3147

### DIFF
--- a/src/cloud-api-adaptor/Makefile.defaults
+++ b/src/cloud-api-adaptor/Makefile.defaults
@@ -82,7 +82,7 @@ done), \
 	$(error $(1) is not set, and could not be automatically set from $(VERSIONS_SRC)))
 endef
 
-ubuntu_RELEASE ?= noble
+ubuntu_RELEASE ?= resolute
 rhel_RELEASE ?= 9
 alinux_RELEASE ?= 3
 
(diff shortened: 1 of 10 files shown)


Xynnn007:
duplicated with https://github.com/confidential-containers/cloud-api-adaptor/pull/3154

PR #3186, opened 2026-07-08
libvirt: close libvirt connection on provider teardown

### LINKED ISSUE #1252: libvirt: teardown of the libvirt manager should close the connection
The https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/libvirt/manager.go#L43 method creates a new instance of a `Provider` and as state of this provider it establishes a connection to the libvirt host.

This connection is never closed, it should at least be closed in the `Teardown` method of the manager.

https://github.com/confidential-containers/cloud-api-adaptor/blob/main/pkg/adaptor/cloud/libvirt/provider.go#L98

ajaypvictor:
The libvirtProvider opens a persistent connection to the libvirt host in NewProvider() via NewLibvirtClient(). This connection was held for the entire CAA daemon lifetime but never closed — Teardown() returned nil unconditionally, leaking the connection on the libvirt host side until its timeout.

Fix Teardown() to:

- Guard against a nil libvirtClient in case of a failed provider initialisation
- Call connection.Close() and propagate any error
- Return an error if Close() returns a positive reference count, indicating the connection was not fully released

Replace the hollow TestTeardown (which constructed a nil provider and only proved nil returns nil) with a real integration test that verifies the connection is alive before Teardown() and dead after it. The test skips automatically when LIBVIRT_URI is not set, consistent with all other integration tests in the package.

Fixes: https://github.com/confidential-containers/cloud-api-adaptor/issues/1252

Assisted-by: IBM Bob <noreply@ibm.com>

### DIFF
--- a/src/cloud-providers/libvirt/provider.go
+++ b/src/cloud-providers/libvirt/provider.go
@@ -154,6 +154,16 @@ func (p *libvirtProvider) DeleteInstance(ctx context.Context, instanceID string)
 }
 
 func (p *libvirtProvider) Teardown() error {
+	if p.libvirtClient == nil {
+		return nil
+	}
+	refCount, err := p.libvirtClient.connection.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close libvirt connection: %w", err)
+	}
+	if refCount > 0 {
+		return fmt.Errorf("libvirt connection still has %d remaining references after close", refCount)
+	}
 	return nil
 }
 
--- a/src/cloud-providers/libvirt/provider_test.go
+++ b/src/cloud-providers/libvirt/provider_test.go
@@ -13,6 +13,7 @@ import (
 	provider "github.com/confidential-containers/cloud-api-adaptor/src/cloud-providers"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	libvirt "libvirt.org/go/libvirt"
 )
 
 const (
@@ -158,10 +159,29 @@ func TestGetIPs(t *testing.T) {
 	}
 }
 
-func TestTeardown(t *testing.T) {
+func TestTeardownNilClient(t *testing.T) {
 	p := &libvirtProvider{}
-	err := p.Teardown()
-	assert.NoError(t, err)
+	assert.NoError(t, p.Teardown())
+}
+
+// TestTeardown verifies that Teardown() closes the underlying libvirt
+// connection and that any subsequent operation on it fails.
+func TestTeardown(t *testing.T) {
+	checkConfig(t)
+
+	conn, err := libvirt.NewConnect(testCfg.URI)
+	require.NoError(t, err)
+
+	p := &libvirtProvider{libvirtClient: &libvirtClient{connection: conn}}
+
+	_, err = conn.GetLibVersion()
+	require.NoError(t, err)
+
+	err = p.Teardown()
+	require.NoError(t, err)
+
+	_, err = conn.GetLibVersion()
+	assert.Error(t, err, "connection should be closed after Teardown")
 }
 
 func TestDeleteInstanceEmptyID(t *testing.T) {


src/cloud-providers/libvirt/provider.go
```diff
@@ -154,6 +154,9 @@ func (p *libvirtProvider) DeleteInstance(ctx context.Context, instanceID string)
 }
 
 func (p *libvirtProvider) Teardown() error {
+	if _, err := p.libvirtClient.connection.Close(); err != nil {
+		return fmt.Errorf("failed to close libvirt connection: %w", err)
+	}
```
stevenhorsman: Could we add a nil-guard incase Teardown is called due to the Provider failing during initialisation:
```
    if p.libvirtClient == nil {
        return nil
    }
````

File: src/cloud-providers/libvirt/provider.go [hunk 2]
```diff
@@ -154,6 +154,9 @@ func (p *libvirtProvider) DeleteInstance(ctx context.Context, instanceID string)
 }
 
 func (p *libvirtProvider) Teardown() error {
+	if _, err := p.libvirtClient.connection.Close(); err != nil {
```
stevenhorsman: I think we might need to expand on this check. `libvirt.Connect.Close()` returns `(int, error)`. A non-negative return value is the reference count remaining after this close — a positive count means the connection is not actually closed yet (it is reference-counted), so do we need to check this and assert that the count is zero, or error otherwise

src/cloud-providers/libvirt/provider_test.go
```diff
@@ -159,11 +160,39 @@ func TestGetIPs(t *testing.T) {
...
+	require.NoError(t, err)
+
+	p := &libvirtProvider{libvirtClient: &libvirtClient{connection: conn}}
+	err = p.Teardown()
 	assert.NoError(t, err)
 }
 
+// TestTeardownClosesConnection verifies that after Teardown() the underlying
+// libvirt connection is actually closed — any subsequent operation must fail.
+func TestTeardownClosesConnection(t *testing.T) {
```
stevenhorsman: This is a superset of `TestTeardown`, so I think that one can be removed and/or this renamed to `TestTeardown`?

src/cloud-providers/libvirt/provider_test.go
```diff
@@ -159,11 +160,39 @@ func TestGetIPs(t *testing.T) {
...
+	checkConfig(t)
+
+	conn, err := libvirt.NewConnect(testCfg.URI)
+	require.NoError(t, err)
+
+	p := &libvirtProvider{libvirtClient: &libvirtClient{connection: conn}}
+
+	// Connection must be alive before Teardown
```
stevenhorsman: This comment seems like AI slop and just restates the error message below, so isn't of value

src/cloud-providers/libvirt/provider_test.go
```diff
@@ -159,11 +160,39 @@ func TestGetIPs(t *testing.T) {
...
+	_, err = conn.GetLibVersion()
+	require.NoError(t, err, "connection should be alive before Teardown")
+
+	// Teardown must succeed
+	err = p.Teardown()
+	require.NoError(t, err, "Teardown should not return an error")
+
+	// Connection must be dead after Teardown — any call should fail
```
stevenhorsman: Same here

src/cloud-providers/libvirt/provider_test.go
```diff
@@ -159,11 +160,39 @@ func TestGetIPs(t *testing.T) {
...
+	conn, err := libvirt.NewConnect(testCfg.URI)
+	require.NoError(t, err)
+
+	p := &libvirtProvider{libvirtClient: &libvirtClient{connection: conn}}
+
+	// Connection must be alive before Teardown
+	_, err = conn.GetLibVersion()
+	require.NoError(t, err, "connection should be alive before Teardown")
+
+	// Teardown must succeed
```
stevenhorsman: And here

src/cloud-providers/libvirt/provider.go
ajaypvictor: added, thanks for the suggestion..

src/cloud-providers/libvirt/provider.go
ajaypvictor: updated, thanks for the suggestion..

src/cloud-providers/libvirt/provider_test.go
ajaypvictor: sorry about that, removed it..

src/cloud-providers/libvirt/provider_test.go
ajaypvictor: sorry about that, removed it..

src/cloud-providers/libvirt/provider_test.go
ajaypvictor: sorry about that, removed it..

src/cloud-providers/libvirt/provider_test.go
ajaypvictor: sorry about that, removed it..

stevenhorsman (approved):
LGTM. Thanks for the updates!

src/cloud-providers/libvirt/provider.go
```diff
@@ -154,6 +154,16 @@ func (p *libvirtProvider) DeleteInstance(ctx context.Context, instanceID string)
...
 
 func (p *libvirtProvider) Teardown() error {
+	if p.libvirtClient == nil {
+		return nil
+	}
+	refCount, err := p.libvirtClient.connection.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close libvirt connection: %w", err)
+	}
+	if refCount > 0 {
```
BbolroC, line 164: Do we have a test to cover this?

BbolroC (commented):
LGTM overall, but one nitpick. Thanks!

src/cloud-providers/libvirt/provider.go
ajaypvictor: The nil guard `(libvirtClient == nil)` is covered by `TestTeardownNilClient` added this UT newly with no daemon needed. The close path is covered by `TestTeardown` (integration, skips without LIBVIRT_URI), which also checks the connection is dead after `Teardown()` returns.

The `refCount > 0` branch has no dedicated unit test. `*libvirt.Connect` looks to be a CGo struct, I couldn't find an interface seam to provide a stub that returns a positive ref-count unfortunately. The condition is also mostly unreachable in practice since nothing else in CAA holds a second reference to connection, `Close()` will always `return (0, nil)` here. Would you be okay to go with this approach for now and then figure out a better solution on a later PR?

BbolroC (approved):
LGTM, thanks @ajaypvictor !

PR #3200, opened 2026-07-20
workflows: Remove libvirt_e2e_arch_prep job

stevenhorsman:
This job no longer seems to provide any value, so tidy it up to remove old code and save on the other jobs waiting for it


mkulke (approved):

ldoktor (approved):
Good catch, each e2e workflow does that by itself and this one is not actually doing anything to benefit from the rebase.

ldoktor:
@stevenhorsman could you please resolve the conflict?

PR #3202, opened 2026-07-21
Kata 4.0.0 release bump

stevenhorsman:
Update version pins post Kata 4.0.0 release

### DIFF
--- a/docs/Release-Process.md
+++ b/docs/Release-Process.md
@@ -15,8 +15,7 @@ The flow of releases should roughly be:
    version is picked) and [trustee releases](https://github.com/confidential-containers/trustee/releases).
     - This triggers [kata-containers](https://github.com/kata-containers/kata-containers) to update to these new versions in
       [versions.yaml](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml) under
-      `externals.coco-guest-components.version`, `externals.coco-trustee` and the `image-rs` crate in the agent's
-      [`Cargo.toml`](https://github.com/kata-containers/kata-containers/blob/main/src/agent/Cargo.toml).
+      `externals.coco-guest-components.version` and `externals.coco-trustee`.
     - At this point it makes sense for us to stay in sync, by updating the guest-components and kbs that we use in peer pods,
       by changing the `oci.guest-components.reference` and `git.kbs.tag`values in [versions.yaml](../src/cloud-api-adaptor/versions.yaml).
       We should also bump the kata agent to the latest commit hash by updating `oci.kata-containers.reference` in
--- a/src/cloud-api-adaptor/go.mod
+++ b/src/cloud-api-adaptor/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/coreos/go-iptables v0.6.0
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/google/uuid v1.6.0
-	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20260519201048-cec98e0d976b
+	github.com/kata-containers/kata-containers/src/runtime v0.0.0-20260720141120-cf82bb35c803
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/stretchr/testify v1.11.1
 	github.com/vishvananda/netlink v1.3.1
--- a/src/cloud-api-adaptor/go.mod
+++ b/src/cloud-api-adaptor/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/IBM/ibm-cos-sdk-go v1.14.1
 	github.com/avast/retry-go/v4 v4.6.1
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f
-	github.com/golang-jwt/jwt/v5 v5.3.0 // indirect
+	github.com/golang-jwt/jwt/v5 v5.3.0
 	github.com/klauspost/cpuid/v2 v2.3.0
 	github.com/moby/sys/mountinfo v0.7.2
 	github.com/pelletier/go-toml/v2 v2.1.0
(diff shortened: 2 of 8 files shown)


stevenhorsman:
Ok, so I guess there is a problem with the guest-components bump as guest-pull fails with permissions errors:
```
failed to create containerd task: failed to create shim task: rpc status: Status { code: INTERNAL, message: "[CDH] [ERROR]: Image Pull error: Failed to pull image quay.io/nginx/nginx-unprivileged:1.27.4 from all mirror/mapping locations or original location: image: quay.io/nginx/nginx-unprivileged:1.27.4, error: Errors happened when pulling image: Failed to decode layer data stream: Failed to unpack layer: Failed to set permissions for path: /run/kata-containers/image/layers/7", details: [], special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }
```
maybe related to https://github.com/confidential-containers/guest-components/commit/920b4b3a ?

stevenhorsman:
My AI tool tells me the cause is a bug in that commit:
> [image-rs/src/stream/unpack.rs:500-511](https://github.com/confidential-containers/guest-components/blob/da8d93f2/image-rs/src/stream/unpack.rs#L500-L511)): fchmodat(..., c"", mode, AT_EMPTY_PATH) — AT_EMPTY_PATH is not a valid flag for fchmodat on Linux. The kernel returns EINVAL. AT_EMPTY_PATH is only valid for fchownat. This is the actual crash.

and that the guest-components code needs changing to:
```
fn fchmodat(fd: BorrowedFd<'_>, mode: u32) -> Result<()> {
    // fchmodat does not support AT_EMPTY_PATH on Linux.
    // Use /proc/self/fd/<N> to reach the inode from an O_PATH fd.
    let path = format!("/proc/self/fd/{}", fd.as_raw_fd());
    let ret = unsafe {
        libc::chmod(
            std::ffi::CString::new(path)?.as_ptr(),
            mode as libc::mode_t,
        )
    };
    if ret != 0 {
        bail!("failed to chmod: {:?}", io::Error::last_os_error());
    }
    Ok(())
}
```
I'm not sure why if this is true that it's not causing problems in bare-metal. @Xynnn007 - any thoughts from you on this problem?

Xynnn007:
@stevenhorsman seems like the guest kernel version issue? I got some documents

- https://man7.org/linux/man-pages/man2/chmod.2.html: says kernel >= 6.6, it supports `AT_EMPTY_PATH` for `fchmodat`
- [ibm zos](https://www.ibm.com/docs/en/zos/3.2.0?topic=lf-fchmodat-change-mode-file-relative-directory-file-descriptor): does not support `AT_EMPTY_PATH` for `fchmodat`
- [freebsd](https://man.freebsd.org/cgi/man.cgi?query=fchmodat&sektion=2&n=1): supports `AT_EMPTY_PATH` for `fchmodat`

stevenhorsman:
> @stevenhorsman seems like the guest kernel version issue? I got some documents
> 
>     * https://man7.org/linux/man-pages/man2/chmod.2.html: says kernel >= 6.6, it supports `AT_EMPTY_PATH` for `fchmodat`
> 
>     * [ibm zos](https://www.ibm.com/docs/en/zos/3.2.0?topic=lf-fchmodat-change-mode-file-relative-directory-file-descriptor): does not support `AT_EMPTY_PATH` for `fchmodat`
> 
>     * [freebsd](https://man.freebsd.org/cgi/man.cgi?query=fchmodat&sektion=2&n=1): supports `AT_EMPTY_PATH` for `fchmodat`

Ok, thanks for the suggestion. I wonder if the 26.04 OS bump would resolve it then? I will try...

I thought that Ubuntu 24.04 was on a newer version than 6.6 though and I think we still use ubuntu 24 as the base in bare-metal, so I'm not super confident that is the cause

stevenhorsman:
26.04 didn't seem to help: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/29901510160/job/88863079276 as expected

Xynnn007:
> 26.04 didn't seem to help: https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/29901510160/job/88863079276 as expected

Ok. Let me get the fd-independent way as Bob suggested.

stevenhorsman:
So the good news is that the pull image problem doesn't seem to be there anymore - huge thanks to @Xynnn007. We have some other issues though:
- On the s390x runner:
```
=== RUN   TestLibvirtSealedSecret
time="2026-07-22T16:31:48Z" level=info msg="set key resource: default/signing-key/sealed-secret"
    libvirt_test.go:159: Set signing key Secret failed with: exit status 2
--- FAIL: TestLibvirtSealedSecret (0.05s)
=== RUN   TestLibvirtKbsKeyRelease
time="2026-07-22T16:31:48Z" level=info msg="set key resource: caa/workload_key/test_key.bin"
    libvirt_test.go:187: SetSecret failed with: exit status 2
--- FAIL: TestLibvirtKbsKeyRelease (0.00s)
```
so it looks like we might have any issue with the new trustee?
- AWS is timing out on the cluster create
- x86 libvirt is timing out on the cluster create
It looks like these have failures pulling packages from ubuntu, and canonical has a partial outage, so we might have to wait to re-try them.

stevenhorsman:
Ok, the ubuntu isssues are resolved, so I think it's just some trustee problems. I'm hoping that there isn't a breaking change between trustee and guest-component since the "tagged" versions that we've picked up with the GC bump to latest?

Xynnn007:
During the two version, IBM SE report format has been changed. https://github.com/confidential-containers/guest-components/commit/9488375628761607636190f8b95fc31979b7107f

others looks same

stevenhorsman:
Thanks Ding, maybe it's related to the change to use admin token rather than admin key to interact with the KBS? I don't remember seeing a change needed on the Kata side for this though?

stevenhorsman:
Ah - it was added as part of the helm chart migration, so there was no separate commit, so I think this is likely to be the problem. I will try and do the switch...

stevenhorsman:
The api token change seems to have works, so this is all passing now 🥳

mkulke (approved):

bpradipt (approved):

PR #3211, opened 2026-07-22
fix: GetLaunchSecurityType opens a redundant libvirt connection

### LINKED ISSUE #3210: libvirt: GetLaunchSecurityType opens a redundant libvirt connection
## Problem

`GetLaunchSecurityType` in `src/cloud-providers/libvirt/libvirt.go` opens a brand-new libvirt connection (`libvirt.NewConnect(uri)`) solely to call `GetNodeInfo()` and inspect the CPU model:

```go
func GetLaunchSecurityType(uri string) (LaunchSecurityType, error) {
    conn, _ := libvirt.NewConnect(uri)       // ← new connection
    nodeInfo, _ := conn.GetNodeInfo()

    switch nodeInfo.Model {
    case archS390x:
        return S390PV, nil
    default:
        return NoLaunchSecurity, nil
    }
}


This connection is opened and immediately abandoned on every `CreateInstance` call that relies on auto-detection (i.e. `DisableCVM=false` and `LaunchSecurity=""`).

The libvirt client created by `NewLibvirtClient` already establishes a connection and caches node info in `libvirtClient.nodeInfo` before any instance creation begins. `GetLaunchSecurityType` has no reason to create a second connection.

## Proposed Fix

Change the signature of `GetLaunchSecurityType` to accept a `*libvirtClient` and read `client.nodeInfo.Model` directly, reusing the already-established connection and cached node info.

```go
func GetLaunchSecurityType(client *libvirtClient) (LaunchSecurityType, error) {
    if client.nodeInfo == nil {
        return NoLaunchSecurity, fmt.Errorf("node info is not available in libvirt client")
    }
    switch client.nodeInfo.Model {
    case archS390x:
        return S390PV, nil
    default:
        return NoLaunchSecurity, nil
    }
}
```

Update the call site in `CreateInstance` (`provider.go`) to pass `p.libvirtClient` instead of `p.serviceConfig.URI`.

### Impact
     - Eliminates one unnecessary libvirt connection per `CreateInstance` call.
     - No behaviour change: the detection logic (`s390x → S390PV`, everything else → `NoLaunchSecurity`) is unchanged.
### Related
Original implementation: [#1397](https://github.com/confidential-containers/cloud-api-adaptor/pull/1397)

chathuryaadapa:
Previously, GetLaunchSecurityType(uri string) opened a brand-new libvirt connection (libvirt.NewConnect) purely to call GetNodeInfo() and inspect the CPU model. This was redundant because NewLibvirtClient already establishes a connection and caches the node info in libvirtClient.nodeInfo before any instance creation begins.

As a result, every call to CreateInstance that relied on auto-detection (i.e. DisableCVM=false and LaunchSecurity="") caused a second, short-lived libvirt connection to be created and immediately abandoned.

- Change the signature of GetLaunchSecurityType from accepting a URI string to accepting a *libvirtClient. The function now reads client.nodeInfo.Model directly, eliminating the extra libvirt connection that was opened on every CreateInstance call.
- Add a nil-guard on client.nodeInfo so a partially constructed client returns a clear error rather than a silent panic.
- Extract archX86_64 = "x86_64" as a named constant alongside the existing arch constants; fix typo "architecutre" -> "architecture" in the archAArch64 comment.
- Collapse case archX86_64/archAArch64 into a single default — both returned NoLaunchSecurity, so explicit arms added no value.
- Update the call site in CreateInstance (provider.go) to pass p.libvirtClient instead of p.serviceConfig.URI.
- Replace TestGetLaunchSecurityTypeInvalidURI (which tested URI-parsing error handling no longer applicable) with:
     - TestGetLaunchSecurityType: table-driven test covering s390x, x86_64, and an unknown arch against their expected security types.
     - TestGetLaunchSecurityTypeNilNodeInfo: verifies that a client with a nil nodeInfo returns an error instead of panicking.
No behaviour change for callers: the launch security detection logic
(s390x -> S390PV, everything else -> NoLaunchSecurity) is unchanged.

Fixes: #3210

### DIFF
--- a/src/cloud-providers/libvirt/libvirt.go
+++ b/src/cloud-providers/libvirt/libvirt.go
@@ -23,8 +23,10 @@ import (
 const (
 	// architecture value for the s390x architecture
 	archS390x = "s390x"
-	// architecutre value for aarch64/arm64
+	// architecture value for aarch64/arm64
 	archAArch64 = "aarch64"
+	// architecture value for x86_64
+	archX86_64 = "x86_64"
 	// hvm indicates that the OS is one designed to run on bare metal, so requires full virtualization.
 	typeHardwareVirtualMachine = "hvm"
 	// The amount of retries to get the domain IP addresses
@@ -826,24 +828,17 @@ func freeDomain(domain *libvirt.Domain, errCtx *error) {
 	}
 }
 
-// Attempts to determine launchSecurity Type from domain capabilities and hardware
-// Currently only supports S390PV
-func GetLaunchSecurityType(uri string) (LaunchSecurityType, error) {
-	conn, err := libvirt.NewConnect(uri)
-	if err != nil {
-		return NoLaunchSecurity, fmt.Errorf("unable to get libvirt connection [%v]", err)
+// GetLaunchSecurityType determines the launch security type from the node info
+// already retrieved by the libvirt client, avoiding an extra connection.
+// Currently only supports S390PV.
+func GetLaunchSecurityType(client *libvirtClient) (LaunchSecurityType, error) {
+	if client.nodeInfo == nil {
+		return NoLaunchSecurity, fmt.Errorf("node info is not available in libvirt client")
 	}
 
-	nodeInfo, err := conn.GetNodeInfo()
-	if err != nil {
-		return NoLaunchSecurity, fmt.Errorf("error retrieving node info: %v", err)
-	}
-
-	switch nodeInfo.Model {
+	switch client.nodeInfo.Model {
 	case archS390x:
 		return S390PV, nil
-	case "x86_64":
-		return NoLaunchSecurity, nil
 	default:
 		return NoLaunchSecurity, nil
 	}
--- a/src/cloud-providers/libvirt/libvirt_test.go
+++ b/src/cloud-providers/libvirt/libvirt_test.go
@@ -255,33 +255,33 @@ func TestGetGuestForArchType(t *testing.T) {
 	}{
 		{
 			name:        "find x86_64 guest",
-			caps:        createMockCaps("x86_64", ""),
-			arch:        "x86_64",
-			ostype:      "hvm",
+			caps:        createMockCaps(archX86_64, ""),
+			arch:        archX86_64,
+			ostype:      typeHardwareVirtualMachine,
 			expectError: false,
-			expectArch:  "x86_64",
+			expectArch:  archX86_64,
 		},
 		{
 			name:        "find s390x guest",
-			caps:        createMockCaps("s390x", ""),
-			arch:        "s390x",
-			ostype:      "hvm",
+			caps:        createMockCaps(archS390x, ""),
+			arch:        archS390x,
+			ostype:      typeHardwareVirtualMachine,
 			expectError: false,
-			expectArch:  "s390x",
+			expectArch:  archS390x,
 		},
 		{
 			name:        "find aarch64 guest",
-			caps:        createMockCaps("aarch64", ""),
-			arch:        "aarch64",
-			ostype:      "hvm",
+			caps:        createMockCaps(archAArch64, ""),
+			arch:        archAArch64,
+			ostype:      typeHardwareVirtualMachine,
 			expectError: false,
-			expectArch:  "aarch64",
+			expectArch:  archAArch64,
 		},
 		{
 			name:        "architecture not found",
-			caps:        createMockCaps("x86_64", ""),
+			caps:        createMockCaps(archX86_64, ""),
 			arch:        "invalid-arch",
-			ostype:      "hvm",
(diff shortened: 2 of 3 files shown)


mkulke (approved):

src/cloud-providers/libvirt/libvirt.go
```diff
@@ -826,23 +828,14 @@ func freeDomain(domain *libvirt.Domain, errCtx *error) {
...
-	if err != nil {
-		return NoLaunchSecurity, fmt.Errorf("error retrieving node info: %v", err)
-	}
-
-	switch nodeInfo.Model {
+// GetLaunchSecurityType determines the launch security type from the node info
+// already retrieved by the libvirt client, avoiding an extra connection.
+// Currently only supports S390PV.
+func GetLaunchSecurityType(client *libvirtClient) (LaunchSecurityType, error) {
+	switch client.nodeInfo.Model {
```
stevenhorsman, line 839: The PR description says: "Add a nil-guard on client.nodeInfo so the function returns a clear error if node info was somehow not populated by the client."?

src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -819,9 +819,26 @@ func TestCPUSetXMLMarshaling(t *testing.T) {
...
-	assert.Error(t, err)
+func TestGetLaunchSecurityType(t *testing.T) {
+	tests := []struct {
+		name     string
+		model    string
+		expected LaunchSecurityType
+	}{
+		{name: "s390x returns S390PV", model: "s390x", expected: S390PV},
+		{name: "x86_64 returns NoLaunchSecurity", model: "x86_64", expected: NoLaunchSecurity},
+		{name: "unknown arch returns NoLaunchSecurity", model: "aarch64", expected: NoLaunchSecurity},
```
stevenhorsman: aarch64 is not an unknown arch? Could we correct this and then add an actual unknown arch test e.g. `riscv64`?

src/cloud-providers/libvirt/libvirt.go
```diff
@@ -826,23 +828,14 @@ func freeDomain(domain *libvirt.Domain, errCtx *error) {
...
 	case archS390x:
 		return S390PV, nil
-	case "x86_64":
+	case archX86_64, archAArch64:
 		return NoLaunchSecurity, nil
 	default:
 		return NoLaunchSecurity, nil
```
stevenhorsman: The PR description says
> Collapse the redundant "x86_64" and "default" switch arms into a single default case, since both returned NoLaunchSecurity.

but this hasn't been done?

src/cloud-providers/libvirt/libvirt_test.go
stevenhorsman: Please use the new constants in the tests rather than literals to prevent future drift

stevenhorsman (changes requested):
I think this needs more attention, especially the lack of nil-guard - it's a bit too much AI generated to me without a human filter, which is causing issues and mismatch between claims and implementation

src/cloud-providers/libvirt/libvirt.go
chathuryaadapa: I updated the commit message missed updating pr description.
That was my first draft changes, apologies will update the pr description.

src/cloud-providers/libvirt/libvirt.go
chathuryaadapa: apologies, the pr description was out-of-date

src/cloud-providers/libvirt/libvirt.go
stevenhorsman: Why not implement the nil-guard?

src/cloud-providers/libvirt/libvirt.go
stevenhorsman: Why did you not implement this though?

src/cloud-providers/libvirt/libvirt.go
chathuryaadapa: yes sure steve, I can go ahead with the implementation of nil-guard.
I removed because thinking [NewLibvirtClient](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-providers/libvirt/libvirt.go#L797) calls conn.GetNodeInfo() and returns an error if it fails — meaning a successfully constructed *libvirtClient always has a non-nil nodeInfo.

I will restore to original flow to handle this.

Thanks for the review!!

src/cloud-providers/libvirt/libvirt.go
chathuryaadapa: Yes, I kept it as is for code readability. There wasn't any other particular reason.

will update this as well!!

stevenhorsman (approved):
LGTM. Thanks for the update.

PR #3222, opened 2026-07-27
podvm: various smaller fixes

mkulke:
Those are in preparation for re-adding the azure e2e tests


src/cloud-api-adaptor/podvm/files/etc/systemd/system/run-kata\x2dcontainers.mount
Amulyam24, line 1: @mkulke, PowerVS provider has not yet adopted the mkosi builds, it would be needing this file, IINW.

Amulyam24 (commented):
One comment, otherwise LGTM!

.github/workflows/podvm_mkosi.yaml
```diff
@@ -190,10 +197,11 @@ jobs:
           echo "digest=${digest}" >> "$GITHUB_OUTPUT"
           echo "tag=${tag}" >> "$GITHUB_OUTPUT"
         env:
-          DEBUG: ${{ inputs.debug == true && 'true' || 'false' }}
+          DEBUG: ${{ inputs.debug == true && '1' || '' }}
           REGISTRY: ${{ inputs.registry }}
           ARCH: ${{ inputs.arch }}
           IMAGE_TAG: ${{ inputs.image_tag }}
+          TEE_PLATFORM: ${{ inputs.tee-platform || '' }}
```
stevenhorsman: input.tee-platform has a default value of none, so I don't think the `|| ''`  can ever be produced?

.github/workflows/podvm_mkosi.yaml
```diff
@@ -171,12 +182,8 @@ jobs:
           cd oras
           cp ../build/podvm-*.qcow2 .
           tar cJf podvm.tar.xz podvm-*.qcow2
-          image=${REGISTRY}/podvm-generic-ubuntu
-          if [ "${DEBUG}" = "true" ]; then
-            image=${image}-debug
-          fi
-          image=${image}-${ARCH}
-          tag=$(git rev-parse --short HEAD)
+          image="${REGISTRY}/podvm-generic-ubuntu-${ARCH}${TEE_PLATFORM:+-${TEE_PLATFORM}}${DEBUG:+-debug}"
```
stevenhorsman: Did you intend to add the `-none` suffix here, or should the env value actually be: `${{ inputs.tee-platform != 'none' && inputs.tee-platform || '' }}`?

stevenhorsman (commented):
Sorry for missing this earlier - I have a question about the naming in the workflow...

src/cloud-api-adaptor/podvm/files/etc/systemd/system/run-kata\x2dcontainers.mount
mkulke: is this consumed downstream? because I think we don't have packer logic left in this repo, no? if we don't remove this unit then an image will have a failed unit, because that mount unit cannot work without the /kata-containers folder. we can create it of course, but it really looks like technical debt that we string along. are there other ways to adjust powervs to not rely on a /kata-containers bind mounbt?

.github/workflows/podvm_mkosi.yaml
mkulke: yes, this was the intention I think, I'll fix it.

.github/workflows/podvm_mkosi.yaml
mkulke: see other remark, yes this was meant to replace a -none suffix with an empty string

stevenhorsman (approved):
LGTM. Thanks

src/cloud-api-adaptor/podvm/files/etc/systemd/system/run-kata\x2dcontainers.mount
Amulyam24: PowerVS uses [automation](https://github.com/confidential-containers/cloud-api-adaptor/tree/main/src/cloud-api-adaptor/ibmcloud-powervs#image-build) to build the ova images consuming the files from podvm dir. It makes sense to remove this from here, I will modify the build to incorporate this change for PowerVS builds.

Amulyam24 (approved):

PR #3223, opened 2026-07-28
gha: fix podvm byom binaries job

Amulyam24:
The byom binaries image uses the podvm binaries image as the base. As we do not push the podvm binaries image or have it available locally during the mkosi builds, build it explicitly for byom binaries image.

### DIFF
--- a/.github/workflows/podvm_byom_binaries_publish.yaml
+++ b/.github/workflows/podvm_byom_binaries_publish.yaml
@@ -77,12 +77,29 @@ jobs:
           ref: ${{ inputs.git_ref || github.sha }}
           persist-credentials: false
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      - name: Install build dependencies
+        run: |
+          command -v yq || sudo snap install yq
+          yq --version | grep -qE 'v4\.[0-9]+' || { echo "yq version is not 4.x, please install yq 4.x"; exit 1; }
+
       - name: Login to Container Registry
         uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # v4.4.0
         with:
           registry: quay.io
           username: ${{ vars.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build podvm binaries image
+        working-directory: src/cloud-api-adaptor/podvm
+        run: make podvm-binaries
+        env:
+          REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
+          PODVM_TAG: ${{ inputs.image_tag || github.sha }}
+          ARCH: ${{ inputs.arch }}
+          PUSH: "true"
 
       - name: Build BYOM binaries image
         working-directory: src/cloud-api-adaptor/podvm
--- a/src/cloud-api-adaptor/podvm/Dockerfile.podvm_byom_binaries
+++ b/src/cloud-api-adaptor/podvm/Dockerfile.podvm_byom_binaries
@@ -11,7 +11,7 @@ ARG BINARIES_IMG="quay.io/confidential-containers/podvm-binaries-ubuntu-amd64"
 FROM ${BINARIES_IMG} AS podvm_binaries
 
 # ubuntu:24.04
-FROM ubuntu@sha256:0d39fcc8335d6d74d5502f6df2d30119ff4790ebbb60b364818d5112d9e3e932 as builder
+FROM ubuntu@sha256:0d39fcc8335d6d74d5502f6df2d30119ff4790ebbb60b364818d5112d9e3e932 AS builder
 
 COPY --from=podvm_binaries /podvm-binaries.tar.gz /src/podvm-binaries.tar.gz
 RUN mkdir /src/files && tar xvf /src/podvm-binaries.tar.gz -C /src/files


.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -77,13 +77,30 @@ jobs:
...
           password: ${{ secrets.QUAY_PASSWORD }}
 
+      - name: Build podvm binaries image
+        working-directory: src/cloud-api-adaptor/podvm
+        run: make podvm-binaries
+        env:
+          REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
+          PODVM_TAG: ${{ inputs.image_tag || github.sha }}
+          ARCH: ${{ inputs.arch }}
+          PUSH: "true"
```
stevenhorsman, line 102: Do we need to push this as it's just an intermediate build, or can you consume it locally?

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -77,13 +77,30 @@ jobs:
           ref: ${{ inputs.git_ref || github.sha }}
           persist-credentials: false
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      - name: Install build dependencies
+        run: |
+          command -v yq || sudo snap install yq
```
mkulke, line 85: why is a yq installation requited?

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: IIUC, the PUSH=true determines if the [output flag](https://github.com/confidential-containers/cloud-api-adaptor/blob/289bea7cdceef4b773709ea08585e690218f2366/src/cloud-api-adaptor/podvm/Makefile#L85) has to be added and if true, the image would be available locally.

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: As we added `make podvm-binaries`, this check would be needed to ensure `yq` is installed to query for the component versions during the build.

mkulke: afaiu this is done in a docker container, no?

```bash
$ git grep YQ_VERSION src/cloud-api-adaptor/Makefile.defaults
src/cloud-api-adaptor/Makefile.defaults:YQ_VERSION := v4.35.1
$ git grep YQ_VERSION src/cloud-api-adaptor/podvm/Makefile
src/cloud-api-adaptor/podvm/Makefile:           --build-arg YQ_VERSION=$(YQ_VERSION) \
```

.github/workflows/podvm_byom_binaries_publish.yaml
stevenhorsman: Ok, so it's a local push rather than binary build (which I guess is more difficult for you to consume). Fair enough

mkulke: it looks like the PUSH env is a misnomer here, since it will not actually push an image? Maybe we can rework the check to consume an `BINARIES_OUTPUT ` variable for clarity?

```
...
BINARIES_OUTPUT =? local
...
		$(if $(filter $(BINARIES_OUTPUT ),local),-o type=local,dest="./resources/binaries-tree",,) \
```

It's probably an artifact and doesn't seem to be used somewhere else:

```bash
$ git grep PUSH .github/workflows/
$
$ git grep PUSH src/cloud-api-adaptor/Makefile.defaults
$
```

Amulyam24: Yes, I'll create a follow up PR for it. It definitely gives a different meaning :)

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: The versions are computed using yq before the docker build right?
Similar to the mkosi workflow - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/workflows/podvm_mkosi.yaml#L131-L133

mkulke: but unlike this workflow, we don't set any versions in this workflow prior to "make podvm-binaries", no? 

yq is installed in the build container:

https://github.com/confidential-containers/cloud-api-adaptor/blob/86589205af144eb1faa0a543b561d6fa629b414c/src/cloud-api-adaptor/podvm/Dockerfile.podvm_binaries#L50-L52

Amulyam24: That is right, but when I test it locally, to pass the validation added in the Makefile, it needs yq, is this intentional?

```
# yq
-bash: /usr/local/bin/yq: No such file or directory
# pwd
/root/cloud-api-adaptor/src/cloud-api-adaptor
# make podvm-binaries
Makefile.defaults:8: *** "yq not found, consider doing snap install yq".  Stop.
```

mkulke: oof, I don't think it is, but fixing this is probably out-of-scope for this PR, thanks for looking into it.

mkulke (approved):

Amulyam24:
Thanks, @mkulke! I'll creating follow up PRs to handle the 2 cases we discussed - yq and binaries output.

stevenhorsman (approved):

PR #3229, opened 2026-08-03
podvm: rename PUSH to EXPORT_BINARIES

Amulyam24:
The environment variable `PUSH` is confusing as if set, it suppresses local export of binaries rather than pushing to a registry, contrary to what the name implies. Rename to `EXPORT_BINARIES` with inverted logic.


stevenhorsman (approved):
Looks okay to me. Thanks!

mkulke (approved):

PR #3233, opened 2026-08-06
azure: add multi-NIC support for external pod networking

manjotsidhu:
## Summary

Attaches a second NIC to the Azure Pod VM so that pod traffic to destinations
outside the cluster can go directly over the Azure VNet instead of through
the worker node's VXLAN tunnel. This builds on the existing
`EXTERNAL_NETWORK_VIA_PODVM` mechanism already used by the AWS and Alibaba
Cloud providers — see [docs/azure-multi-nic.md](../blob/main/docs/azure-multi-nic.md)
for configuration and design details.

- Declares both NICs in the VM creation request (new `AZURE_SECONDARY_SUBNET_ID`
  config), since Azure doesn't support hot-attaching a NIC to a running VM
  without first deallocating it, unlike AWS's post-launch ENI attach.
- Orders the IPs returned from `getIPs` so the primary (control-plane) NIC's
  IPs always come first, regardless of the order Azure returns NICs in.
- `CreateInstance` fails fast with a clear error if multi-NIC pod networking
  is requested (`EXTERNAL_NETWORK_VIA_PODVM=true`) but
  `AZURE_SECONDARY_SUBNET_ID` isn't configured, instead of deferring to a
  runtime failure deep in pod network setup.
- Extends `pkg/podnetwork`'s secondary-interface discovery (shared with the
  AWS/Alibaba providers) to infer a gateway when the secondary NIC has no
  DHCP-assigned default route and isn't on the primary's subnet — the case
  on Azure, since its two NICs sit on different subnets by design. This is
  purely additive: existing discovery paths (own default route, or same
  subnet as primary) are checked first and unchanged.

## Test plan

- [x] `go build`/`go vet` clean for `src/cloud-providers/azure` and
      `src/cloud-api-adaptor/pkg/podnetwork` (Linux target)
- [x] `golangci-lint run` clean on both changed packages
- [x] `go mod tidy` produces no diff in either module
- [x] New unit tests added and passing:
  - `src/cloud-providers/azure/multinic_test.go`: NIC config building
    (single/multi-NIC, public IP placement), `orderByPrimary` IP ordering,
    `ConfigVerifier` subnet validation
  - `src/cloud-api-adaptor/pkg/podnetwork/common_test.go`: gateway
    inference, cross-subnet/same-subnet secondary interface discovery,
    end-to-end `setupExternalNetwork` via network namespaces (root-only,
    skipped locally on this dev machine — matches existing `podnetwork_test.go`
    convention, exercised by CI's `sudo make test`)
- [x] Live Azure e2e validation (VM creation with two NICs, actual traffic
      routing) — not exercised here after merging with latest code, the prototype was tested against 1.12 version of OSC on ARO.

## DCO

This PR was co-developed with @devansh1109, who prototyped the original
implementation; I reworked it against current `main` (which had diverged
significantly — CSI volume support landed since the prototype was written),
removed debug-only logging, added the test coverage and doc above, and am
submitting it with their permission. Both commits are signed off per DCO 1.1.

### DIFF
--- a/docs/azure-multi-nic.md
+++ b/docs/azure-multi-nic.md
@@ -1,0 +1,92 @@
+# Azure multi-NIC pod networking
+
+## Overview
+
+By default a peer-pod VM on Azure has a single NIC that carries both the
+VXLAN tunnel back to the worker node (pod/cluster traffic) and any traffic
+the pod sends to the outside world. This feature attaches a second NIC to
+the Pod VM so that:
+
+- `eth0` (primary NIC) carries pod/cluster traffic over the VXLAN tunnel, as
+  before.
+- `eth1` (secondary NIC) carries the pod's traffic to destinations outside
+  the cluster directly over the Azure VNet, bypassing the worker node.
+
+This builds on the existing `EXTERNAL_NETWORK_VIA_PODVM` mechanism (already
+used by the AWS and Alibaba Cloud providers) which discovers a second NIC
+inside the Pod VM's network namespace and moves it into the pod's namespace
+as its default route.
+
+## Configuration
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: peer-pods-cm
+  namespace: confidential-containers-system
+data:
+  AZURE_SUBNET_ID: "/subscriptions/<sub-id>/resourceGroups/<rg>/providers/Microsoft.Network/virtualNetworks/<vnet>/subnets/<primary-subnet>"
+  AZURE_SECONDARY_SUBNET_ID: "/subscriptions/<sub-id>/resourceGroups/<rg>/providers/Microsoft.Network/virtualNetworks/<vnet>/subnets/<secondary-subnet>"
+  EXTERNAL_NETWORK_VIA_PODVM: "true"
+  POD_SUBNET_CIDRS: "10.128.0.0/14,172.30.0.0/16,100.64.0.0/16"
+```
+
+- `AZURE_SECONDARY_SUBNET_ID` is required to enable multi-NIC. It must be a
+  different subnet than `AZURE_SUBNET_ID`, in the same VNet — Azure requires
+  all NICs on a VM to share a VNet.
+- `EXTERNAL_NETWORK_VIA_PODVM` is the existing flag (shared with other
+  providers) that tells cloud-api-adaptor to move a second NIC into the pod
+  namespace.
+- `POD_SUBNET_CIDRS` should list the cluster's pod, service, and node CIDRs.
+  These stay routed over the VXLAN tunnel (`eth0`); everything else follows
+  the pod's default route, which points at the secondary NIC (`eth1`).
+
+If `EXTERNAL_NETWORK_VIA_PODVM` is enabled without `AZURE_SECONDARY_SUBNET_ID`
+set, `CreateInstance` fails fast with a clear error instead of creating a
+single-NIC VM that would later fail pod networking setup.
+
+## How traffic is routed
+
+The Pod VM's kernel picks the most specific matching route:
+
+| Destination | Device | Notes |
+|---|---|---|
+| CIDRs in `POD_SUBNET_CIDRS` | `eth0` | routed over the VXLAN tunnel to the worker node |
+| everything else (default route) | `eth1` | routed directly over the Azure VNet |
+
+## Secondary NIC discovery
+
+Azure does not assign a default route to a secondary NIC via DHCP, so the
+gateway can't be read off the interface directly. `getSecondaryInterfaceDetails`
+in `pkg/podnetwork/common.go` picks the first non-primary, non-filtered
+interface with a valid IPv4 address using, in order:
+
+1. its own default route (used by other providers whose secondary NIC does
+   get one), or
+2. the primary interface's gateway, if the secondary shares its subnet, or
+3. an inferred gateway at the first usable address of its subnet
+   (`x.x.x.1`), Azure's convention for subnets it manages — this is the path
+   Azure multi-NIC normally takes, since the primary and secondary NICs sit
+   on different subnets by design.
+
+## Known constraints
+
+- Both subnets must be in the same Azure VNet (`SubnetsNotInSameVnet` if
+  not).
+- Both NICs are declared in the VM creation request up front. Azure does not
+  support attaching a NIC to a running VM without first deallocating it, so
+  the secondary NIC can't be added after the fact the way AWS attaches its
+  addon ENI post-launch.
+- `AZURE_USE_PUBLIC_IP`, if enabled together with multi-NIC, attaches the
+  public IP to the secondary (external) NIC rather than the primary
+  (control-plane) NIC.
+
+## Troubleshooting
+
+```bash
+ip addr show
+ip route show
+ip route get 8.8.8.8          # should go via eth1
+ip route get <a pod/service IP>  # should go via eth0
+```
--- a/src/cloud-providers/azure/provider.go
+++ b/src/cloud-providers/azure/provider.go
@@ -112,6 +112,29 @@ func generateSSHPublicKey() ([]byte, error) {
 	return publicKeyBytes, nil
 }
 
+// orderByPrimary returns the IP configurations of the given NICs with the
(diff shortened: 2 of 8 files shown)


manjotsidhu:
@stevenhorsman or anyone from maintainers can help approve workflow checks ?

stevenhorsman:
> @stevenhorsman or anyone from maintainers can help approve workflow checks ?

Pradipta approved the check, but `TestSetupExternalNetworkCrossSubnet` is failing across a few re-runs, so needs attention.

manjotsidhu:
This PR is ready for review.

docs/azure-multi-nic.md
```diff
@@ -0,0 +1,92 @@
...
+  name: peer-pods-cm
+  namespace: confidential-containers-system
+data:
+  AZURE_SUBNET_ID: "/subscriptions/<sub-id>/resourceGroups/<rg>/providers/Microsoft.Network/virtualNetworks/<vnet>/subnets/<primary-subnet>"
+  AZURE_SECONDARY_SUBNET_ID: "/subscriptions/<sub-id>/resourceGroups/<rg>/providers/Microsoft.Network/virtualNetworks/<vnet>/subnets/<secondary-subnet>"
+  EXTERNAL_NETWORK_VIA_PODVM: "true"
+  POD_SUBNET_CIDRS: "10.128.0.0/14,172.30.0.0/16,100.64.0.0/16"
+```
+
+- `AZURE_SECONDARY_SUBNET_ID` is required to enable multi-NIC. It must be a
```
mkulke: is it maybe better to call this `AZURE_EXTERNAL_SUBNET_ID` to indicate what it's used for?

src/cloud-providers/azure/provider.go
```diff
@@ -239,6 +271,27 @@ func (p *azureProvider) buildNetworkConfig(nicName string) *armcompute.VirtualMa
...
+// buildNetworkConfigs returns the NIC configurations to attach to the VM at
+// creation time. Azure requires all NICs to be declared upfront in the VM
+// create request (unlike AWS, which attaches the secondary ENI after the
+// instance is running), so both NICs are built here when multiNic is set.
+//
+// In multi-NIC mode the primary NIC (on SubnetID) carries pod/cluster
+// traffic over the VXLAN tunnel and never gets a public IP; the secondary
+// NIC (on SecondarySubnetID) carries the Pod VM's external traffic and
+// optionally gets a public IP if UsePublicIP is set.
+func (p *azureProvider) buildNetworkConfigs(instanceName string, multiNic bool) []*armcompute.VirtualMachineNetworkInterfaceConfiguration {
```
mkulke, line 285: since we're expecting 2 network configs, does it make sense to return `[2]*armcompute.VirtualMachineNetworkInterfaceConfiguration`? we can later in the azure sdk call re-slice it with `configs[:]`, I think

devansh1109: since the function is handling both the single and multinic, if we return [2]*armcompute.VirtualMachineNetworkInterfaceConfiguration we will have to pass a nil pointer to the azure sdk which might cause some issues?

mkulke: yeah, this requires more ceremony on sdk surface, so it's probably not worth it

mkulke (commented):
I don't love this feature, this breaks certain kubernetes contracts/expecations for pods and has the potential to be a security footgun, for example network policies will not be respected, e.g. if I restrict egress traffic for my pod, this will undermine it. A more sophisticated version of this could maybe involve a controller that attempts to reconcile k8s network policies with security groups on the secondary nic.

In any case, the feature exists right now for other providers, and there's no good reason to not have it for azure.

we probably want to amend the comment here https://github.com/confidential-containers/cloud-api-adaptor/blob/a4b98f5e69e6ca42bc34b9df1c8114155231e3ab/src/cloud-api-adaptor/docs/external-network.md?plain=1#L18

(and maybe a remark about the security implications)

manjotsidhu:
Updated the PR with the changes as per your review.

src/cloud-api-adaptor/docs/azure-multi-nic.md
```diff
@@ -0,0 +1,96 @@
+# Azure multi-NIC pod networking
```
bpradipt, line 1: nit: I think this doc should go under src/cloud-api-adaptor/docs

src/cloud-providers/azure/provider.go
```diff
@@ -239,6 +271,27 @@ func (p *azureProvider) buildNetworkConfig(nicName string) *armcompute.VirtualMa
...
+// creation time. Azure requires all NICs to be declared upfront in the VM
+// create request (unlike AWS, which attaches the secondary ENI after the
+// instance is running), so both NICs are built here when multiNic is set.
+//
+// In multi-NIC mode the primary NIC (on SubnetID) carries pod/cluster
+// traffic over the VXLAN tunnel and never gets a public IP; the secondary
+// NIC (on ExternalSubnetID) carries the Pod VM's external traffic and
+// optionally gets a public IP if UsePublicIP is set.
+func (p *azureProvider) buildNetworkConfigs(instanceName string, multiNic bool) []*armcompute.VirtualMachineNetworkInterfaceConfiguration {
+	primary := p.buildNetworkConfig(fmt.Sprintf("%s-net", instanceName), p.serviceConfig.SubnetID, true, p.serviceConfig.UsePublicIP && !multiNic)
```
bpradipt: What happens when  `EXTERNAL_NETWORK_VIA_PODVM=true` and `AZURE_USE_PUBLIC_IP=true`? I think the worker node ends up dialing the secondary (external) NIC instead of the primary.

manjotsidhu: @bpradipt You are right, that's a problem, I am working on fixing that. We will test with both public ip on and off with multi-nic and get back.

manjotsidhu: We have made the changes. @devansh1109 has tested as well.

devansh1109:
@bpradipt @manjotsidhu 

Tested with UsePublicIP=true and UsePublicIP=false — both working.

UsePublicIP=true:

pod vm IP[0][public]=20.120.113.215
pod vm IP[0][private]=192.168.0.11
pod vm IP[1][private]=192.168.2.11

UsePublicIP=false:

pod vm IP[0][private]=192.168.0.11
pod vm IP[1][private]=192.168.2.11

In both cases:

- Two-NIC VM created successfully (multi-NIC: true confirmed in logs).
- Primary NIC on 192.168.0.x (control plane), secondary on 192.168.2.x (external traffic).
- UsePublicIP=true: public IP correctly attached only to the primary NIC; worker dialed IPs[0] (public IP) to reach agent-protocol-forwarder and VXLAN tunnel endpoint — both on the primary NIC. 
- UsePublicIP=false: worker dialed primary private IP IPs[0]=192.168.0.11 correctly. 
- Secondary NIC never received a public IP in either mode.

mkulke:
@manjotsidhu can you get rid of the merge-commits and rebase your changes on main?

manjotsidhu:
@mkulke Done. Also, can you approve the workflow checks as well please ?

mkulke (approved):

bpradipt (approved):
/lgtm
Thanks @manjotsidhu

bpradipt:
@stevenhorsman can this go into the upcoming 0.23.0 release ?

stevenhorsman:
> @stevenhorsman can this go into the upcoming 0.23.0 release ?

We are in the middle of the release and the code has already been pinned in https://github.com/confidential-containers/cloud-api-adaptor/pull/3311. That isn't merged yet, so if someone wants to make a case for this being urgent then in could be merged and that PR updated I guess?

stevenhorsman:
@bpradipt - any call on this? Do we want to merge it and also merge https://github.com/confidential-containers/cloud-api-adaptor/pull/3323 given that that vuln has crept in since the initial "release candidate" was proposed and then I can update to pick up a new version of the CAA?

bpradipt:
> @bpradipt - any call on this? Do we want to merge it and also merge #3323 given that that vuln has crept in since the initial "release candidate" was proposed and then I can update to pick up a new version of the CAA?

@stevenhorsman let's merge this post the release.

PR #3235, opened 2026-08-07
config-extractor: verify provider name against registered name in manager

Amulyam24:
The provider name in generated chart files was set from the directory name. It would be ideal to respect the registered name at runtime via `AddCloudProvider()` or fall back to the directory name.


src/cloud-providers/cmd/config-extractor/main.go
stevenhorsman, line 1: Bob has found an issue here due to the output still using the directory name?
> ⚠️ Issue: output filename still derives from the directory name, not the registered name

> 
> The Makefile loop passes the directory name as the argument to config-extractor and uses it for the output filename:
> ```
> # Makefile line 95 — $$p is the directory name, e.g. "ibmcloudpowervs"
> for p in $(find . -maxdepth 2 -name 'manager.go' -printf '%h\n' | sed 's|./||'); do
>     $(call gen-provider-values,$$p);  # writes to ibmcloudpowervs.yaml
> ```
> This PR manually commits ibmcloud-powervs.yaml (the correct registered name), but the next time anyone runs make sync-chart-values it will generate ibmcloudpowervs.yaml (the directory name) — a different file. The two will diverge silently.
> 
> ⚠️ Issue: the CI drift check doesn't catch this
> 
> The drift check in validate-config-extractor.yaml uses:
> ```
> git diff --exit-code ../cloud-api-adaptor/install/charts/peerpods/providers/
> ```
> git diff only reports changes to tracked files. When make sync-chart-values generates ibmcloudpowervs.yaml, it's a new untracked file — git diff doesn't see it, so CI passes. The check needs to also catch new untracked files, e.g.:
> ```
> git diff --exit-code .../providers/
> git ls-files --others --exclude-standard .../providers/ | grep -q . && echo "Untracked files found" && exit 1
> ```
> Suggested fix
> 
> The complete fix needs two parts:
> 
> Have config-extractor output the registered name as part of the JSON (it already does — .provider field), and update the Makefile to use that as the output filename rather than $(1).
> Update the drift check to also detect new untracked files in the providers directory.
> Without those, the filename drift will silently recur on the next make sync-chart-values run.

src/cloud-providers/cmd/config-extractor/main.go
```diff
@@ -97,6 +104,41 @@ func main() {
...
 }
 
+// extractProviderName reads manager.go and returns the string literal passed to
+// AddCloudProvider(). Returns an empty string if the call is not found.
+func extractProviderName(managerPath string) string {
+	fset := token.NewFileSet()
+	node, err := parser.ParseFile(fset, managerPath, nil, 0)
+	if err != nil {
+		return ""
+	}
```
stevenhorsman: Could we make this function accept `node *ast.File`, so we don't need to do this double parse?

src/cloud-providers/cmd/config-extractor/main.go
```diff
@@ -51,7 +51,21 @@ func main() {
 	}
 	execDir := filepath.Dir(execPath)
 
-	config := &ProviderConfig{Provider: provider, Flags: []FlagInfo{}}
+	// Parse provider-specific flags from manager.go
+	managerPath := filepath.Join(execDir, "..", provider, "manager.go")
+	providerFlags, err := parseFile(managerPath)
```
stevenhorsman: If we also return the ast.File here, then we can call `registeredName := extractProviderName(file)` with it?

stevenhorsman (commented):
A bug that Bob has suggested and a possible optimisation

src/cloud-providers/cmd/config-extractor/main.go
Amulyam24: Addressed the bugs, PTAL!

src/cloud-providers/cmd/config-extractor/main.go
Amulyam24: Good point, addressed it.

src/cloud-providers/Makefile
```diff
@@ -6,7 +6,9 @@ all: build test clean
 CHART_PROVIDERS_DIR ?= ../cloud-api-adaptor/install/charts/peerpods/providers
 
 define gen-provider-values
-	./bin/config-extractor -o json -no-secrets -include-shared $(1) | jq -r ' \
+	json=$$(./bin/config-extractor -o json -no-secrets -include-shared $(1)); \
```
stevenhorsman: This isn't introduced by you, but Bob has suggested adding `|| exit` to this to ensure errors are propagated correctly: `json=$$(./bin/config-extractor ...)`?

stevenhorsman: I think this will be the same in gen-provider-secrets too?

src/cloud-api-adaptor/ibmcloud-powervs/README.md
```diff
@@ -82,7 +82,7 @@ pvsadm image import -n <service-instance-name> -b <bucket-name> -o <file-name> -
 1. Setup necessary cloud resources such as PowerVS Service instance, network, API Key etc..
 
 2. Create `ibmcloudpowervs-secrets.yaml` based on
- [`ibmcloud-secrets.yaml.template`](../install/charts/peerpods/providers/ibmcloudpowervs-secrets.yaml.template)
+ [`ibmcloudpowervs-secrets.yaml.template`](../install/charts/peerpods/providers/ibmcloudpowervs-secrets.yaml.template)
  with the IBM Cloud API Key filled in.
 
 3. Update [ibmcloudpowervs.yaml](../install/charts/peerpods/providers/ibmcloudpowervs.yaml) with the required details
```
stevenhorsman: These aren't linking to the correct names?

Amulyam24: Rectified it.

src/cloud-providers/cmd/config-extractor/main.go
```diff
@@ -97,11 +104,41 @@ func main() {
...
+		var fn string
+		switch f := call.Fun.(type) {
+		case *ast.SelectorExpr:
+			fn = f.Sel.Name
+		case *ast.Ident:
+			fn = f.Name
+		}
+		if fn == "AddCloudProvider" && len(call.Args) >= 1 {
+			if lit, ok := call.Args[0].(*ast.BasicLit); ok && lit.Kind == token.STRING {
+				name = strings.Trim(lit.Value, `"`)
```
stevenhorsman: Sorry, could we use `strconv.Unquote(lit.Value)` as a safer option here?

stevenhorsman (approved):

src/cloud-providers/cmd/config-extractor/main.go
Amulyam24: Sure, added it!

chathuryaadapa (approved):

Amulyam24:
Hey @stevenhorsman, can this PR be merged?

stevenhorsman:
> Hey @stevenhorsman, can this PR be merged?

Not without second committer's review

.github/workflows/validate-config-extractor.yaml
```diff
@@ -103,4 +103,20 @@ jobs:
             echo ""
             echo "Then commit the updated files."
             exit 1
+          }
+
+          # Check for newly generated files that are not yet tracked by git
```
bpradipt, line 108: So the expectation is to catch any mismatch with provider name and directory name ?

Amulyam24: yes, that correct,

bpradipt (approved):
/lgtm

PR #3239, opened 2026-08-09
azure: add support for ubuntu podvms

mkulke:
The switch to ubuntu images required some changes to support azure. The individual commits describe the changes in more detail, but the gist is:
- azure-ready tool is introduced, a small tool that implements phone-home functionality required on azure, that is either implemented by cloud-init or afterburn, which we do not use on those images.
- workflows have been introduced that re-use/mirror existing logic as much as possible
- image publishing logic has been encapsulated in a Makefile
- fixes to the azure e2e test suite to allow testing SNP and TDX TEEs in the same cluster
- fix order of deletion in the orphaned vm cleanup logic of the test provisioner


src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.conf
```diff
@@ -18,6 +18,11 @@ KernelInitrdModules=dm-mod
                     ata_piix
                     libata
                     sd_mod
+                    hv_vmbus
```
bpradipt: nit: Adding a comment here to indicate these are specific for Azure/hyperv will be helpful when going through the configs later.

bpradipt (approved):

src/cloud-api-adaptor/podvm/mkosi.images/system/mkosi.conf
mkulke: right, we can also put it into its own mkosi.conf.d/ file I think

src/cloud-api-adaptor/cmd/azure-ready/main.go
```diff
@@ -0,0 +1,230 @@
...
+		},
+	}
+
+	ctx := context.Background()
+	var lastErr error
+	for attempt := 1; attempt <= *retries; attempt++ {
+		if err := report(ctx, client, *endpoint, *verbose); err != nil {
+			lastErr = err
+			log.Printf("attempt %d/%d failed: %v", attempt, *retries, err)
+			time.Sleep(retrySleep)
```
stevenhorsman: It looks like this service has a systemd timeout of 5mins, but this has 30 attempts at 30+5 seconds each (if my maths is right), so if there is an issue it could get SIGKILL'd mid-sleep which can't respond and therefore continue to the logging.

AI suggests using a separate parent context with an overall timeout less that the systemd:
```
ctx, cancel := context.WithTimeout(context.Background(), 4*time.Minute)
defer cancel()
```
and/or using select to make the sleep interruptible:
```
select {
        case <-time.After(retrySleep):
            // Sleep finished, continue loop
        case <-ctx.Done():
            // Context cancelled or timed out (e.g. SIGTERM received)
            log.Fatalf("cancelled during retry sleep: %v", ctx.Err())
        }
```

src/cloud-api-adaptor/cmd/azure-ready/main.go
```diff
@@ -0,0 +1,230 @@
...
+	if err := xml.Unmarshal(body, &v); err != nil {
+		return "", fmt.Errorf("parse versions: %w", err)
+	}
+	for _, candidate := range v.Supported.Version {
+		if candidate == defaultVersion {
+			return candidate, nil
+		}
+	}
+	if len(v.Supported.Version) > 0 {
+		return v.Supported.Version[0], nil
```
stevenhorsman: optional: Maybe a log in this circumstance might be helpful for debug?
```
log.Printf("preferred version %q not found, falling back to %q", defaultVersion, v.Supported.Version[0])
```

src/cloud-api-adaptor/azure/Makefile
```diff
@@ -0,0 +1,97 @@
+# (C) Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+VHD_NAME = "podvm-$(subst .,_,$(IMAGE_VERSION)).vhd"
```
stevenhorsman: I don't think the double quotes are needed, as this will set the literal to include the quotes, which I'm not sure is wanted?

.github/workflows/e2e_run_all.yaml
```diff
@@ -236,3 +260,24 @@ jobs:
...
       peerpod_ctrl_image: ${{ inputs.registry }}/peerpod-ctrl:${{ inputs.caa_image_tag }}-amd64
       webhook_image: ${{ inputs.registry }}/peer-pods-webhook:${{ inputs.caa_image_tag }}
+
+  # Run Azure e2e tests when triggered by schedule or workflow dispatch
+  azure:
+    name: azure
+    permissions:
+      contents: read
+      id-token: write # required for Azure login
+    # The azure e2e test must not run untrusted code via pull_rwquest_target, even if gated by a label.
```
stevenhorsman: nit: Typo: `pull_rwquest_target`

stevenhorsman (commented):
A few suggestions, but this looks okay to me. Is there a test run referenced that I missed?

src/cloud-api-adaptor/azure/Makefile
mkulke: right, this didn't cause problems, but it should not be required in a makefile

mkulke:
> A few suggestions, but this looks okay to me. Is there a test run referenced that I missed?

there are runs on my fork, using a different trigger, though. the full-machinery is tricky to test, so it might require follow-up fixes: https://github.com/mkulke/cloud-api-adaptor/actions/runs/30973384611

src/cloud-api-adaptor/cmd/azure-ready/main.go
mkulke: since this is one-off service, I think we can limit the retries to systemd, this generally should work at the first attempt, if we're in userland the call-home endpoint should be available.

stevenhorsman (approved):
LGTM. Thanks

stevenhorsman:
> > A few suggestions, but this looks okay to me. Is there a test run referenced that I missed?
> 
> there are runs on my fork, using a different trigger, though. the full-machinery is tricky to test, so it might require follow-up fixes: https://github.com/mkulke/cloud-api-adaptor/actions/runs/30973384611

Understood. We are not close to a release, so I think it's fine to merge and give it a try

PR #3240, opened 2026-08-10
libvirt: thread context through volume operations to waitForSuccess

### LINKED ISSUE #3231: Libvirt: add context.Context support to waitForSuccess retry loop
Problem

The libvirt helper waitForSuccess() retries until either the operation succeeds or the internal timeout is reached.

Currently, the retry loop is not context-aware, so callers cannot cancel an in-progress operation when:

a parent request is cancelled,
a reconciliation loop is terminated,
a VM creation flow times out at a higher layer,
the provider is shutting down.

As a result, waitForSuccess() may continue polling until waitTimeout expires even though the caller has already abandoned the operation.

Proposed Change

Add a context.Context parameter to waitForSuccess() and check for cancellation during the retry loop.

chathuryaadapa:
uploadVolume, createVolume, deleteVolume, and deleteVolumeByPath had no context parameter, leaving the ctx argument to waitForSuccess undefined. This meant cancellation signals from the gRPC caller (e.g. kubelet deadline expiry) were never propagated — waitForSuccess would continue retrying libvirt pool refreshes for the full 5-minute waitTimeout even after the caller had given up.

Thread context.Context from CreateDomain and DeleteDomain down through uploadIso → uploadVolume, createVolume, and deleteVolumeByPath → deleteVolume so that waitForSuccess can exit immediately on cancellation. The waitTimeout floor is retained as a safety net when no deadline is set. 
fixes: #3231

### DIFF
--- a/src/cloud-providers/libvirt/volume.go
+++ b/src/cloud-providers/libvirt/volume.go
@@ -8,6 +8,7 @@ package libvirt
 // Code copied from https://github.com/openshift/cluster-api-provider-libvirt
 
 import (
+	"context"
 	"encoding/xml"
 	"errors"
 	"fmt"
@@ -27,19 +28,38 @@ var waitSleepInterval = 1 * time.Second
 // waitTimeout time
 var waitTimeout = 5 * time.Minute
 
-// waitForSuccess wait for success and timeout after 5 minutes.
-func waitForSuccess(errorMessage string, f func() error) error {
+// waitForSuccess calls f repeatedly until it returns nil,
+// the context is cancelled, or waitTimeout elapses (whichever comes first).
+func waitForSuccess(ctx context.Context, errorMessage string, f func() error) error {
 	start := time.Now()
+
+	timer := time.NewTimer(0)
+	if !timer.Stop() {
+		<-timer.C
+	}
+	defer timer.Stop()
+
+	var lastErr error
 	for {
-		err := f()
-		if err == nil {
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("%s: %w", errorMessage, err)
+		}
+
+		lastErr = f()
+		if lastErr == nil {
 			return nil
 		}
-		logger.Printf("%s. Re-trying.\n", err)
+		logger.Printf("%s. Re-trying.\n", lastErr)
 
-		time.Sleep(waitSleepInterval)
 		if time.Since(start) > waitTimeout {
-			return fmt.Errorf("%s: %s", errorMessage, err)
+			return fmt.Errorf("%s: %w", errorMessage, lastErr)
+		}
+
+		timer.Reset(waitSleepInterval)
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("%s: %w", errorMessage, ctx.Err())
+		case <-timer.C:
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/cloud-providers/libvirt/volume.go
```diff
@@ -28,16 +29,25 @@ var waitSleepInterval = 1 * time.Second
...
 		if err == nil {
 			return nil
 		}
 		logger.Printf("%s. Re-trying.\n", err)
 
-		time.Sleep(waitSleepInterval)
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("%s: %w", errorMessage, ctx.Err())
+		case <-time.After(waitSleepInterval):
```
stevenhorsman: time.After create a new Timer, so this will aggregate timer object, creating a memory leak. This is the approach that AI suggests:
```
// waitForSuccess waits for success and times out after waitTimeout if the context has no deadline.
func waitForSuccess(ctx context.Context, errorMessage string, f func() error) error {
	start := time.Now()

	// Create a single timer. Start it stopped so it doesn't fire early.
	timer := time.NewTimer(0)
	if !timer.Stop() {
		<-timer.C
	}
	defer timer.Stop()

	// Keep track of the last error to return if the hard timeout hits
	var lastErr error

	for {
		if err := ctx.Err(); err != nil {
			return fmt.Errorf("%s: %w", errorMessage, err)
		}

		lastErr = f()
		if lastErr == nil {
			return nil
		}
		logger.Printf("%s. Re-trying.\n", lastErr)

		// Hard safety net timeout check
		if time.Since(start) > waitTimeout {
			return fmt.Errorf("%s: %w", errorMessage, lastErr)
		}

		// Safely reset the single timer instance
		timer.Reset(waitSleepInterval)

		select {
		case <-ctx.Done():
			return fmt.Errorf("%s: %w", errorMessage, ctx.Err())
		case <-timer.C:
			// Timer fired safely, loop continues
		}
	}
}
```

src/cloud-providers/libvirt/volume.go
```diff
@@ -28,16 +29,25 @@ var waitSleepInterval = 1 * time.Second
 var waitTimeout = 5 * time.Minute
 
 // waitForSuccess wait for success and timeout after 5 minutes.
```
stevenhorsman: This comment needs updating.

src/cloud-providers/libvirt/volume_test.go
```diff
@@ -209,14 +211,27 @@ func TestWaitForSuccess(t *testing.T) {
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			setWaitTimers(t, tt.timeout, tt.interval)
-
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if tt.cancelCtx {
+				cancel()
```
stevenhorsman: This is only testing pre-cancelled context, so I'm not sure it's covering the code as is? I'm not sure that splicing this code into the existing table driven tests is clean either?

stevenhorsman (approved):
Looks fine once the CI passes (which is currently blocked by a canonical issue)

mkulke (approved):

PR #3242, opened 2026-08-11
azure: enable running also normal instances when DISABLECVM=false

esposem:
In Azure, we could be able to run the CoCo pods together with the non-CoCo (peer pods) without any major change. 
The requirements are that:
1. the image is coming from a gallery that supports Confidential and Trusted launch (no legacy bios images). This can be achieved with `SecurityType=TrustedLaunchAndConfidentialVMSupported`
2. By checking the instance type itself, we can figure the additional fields for the disk in order to make Azure happy.

This of course assumes an image with uefi that has also coco bits inside.


mkulke:
the naming heuristic is not pretty, but I don't have good ideas to avoid them for that use-case at the moment, maybe we can add some explicit qualifiers to instance-size values, `Standard_EC96as_v6::non-confidential` or something. In any case, NCC would also be a confidential VM signifier.

esposem:
@mkulke I added a bit of explanation. Basically with the right gallery and this flag away, we can run coco and non-coco on azure peer pods at the same time

esposem:
I think the failed test has nothing to do with the PR: 
> Pulling image separately not support on main. It is required to use the nydus-snapshotter, which isn't configured properly here.

mkulke (commented):
are we conflating 2 things in this change?

1) non-confidential deployments should be able to use trusted launch, i.e. always a vTPM and optionally SecureBoot, because we assume the image is UEFI only.

2) A heuristic to default to override DisableCVM with the name of the instance-type

I think 1) makes sense. with 2) I am not so sure it's a good idea, at least it seem a bit magical to me: if I e.g. specify DisableCVM=false to never launch non-confidential podvms, but when I choose a specific instance-type (maybe by accident) the instance is launched as non-confidential anyway. I think at best it should produce an error in those cases, but probably the boolean isn't expressive enough to support the use case, maybe an enum ConfidentialVM = Always|Never|InstanceSize would be more fitting? (defaulting to Always)

esposem:
Yes 1. is correct 

I think for 2. you are right, but I don't really want to add a new flag or modify too much as disablecvm is used in also other cloud providers so I don't want to add confusion for the azure use case. 

My idea was to keep leaving disablecvm as a trigger. If we don't want to enable cvms, set it to true and cvms are never picked. 
If instead is not set or false, figure it automatically. 
The logic in the commit is probably wrong, simply because I didn't want to duplicate the code. 
If it's ok for you I'll adjust it as mentioned above

mkulke:
> My idea was to keep leaving disablecvm as a trigger. If we don't want to enable cvms, set it to true and cvms are never picked. If instead is not set or false, figure it automatically. The logic in the commit is probably wrong, simply because I didn't want to duplicate the code. If it's ok for you I'll adjust it as mentioned above

I don‘t think it’s a good idea to disable confidentiality without an explicit flag that a user configured. The instance-size might be a misconfiguration. The golang bool flag cannot distinguish between false and unset, and working around it (eg by checking an env) would be surprising behaviour.

I don‘t think this is a security issue, since every confidential workload must depend on a secret released after remote attestation, but this is often not intuitive for users, so its a ux problem.

if we want to extend the scope of the project to support non-confidential VMs for legit use cases (not just for testing purposes) we should make this obvious in documentation and configuration instead of bolting it on, IMO.

esposem:
@mkulke apologies, I was on PTO. 

My point is that I don't want to overcomplicate this by adding new flags or extending the original one, because we then divert from the other cloud providers.

I changed the PR to make this simpler (It's actually a `&&` instead of `||`):
* If DisableCVM is true, forget about CVMs. Don't even look at the instance size.
* Otherwise, look at the instance size and figure if it's confidential or not.

mkulke:
> @mkulke apologies, I was on PTO.

no worries, same :) 

> My point is that I don't want to overcomplicate this by adding new flags or extending the original one, because we then divert from the other cloud providers.
> 
> I changed the PR to make this simpler (It's actually a `&&` instead of `||`):
> 
> * If DisableCVM is true, forget about CVMs. Don't even look at the instance size.
> * Otherwise, look at the instance size and figure if it's confidential or not.

I understand the intent of making this simple, but I don't think relying on a heuristic is a good idea for the use case. If people want to turn off confidentiality it should be an explicit setting, not something that is autodetected. If someone uses a wrong instance-type without indicating that they do not want confidentiality, podvm creation should fail.

esposem:
> I understand the intent of making this simple, but I don't think relying on a heuristic is a good idea for the use case. If people want to turn off confidentiality it should be an explicit setting, not something that is autodetected. If someone uses a wrong instance-type without indicating that they do not want confidentiality, podvm creation should fail.

And isn't this what I am achieving here? Sorry I am missing something then
Currently if people turn off confidentiality (disableCVM = true), the behavior is the same: the disk security profile is not confidential, and deploying a CVM instance fails (azure rejects it). Otherwise, it adds this special logic to understand the instance type

mkulke:
> And isn't this what I am achieving here? Sorry I am missing something then Currently if people turn off confidentiality (disableCVM = true), the behavior is the same: the disk security profile is not confidential, and deploying a CVM instance fails (azure rejects it). Otherwise, it adds this special logic to understand the instance type

```
	// If disableCVM is true, do not set the security type to ConfidentialVM
	// otherwise, check if the instance size is a confidential VM size to determine if we
	// need to set the security type to ConfidentialVM
```

I understand it like this: if disableCVM is unset (i.e. false, _the default_) we would check the instance-size and determine if we want to have a confidential VM, no?

If that's the case, then I think we shouldn't do this. By default, we should not launch a non-confidential virtual machine, even if the user has (accidentally) configured an instance-type that doesn't support confidentiality.

That said, if all the other providers have a behaviour like this (allowing to downgrade confidentiality implicitly), I think it makes sense to have such a heuristic also here for consistency's sake, albeit this would be an unfortunate design IMO. maybe some other reviewers who see it differently can chime in.

esposem:
> That said, if all the other providers have a behaviour like this (allowing to downgrade confidentiality implicitly), I think it makes sense to have such a heuristic also here for consistency's sake, albeit this would be an unfortunate design IMO. maybe some other reviewers who see it differently can chime in.

The way I understood disableCVM flag was to mainly prevent CVMs to be used, and not the other way around (ie prevent non-cvms from being used). But security-wise I do get that it's not good if one thinks of running a cvm when instead it's a non-cvm instance. Even though attestation should pick this immediately.

I don't know how many other cloud providers actually support coco (but my knowledge is mostly downstream though).
Anyways my goal with this PR would be to finally enable coco and kata-remote together, without the need of switching these flags back and forth.

@snir911  @bpradipt  if you have some suggestion either pls let me know

bpradipt:
I think a better option for per-instance override of the global default is via pod annotation - similar to instance sizes and other annotation we have. Imho explicit config is better w.r.to whether to use cvm or not.

c3d:
> If that's the case, then I think we shouldn't do this. By default, we should not launch a non-confidential virtual machine, even if the user has (accidentally) configured an instance-type that doesn't support confidentiality.

Hey @mkulke, I'm curious why you think this is a bad thing? Let me reverse the argument: if the user purposefully selected a non-confidential instance type, why should we also mandate to explicitly set the `DISABLECVM` flag? Don't you think a user would be surprised that a "normal" VM requires an explicit flag? This feels unintuitive to me, but maybe I misunderstood your point.

IOW, why should we consider the instance type as misconfigured by default?

Also @esposem I think it might be useful to expose your rationale for this change in the PR description (namely being able to switch between confidential and non-confidential without having to reinstall/reconfigure Kata) and mention the other related PRs (e.g. https://github.com/openshift/sandboxed-containers-operator/pull/2628).

stevenhorsman:
I guess the theory is that we what to have a "secure default" and it's better for something trying to request a non-CVM to have a bit of user friction than a use thinking they have a confidential system when they don't?

mkulke:
> Hey @mkulke, I'm curious why you think this is a bad thing? Let me reverse the argument: if the user purposefully selected a non-confidential instance type, why should we also mandate to explicitly set the `DISABLECVM` flag? Don't you think a user would be surprised that a "normal" VM requires an explicit flag? This feels unintuitive to me, but maybe I misunderstood your point.

No, that's my point. My assumption being: the user expects to launch confidential VMs with CAA (sort of supported by the project being part of the CoCo github org, the flag named DISABLECVM vs ENABLECVM, etc). The azure instance-size naming scheme can be tricky and if a user attempts to upgrade their IOPS or something they might choose an instance-size that is not confidential. Now since we still ship an attestation stack in the podvm the attester will fall back to sample attester or possibly tpm attester and at best things will fail at the policy level in kbs, at worst a permissively misconfigured kbs will just verify sample evidence and release a secret.

I am also not sure about having that heuristic being exercised in the default path, without a user explicitly opting in. It's a bit brittle, once azure changes the naming scheme or introduces a new type of CVMs code changes are required.

But as I said, if that is the current behaviour for alibaba, aws, gcp, ibmcloud, etc. and the azure provider is the only odd one that doesn't perform this cvm auto detection scheme, it makes sense to align it.

What about turning that DISABLECVM boolean flag into a string that can be `true|false|auto`? we'd default to `false` as before, but if a user picks `auto` the heuristic is chosen?

esposem:
>But as I said, if that is the current behaviour for alibaba, aws, gcp, ibmcloud, etc. and the azure provider is the only odd one that doesn't perform this cvm auto detection scheme, it makes sense to align it.

I think Azure on the opposite will be the first to support this. Which shouldn't be bad I guess :)

> What about turning that DISABLECVM boolean flag into a string that can be `true|false|auto`? we'd default to false as before, but if a user picks auto the heuristic is chosen?

I think this does not solve your point. You are absolutely right, if the user (or we in the operator) set DISABLECVM: auto then it's easy to confuse the user.

I instead am leaning torward @bpradipt suggestion: if we add `io.katacontainers.config.hypervisor.confidential_vm:` annotation, then the user is free to pick the instance. 

The flow would be:
* DisableCVM defaults to false. If true, don't enable ever CVMs. 
* If DisableCVM is false, look at the annotation: 
** no annotation = assume confidential
** annotation is set to false = non-confidential
** annotation is set to true = confidential

It's a bit more work but this might work without confusing the user

mkulke:
> > What about turning that DISABLECVM boolean flag into a string that can be `true|false|auto`? we'd default to false as before, but if a user picks auto the heuristic is chosen?
> 
> I think this does not solve your point. You are absolutely right, if the user (or we in the operator) set DISABLECVM: auto then it's easy to confuse the user.

why is that? that would be an explicit config option to choose a heuristic to determine whether to use a CVM or not. A default coco user would not set DISABLECVM=auto. That would only be done by someone who knows what they are doing, mostly because they want to run non-confidential VMs in CAA alongside confidential VMs.

> I instead am leaning torward @bpradipt suggestion: if we add `io.katacontainers.config.hypervisor.confidential_vm:` annotation, then the user is free to pick the instance.
> 
> The flow would be:
> 
> * DisableCVM defaults to false. If true, don't enable ever CVMs.
> * If DisableCVM is false, look at the annotation:
>   ** no annotation = assume confidential
>   ** annotation is set to false = non-confidential
>   ** annotation is set to true = confidential
> 
> It's a bit more work but this might work without confusing the user

indeed, that is probably the most comprehensive solution, we wouldn't have to rely on a heuristic in any scenario. but yeah, it's more work.

esposem:
@mkulke I tried to implement the full annotation approach. I didn't test it yet so I am tempted to put this PR on draft, just to avoid confusion? 

> > > What about turning that DISABLECVM boolean flag into a string that can be true|false|auto? we'd default to false as before, but if a user picks auto the heuristic is chosen?

> > I think this does not solve your point. You are absolutely right, if the user (or we in the operator) set DISABLECVM: auto then it's easy to confuse the user.

> why is that? that would be an explicit config option to choose a heuristic to determine whether to use a CVM or not. A default coco user would not set DISABLECVM=auto. That would only be done by someone who knows what they are doing, mostly because they want to run non-confidential VMs in CAA alongside confidential VMs.

I still don't like that, especially if we in the operator default to it. But anyways someone could get confused or forget about this setting. I think the annotation is the way to go

esposem:
Basically now we have:
1. use azure SKU API to figure if the instance is confidential or not (changes the existing way to get the instance list)
2. Use TrustedLaunch instead of normal VMs when DisableCVM is true
3. the annotation approach that is considered only when DisableCVM is false

mkulke:
> Basically now we have:
> 
> 1. use azure SKU API to figure if the instance is confidential or not (changes the existing way to get the instance list)

that's better than the heuristic. I assume we will only resolve the SKUs of the instances sizes that are configured globally once, at startup for CAA? I think the SKU api is somewhat slow, and we wouldn't want to do that at runtime.

> 2. Use TrustedLaunch instead of normal VMs when DisableCVM is true

Maybe we can split this out into its own PR? Since this is uncontroversial, this should get merged quickly. Also, if we use the annotation: TrustedLaunch should probably also be used if DISABLECVM=true and the annotation is confidential=false, no?

> 3. the annotation approach that is considered only when DisableCVM is false

maybe there should be an error if the annotation is confidential=true and the global config is DISABLECVM=false?

esposem:
> that's better than the heuristic. I assume we will only resolve the SKUs of the instances sizes that are configured globally once, at startup for CAA? I think the SKU api is somewhat slow, and we wouldn't want to do that at runtime.

Yes this runs only at startup

> Maybe we can split this out into its own PR? Since this is uncontroversial, this should get merged quickly. Also, if we use the annotation: TrustedLaunch should probably also be used if DISABLECVM=true and the annotation is confidential=false, no?

Sure https://github.com/confidential-containers/cloud-api-adaptor/pull/3258

> maybe there should be an error if the annotation is confidential=true and the global config is DISABLECVM=false?

Agree

Will test this and ping you again once I am confident of this fix

esposem:
Ok it works well. I found a bug and now ensure that if one picks a confidential instance, and the instancelist is full of standard only, it doesn't fall back to the standard. 

Few things still to discuss:
1. name of the annotation. @snir911 was suggesting to follow disablecvm convention? 
2. I would need to open a PR kata shim to make the annotation work 
3. Error format: currently if you add something like (DISABLECVM=false)
```
    io.katacontainers.config.hypervisor.confidential_vm: "true"
    io.katacontainers.config.hypervisor.machine_type: Standard_D2as_v5
```
and run it, the pod fails but the error in caa is something like
```
Instance type selected by the cloud provider based on instance type annotation: Standard_D2as_v5
error starting instance: creating an instance : failed to verify instance type: requested instance type ("Standard_D2as_v5") is not part of supported instance types list
```

Which might be difficult to spot?

src/cloud-providers/azure/provider.go
```diff
@@ -374,21 +378,58 @@ func (p *azureProvider) ConfigVerifier() error {
...
+	if annotation != nil {
+		return *annotation
+	}
+	return true
+}
+
 // Add SelectInstanceType method to select an instance type based on the memory and vcpu requirements
-func (p *azureProvider) selectInstanceType(ctx context.Context, spec provider.InstanceTypeSpec) (string, error) {
+func (p *azureProvider) selectInstanceType(ctx context.Context, spec provider.InstanceTypeSpec, confidentialVM bool) (string, error) {
+	specList := p.filterInstanceSizesByConfidentiality(confidentialVM)
```
mkulke:
```suggestion
	specList := p.filterInstanceTypeSpecsByConfidentiality(confidentialVM)
```

src/cloud-providers/azure/provider.go
```diff
@@ -374,21 +378,58 @@ func (p *azureProvider) ConfigVerifier() error {
...
+	validSizes := p.filterValidSizesByConfidentiality(confidentialVM)
+	defaultSize := p.serviceConfig.Size
```
mkulke: chosenSize maybe?

src/cloud-providers/azure/provider.go
```diff
@@ -374,21 +378,58 @@ func (p *azureProvider) ConfigVerifier() error {
...
+	}
+	if annotation != nil {
+		return *annotation
+	}
+	return true
+}
+
 // Add SelectInstanceType method to select an instance type based on the memory and vcpu requirements
-func (p *azureProvider) selectInstanceType(ctx context.Context, spec provider.InstanceTypeSpec) (string, error) {
+func (p *azureProvider) selectInstanceType(ctx context.Context, spec provider.InstanceTypeSpec, confidentialVM bool) (string, error) {
```
mkulke: passing "confidentialVM" through all the signatures is a bit noisy, since we can query this info on the azureProvider struct, we can maybe deeper in the call stack level where it is required.

cvmRequested expresses the intent maybe a bit better

PR #3248, opened 2026-08-21
libvirt: fix empty instanceID when domain already exists

### LINKED ISSUE #3247: libvirt: CreateDomain returns empty instanceID when domain already exists
## Problem

When `CreateDomain` finds an existing domain via `checkDomainExistsByName`, it returns the input `vmConfig` unchanged. At that point `v.instanceID` is still `""` — the UUID is only assigned after `DomainDefineXML`/`GetUUIDString`, which the early-return path never reaches.

The empty ID propagates up through `CreateInstance` into `cloud.go`, where the deferred cleanup guard:

```go
if err != nil && instance != nil && instance.ID != "" {
    s.provider.DeleteInstance(cleanupCtx, instance.ID)
}
```

is always false, so the ghost domain is never cleaned up. Every subsequent create attempt for the same sandbox name hits the same path, making the failure self-reinforcing.

## Root Cause

`checkDomainExistsByName` calls `LookupDomainByName`, gets the domain handle, then immediately frees it and returns only a boolean. The UUID is discarded. The caller has no handle left to call `GetUUIDString` on.

## Steps to Reproduce

1. Start CAA with the libvirt provider
2. Trigger a partial failure after a domain is created (e.g. bad initdata) so the domain exists but cleanup fails
3. Retry pod creation for the same sandbox — `CreateDomain` hits the early-return path and returns `instanceID = ""`
4. Any subsequent error leaves the domain running with no cleanup path

## Fix

Inline `LookupDomainByName` into `CreateDomain`, populate `instanceID` and `ips` from the existing domain before returning, and remove `checkDomainExistsByName` (dead code).

## Related

Sibling of #2963 / PR #2964, which fixed the same class of problem (empty/wrong `instanceID`) on the new-domain path.

ajaypvictor:
## Problem

When `CreateDomain` finds an existing domain via `checkDomainExistsByName`, it returns the input `vmConfig` unchanged. `v.instanceID` is still `""` at that point — the UUID is only assigned after `DomainDefineXML`/`GetUUIDString`, which the early-return path never reaches.

The empty ID propagates up through `CreateInstance` into `cloud.go`, where the deferred cleanup guard:

```go
if err != nil && instance != nil && instance.ID != "" {
    s.provider.DeleteInstance(cleanupCtx, instance.ID)
}
```

is always false, so the ghost domain is never cleaned up. Every subsequent create attempt for the same sandbox name hits the same path — self-reinforcing.

## Root Cause

`checkDomainExistsByName` calls `LookupDomainByName`, gets the domain handle, then immediately frees it and returns only a boolean. The UUID is discarded. The caller has no handle left to call `GetUUIDString` on.

## Fix

Inline `LookupDomainByName` into `CreateDomain`, populate `instanceID` and `ips` from the existing domain before returning, and remove `checkDomainExistsByName` (now dead code). This also eliminates the double libvirt round-trip the old pattern caused.

## Testing

Two integration tests added to `libvirt_test.go`, both gated by `checkConfig(t)` (require `LIBVIRT_URI`):

- `TestCreateDomainExistingDomain`: pre-creates a domain, calls `CreateDomain`, asserts `instanceID` equals the domain's UUID
- `TestCreateDomainNoExistingDomain`: no pre-existing domain, asserts the call falls through to the normal create path

## Related

Sibling of #2963 / #2964, which fixed the same class of problem on the new-domain path.

Fixes: #3247

### DIFF
--- a/src/cloud-providers/libvirt/libvirt.go
+++ b/src/cloud-providers/libvirt/libvirt.go
@@ -79,22 +79,6 @@ func createCloudInitISO(v *vmConfig) ([]byte, error) {
 	return createCloudInit([]byte(userData), []byte(metaData))
 }
 
-func checkDomainExistsByName(name string, libvirtClient *libvirtClient) (exist bool, err error) {
-
-	logger.Printf("Checking if instance (%s) exists", name)
-	domain, err := libvirtClient.connection.LookupDomainByName(name)
-	if err != nil {
-		if err.(libvirt.Error).Code == libvirt.ERR_NO_DOMAIN {
-			return false, nil
-		}
-		return false, err
-	}
-	defer freeDomain(domain, &err)
-
-	return true, nil
-
-}
-
 func uploadIso(ctx context.Context, isoData []byte, isoVolName string, libvirtClient *libvirtClient) (string, error) {
 
 	logger.Printf("Uploading iso file: %s\n", isoVolName)
@@ -573,15 +557,31 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
 
 	v.rootDiskSize = normalizeRootDiskSize(v.rootDiskSize)
 
-	exists, err := checkDomainExistsByName(v.name, libvirtClient)
-	if err != nil {
-		return nil, fmt.Errorf("Error in checking instance: %s", err)
+	logger.Printf("Checking if instance (%s) exists", v.name)
+	domain, err := libvirtClient.connection.LookupDomainByName(v.name)
+	if err == nil {
+		// Domain already exists. Populate instanceID and ips so the caller
+		// has a valid UUID for cleanup if anything goes wrong subsequently.
+		defer freeDomain(domain, &err)
+		uuid, uuidErr := domain.GetUUIDString()
+		if uuidErr != nil {
+			return nil, fmt.Errorf("existing domain UUID retrieval failed: %w", uuidErr)
+		}
+		v.instanceID = uuid
+		ips, ipsErr := getDomainIPs(domain)
+		if ipsErr != nil {
+			// UUID is valid; return the instance so the caller can clean up
+			// even if IP retrieval fails (e.g. domain is shut off).
+			logger.Printf("existing domain IP retrieval failed (uuid=%s): %v", uuid, ipsErr)
+			return &createDomainOutput{instance: v}, fmt.Errorf("existing domain IP retrieval failed: %w", ipsErr)
+		}
+		v.ips = ips
+		logger.Printf("Instance already exists (uuid=%s)", uuid)
+		return &createDomainOutput{instance: v}, nil
 	}
-	if exists {
-		logger.Printf("Instance already exists ")
-		return &createDomainOutput{
-			instance: v,
-		}, nil
+	var libvirtErr libvirt.Error
+	if !errors.As(err, &libvirtErr) || libvirtErr.Code != libvirt.ERR_NO_DOMAIN {
+		return nil, fmt.Errorf("error checking instance: %w", err)
 	}
 
 	rootVolName := v.name + "-root.qcow2"
--- a/src/cloud-providers/libvirt/libvirt_test.go
+++ b/src/cloud-providers/libvirt/libvirt_test.go
@@ -4,6 +4,7 @@
 package libvirt
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


src/cloud-providers/libvirt/libvirt.go
```diff
@@ -573,15 +557,27 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
...
+		if v.ips, err = getDomainIPs(domain); err != nil {
+			// Domain exists and UUID is valid; return the instance so the caller
+			// can clean up even if IP retrieval fails (e.g. domain is shut off).
+			logger.Printf("existing domain IP retrieval failed (uuid=%s): %v", uuid, err)
+			return &createDomainOutput{instance: v}, fmt.Errorf("existing domain IP retrieval failed: %w", err)
+		}
+		logger.Printf("Instance already exists (uuid=%s)", uuid)
+		return &createDomainOutput{instance: v}, nil
 	}
-	if exists {
```
snir911, line 580: earlier if domain existed already, we did nothing, what would happen now? in current implementation, is it possible an existing uuid will be overridden with a wrong one?

mkulke (commented):
Is this desired? Shouldn't we bail out if a domain exists?

src/cloud-providers/libvirt/libvirt.go
ajaypvictor: `vmConfig` is newly allocated on every `CreateInstance` call (provider.go:85) looks like and `instanceID` starts as "", it is never set by the caller before being passed into `CreateDomain`. So `v.instanceID = uuid` is the first write to that field, not an override. It gets the UUID directly from `GetUUIDString` on the existing domain, which is exactly what libvirt assigned to it.

On your second point, I think previously when a domain already existed, `CreateDomain` just returned early without touching `instanceID`, leaving the caller with an empty string and no way to reference or clean up that domain. The fix makes it behave the same way as the new domain path look up the UUID, update it, and return it.

Hope that makes sense?

ajaypvictor:
> Is this desired? Shouldn't we bail out if a domain exists?

The domain already exists path looks to be intentional from the beginning if I understand it correctly, it was added as a check to handle retry scenarios where StartVM is called again for the same sandbox after a partial failure (domain created but instance ID never stored), bailing out there would break recovery.

The bug was not that this path exists, it was that it returned without populating instanceID, so the caller had no way to reference or clean up the domain. This fix keeps the same behaviour, just ensures we return the UUID and IPs correctly before returning.

src/cloud-providers/libvirt/libvirt.go
```diff
@@ -573,15 +557,27 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
...
 	v.rootDiskSize = normalizeRootDiskSize(v.rootDiskSize)
 
-	exists, err := checkDomainExistsByName(v.name, libvirtClient)
-	if err != nil {
-		return nil, fmt.Errorf("Error in checking instance: %s", err)
+	logger.Printf("Checking if instance (%s) exists", v.name)
+	domain, err := libvirtClient.connection.LookupDomainByName(v.name)
+	if err == nil {
+		defer freeDomain(domain, &err)
+		uuid, err := domain.GetUUIDString()
```
snir911: can you add a comment mentioning this is needed in order to allow clean up of the already existing domain when identified..

src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -1316,3 +1317,61 @@ func TestNewDefVolumeCapacity(t *testing.T) {
...
+		d, lookupErr := client.connection.LookupDomainByName(domName)
+		if lookupErr != nil {
+			return
+		}
+		_ = d.Undefine()
+		_ = d.Free()
+	}()
+
+	v := &vmConfig{name: domName}
+	result, err := CreateDomain(context.Background(), client, v)
```
snir911, line 1353: IIUC you should run this twice and see it also returns a valid uuid in the second execution isn't it? as currently it returns the uuid which is being set at the "domain creation path" rather than in the "domain already exist" path

src/cloud-providers/libvirt/libvirt.go
ajaypvictor: done, thanks Snir for the suggestion..

src/cloud-providers/libvirt/libvirt_test.go
ajaypvictor: adjusted the test accordingly, thanks Snir for pointing this..

snir911 (approved):
LGTM, assuming test are passing, thanks

src/cloud-providers/libvirt/libvirt.go
```diff
@@ -573,15 +557,29 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
...
+		}
+		v.instanceID = uuid
+		if v.ips, err = getDomainIPs(domain); err != nil {
+			// Domain exists and UUID is valid; return the instance so the caller
+			// can clean up even if IP retrieval fails (e.g. domain is shut off).
+			logger.Printf("existing domain IP retrieval failed (uuid=%s): %v", uuid, err)
+			return &createDomainOutput{instance: v}, fmt.Errorf("existing domain IP retrieval failed: %w", err)
+		}
+		logger.Printf("Instance already exists (uuid=%s)", uuid)
+		return &createDomainOutput{instance: v}, nil
```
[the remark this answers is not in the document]
ajaypvictor: The existing domain path is intentionally singleshot here. A newly created domain needs time for DHCP to assign an IP, so the retry loop makes sense there. An existing domain has already been running if it has no IPs at this point it is likely shut off, and retrying would not change that I guess? This was discussed in the review and the best solution was chosen here.

src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -1316,3 +1317,74 @@ func TestNewDefVolumeCapacity(t *testing.T) {
...
+// It calls CreateDomain twice: the first call creates the domain and returns the UUID via the normal path,
+// the second call hits the "domain already exists" path and must return the same UUID.
+func TestCreateDomainExistingDomain(t *testing.T) {
+	checkConfig(t)
+
+	client, err := NewLibvirtClient(testCfg)
+	require.NoError(t, err)
+	defer client.connection.Close()
+
+	const domName = "TestCreateDomainExisting"
```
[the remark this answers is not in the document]
ajaypvictor: The existing integration tests in this file uses hardcoded names (`TestCreateDomainS390x`, `TestCreateDomainAArch64`), so keeping it consistent with that convention. The cleanup defer at the top of the test handles any domain left behind from a failed prior run.

src/cloud-providers/libvirt/libvirt.go
```diff
@@ -573,15 +557,29 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
...
-	exists, err := checkDomainExistsByName(v.name, libvirtClient)
-	if err != nil {
-		return nil, fmt.Errorf("Error in checking instance: %s", err)
+	logger.Printf("Checking if instance (%s) exists", v.name)
+	domain, err := libvirtClient.connection.LookupDomainByName(v.name)
+	if err == nil {
+		// Domain already exists. Populate instanceID and ips so the caller
+		// has a valid UUID for cleanup if anything goes wrong subsequently.
+		defer freeDomain(domain, &err)
+		uuid, err := domain.GetUUIDString()
```
stevenhorsman: Is the shadowing of the out err an issue?

src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -1316,3 +1317,73 @@ func TestNewDefVolumeCapacity(t *testing.T) {
...
+	defer client.connection.Close()
+
+	v := &vmConfig{
+		name:    "TestCreateDomainNonExistent",
+		volName: "nonexistent-backing.qcow2",
+	}
+
+	_, err = CreateDomain(context.Background(), client, v)
+	require.Error(t, err, "expected error from missing volume, not from domain lookup")
+	assert.Contains(t, err.Error(), "volume",
```
stevenhorsman: Can you use assert.ErrorContains here please

stevenhorsman (approved):
A couple of nits, but it generally looks ok to me. Thanks

src/cloud-providers/libvirt/libvirt.go
ajaypvictor: thanks, good catch, fixed it by using different variable names `uuidErr` and `ipsErr` throughout the block.

src/cloud-providers/libvirt/libvirt_test.go
ajaypvictor: fixed, thanks..

stevenhorsman (approved):

PR #3249, opened 2026-08-21
podvm: remove quay.io secret from podvm builds

mkulke:
Currently the builds pass through a secret to quay.io. This will prohibit the building of the images on forks for pull request builds, forcing the use of the insecure pull-request-target mechanism in e2e tests, even though podvm builds can be executed safely on forks.

This change removes the secret and its uses at the call site and hard codes it to ghcr.io, which can be used in forks.

We can specify an additional registry in the podvm_publish step that will mirror the built images from ghcr.io to that registry. If it's a quay.io registry we login to that registry.

related to #2214

### DIFF
--- a/.github/workflows/podvm_publish.yaml
+++ b/.github/workflows/podvm_publish.yaml
@@ -29,9 +29,6 @@ on:
         required: false
         type: string
         default: ''
-    secrets:
-      QUAY_PASSWORD:
-        required: true
 
 permissions: {}
 
@@ -53,6 +50,3 @@ jobs:
       image_tag: ${{ inputs.image_tag || github.sha }}
       arch: ${{ matrix.arch}}
       debug: false
-      registry: ${{ inputs.registry != '' && inputs.registry || (github.repository == 'confidential-containers/cloud-api-adaptor' && 'quay.io/confidential-containers' || format('ghcr.io/{0}', github.repository_owner)) }}
-    secrets:
-      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
--- a/.github/workflows/podvm_publish.yaml
+++ b/.github/workflows/podvm_publish.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
     inputs:
       registry:
-        description: 'Container registry to push images to (e.g., quay.io/your-username)'
+        description: 'Alternative container registry to push images to (e.g., quay.io/your-username)'
         required: false
         type: string
         default: ''
@@ -25,12 +25,19 @@ on:
         type: string
         default: ''
       registry:
-        description: 'Container registry to push images to (e.g., quay.io/your-username)'
+        description: 'Alternative container registry to push images to (e.g., quay.io/your-username)'
         required: false
         type: string
         default: ''
+    secrets:
+      QUAY_PASSWORD:
+        required: false
 
 permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || inputs.git_ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
   podvm:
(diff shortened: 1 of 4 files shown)


stevenhorsman:
I'm fine to change the registries for the PR runs, but I don't think this change as is make sense for the release as it looks like we'll have some of our release images in quay and the podvm in ghcr? Can you leave the secrets in the release flow and keep quay.io for the release builds, but ghcr for pr/nightly ones? Either that or we should move all our images to ghcr.io, rather than having half and half?

mkulke:
> I'm fine to change the registries for the PR runs, but I don't think this change as is make sense for the release as it looks like we'll have some of our release images in quay and the podvm in ghcr? Can you leave the secrets in the release flow and keep quay.io for the release builds, but ghcr for pr/nightly ones? Either that or we should move all our images to ghcr.io, rather than having half and half?

If we want to enable runs of e2e on pull-requests workflows `podvm_mkosi.yaml` should never have access to secrets for pushing to the orgs OCI repo. IMO this is important security hygiene, having a secret but overriding it with `QUAY_PASSWORD: ''` would be a footgun.

If we need to also push to quay.io in the release workflow, I think the correct thing to do is to skopeo cp oci://ghcr.io/... oci://quay.io/... in the release wf. would that work?

stevenhorsman:
> > I'm fine to change the registries for the PR runs, but I don't think this change as is make sense for the release as it looks like we'll have some of our release images in quay and the podvm in ghcr? Can you leave the secrets in the release flow and keep quay.io for the release builds, but ghcr for pr/nightly ones? Either that or we should move all our images to ghcr.io, rather than having half and half?
> 
> If we want to enable runs of e2e on pull-requests workflows `podvm_mkosi.yaml` should never have access to secrets for pushing to the orgs OCI repo. IMO this is important security hygiene, having a secret but overriding it with `QUAY_PASSWORD: ''` would be a footgun.
> 
> If we need to also push to quay.io in the release workflow, I think the correct thing to do is to skopeo cp oci://ghcr.io/... oci://quay.io/... in the release wf. would that work?

I don't think we need to push to quay.io in the release workflow, but that means updating the other image build and pushes (e.g. caa, peerpodctrl, webhook) to switch to ghcr.io as well, which we need to do anyway based on your logic that we shouldn't have the secrets in e2e_run_all? Or is this part 1 and those changes will follow before we next release?

mkulke:
> I don't think we need to push to quay.io in the release workflow, but that means updating the other image build and pushes (e.g. caa, peerpodctrl, webhook) to switch to ghcr.io as well, which we need to do anyway based on your logic that we shouldn't have the secrets in e2e_run_all? Or is this part 1 and those changes will follow before we next release?

hmm, good point. I think those default to ghcr atm, otherwise e2e  tests wouldn't work on a fork, which they do for me (I do not have the quay secret). If that is correct, we can do it iteratively. I'm open about dropping the quay pushes, but maybe it breaks something downstream. I think we can copy-push in the release workflow for now

.github/workflows/podvm_publish.yaml
```diff
@@ -53,6 +57,42 @@ jobs:
...
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Login to quay.io container registry
+      uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # v4.4.0
+      with:
+        registry: quay.io
+        username: ${{ vars.QUAY_USERNAME }}
+        password: ${{ secrets.QUAY_PASSWORD }}
+
+    - uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
```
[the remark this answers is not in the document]
mkulke: that yaml is indented fine, actionlint doesn't complain

.github/workflows/podvm_publish.yaml
```diff
@@ -50,3 +57,42 @@ jobs:
...
+    name: Push podvm image to quay.io
+    permissions:
+      packages: read # Required to copy container images
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64, s390x, arm64]
+    needs: podvm
+    if: ${{ startsWith(inputs.registry, 'quay.io') }}
+    runs-on: ubuntu-latest
```
stevenhorsman: Can we pin this runner?

stevenhorsman (approved):
I think this is fine to try and we should consider whether we can migrate everything over to ghcr.io as a follow on step, which I guess we need to check with downstream, or do as a deprecation notice?

.github/workflows/podvm_publish.yaml
stevenhorsman: FYI actionlint has gone unmaintained, so in the short term - if you want to use `ubuntu-26.04`, it will need to be added to the self-hosted runner list: https://github.com/confidential-containers/cloud-api-adaptor/blob/main/.github/actionlint.yaml 😞

bpradipt (approved):

PR #3256, opened 2026-08-26
Minor Powervs fixes

Amulyam24:
- Update default systype from s922 to s1022
- Validate instance type before VM creation
- Add configurable DHCP timeout
- Validate required fields in ConfigVerifier

Assisted-by: IBM Bob

### DIFF
--- a/src/cloud-providers/ibmcloudpowervs/provider.go
+++ b/src/cloud-providers/ibmcloudpowervs/provider.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"net/netip"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


src/cloud-providers/ibmcloudpowervs/powervs.go
```diff
@@ -40,19 +40,11 @@ func newPowervsClient(apikey, serviceinstanceID, zone string) (*powervsService,
 	}
 
 	return &powervsService{
-		session:           piSession,
-		serviceInstanceID: serviceinstanceID,
+		instanceClient: instance.NewIBMPIInstanceClient(context.Background(), piSession, serviceinstanceID),
```
stevenhorsman: Is there an issue here about the context being removed? Originally all the methods passed in their context, but now they share a background one.

src/cloud-providers/ibmcloudpowervs/provider.go
```diff
@@ -200,14 +200,46 @@ func (p *ibmcloudPowerVSProvider) Teardown() error {
...
+		errs = append(errs, fmt.Sprintf("SystemType %q is invalid, must be one of %v", p.serviceConfig.SystemType, validSystemTypes))
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("invalid PowerVS config: %s", strings.Join(errs, "; "))
 	}
 	return nil
 }
 
+func containsStr(slice []string, val string) bool {
```
stevenhorsman: Can we just use `slices.Contains` rather than creating our own function for this?

src/cloud-providers/ibmcloudpowervs/provider.go
```diff
@@ -200,14 +200,46 @@ func (p *ibmcloudPowerVSProvider) Teardown() error {
 	return nil
 }
 
+var validProcessorTypes = []string{"shared", "dedicated", "capped"}
+var validSystemTypes = []string{"s922", "s1022", "s1122", "e980", "e1080"}
```
stevenhorsman: Could we move this into the `ConfigVerifier`, or are they purposely at package scope for use elsewhere?

stevenhorsman (commented):
Some questions/suggestions...

src/cloud-providers/ibmcloudpowervs/provider.go
Amulyam24: Yes definitely, cleaned up a bit of the AI slop.

src/cloud-providers/ibmcloudpowervs/provider.go
Amulyam24: I don't foresee using it outside the function for now, moving it.

mkulke (approved):

src/cloud-providers/ibmcloudpowervs/powervs.go
Amulyam24: Initial thought was to avoid constructing the client on every call and refactor it further to use the root context - this doesn't seem to fit in the flow. Retaining the original flow is better for now, dropped the commit, Thanks!

stevenhorsman (approved):
LGTM. Thanks

PR #3269, opened 2026-09-02
libvirt: emit launchSecurity for s390-pv in domain XML

### LINKED ISSUE #3268: libvirt: s390-pv launch security type is detected but not written to domain XML
Problem

When the libvirt provider is configured with `--launch-security s390-pv`, or
auto-detects S390PV on an s390x host, it correctly resolves and logs the launch
security type. However, `createDomainXMLs390x` never reads `vm.launchSecurityType`
and returns the domain struct with no `<launchSecurity>` element set.

The VM is therefore defined and started as a plain KVM guest instead of an IBM
Secure Execution protected guest, with no error or warning to the caller.

By contrast, `createDomainXMLx86_64` already switches on `vm.launchSecurityType`
and returns an error for any type it does not handle. The s390x path has no
equivalent guard, so the silent discard also goes undetected.

The `libvirtxml` library already provides
`DomainLaunchSecurity.S390PV` and the `MarshalXML` implementation that emits
`<launchSecurity type='s390-pv'/>`. No additional library changes are needed.

Proposed Change

Add a switch on `vm.launchSecurityType` at the end of `createDomainXMLs390x`
that sets `domain.LaunchSecurity` when `S390PV` is requested, returns the domain
unchanged for `NoLaunchSecurity`, and returns an error for any unrecognised type,
mirroring the pattern already used by `createDomainXMLx86_64`.

chathuryaadapa:
createDomainXMLs390x accepted S390PV as the launch security type (set via --launch-security s390-pv or auto-detected on s390x hosts) but never wrote the <launchSecurity type='s390-pv'/> element to the domain XML. The VM was started as a plain KVM guest instead of an IBM Secure Execution protected guest.

Apply the same switch-on-launchSecurityType pattern used by createDomainXMLx86_64: emit the libvirtxml.DomainLaunchSecurity S390PV element when S390PV is requested, return the domain unchanged for NoLaunchSecurity, and return an error for any unrecognised type.

The libvirtxml library (v1.12002.0) already provides DomainLaunchSecurityS390PV and the correct MarshalXML implementation that emits type='s390-pv'; no extra fields are required.

Add TestCreateDomainXMLs390xLaunchSecurity covering all three cases.

fixes: #3268

### DIFF
--- a/src/cloud-providers/libvirt/libvirt_test.go
+++ b/src/cloud-providers/libvirt/libvirt_test.go
@@ -968,6 +968,64 @@ func TestCreateDomainXMLArchitecturesWithMocks(t *testing.T) {
 	}
 }
 
+func TestCreateDomainXMLs390xLaunchSecurity(t *testing.T) {
+	mockCaps := createMockCaps(archS390x, "/usr/bin/qemu-system-s390x", libvirtxml.CapsGuestMachine{
+		Name:      "s390-ccw-virtio",
+		Canonical: "s390-ccw-virtio-rhel9.0.0",
+	})
+	mockClient := &libvirtClient{
+		caps:        mockCaps,
+		networkName: testNetworkName,
+	}
+	cfg := createTestDomainConfig("test-s390x-sec", 2, 2048, testNetworkName, testCiDataISO)
+
+	tests := []struct {
+		name           string
+		launchSecurity LaunchSecurityType
+		expectedError  string
+		expectS390PV   bool
+	}{
+		{
+			name:           "S390PV emits launchSecurity element",
+			launchSecurity: S390PV,
+			expectS390PV:   true,
+		},
+		{
+			name:           "NoLaunchSecurity omits launchSecurity element",
+			launchSecurity: NoLaunchSecurity,
+			expectS390PV:   false,
+		},
+		{
+			name:           "unknown security type returns error",
+			launchSecurity: LaunchSecurityType(99),
+			expectedError:  "launch security type unknown is not supported for s390x",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -968,6 +968,69 @@ func TestCreateDomainXMLArchitecturesWithMocks(t *testing.T) {
 	}
 }
 
+// TestCreateDomainXMLs390xLaunchSecurity tests that createDomainXMLs390x correctly
+// emits <launchSecurity type='s390-pv'/> when S390PV is requested, passes through
+// with no launchSecurity element when NoLaunchSecurity is set, and rejects unknown types.
```
stevenhorsman: I'm not sure if we need this comment as I think the test cases are pretty self-describing?

src/cloud-providers/libvirt/libvirt_test.go
```diff
@@ -968,6 +968,69 @@ func TestCreateDomainXMLArchitecturesWithMocks(t *testing.T) {
...
+		{
+			name:           "NoLaunchSecurity omits launchSecurity element",
+			launchSecurity: NoLaunchSecurity,
+			expectError:    false,
+			expectS390PV:   false,
+		},
+		{
+			name:           "unknown security type returns error",
+			launchSecurity: LaunchSecurityType(99),
+			expectError:    true,
```
stevenhorsman: Could we have an expectedMessage checked please - probably using something like `expectedError: "launch security type unknown is not supported for s390x"` and then empty string if there isn't an error expected?

stevenhorsman (commented):
Minor suggestions

stevenhorsman (approved):
LGTM - thanks!

mkulke (approved):

PR #3272, opened 2026-09-02
workflows: Add extra trigger to validate-config-extractor

stevenhorsman:
In #3257 we found that the change made in #3261 caused a mis-match with the generated provider yaml. The cause was that the workflow job that checks it, didn't run on that PR, so add an extra path trigger to plug this gap.


Amulyam24 (commented):
LGTM, Thank you @stevenhorsman!

Amulyam24 (approved):

bpradipt (approved):

PR #3278, opened 2026-09-04
ci: allow webhook build without quay secret

mkulke:
Currently a webhook builds require a secret to quay.io. This will prohibit the building of the images on forks. This is part of an effort to allow running full e2e tests on forks, thus reducing the security problems of sharing repo secrets via pull_request_target.

This change removes the secret from the build step. We decouple the quay.io pushes into a new "webhook_image_publish" workflow, that will mirror and image from ghcr to quay.io. The publish workflow is invoked from both "release" and "publish_images_on_push" workflows.

### DIFF
--- a/.github/workflows/webhook_image.yaml
+++ b/.github/workflows/webhook_image.yaml
@@ -7,11 +7,6 @@ name: (Callable) Build and push webhook image
 on:
   workflow_call:
     inputs:
-      registry:
-        default: 'quay.io/confidential-containers'
-        description: 'Image registry (e.g. "ghcr.io/confidential-containers") where the built image will be pushed to'
-        required: false
-        type: string
       image_tags:
         description: 'Comma-separated list of tags for the dev built image (e.g. latest,ci-dev). By default uses the values from hack/build.sh'
         required: true
@@ -21,9 +16,10 @@ on:
         description: Git ref to checkout the cloud-api-adaptor repository. Defaults to main.
         required: false
         type: string
-    secrets:
-      QUAY_PASSWORD:
-        required: true
+    outputs:
+      image-name:
+        description: 'The name of the built image'
+        value: ${{ jobs.build_push_webhook.outputs.image-name }}
 
 permissions: {}
 
@@ -37,6 +33,8 @@ jobs:
     permissions:
       contents: read  # Required to checkout code
       packages: write  # Required to push container images
+    outputs:
+      image-name: ${{ steps.build-and-push-image.outputs.image-name }}
     steps:
       - name: Checkout the code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -58,24 +56,18 @@ jobs:
           cache: false
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
-      - name: Login to quay Container Registry
-        if: ${{ startsWith(inputs.registry, 'quay.io') }}
-        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
-        with:
-          registry: quay.io
-          username: ${{ vars.QUAY_USERNAME }}
-          password: ${{ secrets.QUAY_PASSWORD }}
       - name: Login to Github Container Registry
-        if: ${{ startsWith(inputs.registry, 'ghcr.io') }}
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
+          username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push image
+        id: build-and-push-image
         env:
           IMAGE_TAGS: ${{ inputs.image_tags }}
-          REGISTRY: ${{ inputs.registry }}
+          REGISTRY: ghcr.io/${{ github.repository_owner }}
+          IMAGE_NAME: peer-pods-webhook
         run: |
           tags="${IMAGE_TAGS}"
           latest=0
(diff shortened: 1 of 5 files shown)


.github/workflows/webhook_image.yaml
```diff
@@ -58,24 +56,18 @@ jobs:
...
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push image
+        id: build-and-push-image
         env:
-          IMAGE_TAGS: ${{ inputs.image_tags }}
-          REGISTRY: ${{ inputs.registry }}
+          IMAGE_TAGS: ${{ inputs.image_tag }},latest
```
stevenhorsman: Just to check - in the commit message it says:
> we always push to latest, but not when pushing a release

what's the mechanism that stops the image being pushed to ghcr at release time  as we seem to have the latest tag here still?

stevenhorsman (commented):
One question on the workflow and have you tested in with a branch on upstream, or are you planning to just test post-merge?

.github/workflows/webhook_image_publish.yaml
```diff
@@ -0,0 +1,85 @@
...
+    permissions:
+      contents: read  # Required to checkout code
+      packages: write  # Required to push container images
+
+  push-to-quay:
+    name: Push webhook image to quay.io
+    permissions:
+      packages: read # Required to copy container images
+    needs: webhook
+    if: ${{ startsWith(inputs.registry, 'quay.io') }}
```
ldoktor, line 57: Shouldn't the `registry` be passed in the `release.yaml` in order for this to build automatically?

.github/workflows/webhook_image.yaml
ldoktor: Also unless we remove the current `latest`, it will stay there and people might falsely assume it's the latest one. Eg. I often rely on `latest`, not consuming it directly but instead resolving it to it's hash at the beginning of the pipeline and then using the hash in nightly testing. I do find `latest` tag useful.

mkulke:
> One question on the workflow and have you tested in with a branch on upstream, or are you planning to just test post-merge?

haven't tested on upstream with quay secret. I think I can do that for the push workflows, it will be tricker for the release workflow...

.github/workflows/webhook_image.yaml
mkulke: I see. so we should push the latest tag for webhook on push-to-main and release? (but not PRs I assume)

.github/workflows/webhook_image_publish.yaml
mkulke: indeed, good catch. we can also set it as a default here, but it's better to set it on the release workflow as input.

.github/workflows/webhook_image.yaml
mkulke: i just kept the plural ,input arg and the latest tag. its not really consistent with other artifacts, but we don't have to deal with that in this PR.

mkulke:
> > One question on the workflow and have you tested in with a branch on upstream, or are you planning to just test post-merge?
> 
> haven't tested on upstream with quay secret. I think I can do that for the push workflows, it will be tricker for the release workflow...

[release wf](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/34228946027/job/102071444467)

[push-to-main wf](https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/34228945935/job/102071503409)

.github/workflows/webhook_image_publish.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Push webhook image to quay.io
+      env:
+        QUAY_REGISTRY: ${{ inputs.registry }}
+        GHCR_REGISTRY: ghcr.io/${{ github.repository_owner }}
+        IMAGE_TAGS: ${{ inputs.image_tags || github.sha }}
+        IMAGE_NAME: ${{ needs.webhook.outputs.image-name }}
+      run: |
+        tags="${IMAGE_TAGS}"
+        latest=0
+        for t in ${tags/,/ }; do
```
stevenhorsman: Bob tells me that is only replaces the first instance of the comma, so suggests it should be `for t in ${tags//,/ }; do` instead to update them all. I don't think we have anything that uses three tags at the moment, but we don't validate that, so it's probably worth changing?

.github/workflows/webhook_image_publish.yaml
```diff
@@ -0,0 +1,102 @@
...
+      packages: read # Required to copy container images
+    needs: webhook
+    if: ${{ startsWith(inputs.registry, 'quay.io') }}
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Login to ghcr.io container registry
+      uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
```
stevenhorsman, line 64: In webhook_image.yaml we use `username: ${{ github.repository_owner }}`. I think actor is probably better, so maybe we should update that to match this?

.github/workflows/webhook_image_publish.yaml
```diff
@@ -0,0 +1,102 @@
+# Copyright Confidential Containers Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Build and push the pod VM images.
```
stevenhorsman:
```suggestion
# Build and push the peerpod webhook images.
```

stevenhorsman (commented):
A few minor observations

.github/workflows/webhook_image_publish.yaml
mkulke: I copied this verbatim from the other snippet that does multiples tags, but Bob is right, I don't think this will break anything.

stevenhorsman (approved):
LGTM. Thanks

bpradipt (approved):

PR #3280, opened 2026-09-04
podnetwork: skip VNIs still in use after restart

dnwe:
The pod index that each pod VM's VXLAN VNI is derived from was kept in memory only, so a restarted cloud-api-adaptor started again from zero and handed new pods the VNIs of pods still running on the node, where creating the host vxlan device then fails with EEXIST

Instead:

- scan the pod network namespaces under /run/netns at startup for vxlan1 devices on the configured port and range and seed the index past the highest VNI found (skipping any entries that cannot be opened or listed, such as a stale mount point)
- reject a minimum ID outside the 24-bit field or a pod index past it


src/cloud-api-adaptor/pkg/podnetwork/workernode.go
```diff
@@ -25,13 +33,13 @@ type workerNode struct {
 	tunneler tunneler.TunnelerConfigurator
 }
 
-// TODO: Pod index is reset when this process restarts.
-// We need to manage a persistent unique index number for each pod VM
+// The pod index lives in this process only; NewWorkerNode seeds it past the
+// VNIs still in use on the node so a restart does not reuse them.
```
stevenhorsman, line 37: Hey @dnwe - Bob is quite worried about podIndexManager being package-level now as each call to NewWorkerNode would re-invoke the scan and call SetMin again and they could skip some indices not in use. It is recommending moving podIndex into the workerNode struct to reduce the scope. I'm not sure this would impact us in production, the severity of hitting it seems low and it was package-level before, (allbeit with less impact without the seeding code), so it's not a blocker for me, but I thought I'd share it as you've obviously thought about this code a lot more than I have recently. Thanks

stevenhorsman (approved):
One comment shared, but not a blocker, so as long as the CI is passing this seems fine to me. Thanks

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
dnwe: @stevenhorsman thanks - not sure why Bob is worried as NewWorkerNode is only called once at startup as a singleton. The only time it would get repeat called is in unittests.

Regardless, the index can never go backward as we max(…) it against the current value and it would only go forward if the scan found in-use devices at and beyond the current index which wouldn’t be undesirable behaviour either

bpradipt:
@dnwe It would be good to understand your setup where you hit the issue.
Is this someway related to https://github.com/confidential-containers/cloud-api-adaptor/pull/3055 ?

dnwe:
@bpradipt yes #3055 is working towards solving the larger underlying problem (when you restart CAA it should be able to recover the peerpods it owns). This PR is fixing the related issue that a newly restarted CAA should be able to provision new peerpods without colliding with the orphaned ones that have been left around

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
```diff
@@ -43,6 +51,68 @@ func (p *podIndex) Get() int {
...
+			logger.Printf("skipping %s: %v", nsPath, err)
+			continue
+		}
+		for _, link := range links {
+			if link.Name() != vxlan.PodInterfaceName || link.Type() != "vxlan" {
+				continue
+			}
+			device, err := link.GetDevice()
+			if err != nil {
+				return 0, fmt.Errorf("failed to inspect %s in %s: %w", link.Name(), nsPath, err)
```
bpradipt: Should this log and continue ?

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
```diff
@@ -43,6 +51,68 @@ func (p *podIndex) Get() int {
...
+		links, err := ns.LinkList()
+		if closeErr := ns.Close(); closeErr != nil {
+			logger.Printf("failed to close network namespace %s: %v", nsPath, closeErr)
+		}
+		if err != nil {
+			logger.Printf("skipping %s: %v", nsPath, err)
+			continue
+		}
+		for _, link := range links {
+			if link.Name() != vxlan.PodInterfaceName || link.Type() != "vxlan" {
```
bpradipt: What about ppvxlanN devices in the host namespace? Won't similar issue occur if a ppvxlanN device is left in the host ns ?

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
```diff
@@ -60,6 +130,17 @@ func NewWorkerNode(networkConfig *tunneler.NetworkConfig) (WorkerNode, error) {
 		tunneler:      tun,
 	}
 
+	if networkConfig.TunnelType == "vxlan" {
```
bpradipt: nit: s/vxlan/DefaultTunnelType

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
dnwe: GetDevice() doesn't make any kernel calls, it just reads the data from the links we already discovered by LinkList() and since we already checked `link.Type() == "vxlan"` there shouldn't actually be any scenario where we can get back an err from `GetDevice()`. So if an err ever was returned, I think it would be more of a panic / internal state style scenario that we'd want to bail out for, rather than logging it and carrying on

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
dnwe: Yes you're probably right. I originally code for the clean restart case, but a failed setup might leave a ppvxlanN device in the host namespace with its VNI still in use before being moved to the pod namespace. Will add a commit to handle these too

src/cloud-api-adaptor/pkg/podnetwork/workernode.go
dnwe: 👍🏻

bpradipt (approved):
/lgtm
Needs a rebase though

dnwe:
@bpradipt thanks, have rebased, but it'll need the workflow runs re-approving

PR #3285, opened 2026-09-07
byom: load podvm binaries image locally for the BYOM build

Amulyam24:
The BYOM binaries workflow needs the podvm-binaries image available as the base image for the BYOM image build.

Load the image into the local Docker daemon via `DOCKER_OPTS=--load`, and set `EXPORT_BINARIES=false` so the default buildx driver has a single output.

This fixes the failing workflow to publish the BYOM binaries image- https://github.com/confidential-containers/cloud-api-adaptor/actions/runs/33856485393/job/100970893431

### DIFF
--- a/.github/workflows/podvm_byom_binaries_publish.yaml
+++ b/.github/workflows/podvm_byom_binaries_publish.yaml
@@ -99,7 +99,8 @@ jobs:
           REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
           PODVM_TAG: ${{ inputs.image_tag || github.sha }}
           ARCH: ${{ inputs.arch }}
-          PUSH: "true"
+          # Load the binaries image locally for the BYOM build to use as base.
+          EXPORT_BINARIES: "false"
 
       - name: Build BYOM binaries image
         working-directory: src/cloud-api-adaptor/podvm
--- a/src/cloud-api-adaptor/podvm/Makefile
+++ b/src/cloud-api-adaptor/podvm/Makefile
@@ -83,7 +83,7 @@ endif
 		--build-arg VERIFY_PROVENANCE=$(VERIFY_PROVENANCE) \
 		$(if $(AUTHFILE),--build-arg AUTHFILE=$(AUTHFILE),) \
 		$(if $(DEFAULT_AGENT_POLICY_FILE),--build-arg DEFAULT_AGENT_POLICY_FILE=$(DEFAULT_AGENT_POLICY_FILE),) \
-		$(if $(filter $(EXPORT_BINARIES),false),,-o type=local,dest="./resources/binaries-tree") \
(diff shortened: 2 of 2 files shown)


stevenhorsman (commented):
Do we need to update the name of this workflow and the `build-and-push` step, given that we aren't publishing, or pushing the byom binaries any more?

Amulyam24:
> Do we need to update the name of this workflow and the `build-and-push` step, given that we aren't publishing, or pushing the byom binaries any more?

Do you mean pushing podvm binaries image?

.github/workflows/podvm_byom_binaries_publish.yaml
ldoktor, line 111: Aren't we going to end-up with the same situation in this step? The `Push BYOM binaries image` will require it to be present in there as well, right? (I'm not certain about this, just trying to apply what helped in the previous step)

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -99,7 +99,10 @@ jobs:
           REGISTRY: ${{ inputs.registry != '' && inputs.registry || 'quay.io/confidential-containers' }}
           PODVM_TAG: ${{ inputs.image_tag || github.sha }}
           ARCH: ${{ inputs.arch }}
-          PUSH: "true"
+          # Load the binaries image into the local Docker daemon so the BYOM
+          # image build below can reference as the base image.
+          EXPORT_BINARIES: "false"
+          DOCKER_OPTS: "--load"
```
mkulke: can we maybe make this implicit in the Makefile? if export binaries == false then we want the docker builds process to load the image, so it can be used in follow-up steps? (I'm suprised that isn't the default)

Amulyam24: Agree, this would be much better, modified the condition.

.github/workflows/podvm_byom_binaries_publish.yaml
Amulyam24: The default docker driver is being changed in the earlier docker build step which should have this image loaded and available for push, I don't think it would be required here. I tested this scenario locally to verify it.

mkulke (approved):

stevenhorsman (approved):
I think this is worth trying

ldoktor (approved):
makes sense to me, let's see how that behaves.

Amulyam24:
Hi, can this PR be merged?

stevenhorsman:
> Hi, can this PR be merged?

I think so

Amulyam24:
> > Hi, can this PR be merged?
> 
> I think so

I'm not sure if I have permissions to do it, do you mind merging it?
I'll trigger a run to check if it works.

stevenhorsman:
> I'm not sure if I have permissions to do it, do you mind merging it? I'll trigger a run to check if it works.

You are a committer, so you should do, but I can merge

PR #3305, opened 2026-09-15
Kata 4.2 release bumps

stevenhorsman:
Update version pins post Kata 4.2.0 release


stevenhorsman:
Hmm, the `TestLibvirtPodsMTLSCommunication` tests are failing now. I will try and look into this more tomorrow once the daily e2e test has runs to rule out it being a different change that's caused this.

wainersm:
Hi @stevenhorsman ,
don't we need to bump the appVersion in the Chart.yaml of the charts?

stevenhorsman:
> Hi @stevenhorsman , don't we need to bump the appVersion in the Chart.yaml of the charts?

Sorry, my PR description was misleading - this is just the kata bump, which needs merging before the release pinning gets done. e.g. this is the same as https://github.com/confidential-containers/cloud-api-adaptor/pull/3202, the release PR will be like https://github.com/confidential-containers/cloud-api-adaptor/pull/3219/changes

stevenhorsman:
The test failure is:
```
[CDH] [ERROR]: Image Pull error: Failed to pull image quay.io/curl/curl:8.4.0 from all mirror/mapping locations or original location: image: quay.io/curl/curl:8.4.0, error: Errors happened when pulling image: Failed to decode layer data stream: Failed to unpack layer: Failed to read entries from the tar
```
IBM Bob thinks it's due to an `astral-tokio-tar` bump (0.6.2->0.6.4) in image-rs that broke it. Investigating further...

wainersm:
> > Hi @stevenhorsman , don't we need to bump the appVersion in the Chart.yaml of the charts?
> 
> Sorry, my PR description was misleading - this is just the kata bump, which needs merging before the release pinning gets done. e.g. this is the same as #3202, the release PR will be like https://github.com/confidential-containers/cloud-api-adaptor/pull/3219/changes

I was not sure, it's being a while since I last participated on the release and I forgot.

stevenhorsman:
I've raised the issue to guest-components for reference: https://github.com/confidential-containers/guest-components/issues/1737, but a bump of the curl image fixed the problem.

stevenhorsman:
Ah - curl:8.22 does not have s390x and 8.4 (the prior version that failed is the latest which does), so x86 is passing, but s390x failing now :(. We will need to find an s390x supported curl image.

wainersm (commented):
Thanks @stevenhorsman !

wainersm (approved):

mkulke (approved):

PR #3306, opened 2026-09-16
byom: minor improvements to setup-podvm-byom.sh

Amulyam24:
Improve `setup_podvm_byom.sh` script by
-  Adding root and Ubuntu OS check
- Make SSH configuration and `authorized_keys` insertion idempotent on repeat runs
- Add sshd config syntax validation
- Switch BYOM binaries image from from Quay to GHCR


src/cloud-api-adaptor/hack/setup-podvm-byom.sh
```diff
@@ -11,25 +26,76 @@ if [[ "$(id -u)" -ne 0 ]]; then
...
 fi
 
-PODVM_BYOM_BINARIES_IMAGE=${PODVM_BYOM_BINARIES_IMAGE:-"podvm-byom-binaries-ubuntu-amd64:latest"}
+if [[ "$(. /etc/os-release 2>/dev/null && echo "${ID}")" != "ubuntu" ]]; then
+    echo "Error: this script requires Ubuntu."
+    exit 1
+fi
+
+ARCH="$(dpkg --print-architecture)"
+PODVM_BYOM_BINARIES_IMAGE=${PODVM_BYOM_BINARIES_IMAGE:-"quay.io/confidential-containers/podvm-byom-binaries-ubuntu-${ARCH}:latest"}
```
mkulke: let's default to GHCR here, as we decided in the last community call.

src/cloud-api-adaptor/hack/setup-podvm-byom.sh
```diff
@@ -11,25 +26,76 @@ if [[ "$(id -u)" -ne 0 ]]; then
...
+    command -v gpg  &>/dev/null || missing+=(gnupg)
+    command -v lsb_release &>/dev/null || missing+=(lsb-release)
+    if ((${#missing[@]})); then
+        echo "Installing prerequisites: ${missing[*]}"
+        apt-get update -qq
+        apt-get install -y "${missing[@]}"
+    fi
+}
+
+install_docker() {
```
mkulke: I'm not sure whether installing docker on people's machines is in scope for the project and we probably should also not (silently) enable the docker service on people's machines either. It's fine to flag the absence of it, we can describe install steps in the README.

mkulke (commented):
If we want to ship a setup-podvm-byom script to users, so they don't have to clone the repo or download a release maybe can make it as part of a release. the main branch is a moving target and there can be drift between a release and the content of this url at a given time.

src/cloud-api-adaptor/hack/setup-podvm-byom.sh
Amulyam24: Makes sense, removed it.

src/cloud-api-adaptor/hack/setup-podvm-byom.sh
Amulyam24: Ah okay, was not aware of it, moved it to GHCR, Thanks!

Amulyam24:
> If we want to ship a setup-podvm-byom script to users, so they don't have to clone the repo or download a release maybe can make it as part of a release. the main branch is a moving target and there can be drift between a release and the content of this url at a given time.

Agree, that would be much better, made the changes, PTAL!

Team: If there are alternate opinions/approaches to shipping the setup script, feel free to share them.

mkulke (approved):

src/cloud-api-adaptor/hack/setup-podvm-byom.sh
```diff
@@ -41,16 +78,20 @@ PubkeyAuthentication yes
...
-chmod 600 /home/${USER_NAME}/.ssh/authorized_keys
-chown -R ${USER_NAME}:${USER_NAME} /home/${USER_NAME}
+mkdir -p "/home/${USER_NAME}/.ssh" && chmod 700 "/home/${USER_NAME}/.ssh"
+# Append the key only if it isn't already authorized, so re-runs stay idempotent.
+pub_key="$(cat "${SSH_PUBLIC_KEY_PATH}")"
+touch "/home/${USER_NAME}/.ssh/authorized_keys"
+grep -qxF "${pub_key}" "/home/${USER_NAME}/.ssh/authorized_keys" \
+    || echo "${pub_key}" >> "/home/${USER_NAME}/.ssh/authorized_keys"
+chmod 600 "/home/${USER_NAME}/.ssh/authorized_keys"
+chown -R "${USER_NAME}:${USER_NAME}" "/home/${USER_NAME}"
```
stevenhorsman, line 94: The commit message says `sshd config syntax validation`, so I was expecting something like:
```
sshd -t || { echo "Error: sshd config syntax check failed"; exit 1; }
```
somewhere? Am I missing something?

.github/workflows/release.yaml
```diff
@@ -10,6 +10,10 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.release.tag_name || github.ref }}
+  cancel-in-progress: true
```
stevenhorsman, line 15: I don't think we want cancel-in-progress for the release as we could get into partial release, so this is better as false so it waits if re-run?

src/cloud-api-adaptor/hack/setup-podvm-byom.sh
```diff
@@ -11,25 +26,47 @@ if [[ "$(id -u)" -ne 0 ]]; then
...
 # Backup existing sshd_config
-cp "${SSH_CONFIG}" "/etc/ssh/sshd_config.bak.$(date +%F_%T)"
+if ! ls /etc/ssh/sshd_config.bak.* &>/dev/null; then
+    cp "${SSH_CONFIG}" "/etc/ssh/sshd_config.bak.$(date +%F_%T)"
+fi
 
-# Append SSH config to disable login
-if [[ "$DISABLE_SSH_LOGIN" == "true" ]]; then
+# Append SSH config to disable login (guarded so re-runs don't duplicate it)
+if [[ "$DISABLE_SSH_LOGIN" == "true" ]] && ! grep -qE "^\s*#\s*Disable all forms of login" "$SSH_CONFIG"; then
```
ldoktor, line 69: How about using `/etc/ssh/sshd_config.d/00_caa_byom.conf` instead of backup and direct modification?

.github/workflows/release.yaml
```diff
@@ -55,9 +59,29 @@ jobs:
...
+      - name: Checkout Code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.ref }}
+          persist-credentials: false
+
+      - name: Upload setup script to release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ github.event.release.tag_name }}
```
ldoktor, line 80: The `setup-podvm-byom.sh` defaults to `latest`, shouldn't we publish it as well?

.github/workflows/podvm_byom_binaries_publish.yaml
```diff
@@ -88,15 +86,15 @@ jobs:
       - name: Login to Container Registry
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
-          registry: quay.io
-          username: ${{ vars.QUAY_USERNAME }}
-          password: ${{ secrets.QUAY_PASSWORD }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
ldoktor, line 91: The use of `ghcr` might become problematic as users need to authenticate first. Shouldn't we document the process? I mean as a user I'd either have to deploy GH token to all my VMs or pre-pull the image first and publish it elsewhere to consume it from my VMs.

ldoktor (approved):
In general I like this, left a few comments for your consideration.

.github/workflows/release.yaml
ldoktor: ah, very good catch +1 on that

PR #3307, opened 2026-09-17
fix(libvirt): free domain handle and correct IP retry log format in CreateDomain

ajaypvictor:
Two correctness fixes in `CreateDomain` in the libvirt provider.

`DomainDefineXML` returns a `*libvirt.Domain` backed by a C `virDomainPtr` that must be explicitly freed once the caller is done with it. The handle was never released in `CreateDomain`, leaking one reference per pod creation regardless of whether the start succeeded or failed. A log-only defer is used rather than the `freeDomain(dom, &err)` helper, because `freeDomain` writes a `Free()` failure into the named return error — on the success path that would overwrite a nil error and cause the caller to delete a running VM. Freeing the handle only releases the client-side C pointer; the domain keeps running regardless.

The IP retry timeout log formatted `GetDomainIPsSleep` with `%d`, treating the `time.Duration` as a raw `int64` and printing nanoseconds with a literal `s` suffix appended by the format string — so `3 * time.Second` came out as `3000000000s`. Switching to `%v` lets `Duration.String()` handle the formatting and prints `3s` as intended.


src/cloud-providers/libvirt/libvirt.go
```diff
@@ -652,6 +652,7 @@ func CreateDomain(ctx context.Context, libvirtClient *libvirtClient, v *vmConfig
 	if err != nil {
 		return nil, fmt.Errorf("Failed to define domain: %s", err)
 	}
+	defer freeDomain(dom, &err)
```
wainersm: HI @ajaypvictor , one comment made by Claude that makes sense to me:
```
freeDomain(dom, &err) writes a dom.Free() failure into the named return err. On the success path, err is nil when the defer fires — so any Free() error overwrites it. The caller sees (result with UUID, non-nil error) and deletes the running VM.

Root cause: dom.Free() only releases a client-side C pointer — the domain keeps running regardless. A Free() failure is irrelevant to whether creation succeeded, but the current code elevates it to "creation failed."

---

Solutions

1. Log-only defer (simplest):
defer func() {
    if freeErr := dom.Free(); freeErr != nil {
        logger.Printf("Warning: failed to free domain handle: %v", freeErr)
    }
}()

2. Add a logFreeDomain helper alongside freeDomain, for use in contexts where Free errors should not propagate. Use it here; keep freeDomain(domain, &err) in DeleteDomain/lookup paths.
```

The solution 1) works for me.

ajaypvictor: Updated to use the suggested log-only defer apprach. Thanks Wainer..

stevenhorsman (approved):
LGTM. Thanks!

wainersm (approved):
Thanks @ajaypvictor !

PR #3135, opened 2026-06-04
add configurable TLS min version and cipher suite support

thejasn:
Post-quantum cryptography key exchange mechanisms such as ML-KEM (X25519MLKEM768) require TLS 1.3 — TLS 1.2 does not support the key exchange groups needed for quantum-resistant handshakes. Some platforms require all components to consume TLS settings from central configuration sources and respect custom TLS profiles, so customers can opt into PQC-resilient ciphers across the entire platform via documented knobs.

CAA hardcodes TLS 1.2 with no cipher suite control and the peer-pods webhook ignores TLS configuration entirely, making both non-compliant with this requirement.

Add TLS_MIN_VERSION and TLS_CIPHER_SUITES env vars to CAA and the peer-pods webhook. Warn when insecure TLS options are in use, and when TLS settings are changed — since they are baked into peer pod VMs at creation time and require pod recreation to take effect.

## Testing

Tested on libvirt (KVM) with a 2-node Kubernetes v1.30.0 cluster (Ubuntu 22.04, kernel 5.15.0).
  - Deployed CAA with TLS_MIN_VERSION=VersionTLS12 and IANA Intermediate profile cipher
  suites; confirmed startup warning in logs.
  - Created a peer pod; confirmed tls-min-version and tls-cipher-suites in apf.json on the
  worker node and inside the peer pod VM.
  - Pod reached Running state, confirming agent proxy to APF TLS handshake succeeded.
  - TLS_SKIP_VERIFY warning fires when set.
  - Webhook TLS 1.3 enforcement verified: TLS 1.2 rejected with alert 70 (protocol_version),
  TLS 1.3 accepted with TLS_AES_128_GCM_SHA256.

Fixes: rhjira#KATA-5129
Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #3314, opened 2026-09-19
peerpod-ctrl: replace kube-rbac-proxy with native metrics SecureServing

### LINKED ISSUE #3313: peerpod-ctrl: replace the kube-rbac-proxy sidecar with controller-runtime metrics SecureServing
The peerpod-ctrl chart still ships a kube-rbac-proxy sidecar pinned to `registry.k8s.io/kubebuilder/kube-rbac-proxy:v0.14.0`. Kubebuilder stopped scaffolding that sidecar and the image has already moved registries once, which broke installs of older charts, and #2921 (the registry bump for this chart) never landed. #3135 already replaced the sidecar in the webhook with controller-runtime's SecureServing plus the authn/authz filter, so I'd like to do the same for peerpod-ctrl: drop the container, serve /metrics over TLS on 8443 from the manager, keep the metrics-reader ClusterRole, and remove the authProxy values. The one behaviour change is that `authProxy.enabled=false` (plaintext metrics on 8080) goes away. Happy to open the PR if that sounds right.

david-long1:
kubebuilder has discontinued its kube-rbac-proxy images: the `gcr.io/kubebuilder` copies went away in March 2025 and the `registry.k8s.io/kubebuilder/kube-rbac-proxy` mirror this chart pins at v0.14.0 is explicitly unsupported and not recommended (kubernetes-sigs/kubebuilder#3907), so the sidecar has no upgrade path. controller-runtime serves the same authn/authz-gated TLS metrics endpoint natively via `SecureServing` with the `WithAuthenticationAndAuthorization` filter, which is what kubebuilder now scaffolds and what the webhook already switched to in #3135.

This PR mirrors #3135 for peerpod-ctrl. The sidecar container and the `authProxy` chart values are removed, the manager binds metrics on `:8443` with `SecureServing` enabled, and the metrics-auth RBAC (tokenreviews + subjectaccessreviews) becomes unconditional and is renamed from `proxy-role`/`proxy-rolebinding` to `metrics-auth-role`/`metrics-auth-rolebinding`. The install docs that still showed `2/2` READY for the peerpod-ctrl and webhook pods are corrected to `1/1`.

Fixes: #3313

## Compatibility

- The `authProxy.*` chart values (`authProxy.enabled`, `authProxy.image`) are removed. Setting them has no effect. The `authProxy.enabled=false` path that served plaintext metrics on port 8080 no longer exists.
- Metrics are always served over TLS on port 8443 and require a bearer token whose subject is bound to the `metrics-reader` ClusterRole. The `tokenreviews`/`subjectaccessreviews` rules the proxy used are unchanged, only renamed.
- The metrics Service name (`<namePrefix>controller-manager-metrics-service`), its `https` port (8443), and the `metrics-reader` ClusterRole are unchanged, so existing scrape configs keep working.
- On upgrade Helm deletes the old `proxy-role`/`proxy-rolebinding` objects and creates the renamed ones.
- The serving certificate stays self-signed: controller-runtime generates one at startup, as kube-rbac-proxy did when no cert was supplied. Scrapers still need to skip verification or trust that certificate.

## Testing

Toolchain: go 1.27.1, helm v4.3.0, GNU sed 4.10.

```
$ cd src/peerpod-ctrl
$ go mod tidy && go build ./... && go vet ./...
build ok
vet ok
# go.mod gains k8s.io/apiserver, k8s.io/component-base, cel-go, otel
# exporters and friends as indirect deps (same set the webhook carries)

$ go test ./...
?   	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl	[no test files]
?   	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl/api/v1alpha1	[no test files]
ok  	github.com/confidential-containers/cloud-api-adaptor/src/peerpod-ctrl/controllers	0.629s

$ make verify            # verify-modules, verify-gen, verify-manifests
make verify exit=0       # working tree clean afterwards

$ make static-helm-check
Linting helm chart...
==> Linting chart/
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
Rendering helm chart templates...
Static helm checks passed

$ helm template x chart | grep -cE 'kube-rbac-proxy|authProxy'
0
$ helm template x chart | grep -nE 'metrics-bind-address|containerPort: 8443|name: https|targetPort: https'
240:  - name: https
243:    targetPort: https
295:        - --metrics-bind-address=:8443
300:        - containerPort: 8443
301:          name: https
$ helm template x chart | grep -nE '^  name: peerpodctrl-(metrics-auth-role|metrics-auth-rolebinding|metrics-reader|controller-manager-metrics-service)$'
26:  name: peerpodctrl-metrics-auth-role
52:  name: peerpodctrl-metrics-reader
145:  name: peerpodctrl-metrics-auth-rolebinding
236:  name: peerpodctrl-controller-manager-metrics-service

$ cd ../cloud-api-adaptor
$ make static-helm-check   # umbrella chart, every provider values file
...
  libvirt.yaml
==> Linting install/charts/peerpods/
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
  vsphere.yaml
==> Linting install/charts/peerpods/
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
Static helm checks passed
# Chart.lock unchanged
```


bpradipt (approved):
/lgtm
Thanks @david-long1

david-long1:
Thanks for the review and the CI approval, @bpradipt. CI is green now. As I understand it the repo needs a second approval to merge — @stevenhorsman, since you reviewed the webhook side of this in #3135, would you have a moment to take a look? Happy to address anything.

stevenhorsman (approved):
The code seems reasonable to me, but the PR will need rebasing as I think some of the dependabot bumps have clashed with it. It would be interesting to see if @thejasn had any thought about the TLS minimum versions here and if we should consider a follow up PR related to that?

david-long1:
Rebased onto current `main`. The only conflict was in `go.sum` from the recent libvirt bump, and after re-running tidy the `cloud-api-adaptor/go.mod` change is no longer needed, so I dropped that part.

On the TLS question: this PR doesn't change the minimum TLS version directly. The webhook already picks up the operator-provided `TLS_MIN_VERSION` / `TLS_CIPHER_SUITES` settings and applies them to both the webhook and metrics servers. `peerpod-ctrl` doesn't have that plumbing today, so its new metrics server just uses Go's default TLS settings, with TLS 1.2 as the minimum.

I'm happy to follow up and add the same `tlsMinVersion` / `tlsCipherSuites` handling to `peerpod-ctrl` after this lands, if that makes sense to you and @thejasn.

stevenhorsman:
Hmm, this introduced multiple go modules with security vulns in, which wasn't picked up by our CI. Maybe because they are indirect? I think we have a gap here as we shouldn't have merged it due to this.

david-long1:
@stevenhorsman Thanks for flagging this. I dug into what changed here.

The new `metrics/filters` import pulls `k8s.io/apiserver/pkg/authorization/cel` into the `peerpod-ctrl` build, which in turn brings in `cel-go` and the otel SDK. That's why these alerts showed up now.

On the CI gap: `govulncheck` is running in source mode, so it only fails when the vulnerable code is actually reachable. It sees the `cel-go` advisory, but reports the affected functions as not called from `peerpod-ctrl`. The otel advisory isn't in Go's vuln DB yet, so `govulncheck` doesn't see that one at all. Dependabot is version-based, so it flags both regardless of reachability.

#3323 already covers the `cel-go` bump. I also tested `cel-go` 0.30.0 together with otel 1.45.0 locally and the build/tests are clean. Once #3323 lands, I'm happy to send the otel bump separately if that's useful.

Longer term, if we want CI to catch this class of issue before merge, we could add a module-level scanner such as `osv-scanner` alongside `govulncheck`.

stevenhorsman:
> @stevenhorsman Thanks for flagging this. I dug into what changed here.
> 
> The new `metrics/filters` import pulls `k8s.io/apiserver/pkg/authorization/cel` into the `peerpod-ctrl` build, which in turn brings in `cel-go` and the otel SDK. That's why these alerts showed up now.
> 
> On the CI gap: `govulncheck` is running in source mode, so it only fails when the vulnerable code is actually reachable. It sees the `cel-go` advisory, but reports the affected functions as not called from `peerpod-ctrl`. The otel advisory isn't in Go's vuln DB yet, so `govulncheck` doesn't see that one at all. Dependabot is version-based, so it flags both regardless of reachability.
> 
> #3323 already covers the `cel-go` bump. I also tested `cel-go` 0.30.0 together with otel 1.45.0 locally and the build/tests are clean. Once #3323 lands, I'm happy to send the otel bump separately if that's useful.
> 
> Longer term, if we want CI to catch this class of issue before merge, we could add a module-level scanner such as `osv-scanner` alongside `govulncheck`.

Thanks for looking into it - I didn't mean to suggest it was your problem. I think I need to consider OSV-scanner, I'm not sure why I didn't add it here when I added it to Kata, probably just was waiting to see how it performed there and then forgot!

PR #3322, opened 2026-09-21
libvirt: use fixed delay and context propagation for IP retries

### LINKED ISSUE #3321: libvirt: IP retry in CreateDomain uses exponential backoff and ignores context
### Description
In `src/cloud-providers/libvirt/libvirt.go`, `CreateDomain` calls `retry.Do` to poll for domain IPs with:
- `retry.Attempts(GetDomainIPsRetries)` (20)
- `retry.Delay(GetDomainIPsSleep)` (3s)

Because `retry.DelayType` is not specified, `retry-go/v4` defaults to exponential backoff without an upper bound (`3s << n`). A transient or permanent network failure causes retry delays to escalate dramatically (e.g. ~51m on attempt 10, ~18 days on attempt 20), hanging domain creation rather than timing out after the intended ~60s.

Additionally, `retry.Context(ctx)` is not passed to `retry.Do`. If the caller's context is cancelled or hits a deadline, the retry loop continues sleeping and retrying in the background instead of aborting promptly.

### Expected Behavior
- IP polling should use fixed delay between attempts (~60s total for 20 attempts of 3s).
- Context cancellation should abort the wait loop immediately.

chathuryaadapa:
CreateDomain called `retry.Do` without `retry.DelayType` or `retry.Context`. Because `retry-go/v4` defaults to exponential backoff with uncapped delay, retries doubled on each attempt rather than staying at a fixed 3s interval, and cancellation of the caller context did not abort the retry loop.

Use `retry.FixedDelay` and `retry.Context(ctx)` via `waitForDomainIPs`, and assign the retrieved IP slice directly without a redundant getDomainIPs call on success.

Add `//go:build cgo` tag to `libvirt_test.go` so the file compiles under CGO_ENABLED=0 (the Makefile default outside the libvirt build).

Add unit tests covering immediate success, retries, retry exhaustion, unrecoverable errors, and context cancellation.

Fixes: #3321
Assisted-By: IBM Bob <noreply@ibm.com>


stevenhorsman:
Is there are reason why exponential backoff isn't suitable here. I thought it was often considered best practice to avoid spamming?

chathuryaadapa:
> Is there are reason why exponential backoff isn't suitable here. I thought it was often considered best practice to avoid spamming?

Yes steve, Exponential backoff makes sense when retrying against a external service and backing off reduces load on the remote. Here we're just polling the local libvirt daemon waiting for an IP to be assigned to the vm
The problem is that `GetDomainIPsRetries = 20` and `GetDomainIPsSleep = 3s` results in ~60s budget. Exponential backoff without a upper limit - attempt 10 alone sleeps ~51 minutes. Fixed delay keeps the behaviour consistent with what the constants say.

If you prefer backoff style, maybe I can update `retry.MaxDelay(GetDomainIPsSleep)` upper limit it at 3s.

stevenhorsman:
> > Is there are reason why exponential backoff isn't suitable here. I thought it was often considered best practice to avoid spamming?
> 
> Yes steve, Exponential backoff makes sense when retrying against a external service and backing off reduces load on the remote. Here we're just polling the local libvirt daemon waiting for an IP to be assigned to the vm The problem is that `GetDomainIPsRetries = 20` and `GetDomainIPsSleep = 3s` results in ~60s budget. Exponential backoff without a upper limit - attempt 10 alone sleeps ~51 minutes. Fixed delay keeps the behaviour consistent with what the constants say.
> 
> If you prefer backoff style, maybe I can update `retry.MaxDelay(GetDomainIPsSleep)` upper limit it at 3s.

Thanks for the explanation. I think the fixed delay sounds okay with 3s poll as that shouldn't be too "spammy"

stevenhorsman (approved):
One minor comment, otherwise it seems okay to me. Thanks

bpradipt (approved):
/lgtm
Makes sense. 
Thanks @chathuryaadapa

PR #3329, opened 2026-09-22
peerpod-ctrl: honour the cluster TLS profile on the metrics server

### LINKED ISSUE #3328: peerpod-ctrl: metrics server does not honour the cluster TLS profile (TLS_MIN_VERSION / TLS_CIPHER_SUITES)
Since #3314 peerpod-ctrl serves its metrics over TLS from controller-runtime's metrics server, but that server runs with Go's default TLS settings. The webhook, which made the same switch in #3135, additionally reads the operator-injected `TLS_MIN_VERSION` and `TLS_CIPHER_SUITES` (`src/webhook/tlsconfig.go`) and applies them to both its webhook server and its metrics server, so a cluster TLS profile (for example TLS 1.3 only) reaches the webhook but not peerpod-ctrl. @stevenhorsman raised this on #3314; opening it here so it has a home.

Proposal: mirror the webhook. Read the same two variables in peerpod-ctrl, parse them with the same rules (`VersionTLS12`/`VersionTLS13`, IANA cipher suite names, no cipher suites with TLS 1.3), and pass them as `TLSOpts` on the metrics server; add `tlsMinVersion` / `tlsCipherSuites` to the peerpod-ctrl chart with the same env injection the webhook chart uses. Operator-side injection for peerpod-ctrl is a separate change in the operator repo.

cc @thejasn, since the webhook side is yours.

david-long1:
Fixes #3328. Follow-up to #3314, as offered there.

The metrics server that replaced kube-rbac-proxy runs with Go's default TLS settings, while the webhook reads the operator-injected `TLS_MIN_VERSION` / `TLS_CIPHER_SUITES` and applies them to its servers, so a cluster TLS profile reached the webhook but not peerpod-ctrl.

This reads the same two variables in peerpod-ctrl, parses them with the webhook's rules — `tlsconfig.go` and its table test are copied unchanged from `src/webhook` — and passes the result as `TLSOpts` on the metrics server. Go's defaults still apply when both are unset. The chart gains `tlsMinVersion` / `tlsCipherSuites` with the same conditional env injection as the webhook chart, documented in the chart README. `k8s.io/component-base` moves from indirect to direct in `go.mod`; nothing else in the dependency graph changes.

Not in this PR: the operator injecting the variables for peerpod-ctrl the way it does for the webhook — that lives in the operator repo.

Testing: `go test ./...` in `src/peerpod-ctrl` (the copied `TestParseTLSOptions` table passes), `GOOS=linux go vet ./...`, `make tidy-check`, `helm lint`, and `helm template --set tlsMinVersion=VersionTLS13` renders only `TLS_MIN_VERSION`.


david-long1:
The earlier red build here was worth a look on its own: `peerpod-ctrl`'s build target compiled `main.go` alone, so the new `tlsconfig.go` was invisible to the compiler and the build failed with `undefined: parseTLSOptions`, while `go vet ./...` and `go test ./...` in the same target passed. The webhook's Makefile already builds the package, so this now does the same. Reproduced both ways locally: single file fails with the CI error, package builds clean. Rebased into one commit.

@snir911 thanks for approving the run yesterday, would you mind doing the same on the new head?

@stevenhorsman this is the follow-up I offered on #3314, mirroring the webhook's `TLS_MIN_VERSION` / `TLS_CIPHER_SUITES` handling onto the peerpod-ctrl metrics server, which otherwise sits at Go's default TLS 1.2. @bpradipt would you take the second review?

thejasn:
LGTM
thanks @david-long1!

You'll also need to update the Dockerfile (triggered from `make docker-build`) for the CI to pass.

david-long1:
Thanks @thejasn, you were right. The Dockerfile had the same shape in two places: it copied `main.go` by name, so `tlsconfig.go` never reached the image, and then built `go build ... -o manager main.go`. Mirrored `src/webhook/Dockerfile`, which already handles its own `tlsconfig.go`: copy the file, build the package.

Verified with `make docker-build` locally, and separately through the container with the full `-tags=...,libvirt` set, since the builder stage installs `libvirt-dev` itself and my host cannot. Folded into the single commit.

src/peerpod-ctrl/tlsconfig.go
stevenhorsman: I don't love that this is copied from a maintenance PoV, but I don't necessarily have a better solution, other that creating a shared pkg and moving it there?

stevenhorsman (approved):
I think this is okay. I am concerned about the duplication of code though.

src/peerpod-ctrl/tlsconfig.go
david-long1: @stevenhorsman Fair point. I copied it to keep this PR to the TLS fix. The obvious home is `src/cloud-providers/util`, since peerpod-ctrl already imports cloud-providers, but the webhook doesn't: it'd need a `replace ../cloud-providers` and its Dockerfile/Makefile/CI build context moved up to `src/` like peerpod-ctrl's. I can do that as a follow-up right after this lands — does that work, or would you rather it go in here?

